PR #325, opened 2017-05-12
Android readonly

bekao:
Android readonly components.  Note - separation is only implemented for TextBlock and Image.  Separation on other components are currently commented out.


source/shared/cpp/ObjectModel/AdaptiveCard.cpp
```diff
@@ -103,9 +103,14 @@ std::shared_ptr<AdaptiveCard> AdaptiveCard::Deserialize(const Json::Value& json)
     return result;
 }
 
-std::shared_ptr<AdaptiveCard> AdaptiveCard::DeserializeFromString(const std::string& jsonString)
+std::shared_ptr<AdaptiveCard> AdaptiveCard::DeserializeFromString(const std::string& jsonString) throw(AdaptiveCards::AdaptiveCardParseException)
 {
-    Json::Value jsonValue(jsonString);
+    Json::Reader reader;
+    Json::Value jsonValue;
+    if (!reader.parse(jsonString.c_str(), jsonValue))
```
riarenas: Thanks for fixing this. I will also send a PR for the FromString for the rest of the card elements.

source/shared/cpp/ObjectModel/HostOptions.h
```diff
@@ -52,11 +52,11 @@ struct SeparationOptions
 {
     unsigned int spacing = 10;
     unsigned int lineThickness = 0;
-    std::string lineColor;
+    std::string lineColor = "#FF101010";
 };
 struct ImageSizeOptions
 {
-    unsigned int smallSize = 60;
+    unsigned int smallSize = 80;
```
riarenas: >unsigned int smallSize = 80; [](start = 4, length = 28)

Changing this will affect the UWP renderer as well. If sizes are overall smaller in android, does it make sense to add an offset during image rendering directly rather than modifying the default host config?

PR #329, opened 2017-05-15
Add action events to the uwp renderer 

RebeccaAnne:


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -623,6 +631,7 @@ namespace AdaptiveCards { namespace XamlCardRenderer
 
         UINT currentAction = 0;
 
+        ComPtr<XamlCardRenderer> strongRenderer(renderer);
```
riarenas: [nit] we've been calling these local in the rest of the code.

RebeccaAnne: Chris had asked that i call it "strong" rather than "local" on a previous change where i had defined the com pointer for the purpose of taking a strong reference to pass to a lambda.  

---
In reply to: [116594514](https://github.com/Microsoft/AdaptiveCards/pull/329#discussion_r116594514) [](ancestors = 116594514)

anbare:
@RebeccaAnne how do you retrieve the inputs that the user typed into text boxes? Will there be a future separate GetInputValues method that consumers of the renderer can call at any point in time? Or will this action event be updated to include the current input parameters?

A separate GetInputValues method might be nice since it could help allow hosts to persist and store what a user typed. For example if a user started typing into the card and then leaves the app, the app needs to persist what the user typed (even if the app gets terminated) so that when the user returns, what they typed is still there.

PR #341, opened 2017-05-17
added hostConfig tools for SCSS

danmarshall:
These tools will allow SCSS developers to integrate HostConfig styling into their SCSS pipeline.
There are 2 tools, *json-to-scss* and *scss-to-json* for a complete roundtrip.


source/tools/hostConfig/json-to-scss/test.js
```diff
@@ -0,0 +1,179 @@
...
+            right: 8,
+            bottom: 8
+        }
+    },
+    container: {
+        separation: {
+            spacing: 8
+        },
+        normal: {
+        },
```
danmarshall: I need to add a default into normal so it isn't empty

Issue #346, opened 2017-05-17
Remove isSelected from schema


dsanvita:
The Choice schema has an isSelected property, but this is not implemented in the HTML renderer.

https://github.com/Microsoft/AdaptiveCards/pull/347

dclaux:
There shouldn't be an isSelected property. This has already been discussed; just like any other input, the only way to pre-select a value (or several in the case of ChoiceSet) is to include the "value" property in the payload in the same format a client application would expect it at runtime, e.g. a comma-separated list of values. For instance, "1,2,3" where 1, 2 and 3 are three valid choice values in the ChoiceSet.

This issue should be changed into "Remove isSelected from schema."

matthidinger:
Thanks David, I'll take this on myself to update

matthidinger:
This is done

PR #347, opened 2017-05-17
Set isSelected property of choice input in html renderer

dsanvita:
https://github.com/Microsoft/AdaptiveCards/issues/346

### DIFF
--- a/source/html/renderer/src/card-elements.ts
+++ b/source/html/renderer/src/card-elements.ts
@@ -470,7 +470,7 @@ export class FactSet extends CardElement {
             return this.speak + '\n';
         }
 
-        // render each fact 
+        // render each fact
         let speak = null;
 
         if (this.facts.length > 0) {
@@ -839,6 +839,7 @@ export class ToggleInput extends Input {
 export class Choice {
     title: string;
     value: string;
+    isSelected: boolean;
 }
 
 export class ChoiceSetInput extends Input {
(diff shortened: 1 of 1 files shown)


dclaux:
This is by design. To pre-select a choice, you need to set the value of the input in the JSON payload to one of the Choice input's values. There is no isSelected property on choice.

source/html/renderer/src/card-elements.ts
```diff
@@ -839,6 +839,7 @@ export class ToggleInput extends Input {
 export class Choice {
     title: string;
     value: string;
+    isSelected: boolean;
```
dclaux: This is not the design. To pre-select a choice, set the Choice input's value property to one of it's choices value. This is the value pre-selection/default value is done for all input types.

dsanvita: I found the isSelected property in the Docs: http://adaptivecards.io/explorer/#InputChoiceSet
The docs do not mention ChoiceSet as having a value.

Do the docs need to be updated?

source/html/renderer/src/card-elements.ts
```diff
@@ -470,7 +470,7 @@ export class FactSet extends CardElement {
             return this.speak + '\n';
```
dclaux: Yes, I think there is a mistake in the doc. I'll discuss with the team, thanks for reporting.

dsanvita: Also, I don't think the default value is implemented yet in the HTML renderer for Input.ChoiceSet

dclaux: You are correct. Let me double-check that isSelected is indeed not the design as the implementation will depend on that. I'll get back to you.

dclaux: Still checking with the team, but in the meantime I made the change to honor the default value property. Again thanks for noticing!

PR #351, opened 2017-05-17
Handle Popup action mode and styling for ShowCard Actions.

RebeccaAnne:


source/uwp/Renderer/lib/AdaptiveActionEventArgs.h
```diff
@@ -3,18 +3,20 @@
...
         HRESULT RuntimeClassInitialize();
-        IFACEMETHODIMP get_ActionType(_Out_ ABI::AdaptiveCards::XamlCardRenderer::ActionType* actionType);
-    };
+        HRESULT RuntimeClassInitialize(ABI::AdaptiveCards::XamlCardRenderer::IAdaptiveActionElement* action);
+
+        // IAdaptiveActionEventArgs
+        IFACEMETHODIMP get_Action(ABI::AdaptiveCards::XamlCardRenderer::IAdaptiveActionElement** action);
 
-    ActivatableClass(AdaptiveSubmitActionEventArgs);
+        Microsoft::WRL::ComPtr<ABI::AdaptiveCards::XamlCardRenderer::IAdaptiveActionElement> m_action;
```
riarenas: >Microsoft::WRL::ComPtr<ABI::AdaptiveCards::XamlCardRenderer::IAdaptiveActionElement> m_action [](start = 8, length = 93)

Is this public for a reason?

riarenas:
>     void XamlBuilder::BuildActions(

This function is getting rather long. Do you think we'd benefit by splitting into one different method for each action type, and have this one call into those?

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:561 in 0789e1b. [](commit_id = 0789e1b1bc30ef68c883dbcd5eb50ef16d0967de, deletion_comment = False)

riarenas:
>     void XamlBuilder::BuildActions(

I'm fine if that is out of scope for this change and we just create a bug/issue so we can at least discuss it.

---
In reply to: [302740803](https://github.com/Microsoft/AdaptiveCards/pull/351#issuecomment-302740803) [](ancestors = 302740803)

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:561 in 0789e1b. [](commit_id = 0789e1b1bc30ef68c883dbcd5eb50ef16d0967de, deletion_comment = False)

source/uwp/Renderer/lib/AdaptiveActionEventArgs.h
RebeccaAnne: Nope.  Fixed.

---
In reply to: [117513390](https://github.com/Microsoft/AdaptiveCards/pull/351#discussion_r117513390) [](ancestors = 117513390)

RebeccaAnne:
>     void XamlBuilder::BuildActions(

I agree that this function is too long, but much of it is shared.  The most obvious candidate for refactoring is the piece where we render the show card.  I'll go ahead and pull that into a separate function and see how that looks, and i'll keep an eye on further refactoring as i add the rest of the implementation for other actions.

---
In reply to: [302741440](https://github.com/Microsoft/AdaptiveCards/pull/351#issuecomment-302741440) [](ancestors = 302741440,302740803)

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:561 in 0789e1b. [](commit_id = 0789e1b1bc30ef68c883dbcd5eb50ef16d0967de, deletion_comment = False)

Issue #331, opened 2017-05-15
UWP renderer fails to render image with "person" style unless "size" property provided


anbare:
In the UWP renderer, if you're using an image with "person" style, and you don't manually specify the size of the image, it fails to render. Non-person images render fine without the style provided.

```json
{
	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
	"type": "AdaptiveCard",
	"body": [
		{
			"type": "ColumnSet",
			"columns": [
			    {
			        "type": "Column",
			        "items": [
        			    {
        			        "type": "Image",
        			        "url": "https://unsplash.it/64?image=1027",
        			        "style": "person"
        			    }
		            ]
			    }
		    ]
		}
	]
}
```

Notice how the image just completely gets dropped. Adding in `"size": "medium"` makes it render. Web visualizer displays this correctly without the size provided.

anbare:
Fixed!

PR #355, opened 2017-05-18
Fix UWP person images that don't have size set

anbare:
In the UWP renderer, if you're using an image with "person" style, and
you don't manually specify the size of the image, it was failing to
render, as described in issue #331. Non-person images render fine
without the style provided.

Code was forgetting to set the stretch behavior when image size was set
to Default inside the person logic branch of the code. Just had to add
that line so it included Default size.

Verified behavior in Visualizer. Now a "person" image behaves identically to a normal image when size is not specified.


riarenas:
:shipit:

Issue #357, opened 2017-05-18
UWP larger font sizes have extra padding on left side


anbare:
**Expected:** Regardless of font size, the left edge of text should be consistent. If there's a small text followed by a large text, you should be able to draw a line down from the left edge of the small text and it should be flush with the left edge of the large text.

**Actual:** Larger fonts start to creep to the right, rather than staying flush with the imaginary line on the left.

Visualization of the issue today...

![image](https://cloud.githubusercontent.com/assets/13246069/26225426/fbba15b6-3bdb-11e7-8fc7-6e4eef8b8fc1.png)

Fix is to use a specific property on the XAML TextBlock... I'll look up what it is, we used it for tiles previously

mdtauk:
Have you tried this property, it is supposed to fix those issues.
`<TextBlock OpticalMarginAlignment="TrimSideBearings"/>`

anbare:
Yep that's the one, found it from my previous project a while ago, currently adding it to the code. Thanks mdtauk!

anbare:
Fixed!

PR #358, opened 2017-05-18
Ensure UWP text elements are perfectly left aligned

anbare:
Enabling TrimSideBearings to ensure that regardless of the font size,
the left of the text is consistently in line with each other, fixing
issue #357

Simply had to assign that property (and moved the code to obtain
ITextBlock2 to the top since it's used in two places now)

Verified changes using Visualizer, text is now always flush to the left, pixel perfect!

image


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -990,6 +990,10 @@ namespace AdaptiveCards { namespace XamlCardRenderer
 
         ComPtr<ITextBlock> xamlTextBlock = XamlHelpers::CreateXamlClass<ITextBlock>(HStringReference(RuntimeClass_Windows_UI_Xaml_Controls_TextBlock));
 
+        // ITextBlock2 will be used later on
+        ComPtr<ITextBlock2> xamlTextBlock2;
```
riarenas: [nit] I find it easier to follow when a variable is declared closer to the point of first usage. Thoughts?

PR #359, opened 2017-05-19
Fix DeserializeFromString

riarenas:
UWP renderer also now uses the DeserializeFromString method rather than doing its own thing to render a card from a string.


RebeccaAnne:
:shipit:

Issue #361, opened 2017-05-20
HTML renderer doesn't implement isRequired for inputs


dsanvita:
The Input.* card elements all have an isRequired property, but this is not implemented in the HTML renderer.

matthidinger:
We removed isRequired for the v1.0 schema and are not doing client validation at this point, but we will likely revisit in a future release

hobpet:
@matthidinger, I know this one is closed, but the end of your sentence suggests that you will likely revisit in a future release.
Any update on this?
Thanks!

matthidinger:
Hi @hobpet, indeed we are looking at this for the 1.2 release. It's not commited yet but we are doing initial investigations. Once we get closer we will get a 1.2 announcement post published.

https://github.com/Microsoft/AdaptiveCards/issues/808

PR #362, opened 2017-05-20
Set required value of inputs in html renderer

dsanvita:
https://github.com/Microsoft/AdaptiveCards/issues/361

I wasn't sure what to do for checkboxes (Toggle and multiselect ChoiceSets) since the HTML 5 spec for the required attribute behaves strangely in this case.   But I aded code to set this attribute for the inputs where it made sense.

### DIFF
--- a/source/html/renderer/src/card-elements.ts
+++ b/source/html/renderer/src/card-elements.ts
@@ -470,7 +470,7 @@ export class FactSet extends CardElement {
             return this.speak + '\n';
         }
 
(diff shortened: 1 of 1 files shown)


nitinkallikat:
Adaptive cards are also used for capturing user inputs very much like an HTML form. Without having 
 the required property on input elements, even with server side validations, there is no meaningful way to interact with user since the card once posted and displayed cannot be modified.

PR #365, opened 2017-05-23
Shared Parser now understands Optional and required properties in the schema

riarenas:
Parser was being too permissive with missing required properties. Parser will now throw if a required property is missing. It will keep ignoring extra fields, and allow for missing optional fields.


RebeccaAnne:
:shipit:

PR #369, opened 2017-05-24
Serialize inputs to return as part of action events.

RebeccaAnne:


source/shared/cpp/ObjectModel/BaseInputElement.h
```diff
@@ -0,0 +1,39 @@
...
+};
+
+template <typename T>
+std::shared_ptr<T> BaseInputElement::Deserialize(const Json::Value& json)
+{
+    std::shared_ptr<T> baseInputElement = BaseCardElement::Deserialize<T>(json);
+
+    ParseUtil::ThrowIfNotJsonObject(json);
+
+    baseInputElement->SetId(ParseUtil::GetString(json, AdaptiveCardSchemaKey::Id));
```
riarenas: This is a required field

source/uwp/Renderer/lib/AdaptiveActionEventArgs.cpp
```diff
@@ -25,4 +26,8 @@ namespace AdaptiveCards { namespace XamlCardRenderer
         *action = localAction.Detach();
         return S_OK;
     }
+    IFACEMETHODIMP AdaptiveActionEventArgs::get_Inputs(HSTRING * inputs)
```
riarenas: [nit] _Use_decl_annotations_ and extra line

source/uwp/Renderer/lib/AdaptiveActionEventArgs.h
```diff
@@ -12,12 +12,14 @@ namespace AdaptiveCards { namespace XamlCardRenderer
 
     public:
         HRESULT RuntimeClassInitialize();
-        HRESULT RuntimeClassInitialize(ABI::AdaptiveCards::XamlCardRenderer::IAdaptiveActionElement* action);
+        HRESULT RuntimeClassInitialize(ABI::AdaptiveCards::XamlCardRenderer::IAdaptiveActionElement* action, HSTRING inputs);
 
         // IAdaptiveActionEventArgs
         IFACEMETHODIMP get_Action(ABI::AdaptiveCards::XamlCardRenderer::IAdaptiveActionElement** action);
```
riarenas: Can you add SAL to this one?

source/uwp/Renderer/lib/AdaptiveInputChoiceSet.cpp
```diff
@@ -77,9 +77,24 @@ namespace AdaptiveCards { namespace XamlCardRenderer
...
     _Use_decl_annotations_
-    HRESULT AdaptiveInputChoiceSet::get_Choices(IVector<IAdaptiveInputChoice*>** InputChoices)
+    HRESULT AdaptiveInputChoiceSet::get_Choices(IVector<IAdaptiveInputChoice*>** inputChoices)
     {
-        return m_InputChoices.CopyTo(InputChoices);
+        return m_InputChoices.CopyTo(inputChoices);
+    }
+
+    _Use_decl_annotations_
+    HRESULT AdaptiveInputChoiceSet::get_Id(HSTRING* speak)
```
riarenas: >speak [](start = 52, length = 5)

id here and in the putter

source/uwp/Renderer/lib/AdaptiveInputChoiceSet.h
```diff
@@ -34,6 +35,10 @@ namespace AdaptiveCards { namespace XamlCardRenderer
 
         IFACEMETHODIMP get_Choices(_COM_Outptr_ ABI::Windows::Foundation::Collections::IVector<ABI::AdaptiveCards::XamlCardRenderer::IAdaptiveInputChoice*>** columns);
 
+        // IAdaptiveInputElement
+        IFACEMETHODIMP get_Id(_Out_ HSTRING* text);
```
riarenas: >text [](start = 45, length = 4)

id

source/uwp/Renderer/lib/AdaptiveInputChoiceSet.cpp
riarenas: And in the other input classes

---
In reply to: [118370064](https://github.com/Microsoft/AdaptiveCards/pull/369#discussion_r118370064) [](ancestors = 118370064)

source/uwp/Renderer/lib/AdaptiveInputChoiceSet.h
riarenas: Same for the other input classes

---
In reply to: [118370122](https://github.com/Microsoft/AdaptiveCards/pull/369#discussion_r118370122) [](ancestors = 118370122)

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -15,6 +15,9 @@
 #include "XamlBuilder.h"
 #include "XamlHelpers.h"
 #include "XamlStyleKeyGenerators.h"
+#include "json/json.h"
+
+#include "windows.globalization.datetimeformatting.h"
```
riarenas: >"windows.globalization.datetimeformatting.h" [](start = 9, length = 44)

Is there a reason why this is under quotes and not <> ?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -129,8 +133,9 @@ namespace AdaptiveCards { namespace XamlCardRenderer
 
         // Enumerate the child items of the card and build xaml for them
         ComPtr<IVector<IAdaptiveCardElement*>> body;
+        std::shared_ptr<std::vector<InputItem>> inputElements(new std::vector<InputItem>());
```
riarenas: >std::shared_ptr<std::vector<InputItem>> inputElements(new std::vector<InputItem>()); [](start = 8, length = 84)

Can you use std::make_shared instead?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -616,10 +622,272 @@ namespace AdaptiveCards { namespace XamlCardRenderer
...
+        THROW_IF_FAILED(textBox->get_Text(text.GetAddressOf()));
+
+        std::string textString;
+        if (text.IsValid())
+        {
+            THROW_IF_FAILED(HStringToUTF8(text.Get(), textString));
+        }
+
+        jsonValue[idString] = textString.c_str();
+    }
```
riarenas: All the serializeXInput methods feel like they should be part of the input classes. What do you think?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -616,10 +622,272 @@ namespace AdaptiveCards { namespace XamlCardRenderer
...
+            }
+        }
+    }
+
+    void SerializeInputs(
+        std::shared_ptr<std::vector<InputItem>> inputElements,
+        std::string* inputString)
+    {
+        Json::Value jsonValue;
+        for (std::vector<InputItem>::iterator inputItem = inputElements->begin(); inputItem != inputElements->end(); ++inputItem)
```
riarenas: >std::vector<InputItem>::iterator inputItem = inputElements->begin(); inputItem != inputElements->end(); ++inputItem [](start = 13, length = 115)

[nit] I usually prefer range-based for loops over iterators in cases where I go over an entire collection

source/uwp/Renderer/idl/AdaptiveCards.XamlCardRenderer.idl
```diff
@@ -777,6 +793,7 @@ namespace AdaptiveCards
         interface IAdaptiveActionEventArgs : IInspectable
         {
             [propget] HRESULT Action([out, retval] IAdaptiveActionElement** action);
+            [propget] HRESULT Inputs([out, retval] HSTRING* value);
```
anbare: @matthidinger We're returning a string here for the input values? Why not a KeyValuePair of string/string?

source/uwp/Renderer/lib/AdaptiveActionEventArgs.cpp
riarenas: >HSTRING * [](start = 55, length = 9)

[nit] spacing

source/uwp/Renderer/lib/XamlBuilder.cpp
RebeccaAnne: We spoke offline.  These don't belong in the public classes, because they're mostly pulling from the UI element rather than the adaptive card element.  Instead, expanded the InputItem struct into a class with a Serialize() method and moved all of this logic in there.

---
In reply to: [118372526](https://github.com/Microsoft/AdaptiveCards/pull/369#discussion_r118372526) [](ancestors = 118372526)

source/uwp/Renderer/lib/InputItem.h
```diff
@@ -0,0 +1,32 @@
...
+            ABI::AdaptiveCards::XamlCardRenderer::IAdaptiveInputElement* adaptiveInputElement,
+            ABI::Windows::UI::Xaml::IUIElement* uiInputElement) :
+            m_adaptiveInputElement(adaptiveInputElement),
+            m_uiInputElement(uiInputElement)
+        {}
+
+        void Serialize(Json::Value& jsonValue);
+
+    private:
+        void SerializeChoiceSetInput(Json::Value& jsonValue, const char * idString);
```
riarenas: >char * [](start = 67, length = 6)

[nit] spacing

source/uwp/Renderer/idl/AdaptiveCards.XamlCardRenderer.idl
RebeccaAnne: I believe the reasoning behind returning json for the submit inputs is in order to make it easier to send the results back to a theoretical server (in say, the bot case).  The other renderers return json as well.  

@matthidinger can speak to this further and comment on whether we might additionally want to add something more convenient for local callers.

---
In reply to: [118375602](https://github.com/Microsoft/AdaptiveCards/pull/369#discussion_r118375602) [](ancestors = 118375602)

anbare: Hmm, for UWP that makes it much more difficult for apps to process the inputs. Action.Submit would be used for foreground and background activations, where apps need to pull out what the user typed from a text box and send it as a message, for example. Putting it in JSON instead of a KeyValuePair makes it much more difficult to process.

If you want to check this in now you can, but I think we'll need to change it in the future. Hopefully @matthidinger can chime in.

PR #376, opened 2017-05-25
Add Enum map generation functions

riarenas:
* Refactored the enums to generate the String -> enum map based on the Enum -> String map.
* enum -> string maps and AdaptiveCardSchemaKey enum (not used by any of the renderers, only the parser) are now sorted alphabetically for better readability.
* Removed superfluous Schema key entries in the enum that mapped to repeated strings.
* Adds the schema keys required for HostConfig parsing.


RebeccaAnne:
:shipit:

PR #388, opened 2017-06-02
UWP: Added DataJson property to Action.Submit

### LINKED ISSUE #387: UWP "data" property on Action.Submit isn't accessible (data values not returned on action click)
image

There's no Data property on the AdaptiveSubmitAction. The event you get back for a button being clicked is called "Inputs", which would imply that it would NOT contain the Data key/value pairs (and today it does NOT).

Hence exposing the Data property on AdaptiveSubmitAction would allow the host to merge Inputs and Data into a single ValueSet. Plus having the property accessible could be useful for other reasons.

anbare:
Added the DataJson property to the UWP object model. DataJson returns a string of the JSON, consistent with the .NET version of the library that has a DataJson property.

Tested with visualizer app and Notifications Visualizer, confirmed that the "data" property is now being passed through and accessible by renderer hosts.

Fixes #387

### DIFF
--- a/source/uwp/Renderer/lib/AdaptiveSubmitAction.cpp
+++ b/source/uwp/Renderer/lib/AdaptiveSubmitAction.cpp
@@ -56,4 +56,19 @@ namespace AdaptiveCards { namespace XamlCardRenderer
         m_sharedSubmitAction->SetSpeak(out);
         return S_OK;
     }
+
+    _Use_decl_annotations_
+        HRESULT AdaptiveSubmitAction::get_DataJson(HSTRING* data)
+    {
+        return UTF8ToHString(m_sharedSubmitAction->GetDataJson(), data);
+    }
+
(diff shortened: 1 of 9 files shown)


source/uwp/Renderer/lib/AdaptiveSubmitAction.cpp
```diff
@@ -56,4 +56,19 @@ namespace AdaptiveCards { namespace XamlCardRenderer
         m_sharedSubmitAction->SetSpeak(out);
         return S_OK;
     }
+
+    _Use_decl_annotations_
+        HRESULT AdaptiveSubmitAction::get_DataJson(HSTRING* data)
```
riarenas: >      [](start = 3, length = 5)

[nit] spacing

RebeccaAnne:
:shipit:

Issue #399, opened 2017-06-08
UWP column "auto" width doesn't work correctly


anbare:
When assigning size = "auto" on a column in the UWP renderer, it still behaves as a * width, as seen with the example below...

UWP renderer (incorrect)
![image](https://user-images.githubusercontent.com/13246069/26939971-e0a5c3ee-4c2d-11e7-9dd7-e9aba10c1ae0.png)

HTML renderer (correct)
![image](https://user-images.githubusercontent.com/13246069/26939989-f0e13086-4c2d-11e7-996f-c0589c6aafc3.png)

```json
{
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "type": "AdaptiveCard",
    "version": "0.5",
    "body": [
        {
            "type": "ColumnSet",
            "columns": [
                {
                    "type": "Column",
                    "size": "auto",
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "62.30",
                            "size": "extraLarge"
                        }
                    ]
                },
                {
                    "type": "Column",
                    "size": "stretch",
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "This column should stretch, whereas the left column should be auto width"
                        }
                    ]
                }
            ]
        }
    ]
}
```

I'll investigate and take a look

riarenas:
Thanks for the fix Andrew!

PR #401, opened 2017-06-08
Fixed UWP column auto/stretch behavior

anbare:
The boolean logic for auto/stretch was inverted. When dev specified stretch, auto was being assigned, and vice versa.

Fixed the boolean logic (also added check to invalidate negative star size entries) and verified changes using UWP Visualizer and Notifications Visualizer with all of the sample payloads

Before
image

After
image


anbare:
Fixes #399

riarenas:
:shipit:

PR #410, opened 2017-06-09
Serialize shared model

riarenas:


riarenas:
Added ToJsonString method to UWP projection AdaptiveCard class.

RebeccaAnne:
:shipit:

PR #417, opened 2017-06-13
Add IsRequired property to inputs

### LINKED ISSUE #416: C++ Shared model and UWP renderer do not implement IsRequired property for inputs
Currently, only CHoiceSetInput elements include the IsRequire Property. This is shared by every input element and should be moved to the BaseInputElement classes.

riarenas:
fixes #416


RebeccaAnne:
:shipit:

PR #423, opened 2017-06-14
Adaptive cards objc

nesalang:
This changes include image/text block implementation in obj-c, shared object model integration, partial implementations of hostconfig, single view Visualizer app, changes in path-to-header to unix style and addition of dedicated constructors in hostconfig.h to resolve compilation issues.


source/shared/cpp/ObjectModel/HostConfig.h
```diff
@@ -12,6 +12,7 @@ struct SpacingDefinition
     unsigned int right = 0;
     unsigned int top = 0;
     unsigned int bottom = 0;
+    SpacingDefinition(int left = 0, int right = 0, int top = 0, int bottom = 0):left(left), right(right), top(top), bottom(bottom){};
```
riarenas: >SpacingDefinition(int left = 0, int right = 0, int top = 0, int bottom = 0):left(left), right(right), top(top), bottom(bottom){}; [](start = 4, length = 129)

What makes these necessary?

riarenas:
Should the .DS_Store files be in the Repo? It's not recommended checking in executables

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
```diff
@@ -0,0 +1,31 @@
...
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+#import "TextBlock.h"
+#import "Image.h"
+#import "HostConfig.h"
+
+using namespace AdaptiveCards;
+
+@interface ADCIOSAdaptiveHostConfig:NSObject
```
riarenas: >ADCIOS [](start = 11, length = 6)

Why ADCIOS prefix?

the IOS part seems redundant.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.mm
```diff
@@ -0,0 +1,145 @@
...
+
+        default: {
+            NSLog(@"unimplemented");
+        }
+    }
+    CGSize cgSize = CGSizeMake(sz, sz);
+    return cgSize;
+}
+
+- (int) getTextBlockTexSize:(std::shared_ptr<TextBlock>)txtBlock {
```
riarenas: >txtBlock [](start = 56, length = 8)

Just a heads up. Facts also need to be styled using these same sizes, so make sure it's generic enough to not have to repeat this for Facts as well.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSViewController.h
```diff
@@ -0,0 +1,17 @@
...
+//  Created by jwoo on 5/26/17.
+//  Copyright © 2017 Microsoft. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+@interface ADCIOSViewController : UIViewController
+
+@property NSString* jsaonString;
```
riarenas: >jsaonString [](start = 20, length = 11)

[nit] typo

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
```diff
@@ -0,0 +1,31 @@
...
+
+- (CGSize) getImageSize:(std::shared_ptr<Image>) imgBlock;
```
riarenas: >imgBlock [](start = 49, length = 8)

[nit] It's just Image, no block.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
```diff
@@ -0,0 +1,31 @@
...
+
+- (UIColor* ) getTextBlockColor:(std::shared_ptr<AdaptiveCards::TextBlock>) txtBlock;
+
+- (int) getTextBlockTexSize:(std::shared_ptr<TextBlock>)txtBlock;
```
riarenas: >getTextBlockTexSize [](start = 8, length = 19)

A lot of the HostConfig properties apply to more than one element. Maybe this should take an OBJC textview instead of the object model textBlock? That way you can reuse this for fact, and any other element that needs to have its text style changed

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
```diff
@@ -0,0 +1,31 @@
...
+
+- (NSTextAlignment) getTextBlockAlignment:(std::shared_ptr<TextBlock>) txtBlock;
+
+- (NSNumber* ) getTextBlockTextWeight:(std::shared_ptr<TextBlock>) txtBlock;
```
riarenas: Both of these are part of the TextBlock element itself, they do not come from hostConfig. Maybe this is the wrong place for these?

source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ViewController.m
```diff
@@ -0,0 +1,173 @@
...
+    [self.JSLab setFont:[UIFont systemFontOfSize:8]];
+    self.JSLab.lineBreakMode = NSLineBreakByTruncatingTail;
+    self.JSLab.text = self.inputs[self.idx];
+   
+    ADCIOSPresentationViewController* adcVc = [[ADCIOSPresentationViewController alloc] init: self.inputs[self.idx]];
+    
+    //view.translatesAutoresizingMaskIntoConstraints = false;
+   
+    
+
```
riarenas: Remove if not needed, and fix extra lines

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSViewController.mm
```diff
@@ -0,0 +1,127 @@
...
+#import "ADCIOSAdaptiveHostConfig.h"
+#import "AdaptiveCard.h"
+#import "TextBlock.h"
+#import "Image.h"
+#import <AVFoundation/AVFoundation.h>
+
+using namespace AdaptiveCards;
+
+@implementation ADCIOSViewController
+std::shared_ptr<AdaptiveCard> adc;
```
riarenas: >adc [](start = 30, length = 3)

Is using acronym variable names the coding convention for objc? Feels like it becomes hard to read without proper names.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
riarenas: >NSNumber [](start = 3, length = 8)

Why number? Feels like there should be a corresponding enum in the objc side.

source/shared/cpp/ObjectModel/HostConfig.h
nesalang: We use the SpacingDefintions and structures defined here as members of other structures, and use initializes in form of sturctname(val, val, ...). 
This causes compilation errors on iOS side. To fix it, I added dedicated constructors and initialize lists with default values.There are other ways to fix it, but it would require editing instead of adding.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
nesalang: Good point, apple requires at least 4 letters prefix to prevent name collisions with each others.
I couldn't come up with additional one letter.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSViewController.h
nesalang: noted

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
nesalang: noted

source/shared/cpp/ObjectModel/HostConfig.h
riarenas: It turns out this is necessary because the brace initialization that we use is a c++14+ feature, and the project was configured to compile with c++11. We can either add explicit constructors to every struct in the hostConfig, or change the project to use c++14 (like windows and android already do). Any thoughts on which is the better approach?

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
nesalang: Yes, i haven't got into other parts, yet. As I see other types, will make them re-usable.

---
In reply to: [122247625](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122247625) [](ancestors = 122247625)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
nesalang: noted

---
In reply to: [122248026](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122248026) [](ancestors = 122248026)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
riarenas: Just ADC should be fine. Apple recommends using 3-letter prefixes according to https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/Conventions/Conventions.html.

source/shared/cpp/ObjectModel/HostConfig.h
```diff
@@ -170,6 +178,7 @@ struct ChoiceSetConfig
 
 struct HostConfig
 {
+//#if 0
```
riarenas: Remove

source/shared/cpp/ObjectModel/HostConfig.h
nesalang: either ways is good, just concerned that if there would be any side effects of adding compiler flag. let's add Ben Vi from islandwood if there is any side effects of adding c++14 flag.

---
In reply to: [122282237](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122282237) [](ancestors = 122282237)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
riarenas: Renderer - > ACR
Object model - > ACM
Visualizer - > ACV

source/shared/cpp/ObjectModel/HostConfig.h
nesalang: noted

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ADCIOSAdaptiveHostConfig.h
nesalang: iOS API takes NSNumber*

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACFramework.h
```diff
@@ -0,0 +1,19 @@
+//
+//  ADCIOSFramework.h
```
riarenas: >ADCIOSFramework [](start = 4, length = 15)

Update all of these

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRViewController.mm
```diff
@@ -0,0 +1,127 @@
...
+#import "ACRHostConfig.h"
+#import "AdaptiveCard.h"
+#import "TextBlock.h"
+#import "Image.h"
+#import <AVFoundation/AVFoundation.h>
+
+using namespace AdaptiveCards;
+
+@implementation ACRViewController
+std::shared_ptr<AdaptiveCard> adaptiveCards;
```
riarenas: >adaptiveCards [](start = 30, length = 13)

[nit]this should be singular.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRViewController.mm
```diff
@@ -0,0 +1,127 @@
...
+ACRHostConfig* m_hostConfig;
```
riarenas: >m_hostConfig [](start = 15, length = 12)

[nit] remove prefix

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRViewController.mm
```diff
@@ -0,0 +1,127 @@
...
+                std::shared_ptr<Image> iblock = std::dynamic_pointer_cast<Image>(elem);
+                [mainView addArrangedSubview:[self buildImageBlock:iblock]];
+                break;
+            }
+            default:;
+        }
+    }
+}
+
+-(UIImageView* ) buildImageBlock:(std::shared_ptr<Image>) blck{
```
riarenas: >buildImageBlock [](start = 17, length = 15)

[nit] it's just Image, not ImageBlock

source/shared/cpp/ObjectModel/HostConfig.h
bviglietta: No bad side effects from moving to C++14, assuming this is a private header.

---
In reply to: [122286906](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122286906) [](ancestors = 122286906,122282237)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRPresentationViewController.h
```diff
@@ -0,0 +1,18 @@
...
+//
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+@interface ACRPresentationViewController : UIViewController
+
+@property (readonly) NSString* jsonStr;
+
+- (id) init:(NSString*) str;
```
bviglietta: >id [](start = 3, length = 2)

instancetype

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRPresentationViewController.m
```diff
@@ -0,0 +1,44 @@
...
+    
+        [NSLayoutConstraint activateConstraints:@[[acVc.view.centerXAnchor constraintEqualToAnchor:self.view.centerXAnchor], [acVc.view.centerYAnchor constraintEqualToAnchor:self.view.centerYAnchor]]];
+    }
+    // Do any additional setup after loading the view, typically from a nib.
+}
+
+- (void)didReceiveMemoryWarning {
+    [super didReceiveMemoryWarning];
+    // Dispose of any resources that can be recreated.
+}
```
bviglietta: This method isn't doing anything for you.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRPresentationViewController.m
```diff
@@ -0,0 +1,44 @@
...
+    
+    return self;
+}
+
+- (void) viewDidLoad {
+
+    [super viewDidLoad];
+    
+    if(self.jsonStr){
+        ACRViewController* acVc = [[ACRViewController alloc] init: self.jsonStr];
```
bviglietta: >acVc [](start = 27, length = 4)

You should call `[self addChildViewController:acVc]` here, and then after adding the subview call `[acVc didMoveToParentViewController:self]`. This will establish a parent-child relationship between the two view controllers, which ensures that various system notifications will be forwarded from the parent to the child.

bviglietta: Although I'm actually curious, what's the purpose of this view controller? I don't see what functionality it adds to the underlying ACRViewController..

---
In reply to: [122548933](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122548933) [](ancestors = 122548933)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRViewController.mm
bviglietta: These are global variables, not member variables. If you want them to be member variables, you have to surround the declarations with curly braces.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACFramework.h
```diff
@@ -0,0 +1,19 @@
...
+//  ADCIOSFramework
+//
+//  Created by jwoo on 6/2/17.
```
bviglietta: >Created by jwoo on 6/2/17. [](start = 4, length = 26)

We don't normally put file history in the file. Source control keeps track of that.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.h
```diff
@@ -0,0 +1,31 @@
...
+#import <UIKit/UIKit.h>
+#import "TextBlock.h"
+#import "Image.h"
+#import "HostConfig.h"
+
+using namespace AdaptiveCards;
+
+@interface ACRHostConfig:NSObject
+
+- (CGSize) getImageSize:(std::shared_ptr<Image>) img;
```
bviglietta: >std::shared_ptr<Image> [](start = 25, length = 22)

It would be better to pass these by const reference, to avoid making unnecessary copies.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.mm
```diff
@@ -0,0 +1,144 @@
...
+            break;
+        }
+        default: {
+            str = self.m_hostConfig->colors.good.normal;
+            break;
+        }
+    }
+    
+    num = [self retrieveIntFromHostConfigString:[[NSString alloc] initWithCString:str.c_str()
+                                encoding:[NSString defaultCStringEncoding]]];
```
bviglietta: It doesn't seem necessary to convert std::string -> NSString -> int. You can convert without going through NSString:

`std::stoul(str.substr(1), nullptr, 16)`

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.mm
```diff
@@ -0,0 +1,144 @@
...
+
+    return [UIColor colorWithRed:((num & 0x00FF0000) >> 16) / 255
```
bviglietta: >/ 255 [](start = 60, length = 5)

You need to do floating-point division, not integer division.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.h
```diff
@@ -0,0 +1,31 @@
...
+
+- (UIColor* ) getTextBlockColor:(std::shared_ptr<AdaptiveCards::TextBlock>) txtBlock;
+
+- (int) getTextBlockTexSize:(std::shared_ptr<TextBlock>)txtBlock;
```
bviglietta: >Tex [](start = 20, length = 3)

Text?

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.mm
```diff
@@ -0,0 +1,144 @@
...
+            return NSTextAlignmentRight;
+        default:
+            return NSTextAlignmentLeft;
+    }
+}
+     
+-(NSNumber* ) getTextBlockTextWeight:(std::shared_ptr<TextBlock>) txtBlock{
+    switch (txtBlock->GetTextWeight()) { 
+        case TextWeight::Normal:
+            return [[NSNumber alloc] initWithInt:0];
```
bviglietta: >[[NSNumber alloc] initWithInt:0] [](start = 19, length = 32)

`@0`

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRPresentationViewController.m
nesalang: It's a Objective-C wrapper that wraps ACRViewController that wraps C++ codes.

---
In reply to: [122549111](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122549111) [](ancestors = 122549111,122548933)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRViewController.h
```diff
@@ -0,0 +1,17 @@
...
+//  Copyright © 2017 Microsoft. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+@interface ACRViewController : UIViewController
+
+@property NSString* jsonString;
+-(id) init: (NSString*) str;
```
bviglietta: >id [](start = 2, length = 2)

instancetype

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.h
nesalang: is it still true for shared pointer? I thought I was only copying pointers.

---
In reply to: [122550418](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122550418) [](ancestors = 122550418)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.mm
nesalang: great point!

---
In reply to: [122551184](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122551184) [](ancestors = 122551184)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.mm
nesalang: thanks

---
In reply to: [122551451](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122551451) [](ancestors = 122551451)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.h
nesalang: Thanks :-)

---
In reply to: [122551624](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122551624) [](ancestors = 122551624)

source/shared/cpp/ObjectModel/HostConfig.h
nesalang: awesome, thanks

---
In reply to: [122545652](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122545652) [](ancestors = 122545652,122286906,122282237)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRViewController.mm
nesalang: awesome, thanks

---
In reply to: [122549457](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122549457) [](ancestors = 122549457)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.h
bviglietta: It's true that it won't make any copies of the Image resource. It will just make some copies of shared_ptr objects, which will entail some extra reference counting operations. It's not very expensive, but it's unnecessary.

---
In reply to: [122552703](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122552703) [](ancestors = 122552703,122550418)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRPresentationViewController.m
bviglietta: But ACRViewController is already an Objective-C object. Why does it need a wrapper?

---
In reply to: [122552582](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122552582) [](ancestors = 122552582,122549111,122548933)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.h
nesalang: you're right; thanks for insight


---
In reply to: [122553186](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122553186) [](ancestors = 122553186,122552703,122550418)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.mm
nesalang: thanks

---
In reply to: [122551696](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122551696) [](ancestors = 122551696)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRViewController.h
nesalang: thanks

---
In reply to: [122552608](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122552608) [](ancestors = 122552608)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRPresentationViewController.h
nesalang: thanks

---
In reply to: [122546099](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122546099) [](ancestors = 122546099)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRPresentationViewController.m
```diff
@@ -0,0 +1,44 @@
...
+//  Copyright © 2017 Microsoft. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "ACRPresentationViewController.h"
+#import "ACRViewController.h"
+
+@implementation ACRPresentationViewController
+
+- (id) init:(NSString*) str{
```
bviglietta: >init:(NSString*) str [](start = 7, length = 20)

Can this view controller be created via storyboard, or only programmatically? If it's allowed in a storyboard (the common way of creating view controllers), this initializer will never be called.

More generally, the designated initializer for UIViewController is initWithNibName:bundle:, so you should be sure to override that method and make sure that all initializers in this class call that method on the superclass.

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRPresentationViewController.m
nesalang: from my understanding without this wrapper, user needs to have at least one file that has .mm extension  or needs a compilation flag that link c++ std lib, but now you raised this point, i think you might be correct because now i'm using framework, adding the compilation flag may not necessary. is the compilation flag not needed?

---
In reply to: [122553246](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122553246) [](ancestors = 122553246,122552582,122549111,122548933)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACFramework.h
nesalang: thanks

---
In reply to: [122549969](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122549969) [](ancestors = 122549969)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRPresentationViewController.m
nesalang: thanks

---
In reply to: [122553987](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122553987) [](ancestors = 122553987)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRPresentationViewController.m
bviglietta: The client code will only ever see your header files. They won't see your implementation files, because those will already be compiled into a library by the time the client gets them. That means that as long as there's no C++ in your _header_ (and there isn't any in ACRViewController.h), then the client is free to compile their own code as Objective-C using whatever compiler flags they want.

---
In reply to: [122553995](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122553995) [](ancestors = 122553995,122553246,122552582,122549111,122548933)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.mm
nesalang: Ben, thanks so much for your amazing contribution!

---
In reply to: [122553457](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122553457) [](ancestors = 122553457,122551696)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRPresentationViewController.m
nesalang: cool, thanks. I will make amends

---
In reply to: [122554847](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122554847) [](ancestors = 122554847,122553995,122553246,122552582,122549111,122548933)

bviglietta:
Agreed, .DS_Store files shouldn't be checked in.

---
In reply to: [308791517](https://github.com/Microsoft/AdaptiveCards/pull/423#issuecomment-308791517) [](ancestors = 308791517)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.mm
```diff
@@ -0,0 +1,142 @@
...
+
+    if(self)
+    { 
+        _m_hostConfig = std::make_shared<HostConfig>();
+    }
+
+    return self;
+}
+
+- (u_int32_t) retrieveIntFromHostConfigString: (NSString *) str{
```
bviglietta: >retrieveIntFromHostConfigString [](start = 14, length = 31)

This method isn't needed anymore.

bviglietta:
:shipit:

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework/ACRHostConfig.mm
nesalang: Thanks

---
In reply to: [122776553](https://github.com/Microsoft/AdaptiveCards/pull/423#discussion_r122776553) [](ancestors = 122776553)

source/ios/AdaptiveCards/ADCIOSFramework/ADCIOSFramework.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,685 @@
+// !$*UTF8*$!
```
chuchoezequiel67-collab, line 1: f915e02a017cf2dba85e4219adbcd1e851a102bc

chuchoezequiel67-collab (approved):

Issue #418, opened 2017-06-13
Add SelectAction support to Shared model and UWP renderer


riarenas:
Implement the SelectAction property for Container and Image elements.

**UWP implementation status**
- [x] Container (PR #747)
- [x] ColumnSet (PR #747)
- [ ] Column
- [ ] Image

anbare:
@matthidinger this is a key feature for toasts, should we make this milestone 1.1?

I could probably help implement the UWP code for this since it's already in the shared model if everyone else is busy

matthidinger:
Duplicate

PR #441, opened 2017-06-26
Add SelectAction properties to c++ shared model

riarenas:
References #418
Add SelectAction to shared model
Update Android Auto generated files


riarenas:
Sending this now to unblock work on Android renderer. UWP renderer handling of select Actions coming in part 2.

RebeccaAnne:
:shipit:

source/shared/cpp/ObjectModel/Column.cpp
```diff
@@ -112,4 +112,14 @@ std::shared_ptr<Column> Column::Deserialize(const Json::Value& value)
 std::shared_ptr<Column> Column::DeserializeFromString(const std::string& jsonString)
 {
     return Column::Deserialize(ParseUtil::GetJsonValueFromString(jsonString));
+}
+
+std::shared_ptr<BaseActionElement> Column::GetSelectAction() const
```
bekao: General comment:  Where is the serialization and deserialization logic for all three classes?

bekao:
:clock1:

source/shared/cpp/ObjectModel/Column.cpp
riarenas: I was going to add that in a separate PR, but I guess there really is no reason to split it out that much. Adding the fields to serialization logic now.

bekao: Ok.  Thought you forgot since it wasn't mentioned in the description. :)

---
In reply to: [124076652](https://github.com/Microsoft/AdaptiveCards/pull/441#discussion_r124076652) [](ancestors = 124076652)

source/shared/cpp/ObjectModel/Container.cpp
```diff
@@ -108,6 +124,13 @@ std::shared_ptr<Container> Container::Deserialize(const Json::Value& value)
     // Parse Items
     auto cardElements = ParseUtil::GetElementCollection<BaseCardElement>(value, AdaptiveCardSchemaKey::Items, Container::CardElementParsers, true);
     container->m_items = std::move(cardElements);
+
+    Json::Value selectActionValue = ParseUtil::ExtractJsonValue(value, AdaptiveCardSchemaKey::SelectAction, false);
+    if (!selectActionValue.empty())
+    {
+        container->SetSelectAction(ParseUtil::GetActionFromJsonValue<BaseActionElement>(selectActionValue, BaseCardElement::ActionParsers));
+    }
```
bekao: Consider refactoring this code.  Possibly create helper methods in BaseCardElement to parse Card Element and Action.  This would allow you to keep ActionParsers private and remove all the duplicated Card Element Parsers.

riarenas: I can do that for the ActionParsers since there is no plan for specific elements to support different action types, but I believe the CardElementParsers should remain in place, as not all of them are the same.

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
```diff
@@ -91,7 +91,7 @@ std::shared_ptr<AdaptiveCard> AdaptiveCard::Deserialize(const Json::Value& json)
     auto body = ParseUtil::GetElementCollection<BaseCardElement>(json, AdaptiveCardSchemaKey::Body, AdaptiveCard::CardElementParsers, true);
 
     // Parse actions if present
-    auto actions = ParseUtil::GetActionCollection<BaseActionElement>(json, AdaptiveCardSchemaKey::Actions, AdaptiveCard::CardActionParsers);
+    auto actions = ParseUtil::GetActionCollection<BaseActionElement>(json, AdaptiveCardSchemaKey::Actions, AdaptiveCard::ActionParsers);
```
bekao: > auto actions = ParseUtil::GetActionCollection<BaseActionElement>(json, AdaptiveCardSchemaKey::Actions, AdaptiveCard::ActionParsers); [](start = 3, length = 133)

we will need to look at how we can refactor this code in the future.  Can you add an issue in github?  I am not going to block the check-in for this.  Granted, this parses multiple actions instead of one action (selectAction).

riarenas: Created https://github.com/Microsoft/AdaptiveCards/issues/442

PR #448, opened 2017-06-29
Adaptive cards objc

nesalang:
Improved Visualizer:
- user can now select json directly

New Features Support Added:
- container / column / column set

Refactoring:


bviglietta:
This pull request shows the entire ADC framework and visualizer being added. Can you fix it so that it shows only your changes?

Issue #462, opened 2017-07-10
New spacing model


dclaux:
# Implementation status
- [x] TypeScript (PR https://github.com/Microsoft/AdaptiveCards/pull/463)
- [x] UWP
- [x] .NET (PR #718)
- [x] iOS
- [x] Android
- [x] Documentation

# Design proposal
In order to accommodate for more scenarios, we propose to make it possible to select from a wider ranger of spacing between elements, so as to achieve more subtle or more marked separations. At the same time, we propose to separate the concept of "separator" from spacing itself (currently, to get a separator line, one needs to use the "strong" separation.

So we propose the following:
- Remove the "separation" property everywhere
- Remove separation definitions from HostConfig
- Remove strongSeparation from HostConfig
- Introduce a "spacing" enum property available on any card element
- Remove HostConfig.adaptiveCard.padding as it is replaced with one of the named spacing
- Introduce a "separator" property, available on any card element
- Introduce spacing definitions in HostConfig
- Introduce separator thickness and color definitions in HostConfig

### Removed
- All separation properties
- Separation enumeration
### New properties on every element 
|Property|Type|Description
----------|------|-------------
|spacing|Spacing|The amount of space the element should be separated from the previous element. Default value is **"default"**|
|separator|bool|Indicates whether there should be a visible separator (e.g. a line) between the element and the one before it. |

### Spacing enumeration
Spacing = "none" | "small" | "default" | "medium" | "large" | "extraLarge" | "padding"

The **padding** value means "same as card padding". This is to help card authors design cards where, say, two columns are separated by the same spacing as there is padding around the card.

### Separator type
Separator = true | false

# V1.0 parser requirements

For v1.0 I've updated the spec that separator is back to being a bool instead of an object. We will add the object back in a future release.

Parsers should parse the bool and populate the Separator object with an IsVisible property.

"separator": true becomes

"separator": {
   "isVisible": true,
   "thickness": "default",
   "color": "default
}

"separator": false becomes

"separator": {
   "isVisible": false,
   "thickness": "default",
   "color": "default
}

### Removed
- All separation definitions 
- HostConfig.strongSeparation property

Currently some HostConfig elements only contain a separation definition. This change removes these separation definitions and therefore these elements would be empty. They will therefore be removed as a result. The elements in question are:
- All input configurations
- columnSet configuration
- column configuration

### Added
Spacing definitions at the root level:
```
spacing: {
    small: number,
    default: number,
    medium: number,
    large: number,
[...]

PR #463, opened 2017-07-10
New spacing model

dclaux:
See issue https://github.com/Microsoft/AdaptiveCards/issues/462


dclaux:
FYI this is a **breaking change.** Separation in cards will stop working.

danmarshall:
Will this breaking change happen across all platforms? I'm hoping that every platform will have the same featureset for the same number of major version. Also, at a breaking change, we should branch the existing code so that we can continue to fix bugs in the previous major version.

danmarshall:
Perhaps this should not go to the master branch.

dclaux:
Where else? It's already in its own branch (finalize-html)

dclaux:
Yes, eventually all renderers will implement this change. We're starting with the HTML renderer as the reference implementation.
BTW are you suggesting I open a new master_v0.<some number> branch for this? We haven't reached v1 yet, so the assumption should be that there could still be breaking changes at this stage. So I was definitely simply going to push changes to master. If the WebChat control is using the NPM package (v0.6 being the latest), it won't break until we publish a new package and you decide to switch to using it. Do you think that's not good enough?

danmarshall:
Yes, I suppose that master_v0 will suffice. And I don't anticipate there will be many bug fixes in it, but it is good to have in that case. Also true, the WebChat will depend on the NPM v0.6 and will not break.

The bigger issue is supporting our developers, many of whom will want to use the Visualizer for v0 and some will want to move ahead with v1. And they will need to know which of these versions is used in their WebChat, which they typically do not need to be aware of that dependency.

dclaux:
Discussed offline with Dan. Since this is a breaking change, I'm closing this PR, will create a new branch and push my changes in there.

dclaux:
I actually created a v0 branch that now represents the stable version of the pre v1 spec. All new changes will continue to be pushed to master which is the "working" (e.g. cutting edge) branch.

PR #470, opened 2017-07-12
Fix stretch behavior on UWP rendered actions

anbare:
This pull request addresses two issues: (1) Shared model wasn't parsing the actions property in HostConfig, and (2) UWP renderer wasn't properly displaying actions that were set to stretch (and also fixed up the vertical display behaviors too).

1) Parsing the actions property, that was a super simple one. The line of code to call parse actions was simply missing. Also noticed a subsequent bug where the mapping of inlineEdgeToEdge enum to string was missing so added that too.

2) For the buttons, the images below illustrate it best.

When HostConfig set to Horizontal, Stretch, the renderer used to be displaying this...
image

After my first fixes, it correctly displayed this
image

But notice there's extra padding on the left and right of the buttons. That's because for the button spacing, the code is applying a left/right margin to each button, regardless of whether it's the first or last button. One's first solution to fixing this would likely be "Don't apply the margin to the first and the last button" - however, that would result in unequal columns, since the first and last button would actually be larger since they don't have that extra margin included in their size. Therefore, I simply negated the left/right button margins on the outside parent container, so that the parent container actually starts further left and further right to compensate for the margins that the buttons themselves have.

After that, the horizontal stretch behavior was correct
image

Left, center, and right are still correct too and have flush margins on the edges after that fix
image
image
image

The vertical alignment needed some fixing up too so that the buttons would be properly aligned depending on what alignment the host config has set, so after a super minor change there, here's all the possible vertical layouts...
image
image
image
image

### DIFF
--- a/source/shared/cpp/ObjectModel/Enums.cpp
+++ b/source/shared/cpp/ObjectModel/Enums.cpp
@@ -237,6 +237,7 @@ ActionsOrientationNameToEnum = GenerateStringToEnumMap<ActionsOrientation>(Actio
 
 static std::unordered_map<ActionMode, std::string, EnumHash> ActionModeEnumToName =
 {
+    { ActionMode::InlineEdgeToEdge, "InlineEdgeToEdge" },
     { ActionMode::Inline, "Inline" },
     { ActionMode::Popup, "Popup" }
 };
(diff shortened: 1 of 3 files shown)


anbare:
The StackPanel makes all the other scenarios (left/center/right, and vertical) much simpler. Otherwise you would have to add column definitions every time (auto columns in those cases) AND you would have to check "Is this vertical? If so add rows instead of columns and assign the grid row instead of grid column"

I think having the separate implementations would lead to fewer bugs in the future, so it's up to you if you'd like to switch it to only a Grid implementation.

Issue #484, opened 2017-07-17
Allow elements to stretch vertically to make it possible to align content to the bottom of a fixed-height card


dclaux:
# Implementation status
- [X] TypeScript (PR https://github.com/Microsoft/AdaptiveCards/pull/483)
- [x] UWP (PR https://github.com/Microsoft/AdaptiveCards/pull/1377)
- [x] WPF (#1665)
- [x] iOS (#1296)
- [x] Android (#1297)
- [x] Documentation

# Problem
Currently, it is not really possible to align content at the bottom of a fixed-height card. Card authors have to recourse to hacks such as inserting an invisible image of a specific height to reserve some blank space. Such hacks are very fragile because they rely on knowing what the actual height of the target container is; cards authored using such hacks risk breaking if the container's height changes.

# Design
Add a new **height** property to all element types:

| Property name | Type | Description |
|-----|-----|-----|
| height | string. Allowed values are **"auto"** and **"stretch"** | Defines how the element's height is computed within its parent container. When set to **"auto"** (default), the height of the element is what it needs. When set to **"stretch"**, the element stretches to the available height in its parent container. When multiple elements have their height property set to "stretch", they share the available height equally. |

Because we are introducing a "height" property, it makes sense to also rename the "size" property of a Column to "width"

# Example
## Card payload
```
{
	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
	"type": "AdaptiveCard",
	"height": "stretch",
	"body": [
		{
			"type": "Container",
			"items": [
				{
					"type": "TextBlock",
					"text": "This is the top container.",
					"wrap": true
				}
			]
		},
		{
		    "type": "Container",
			"spacing": "large",
			"separator": true,
		    "items": [
		        {
		            "type": "TextBlock",
		            "text": "This is the middle container. Its **height** property is **stretch** so it uses all the remaining vertical space.",
		            "wrap": true
		        }
		    ]
		    
		},
		{
			"type": "ColumnSet",
			"height": "stretch",
			"columns": [
			    {
			        "width": "auto",
			        "items": [
						{
							"type": "Image",
							"url": "http://connectorsdemo.azurewebsites.net/images/MSC12_Oscar_002.jpg",
							"size": "small",
							"style": "person"
						}
			        ]
			    },
			    {
			        "width": "stretch",
			        "spacing": "large",
			        "separator": true,
			        "items": [
	                    {
	                        "type": "TextBlock",
	                        "height": "stretch",
[...]

PR #504, opened 2017-07-21
UWP: Add new height property (and change column size to width)

anbare:
As decided in #484, we're adding a height property to all card element types.

Adding this to the UWP renderer involved swapping the existing stack panel for a Grid, so that when height is set to "stretch", we can use a star-GridLength on the row that the item is being added in.

But the change is super verbose since adding the property to the UWP projection means adding the property to every single class that inherits from the card element. 

I created an AdaptiveHeight class rather than using a string or enum, since in the future you could definitely imagine that there might be pixel heights supported. So by using a class, the class would be able to support both pixel heights and values like "auto" and "stretch" - equivalent to the pattern used for XAML GridLength.

Output matches the HTML renderer which has already been updated. First screenshot is HTML, second one is UWP

image

image

Other card scenarios were validated and confirmed still working.

As part of this, we've also renamed the column "size" property to "width", since size became ambiguous with the introduction of height. Android and iOS utilizing the shared model will have to be updated since the shared model methods have changed.

Also note I didn't add support for setting the Height property programmatically, similar to how HostConfig doesn't support setting HostConfig values programmatically. It's the same problem of having the UWP projection update the actual shared model - whoever adds support for that in HostConfig can also add support for that in the Height property.

### DIFF
--- a/source/shared/cpp/ObjectModel/BaseCardElement.cpp
+++ b/source/shared/cpp/ObjectModel/BaseCardElement.cpp
@@ -34,6 +34,16 @@ AdaptiveCards::BaseCardElement::~BaseCardElement()
 {
 }
 
+Height BaseCardElement::GetHeight() const
+{
+    return m_height;
+}
+
+void BaseCardElement::SetHeight(const Height value)
+{
+    m_height = value;
+}
+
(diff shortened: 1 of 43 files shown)


almedina-ms:
The height property has already been added in a different pull request

PR #530, opened 2017-08-04
Render TextInput with textarea element

zhusongm:
Render the TextInput with <textarea> instead of <p>. The class is ac-input and ac-textinput to match the client renderer.

### DIFF
--- a/source/dotnet/Library/AdaptiveCards.Html/Rendering/HtmlRenderer.cs
+++ b/source/dotnet/Library/AdaptiveCards.Html/Rendering/HtmlRenderer.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Reflection;
 using System.Xml;
 using HtmlTags;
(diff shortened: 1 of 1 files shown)


source/dotnet/Library/AdaptiveCards.Html/Rendering/HtmlRenderer.cs
```diff
@@ -620,16 +621,24 @@ protected static HtmlTag NumberInputRender(TypedElement element, RenderContext c
...
-                    Color = TextColor.Accent,
-                    Wrap = true
-                });
-            return context.Render(container);
+
+            var uiTextInput = new HtmlTag("textarea")
+                .AddClass("ac-textinput")
+                .AddClass("ac-input")
+                .Style("width", "100%")
+                .Style("box-sizing", "border-box");
```
matthidinger: Should this be included in the ac-textinput class or does it need to be an inline style?

likecool21: I'm not sure at this point about styles yet; some styles could be tied to the TextInput itself while some are determined by the host config. I simply copy-pasted the attributes I saw while running the example on adaptivecards.io. I'll definitely revisit the styles with followup commits.

source/dotnet/Library/AdaptiveCards.Html/Rendering/HtmlRenderer.cs
```diff
@@ -620,16 +621,24 @@ protected static HtmlTag NumberInputRender(TypedElement element, RenderContext c
...
+
+            if (!string.IsNullOrEmpty(input.Placeholder))
+            {
+                uiTextInput.Attr("placeholder", WebUtility.HtmlEncode(input.Placeholder));
```
andrewmagee: Don't encode here. The HtmlTag object contains unencoded attribute keys/values. They're encoded at the time they're converted to HTML, not when you set them.

source/dotnet/Library/AdaptiveCards.Html/Rendering/HtmlRenderer.cs
```diff
@@ -620,16 +621,24 @@ protected static HtmlTag NumberInputRender(TypedElement element, RenderContext c
...
+            }
+
+            if (!string.IsNullOrEmpty(input.Value))
+            {
+                uiTextInput.Text = input.Value;
```
andrewmagee: Value should be stored in the uiTextInput's "value" attribute, not on the text field.

zhusongm: I think this is the way to set the text in textarea. According to https://www.w3schools.com/tags/tag_textarea.asp <textarea> tag does not have a value attribute; and I've tried setting value attribute does not have any effect in chrome

andrewmagee: Oops, you're right - I was looking at text input, not textarea.

source/dotnet/Library/AdaptiveCards.Html/Rendering/HtmlRenderer.cs
zhusongm: So I have found the client side renderer implementation and I'm going to follow exact what it does there in following PRs.

zhusongm: I have updated a commit where the renderer matched the behavior with the client renderer. When IsMultiline is set to true we render a <textarea>; otherwise render a <input> with type=text.

Issue #330, opened 2017-05-15
UWP parser requires "type": "Column" when it's not necessary


andrewleader:
Accepted, task list: 

- [x] TypeScript
- [x] UWP (PR #555)
- [x] Android (PR #555)
- [x] iOS (PR #555)
- [x] .NET

The following payload will break the UWP parser

```json
{
   "type":"AdaptiveCard",
   "version":"1.0",
   "body":[
      {
         "type":"ColumnSet",
         "columns":[
            {
               "size":33,
               "items":[
                  {
                     "type":"Image",
                     "url":"https://unsplash.it/64?image=1027",
                     "size":"small",
                     "style":"person"
                  }
               ]
            },
            {
               "size":67,
               "contentVerticalAlignment":"center",
               "items":[
                  {
                     "type":"TextBlock",
                     "text":"Hi,",
                     "size":"large"
                  },
                  {
                     "type":"TextBlock",
                     "text":"MasterHip",
                     "size":"medium",
                     "isSubtle":true
                  }
               ]
            }
         ]
      }
   ]
}
```

It breaks simply because each column item doesn't have `"type": "Column"` specified which is redundant and unnecessary, since ColumnSet's column property is explicitly an array of Column type... the Web visualizer doesn't break on this.

riarenas:
According to http://adaptivecards.io/documentation/#column. The type property is required.

riarenas:
Didn't mean to close.

riarenas:
Adding Schema label for further review. Simpliifying this seems like a good idea.

matthidinger:
My vote is that we keep the Type property required for the time being. It makes readability of a large payload slightly easier

anbare:
Why not just use the Type property for the samples, but make the property itself optional? Who cares what devs are sending when they programmatically create their payloads. Might as well allow them to save space by not providing properties that are meaningless except when visually viewing the payload itself.

dclaux:
If we think we could have a different type of column in the future, then I vote for "type" being required. Otherwise, omitting "type" when the type is implicit is a good simplification. We don't have "type": "Fact" for a fact in a FactSet for instance.

dclaux:
@anbare I notice the contentVerticalAlignment property in the payload above. Is it a proposed new feature?

anbare:
[...]

PR #555, opened 2017-08-14
Shared Model: Optional type property on Set children

anbare:
Updated the shared model so that columns don't require a type property (since their type is implicit) and images inside ImageSet don't require a type property. Facts and ChoiceInputs already didn't require the type property.

Also refactored fact/choicesetinput code to use the new shared method for parsing an array of single type.

Tested/verified all changes in UWP Visualizer app for ChoiceSet, ColumnSet, FactSet, ImageSet, and normal inline images both with and without types specified, all worked correctly.

image

image


anbare:
Fixes #330

PR #564, opened 2017-08-17
Render NumberInput, DateInput, TimeInput, ToggleInput and ChoicesetInput in .net html renderer

zhusongm:


zhusongm:
Render a choice set based on following conditions.
1. IsMultiSelect == false && IsCompact == true => render as a drop down select element
2. IsMultiSelect == false && IsCompact == false => render as a list of radio buttons
3. IsMultiSelect == true => render as a list of toggle inputs

Render toggleInput using an input element with type=checkbox.

Also move the style:width=100% in the textblock renderer from the parent div tag to child p tag. This matches the client renderer behavior and will put the textblock on the same line with the radio button and checkbox when used together.

PR #580, opened 2017-08-24
scheme shared for CLI compliation

nesalang:
fixes compilation issue when compiling for framework


nesalang:
failure is expected on Android

Issue #556, opened 2017-08-14
UWP renderer crashes if HTTP image cannot be found


anbare:
If an image can't be found, like bad HTTP link or anything, the UWP renderer crashes uncatchably and crashes the parent app.

Repro payload

```json
{
	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
	"type": "AdaptiveCard",
	"version": "0.5",
	"body": [
		{
			"type": "Image",
			"url": "http://adaptivecards.io/assets/Closed%20bug%2092x92.png",
			"size": "medium"
		}
	]
}
```

anbare:
@riarenas this is likely that missing THROW_IF_FAILED, did you already fix that or want me to submit a pull request?

```c++
httpClient->GetInputStreamAsync(imageUri, &getStreamOperation);
```

anbare:
Eh, seems it's not simply just that, adding a THROW_IF_FAILED it still crashes, investigating further

anbare:
Figured out the source of the error, inside PopulateImageFromUrlAsync, one of the delegates will throw an error if the image download failed. But since no one is capturing that async result, it ends up as an unhandled exception.

![image](https://user-images.githubusercontent.com/13246069/29293258-53aee3a6-80ff-11e7-99cb-875eb0ed336e.png)

Maybe reporting the error via the imageLoadTracker will be the solution.

anbare:
@matthidinger needs to decide on the final behavior of this.

Options are...

1) Have RenderAsync throw an exception when an HTTP image fails to load

2) Make RenderAsync return some sort of result that contains both the rendered card (without the image) and info like "2 images failed to load"

I've written the code for (1) in branch anbare/uwp-fix-image-errors, but (2) is likely the better solution so it would have to be slightly modified if we want that.

anbare:
@matthidinger your vote on this?

We could always start with option 1 since the hosts like Action Center that we know about would choose that option to not display anything if image download fails, and later on we could add an advanced Render method which supports partial rendering and returning errors like image download failure.

Although another interesting question is whether buttons should still be displayed? Action Center currently will drop the adaptive XML portion if an image fails to download, but since buttons currently aren't part of adaptive toasts, the buttons still get displayed.

While I'm gone next week feel free to use my code in branch _anbare/uwp-fix-image-errors_ if we're going with option 1 since that's already coded up and I've tested that it works!

matthidinger:
[...]

PR #587, opened 2017-08-29
Stop UWP renderer from crashing due to Image loading/XamlBuilder exceptions.

riarenas:
Exceptions thrown while building the Xaml tree were not being caught and reported to the async operation. This resulted in the Async operation never completing.

This change adds two error results to the async operation:
Image Loading related Errors
All other XamlBuilder errors.

This will be overhauled once we switch the renderer to return a RenderResult instead of the Xaml elements directly.

### DIFF
--- a/source/uwp/Renderer/lib/AsyncOperations.h
+++ b/source/uwp/Renderer/lib/AsyncOperations.h
@@ -53,10 +53,14 @@ public:
         return OnXamlImagesLoaded();
     }
 
-    // IXamlBuilderListener
     IFACEMETHODIMP ImagesLoadingHadError()
     {
         return OnXamlImagesHadError();
+    }
+
+    IFACEMETHODIMP XamlBuilderHadError()
+    {
+        return OnXamlBuilderHadError();
     }
 
 protected:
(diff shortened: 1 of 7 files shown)


riarenas:
Fixes #556

source/uwp/Renderer/lib/AsyncOperations.h
```diff
@@ -67,7 +77,15 @@ class RenderAsyncBase :
...
-            m_builder->BuildXamlTreeFromAdaptiveCard(m_card.Get(), &m_rootXamlElement, m_renderer.Get());
+            try
+            {
+                m_builder->BuildXamlTreeFromAdaptiveCard(m_card.Get(), &m_rootXamlElement, m_renderer.Get());
+            }
+            catch (...)
+            {
+                //Catch all non-Image loading related problems.
+                return XamlBuilderHadError();
+            }
```
RebeccaAnne: Perhaps xamlbuilder should return an HRESULT rather than convert all the hresults to exceptions via THROW_IF_FAILED.  Something to consider when we rework to return a "render result"

riarenas: Yep, definitely. This is not the ideal solution, but since we need to do a more extensive rework in this same area, this is meant as a small patch to mitigate the issues we are already seeing.

PR #588, opened 2017-08-29
Fixes Android Lint Errors

nesalang:
fixes lint errors that cause build failures

### DIFF
--- a/source/android/adaptivecards/build.gradle
+++ b/source/android/adaptivecards/build.gradle
@@ -5,7 +5,7 @@ android {
     buildToolsVersion "25.0.3"
 
     defaultConfig {
-        minSdkVersion 14
+        minSdkVersion 21
         targetSdkVersion 25
         versionCode 1
         versionName "1.0"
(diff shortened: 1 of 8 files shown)


source/android/mobile/src/main/AndroidManifest.xml
```diff
@@ -19,7 +19,7 @@
             android:configChanges="orientation|screenSize|keyboardHidden">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
-
+                <action android:name="android.media.action.MEDIA_PLAY_FROM_SEARCH" />
```
riarenas: What is this for? I thought this was for apps that want to play music.

nesalang: "To support voice searches on Android Auto, you should also register an intent-filter for the action android.media.action.MEDIA_PLAY_FROM_SEARCH."

---
In reply to: [135900329](https://github.com/Microsoft/AdaptiveCards/pull/588#discussion_r135900329) [](ancestors = 135900329)

source/android/adaptivecards/build.gradle
```diff
@@ -29,6 +30,19 @@ android {
     }
 }
 
+uploadArchives {
+        repositories.mavenDeployer {
+            repository(url: "https://oss.sonatype.org/service/local/staging/deploy/maven2") {
+                authentication(userName: "adaptivecards", password: "Sird[oshkIt5")
```
riarenas: having the password there seems like a very bad idea.

riarenas: And please change it ASAP before adding the new commits.

Issue #585, opened 2017-08-29
Adaptive Card Schema is missing Speak property


riarenas:
https://docs.microsoft.com/en-us/adaptive-cards/create/cardschema#adaptivecard

has no mention of a speak property, but https://docs.microsoft.com/en-us/adaptive-cards/create/speech mentions there should be a speak property on the card for "On Delivery" reading of the property.
Once it is added, should be implemented in all renderers.

matthidinger:
This has been added  now

Issue #319, opened 2017-05-10
Handle speak properties in the UWP renderer


RebeccaAnne:

riarenas:
@matthidinger Do we need to handle the speak property in any way? Or is it up to the host to handle it?

PR #589, opened 2017-08-29
Add speak property to AdaptiveCard in shared model and UWP renderer

riarenas:
Also fixes the Property name for Adaptive Card Background Image in the WinRT projection


riarenas:
References #585

RebeccaAnne:
Also #319

---
In reply to: [325810517](https://github.com/Microsoft/AdaptiveCards/pull/589#issuecomment-325810517) [](ancestors = 325810517)

Issue #584, opened 2017-08-29
Typescript/UWP renderer are case sensitive for string property values. 


riarenas:
As an example, column Size property is case sensitive:

"size": "auto" :heavy_check_mark:
"size": "Auto" :x:

Should we be this strict with the casing in the value for the size? The documentation says this:

"auto", "stretch", or a number representing relative width of the column in the column Set (Default:Auto)

Although the docs are clear in that we expect those literal strings, the remark about the default might be interpreted as taking Auto as a valid value.

I tend to think we should be case insensitive when evaluating this value. Thoughts?

matthidinger:
I agree case insensitive seems the best way to go. @dclaux any issue?

dclaux:
I don't have a problem with it.

riarenas:
Trying to get some closure on this one. @dclaux Did this make it to the typescript renderer? The published visualizer is still case sensitive for column width.

dclaux:
This has been changed a long time ago.

PR #590, opened 2017-08-29
force column size parsing to lowercase in Shared Model

riarenas:
Do not merge until we have a decision on #584


source/shared/cpp/ObjectModel/Column.cpp
```diff
@@ -59,7 +59,7 @@ std::string Column::GetSize() const
 
 void Column::SetSize(const std::string value)
```
RebeccaAnne: >SetSize [](start = 13, length = 7)

Other similar properties coming in a future change?

PR #601, opened 2017-09-01
string properties comparison should be case insensitive 

QingpingMeng:


source/html/renderer/src/utils.ts
```diff
@@ -27,22 +27,39 @@ export function isNullOrEmpty(value: string): boolean {
...
     }
 }
 
+export function getEnumValueOrDefault(targetEnum: { [s: number]: string }, name: string, defaultValue: number): number {
+    if (isNullOrEmpty(name)) {
+        return defaultValue;
+    }
+
+    for (var key in targetEnum) {
+        let isValueProperty = parseInt(key, 10) >= 0
```
dclaux: I've been trying to use "var" everywhere instead of "let" (although there are still some left.) Both will work just fine, and actually we should probably use let everywhere rather than var.

source/html/renderer/src/utils.ts
```diff
@@ -27,22 +27,39 @@ export function isNullOrEmpty(value: string): boolean {
...
+        if (isValueProperty) {
```
dclaux: Empty line before if

source/html/renderer/src/utils.ts
```diff
@@ -27,22 +27,39 @@ export function isNullOrEmpty(value: string): boolean {
...
+            let value = targetEnum[key];
+            if (typeof value === 'string' || value instanceof String) {
```
dclaux: Empty line before if

source/html/renderer/src/utils.ts
```diff
@@ -27,22 +27,39 @@ export function isNullOrEmpty(value: string): boolean {
...
+                console.log(value, name);
+                if (value.toLowerCase() === name.toLowerCase()) {
```
dclaux: Empty line before if

source/html/visualizer/package.json
```diff
@@ -11,7 +11,7 @@
     "prebuild-css": "node-sass ./scss/ -o ./css --output-style expanded",
     "build-css": "postcss ./css/*.css --use autoprefixer --replace --no-map",
     "watch": "webpack --watch",
-    "start": "http-server -c-1 -o",
+    "start": "http-server -c-1 -o -p 5666",
```
dclaux: What does this do?

source/html/visualizer/src/app.ts
```diff
@@ -43,12 +43,9 @@ function renderCard(): HTMLElement {
...
     var adaptiveCard = new Adaptive.AdaptiveCard();
 
-    try {
-        adaptiveCard.hostConfig = Adaptive.parseHostConfig(currentConfigPayload);
-    }
-    catch (e) {
-        // TODO
-    }
+    
+    adaptiveCard.hostConfig = Adaptive.parseHostConfig(currentConfigPayload);
```
dclaux: Why remove the try...catch? If not caught, this exception will break the visualizer. It is swallowed by design.

source/html/visualizer/src/containers/bing.ts
```diff
@@ -1,32 +1,42 @@
...
-        element.style.overflow = "hidden"; 
- 
-        renderedCard.style.height = "100%"; 
- 
-        element.appendChild(renderedCard); 
- 
-        return element; 
-    } 
+import { HostContainer } from "./host-container";
+import {
```
dclaux: Why do this instead of what was before? Is it just to not have to type "Adaptive." everywhere?

source/html/visualizer/package.json
QingpingMeng: change the localhost http server port number as it conflicts with my other projects. I can revert this back.

source/html/visualizer/src/app.ts
QingpingMeng: the exception will be caught by outer try..catch of tryRenderCard

source/html/visualizer/src/containers/bing.ts
QingpingMeng: Basically only import minimal required object, it's easier for Webpack to do the code splitting work.

source/html/renderer/src/utils.ts
QingpingMeng: done

source/html/renderer/src/utils.ts
QingpingMeng: Agreed. Will gradually change to let/const.

source/html/renderer/src/utils.ts
QingpingMeng: done

source/html/renderer/src/utils.ts
QingpingMeng: done

source/html/renderer/src/card-elements.ts
```diff
@@ -3416,7 +3416,7 @@ var defaultHostConfig: HostConfig.IHostConfig = {
         spacing: Enums.Spacing.Default,
         buttonSpacing: 20,
         showCard: {
-            actionMode: Enums.ShowCardActionMode.Inline,
+            actionMode: Enums.ShowCardActionMode.InlineEdgeToEdge,
```
dclaux: You will want to pull master - inlineEdgeToEdge is gone.

PR #614, opened 2017-09-08
Reorg

nesalang:
changed name of framework & updated .gitignore
added new build target for universal bin generation


nesalang:
iOS check failing is none issue. complains about missing workspace -> workspace name has changed and VSTS build task can't have updated workspace name.

PR #615, opened 2017-09-08
add basic accessibility support

QingpingMeng:


source/html/renderer/src/card-elements.ts
```diff
@@ -3253,12 +3299,16 @@ export class AdaptiveCard extends ContainerWithActions {
...
             renderedCard = document.createElement("div");
             renderedCard.innerHTML = this.fallbackText ? this.fallbackText : "The specified card version is not supported.";
-
-            return renderedCard;
         }
         else {
-            return super.render();
+            renderedCard = super.render();
+            renderedCard.tabIndex = 0;
+            if (!Utils.isNullOrEmpty(this.speak)) {
```
dclaux: Empty line before if

QingpingMeng: No, text block doesn't require aria label. It can be picked up by narrators.

PR #617, opened 2017-09-11
Fix UWP date/time not stretching full width

### LINKED ISSUE #616: [UWP] Date and time picker doesn't stretch full width
image

Simple fix, horizontal alignment needs to be set for them, I'll make the changes.

anbare:
Fixes #616. Simply had to assign horizontal alignment stretch.

**BEFORE**
image

**AFTER**
image

Note that at very wide sizes, the time picker will reach a max width and stop growing. Not sure there's an easy way to fix that (I believe that's a limitation in the time picker itself). Plus it seems like we have a TODO to support displaying placeholder text on the time picker, which would require creating a brand new control, so that new control would support super wide widths... no need to invest time in making the time picker we have today support ultra-wide sizes.

image

### DIFF
--- a/source/uwp/Renderer/lib/XamlBuilder.cpp
+++ b/source/uwp/Renderer/lib/XamlBuilder.cpp
@@ -693,16 +693,16 @@ namespace AdaptiveCards { namespace XamlCardRenderer
             switch (actionAlignment)
             {
             case ABI::AdaptiveCards::XamlCardRenderer::ActionAlignment::Center:
-                actionsFrameworkElement->put_HorizontalAlignment(HorizontalAlignment_Center);
+                THROW_IF_FAILED(actionsFrameworkElement->put_HorizontalAlignment(HorizontalAlignment_Center));
                 break;
             case ABI::AdaptiveCards::XamlCardRenderer::ActionAlignment::Left:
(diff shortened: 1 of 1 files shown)


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1883,6 +1883,11 @@ namespace AdaptiveCards { namespace XamlCardRenderer
         THROW_IF_FAILED(adaptiveDateInput->get_Placeholder(placeHolderText.GetAddressOf()));
         THROW_IF_FAILED(datePicker->put_PlaceholderText(placeHolderText.Get()));
 
+        // Make the picker stretch full width
+        ComPtr<IFrameworkElement> datePickerAsFrameworkElement;
+        THROW_IF_FAILED(datePicker.As(&datePickerAsFrameworkElement));
+        datePickerAsFrameworkElement->put_HorizontalAlignment(ABI::Windows::UI::Xaml::HorizontalAlignment::HorizontalAlignment_Stretch);
```
RebeccaAnne: >put_HorizontalAlignment [](start = 38, length = 23)

Wrap in THROW_IF_FAILED?

anbare: Thanks for catching that! Updating

PR #621, opened 2017-09-12
updated build script / fixed header names

nesalang:


nesalang:
plz ignore iOS failure as it is build target naming issue on VSTS

PR #625, opened 2017-09-14
nuget name changes

nesalang:
updates uwp nuget package names to AdaptiveCards.UWP


nesalang:
I will Leave it to Ricardo to comment

riarenas:
I would keep this change contained to the nuget package. I will follow up separately if there are any issues with changing the dll names, I don't suspect there will be, but just in case. I will send a separate PR for that once I have an answer.

nesalang:
I guess we are agreeing that this PR be merged; and Ricardo follows up with dll name changes later?
or should I wait for the dll name change commit(s) added to this PR?

riarenas:
dll and winmd name change will come as a separate PR. I need to evaluate the impact on the name change int he windows component

matthidinger:
Please keep this PR open for a bit, I'm researching a few other naming/casing guidelines for nuget, as well as the version. We shouldn't be using 1.0.0 yet, it should be something like 1.0.0-alpha.1

riarenas:
@nesalang Please revert the build definition changes until this is checked in. It is causing official build failures.

matthidinger:
This is probably obsolete and can be closed and replaced with the latest decision?

Issue #138, opened 2017-03-06
Support custom image URIs, enabling symbols and URI schemes like ms-appx


matthidinger:
# Implementation status
- [ ] TypeScript
- [x] UWP (PR #638)
- [x] WPF
- [x] iOS
- [x] Android
- [x] Documentation

# Scenarios  
Enables Card authors to use custom URL schemes for Images, e.g., a set of icons that are packaged within a particular app, or specialized form of URL resolution.

## Scenario: Custom Icons
The Skype app could include a custom set of icons that are packaged within their app, e.g., a set of Media player icons. A developer could reference those icons by using a custom `icon:` scheme that Skype controls and documents.

```json
{
"type":"image",
"url":"skype-icon:play",
}
```

## Scenario: ms-appx resolution
Host apps like Timeline and Action Center can resolve ms-appx URLs to the correct package that contains the assets.

```json
{
"type":"image",
"url":"ms-appx://foo.png",
}
```

# Design

Renderers provide an API to register a custom `ResourceResolver` for a specific scheme. Hosts can override http/https images too.

## Host App
```csharp
renderer = new AdaptiveCardRenderer();

// Provide custom resource resolver
renderer.ResourceResolvers.Set("ms-appx", new MsAppxResourceResolver());

// Host implements the resolver like seen below
private class MsAppxResourceResolver : IAdaptiveCardResourceResolver
{
    public IAsyncOperation<IRandomAccessStream> GetResourceStreamAsync(XamlCardGetImageStreamArgs args)
    {
        return GetResourceStreamHelperAsync(args).AsAsyncOperation();
    }

    private async Task<IRandomAccessStream> GetResourceStreamHelperAsync(XamlCardGetImageStreamArgs args)
    {
        StorageFile file = await StorageFile.GetFileFromApplicationUriAsync(args.Url);
        return await file.OpenAsync(FileAccessMode.Read);
    }
}
```

## Interface of IAdaptiveCardResourceResolver
Note that each platform will have their own style of Resolver. For WinRT, it returns an IRandomAccessStream, since that's what WinRT uses. Other platforms might use a byte[] or something else.

```csharp
// Resource resolvers have the following interface
public interface IAdaptiveCardResourceResolver
{
    IAsyncOperation<IRandomAccessStream> GetResourceStreamAsync(XamlCardGetImageStreamArgs args);
}
```

# Fallback URL
Introduce fallbackBack url in case the primary URL doesn't load an image

```json
{
[...]

PR #638, opened 2017-09-18
Add custom resource resolver for UWP renderer

anbare:
Custom resource resolver allows apps to register custom handlers for resolving various URL streams, like custom image handlers and in the future audio/video URLs.

I also added some code in the Visualizer to test/utilize custom resolvers. See the bottom of this for how to test.

## How hosts provide a custom resolver

```csharp
_renderer = new XamlCardRenderer();

// Custom resolvers
IXamlCardResourceResolver httpResolver = new MyHttpResourceResolver();

_renderer.ResourceResolvers.Set("http", httpResolver);
_renderer.ResourceResolvers.Set("https", httpResolver);
_renderer.ResourceResolvers.Set("ms-appx", new MyAppxResourceResolver());
```

Your custom resolver needs to implement the `IXamlCardResourceResolver` interface

```csharp
public interface IXamlCardResourceResolver
{
    IAsyncOperation<IRandomAccessStream> GetResourceStreamAsync(XamlCardGetResourceStreamArgs args);
}
```

A simple implementation is seen below...

```csharp
private class MyAppxResourceResolver : IXamlCardResourceResolver
{
    public IAsyncOperation<IRandomAccessStream> GetResourceStreamAsync(XamlCardGetResourceStreamArgs args)
    {
        return GetResourceStreamHelperAsync(args).AsAsyncOperation();
    }

    private async Task<IRandomAccessStream> GetResourceStreamHelperAsync(XamlCardGetResourceStreamArgs args)
    {
        StorageFile file = await StorageFile.GetFileFromApplicationUriAsync(args.Url);
        return await file.OpenAsync(FileAccessMode.Read);
    }
}
```

You get passed an args object, which allows us to add additional properties in the future. Today the args only contains the image URL, but you can imagine in the future we’ll also possibly want to supply the style of the image (person or normal), or maybe the size of it (small/medium/wide/large) or other values. If we passed those as parameters in the interface, we would never be able to add new parameters since changing the interface would be a breaking change. Adding new properties to this args class isn’t a breaking change though! 

```csharp
public sealed class XamlCardGetResourceStreamArgs
{
    public Uri Url { get; }
}
```

## How to test

1. Build/run/open the UWP Visualizer
2. Paste in the following payload

```json
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "TextBlock",
      "text": "Seattle vs Panthers",
      "isSubtle": true
    },
    {
      "type": "Container",
      "items": [
        {
          "type": "ColumnSet",
          "columns": [
            {
              "type": "Column",
              "size": "auto",
              "items": [
                {
                  "type": "Image",
                  "url": "symbol:down",
                  "size": "small",
                  "horizontalAlignment": "center"
                },
                {
                  "type": "TextBlock",
                  "text": "CAR",
                  "horizontalAlignment": "center",
                  "weight": "bolder"
                }
              ]
            },
            {
              "type": "Column",
              "size": "auto",
              "items": [
                {
                  "type": "Image",
                  "url": "symbol:up",
                  "size": "small",
                  "horizontalAlignment": "center"
                },
                {
                  "type": "TextBlock",
                  "text": "SEA",
                  "horizontalAlignment": "center",
                  "weight": "bolder"
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

image

You should then see the custom down/up icons for the symbol: scheme! Note that down and up are the only custom assets supported.


anbare:
Implements feature #138

anbare:
Updated this to support the non-Async render method, tested both async and non-async using the UWP Visualizer, both paths function!

anbare:
Fixed the merge conflicts that were introduced in master and then re-tested. @RebeccaAnne this is ready for review

PR #649, opened 2017-09-20
Rename column size to width

RebeccaAnne:


RebeccaAnne:
Resending this from a new branch to work around the issue in the staging branch.

PR #663, opened 2017-09-22
Remove speak from everything except adaptive card in shared model and UWP projections

riarenas:


riarenas:
Too many conflicts! closing and reopening once resolved.

PR #668, opened 2017-09-22
Fix UWP host config color parsing and add host config editor to UWP Visualizer

### LINKED ISSUE #667: UWP colors are incorrect due to host config not parsing the hex colors correctly
In UWP, the host config colors get parsed as ARBG, which is incorrect, it's supposed to be ARGB

That results in colors like this
image

```c
    std::string alphaString = colorString.substr(1, 2);
    INT32 alpha = strtol(alphaString.c_str(), nullptr, 16);

    std::string redString = colorString.substr(3, 2);
    INT32 red = strtol(redString.c_str(), nullptr, 16);

    // GREEN should be parsed here
    std::string blueString = colorString.substr(5, 2);
    INT32 blue = strtol(blueString.c_str(), nullptr, 16);

    std::string greenString = colorString.substr(7, 2);
    INT32 green = strtol(greenString.c_str(), nullptr, 16);
```

Simple fix, I'll make it as part of the addition of adding Host Config editor to the UWP Visualizer

anbare:
This fixes #667 where colors specified in host config get rendered incorrectly in UWP renderer.

As part of this I also added a host config editor to the UWP Visualizer so we can test host config changes.

For the incorrect colors, UWP was parsing as ARBG, which is incorrect, it's supposed to be ARGB

That results in colors like this
image

```c
    std::string alphaString = colorString.substr(1, 2);
    INT32 alpha = strtol(alphaString.c_str(), nullptr, 16);

    std::string redString = colorString.substr(3, 2);
    INT32 red = strtol(redString.c_str(), nullptr, 16);

    // GREEN should be parsed here
    std::string blueString = colorString.substr(5, 2);
    INT32 blue = strtol(blueString.c_str(), nullptr, 16);

    std::string greenString = colorString.substr(7, 2);
    INT32 green = strtol(greenString.c_str(), nullptr, 16);
```

After that simple fix, it's displayed correctly

image

As for the new host config editor, you can now edit the host config in the UWP Visualizer. I also made some improvements so that cards only get rendered when you select the tab (rather than rendering ALL tabs at once regardless of which one is actually open).

So lots of edits in the Visualizer code, but only one file modification in the actual renderer code.

### DIFF
--- a/source/uwp/Renderer/lib/Util.cpp
+++ b/source/uwp/Renderer/lib/Util.cpp
@@ -225,16 +225,16 @@ HRESULT GetColorFromString(std::string colorString, Color *color) noexcept try
     std::string redString = colorString.substr(3, 2);
     INT32 red = strtol(redString.c_str(), nullptr, 16);
 
-    std::string blueString = colorString.substr(5, 2);
-    INT32 blue = strtol(blueString.c_str(), nullptr, 16);
-
-    std::string greenString = colorString.substr(7, 2);
+    std::string greenString = colorString.substr(5, 2);
     INT32 green = strtol(greenString.c_str(), nullptr, 16);
+
+    std::string blueString = colorString.substr(7, 2);
+    INT32 blue = strtol(blueString.c_str(), nullptr, 16);
 
     color->A = static_cast<BYTE>(alpha);
     color->R = static_cast<BYTE>(red);
-    color->B = static_cast<BYTE>(blue);
     color->G = static_cast<BYTE>(green);
+    color->B = static_cast<BYTE>(blue);
 
     return S_OK;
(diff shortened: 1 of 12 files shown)


source/uwp/Renderer/lib/Util.cpp
```diff
@@ -225,16 +225,16 @@ HRESULT GetColorFromString(std::string colorString, Color *color) noexcept try
     std::string redString = colorString.substr(3, 2);
     INT32 red = strtol(redString.c_str(), nullptr, 16);
 
-    std::string blueString = colorString.substr(5, 2);
```
riarenas: >std::string blueString = colorString.substr(5, 2); [](start = 4, length = 50)

Good catch!

PR #684, opened 2017-09-25
This will never register a renderer currently

Blahartinger:
Needs to be  && not ||

### DIFF
--- a/source/android/adaptivecards/src/main/java/com/microsoft/adaptivecards/renderer/registration/CardRendererRegistration.java
+++ b/source/android/adaptivecards/src/main/java/com/microsoft/adaptivecards/renderer/registration/CardRendererRegistration.java
@@ -68,7 +68,7 @@ public class CardRendererRegistration
 
     public void registerRenderer(CardElementType cardElementType, IBaseCardElementRenderer renderer)
     {
-        if (cardElementType != null || cardElementType == CardElementType.Unsupported)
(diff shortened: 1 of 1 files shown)


source/android/adaptivecards/src/main/java/com/microsoft/adaptivecards/renderer/registration/CardRendererRegistration.java
```diff
@@ -68,7 +68,7 @@ public static CardRendererRegistration getInstance()
 
     public void registerRenderer(CardElementType cardElementType, IBaseCardElementRenderer renderer)
     {
-        if (cardElementType != null || cardElementType == CardElementType.Unsupported)
+        if (cardElementType != null && cardElementType == CardElementType.Unsupported)
```
bekao: >&& [](start = 36, length = 2)

It should be || and not &&.  However, the first check should be == and not !=

bekao: I have a follow up PR that will include this fix plus passing string as id instead of CardElementType enum since this is final class and cannot be inherited.

---
In reply to: [140910916](https://github.com/Microsoft/AdaptiveCards/pull/684#discussion_r140910916) [](ancestors = 140910916)

PR #688, opened 2017-09-26
Change namespace to AdaptiveCards.Uwp, and rename XamlCard* classes to AdaptiveCard*

riarenas:
Prep work for API changes.

Verified by running the visualizer and looping through the samples after the renames.


riarenas:
Updated all the new renderer/parser files

PR #695, opened 2017-09-26
Add renderer registration and per-element renderer APIs

riarenas:


source/uwp/Renderer/lib/XamlCardRendererComponent.cpp
```diff
@@ -231,4 +265,45 @@ namespace AdaptiveCards { namespace XamlCardRenderer
...
+    HRESULT XamlCardRenderer::get_ActionRenderers(IAdaptiveActionRendererRegistration **value)
+    {
+        *value = m_actionRendererRegistration.Get();
+        return S_OK;
+    }
+
+    _Use_decl_annotations_
+    HRESULT XamlCardRenderer::RegisterDefaultElementRenderers()
+    {
+        RETURN_IF_FAILED(m_elementRendererRegistration->Set(HStringReference(L"Input.ChoiceSet").Get(), ComPtr<IAdaptiveElementRenderer>(Make<AdaptiveChoiceSetInputRenderer>().Get()).Get()));
```
riarenas: > RETURN_IF_FAILED(m_ [](start = 7, length = 20)

Any ideas on another way to do this easily? I'm not a fan of how this ended up looking.

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -390,10 +391,15 @@ HRESULT StringToJsonObject(const string inputString, IJsonObject** result)
...
     ComPtr<IJsonObject> jObject;
-    RETURN_IF_FAILED(jObjectStatics->Parse(inputsHString, &jObject));
-
+    HRESULT hr = jObjectStatics->Parse(inputsHString, &jObject);
+    if (FAILED(hr))
+    {
+        RETURN_IF_FAILED(ActivateInstance(
+            HStringReference(RuntimeClass_Windows_Data_Json_JsonObject).Get(),
+            &jObject));
+    }
```
riarenas: Noticed this bug while adding the new util functions.

riarenas:
>             [propget] HRESULT ElementType([out, retval] ElementType* value);

We need to think about this property here. If someone is creating a custom card element and a custom parser for it, they won't be able to extend the ElementType enum, so they can't really implement this property.

---
Refers to: source/uwp/Renderer/idl/AdaptiveCards.XamlCardRenderer.idl:298 in 8a61c98. [](commit_id = 8a61c98216867962167b5adc32295012ba4d6b06, deletion_comment = False)

source/uwp/Renderer/lib/AdaptiveActionRendererRegistration.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include "AdaptiveCards.XamlCardRenderer.h"
+#include "Util.h"
+
+namespace AdaptiveCards { namespace XamlCardRenderer
+{
+    class AdaptiveActionRendererRegistration :
```
riarenas: >AdaptiveActionRendererRegistration [](start = 10, length = 34)

This class looks exactly the same as the ElementRenderer Registration. Sadly we can't inherit from templated BaseRegistration class that would make this class as easy as creating a BaseRegistration<ABI::AdaptiveCards::XamlCardRenderer::IAdaptiveActionElement>

riarenas: The parser registrations will look the same as well.

---
In reply to: [141201602](https://github.com/Microsoft/AdaptiveCards/pull/695#discussion_r141201602) [](ancestors = 141201602)

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
```diff
@@ -0,0 +1,29 @@
...
+using namespace ABI::Windows::Foundation;
+
+namespace AdaptiveCards { namespace XamlCardRenderer
+{
+    HRESULT AdaptiveChoiceSetInputRenderer::RuntimeClassInitialize() noexcept try
+    {
+        return S_OK;
+    } CATCH_RETURN;
+
+
```
riarenas: Extra lines are the bane of my existence.

source/uwp/Renderer/idl/AdaptiveCards.XamlCardRenderer.idl
```diff
@@ -278,6 +305,8 @@ namespace AdaptiveCards
 
             [propget] HRESULT Id([out, retval] HSTRING* value);
             [propput] HRESULT Id([in] HSTRING value);
+
+            [propget] HRESULT Type([out, retval] HSTRING* value);
```
riarenas: >Type [](start = 30, length = 4)

This will be used when querying for the ElementRenderers

RebeccaAnne:
>             [propget] HRESULT ElementType([out, retval] ElementType* value);

We talked offline.  Lets keep this (so people walking our object model can figure out what they're looking at, and add a "Custom" value to the enum.  Let's also rename Type to ElementTypeString.

---
In reply to: [332353799](https://github.com/Microsoft/AdaptiveCards/pull/695#issuecomment-332353799) [](ancestors = 332353799)

---
Refers to: source/uwp/Renderer/idl/AdaptiveCards.XamlCardRenderer.idl:298 in b6081ad. [](commit_id = b6081adb158436f0290788cad29cf7381b739e0a, deletion_comment = False)

source/uwp/Renderer/idl/AdaptiveCards.XamlCardRenderer.idl
```diff
@@ -1364,6 +1397,8 @@ namespace AdaptiveCards
...
```
RebeccaAnne: >Type [](start = 30, length = 4)

Same thing here as for the card element

source/uwp/Renderer/lib/AdaptiveActionRendererRegistration.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    _Use_decl_annotations_
+    HRESULT AdaptiveActionRendererRegistration::Remove(_In_ HSTRING type)
+    {
+        m_registration->erase(TypeAsRegistrationKey(type));
+        return S_OK;
+    }
+
+    _Use_decl_annotations_
+    std::string AdaptiveActionRendererRegistration::TypeAsRegistrationKey(_In_ HSTRING type)
```
RebeccaAnne: >TypeAsRegistrationKey [](start = 52, length = 21)

We've got this function in three places now, maybe make it a utility?

riarenas: I had it as an util and for some reason decided to make it a class member. I can't remember if I had a good reason for it, so that means I probably didn't.

source/uwp/Renderer/lib/AdaptiveActionRendererRegistration.h
RebeccaAnne: We spoke about this offline.  We can investigate further whether we can do something to reduce the duplicated code during the polish phase.

---
In reply to: [141201650](https://github.com/Microsoft/AdaptiveCards/pull/695#discussion_r141201650) [](ancestors = 141201650,141201602)

source/uwp/Renderer/lib/XamlCardRendererComponent.cpp
RebeccaAnne: I think you've got one more layer of ComPtr than you need.  This seems to work:
RETURN_IF_FAILED(m_elementRendererRegistration->Set(HStringReference(L"Input.ChoiceSet").Get(), (Make<AdaptiveChoiceSetInputRenderer>().Get())));

---
In reply to: [141197285](https://github.com/Microsoft/AdaptiveCards/pull/695#discussion_r141197285) [](ancestors = 141197285)

riarenas: I seem to recall it was complaining about Make Returning a ComPtr of the class rather than the interface. I'll give that another try. thanks for checking!

---
In reply to: [141431783](https://github.com/Microsoft/AdaptiveCards/pull/695#discussion_r141431783) [](ancestors = 141431783,141197285)

riarenas: Yay!. That worked just fine. Not sure what I was doing differently yesterday that made me go this route.

---
In reply to: [141432272](https://github.com/Microsoft/AdaptiveCards/pull/695#discussion_r141432272) [](ancestors = 141432272,141431783,141197285)

source/uwp/Renderer/lib/BaseRegistration.h
```diff
@@ -0,0 +1,54 @@
+#pragma once
```
RebeccaAnne: Looks like this file isn't added to the solution

RebeccaAnne: Looks like this didn't get used, so we should remove it from the PR

---
In reply to: [141433885](https://github.com/Microsoft/AdaptiveCards/pull/695#discussion_r141433885) [](ancestors = 141433885)

riarenas: It's not supposed to even be in the CR. Removing it :P This was one of my bad attempts to getting a templated registration to work.

---
In reply to: [141433885](https://github.com/Microsoft/AdaptiveCards/pull/695#discussion_r141433885) [](ancestors = 141433885)

riarenas:
Reran Android build manually. The failure was an intermittent infra issue.

PR #713, opened 2017-09-28
Remove Async render methods from public API surface

riarenas:


riarenas:
Pushing change so API is ready for beta.

PR #728, opened 2017-09-29
Update .NET nuget packages

anbare:
Main AdaptiveCards library definitely works

AdaptiveCards.Renderer.Html keeps on referencing Microsoft.AdaptiveCards instead of just AdaptiveCards so won't work

Not sure how to build the Wpf Nuget, but I updated the nuspec file to be correct, so if you build it somehow it should be right


matthidinger:
This will be fixed in an upcoming PR

PR #744, opened 2017-10-04
Enable use of Per-element renderers PT1.

riarenas:
This change keeps usage of xamlBuilder internally.
Also adds the RenderContext object that will be passed around.
Only CardElements use the new Registration entries. Actions require a bigger refactor.

### DIFF
--- a/source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl
+++ b/source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl
@@ -72,6 +72,7 @@ namespace AdaptiveCards
         runtimeclass AdaptiveOpenUrlActionRenderer;
         runtimeclass AdaptiveRenderContext;
         runtimeclass AdaptiveRenderArgs;
+        runtimeclass AdaptiveActionInvoker;
         runtimeclass AdaptiveShowCardActionRenderer;
         runtimeclass AdaptiveSubmitActionRenderer;
         runtimeclass AdaptiveTextBlockRenderer;
@@ -138,6 +139,7 @@ namespace AdaptiveCards
         interface IAdaptiveElementRendererRegistration;
         interface IAdaptiveRenderContext;
         interface IAdaptiveRenderArgs;
+        interface IAdaptiveActionInvoker;
 
         [version(NTDDI_WIN10_RS1)]
         typedef [v1_enum] enum TextSize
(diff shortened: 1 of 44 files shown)


source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl
```diff
@@ -1751,6 +1744,12 @@ namespace AdaptiveCards
 
             [propget] HRESULT ActionRenderers([out, retval] AdaptiveActionRendererRegistration** value);
 
+            [propget] HRESULT UserInputs([out, retval] AdaptiveInputs** value);
```
RebeccaAnne: >UserInputs [](start = 30, length = 10)

We don't need this because we pass the inputs in the action event args.

source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl
```diff
@@ -1751,6 +1744,12 @@ namespace AdaptiveCards
...
+
+            HRESULT AddInputItem([in] IAdaptiveCardElement* cardElement, [in] Windows.UI.Xaml.UIElement* uiElement);
+
+            HRESULT SendActionEvent([in] AdaptiveActionEventArgs* eventArgs);
```
RebeccaAnne: >SendActionEvent [](start = 20, length = 15)

We spoke about this offline.  This feels weird to me from an API design perspective because you're not going to actually call this method at render time.  Saving it for the click handler requires keeping the entire render context around for the click handler, at which point the other stuff in there doesn't make as much sense anymore.  Consider moving this method into a separate interface that you can get from the render context.

RebeccaAnne:
>             [propget] HRESULT Inputs([out, retval] Windows.Data.Json.JsonObject** value);

Consider making this the Inputs object rather than the json (either as part of this or your next change)

---
Refers to: source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl:897 in c78585b. [](commit_id = c78585bdf4ca9e47f68cddc1719c6b69598b1160, deletion_comment = False)

source/uwp/Renderer/lib/AdaptiveTextBlockRenderer.h
```diff
@@ -3,24 +3,33 @@
...
+#include "XamlBuilder.h"
 
-namespace AdaptiveCards { namespace Uwp
-{
-    class AdaptiveTextBlockRenderer :
-        public Microsoft::WRL::RuntimeClass<
-        Microsoft::WRL::RuntimeClassFlags<Microsoft::WRL::RuntimeClassType::WinRtClassicComMix>,
-        ABI::AdaptiveCards::Uwp::IAdaptiveElementRenderer>
+namespace AdaptiveCards {
+    namespace Uwp
```
RebeccaAnne: The other classes all mostly have this on one line, so probably makes sense to keep this one that way as well for consistency.

riarenas:
>             [propget] HRESULT Inputs([out, retval] Windows.Data.Json.JsonObject** value);

I will add this as part of my next change that deals with some actions issues.

---
In reply to: [334302062](https://github.com/Microsoft/AdaptiveCards/pull/744#issuecomment-334302062) [](ancestors = 334302062)

---
Refers to: source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl:897 in c78585b. [](commit_id = c78585bdf4ca9e47f68cddc1719c6b69598b1160, deletion_comment = False)

source/uwp/Renderer/lib/AdaptiveActionInvoker.cpp
```diff
@@ -0,0 +1,34 @@
...
+    {
+        m_renderResult = renderResult;
+        return S_OK;
+    } CATCH_RETURN;
+
+    _Use_decl_annotations_
+    HRESULT AdaptiveActionInvoker::SendActionEvent(IAdaptiveActionElement* actionElement)
+    {
+
+        return m_renderResult->SendActionEvent(actionElement);
```
RebeccaAnne: [nit] extra line

PR #745, opened 2017-10-04
GetAdditionalProperties added to shared model and uwp

RebeccaAnne:


source/uwp/Renderer/lib/AdaptiveOpenUrlAction.cpp
```diff
@@ -101,4 +101,12 @@ namespace AdaptiveCards { namespace Uwp
         RETURN_IF_FAILED(get_ActionType(&typeEnum));
         return ProjectedActionTypeToHString(typeEnum, type);
     }
+
+    _Use_decl_annotations_
+    HRESULT AdaptiveOpenUrlAction::GetAdditionalProperty(HSTRING propertyName, ABI::Windows::Data::Json::IJsonObject ** result)
```
riarenas: >IJsonObject ** [](start = 105, length = 14)

[nit] spacing

source/shared/cpp/ObjectModel/BaseActionElement.cpp
```diff
@@ -45,3 +48,15 @@ Json::Value BaseActionElement::SerializeToJsonValue()
     root[AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Id)] = GetId();
     return root;
 }
+
+void BaseActionElement::GetAdditionalProperty(std::string name, Json::Value & json)
```
riarenas: >void [](start = 0, length = 4)

Do you think it would make sense to make this a bool depending on whether the item was found?

riarenas: >Json::Value & [](start = 64, length = 13)

[nit] spacing

RebeccaAnne: I like returning the json value since that's consistent with what the winrt layer will look like.  I'll change to that.

---
In reply to: [142794093](https://github.com/Microsoft/AdaptiveCards/pull/745#discussion_r142794093) [](ancestors = 142794093)

source/shared/cpp/ObjectModel/BaseActionElement.cpp
```diff
@@ -45,3 +48,15 @@ Json::Value BaseActionElement::SerializeToJsonValue()
     root[AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Id)] = GetId();
     return root;
 }
+
+Json::Value BaseActionElement::GetAdditionalProperty(std::string name)
+{
+    if (m_additionalProperties.find(name) != m_additionalProperties.end())
```
riarenas: >m_additionalProperties.find(name) [](start = 8, length = 33)

Pretty irrelevant in the grand scheme of things, but you can create a local iterator variable, and return iterator-> second if found. This means there will only be one lookup to the map instead of two.

source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl
```diff
@@ -1440,14 +1440,15 @@ namespace AdaptiveCards
...
+            [propget] HRESULT ActionTypeString([out, retval] HSTRING* value);
 
             [propget] HRESULT Title([out, retval] HSTRING* value);
             [propput] HRESULT Title([in] HSTRING value);
 
             [propget] HRESULT Id([out, retval] HSTRING* value);
             [propput] HRESULT Id([in] HSTRING value);
 
-            [propget] HRESULT ActionTypeString([out, retval] HSTRING* value);
+            HRESULT GetAdditionalProperty([in] HSTRING propertyName, [out, retval] Windows.Data.Json.JsonObject** json);
```
riarenas: >GetAdditionalProperty [](start = 20, length = 21)

We discussed this offline. Instead of providing the GetAdditionalProperty where they need to specify the key they want to get, this can return a Json object that has all the additional properties. If all the caller cares about is passing the whole bag of additional properties somewhere else, they can do so, and if they just want to extract certain values, they can use the Windows.Data.Json apis to do it as well.

RebeccaAnne: I'm also going to make this a property now...

---
In reply to: [143250600](https://github.com/Microsoft/AdaptiveCards/pull/745#discussion_r143250600) [](ancestors = 143250600)

riarenas: Good catch

PR #747, opened 2017-10-05
UWP selectAction support on Container/ColumnSet

anbare:
Implemented selectAction (hit targets) on Container and ColumnSet for the UWP renderer. This implements part of #418

## How tested
Added a new test file `samples\v1.0\Elements\ColumnSet.SelectAction.json` and also updated `Container.SelectAction.json` and tested/verified both of these in the UWP Visualizer

image
image
image
image

## What was tested

- When selectAction provided on Container/ColumnSet, it becomes clickable
- When not provided, still functions normally
- When clicked on, invokes the action
- Action.OpenUrl works
- Action.Submit works
- Existing buttons (actions) work
- Emphasis behavior on columns/containers
- Keyboard accessibility
- High contrast (mostly seems good, we'll have to analyze high contrast in Action Center later)
- All main scenario payloads (ActivityUpdate, CalendarReminder, FlightItinerary, etc)

## Known not-implemented features

- In UWP object model, trying to programmatically set the SelectAction on Container or ColumnSet will throw not implemented exception... same issue there as constructing HostConfig programmatically... whoever solves that problem for HostConfig can apply the solution here too
- Vertical spacing of the hit target doesn't extend halfway between spacing of the item above and below, that's because we would have to know the spacing of the item below it, which would be very complicated
- Inline ShowCard is not supported (would be complicated), but it simply won't render a hit target if dev uses ShowCard on selectAction (and I tested that this behaves correctly)
- Didn't implement selectAction on columns themselves... will need designer input on how those hit targets should behave in terms of stretching horizontal width, etc
- Didn't implement selectAction on Image, that one should be easy/straightforward but there's no immediate requirements for it so not worth adding right now

## Implementation discussion

- In shared model, made ColumnSet parse the selectAction (it was only parsing it for Container previously)
- In UWP AdaptiveColumnSet and AdaptiveContainer, expose `get_SelectAction` and `put_SelectAction` (with the `put` throwing not implemented exception)
- In `Util.cpp`, genericize a `GenerateActionProjection` method since `AdaptiveColumnSet` and `AdaptiveContainer` need to perform the same logic as the Adaptive Card actions element was performing
- In `XamlBuilder.cpp`, change the background style applying on containers and column to only apply a background if the card author specifically assigned the style property on container/column. This is because if we always apply the background, when the user hovers over and the hit target appears, the background on the column becomes visible (whereas previously it was "invisible" since the background color was identical to its parent). This issue would also appear if the card author used a background image (but now it won't with my changes!)
- In `XamlBuilder.cpp`, add a `WrapInFullWidthTouchTarget` method that will take a UI element and wrap it in a button and wire up the click
- In `XamlBuilder.cpp`, add a `WireButtonClickToAction` method which takes a button and wires up the click handler so that it'll send the action event to hosts when button clicked. Right now only consumer of this method is the `WrapInFullWidthTouchTarget` but you can imagine in the future other methods might consume this (like when we figure out inline show card, we could potentially refactor the existing action button code to also use this method)
- When rendering container/column set, check if SelectAction is provided, and if so, use the `WrapInFullWidthTouchTarget` method
- Also made some updates to UWP Visualizer to allow seeing the Submit data and the OpenUrl url


source/uwp/Renderer/lib/Util.cpp
```diff
@@ -144,29 +144,45 @@ HRESULT GenerateActionsProjection(
...
     return S_OK;
 } CATCH_RETURN;
 
+HRESULT GenerateActionProjection(
+    const std::shared_ptr<AdaptiveCards::BaseActionElement> action,
+    ABI::AdaptiveCards::Uwp::IAdaptiveActionElement** projectedAction) noexcept try
+{
+    if (action == nullptr)
+    {
+        projectedAction = nullptr;
```
riarenas: >projectedAction [](start = 8, length = 15)

I believe what you want to do is *projectedAction = nullptr. That way whatever is inside the ComPtr is null, instead of the smart pointer object itself.

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1379,17 +1397,20 @@ namespace AdaptiveCards { namespace Uwp
...
             THROW_IF_FAILED(renderArgs->get_ContainerStyle(&parentContainerStyle));
             containerStyle = parentContainerStyle;
         }
         ComPtr<IAdaptiveRenderArgs> newRenderArgs;
         THROW_IF_FAILED(MakeAndInitialize<AdaptiveRenderArgs>(&newRenderArgs, containerStyle));
 
+        // If container style was explicitly assigned, apply background
         ABI::Windows::UI::Color backgroundColor;
-        if (SUCCEEDED(GetBackgroundColorFromStyle(containerStyle, m_hostConfig.Get(), &backgroundColor)))
+        if (hasExplicitContainerStyle && SUCCEEDED(GetBackgroundColorFromStyle(containerStyle, m_hostConfig.Get(), &backgroundColor)))
```
riarenas: >(hasExplicitContainerStyle && SUCCEEDED(GetBackgroundColorFromStyle(containerStyle, m_hostConfig.Get(), &backgroundColor))) [](start = 11, length = 123)

I like how this looks. Would you mind changing the one in BuildContainer to behave the same?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -2016,4 +2048,90 @@ namespace AdaptiveCards { namespace Uwp
...
+
+        WireButtonClickToAction(button.Get(), action, renderContext);
+
+        THROW_IF_FAILED(button.CopyTo(finalElement));
+    }
+
+    void XamlBuilder::WireButtonClickToAction(
+        IButton* button,
+        IAdaptiveActionElement* action,
+        IAdaptiveRenderContext* renderContext)
```
riarenas: Do you see a future where this would need the whole context rather than just the ActionInvoker?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -2016,4 +2048,90 @@ namespace AdaptiveCards { namespace Uwp
...
+        THROW_IF_FAILED(button.As(&buttonAsContentControl));
+        THROW_IF_FAILED(buttonAsContentControl->put_Content(elementToWrap));
+
+        ComPtr<IAdaptiveSpacingConfig> spacingConfig;
+        THROW_IF_FAILED(m_hostConfig->get_Spacing(&spacingConfig));
+
+        UINT32 cardPadding;
+        THROW_IF_FAILED(spacingConfig->get_Padding(&cardPadding));
+
+        double negativeCardMargin = cardPadding * -1.0;
```
riarenas: >double negativeCardMargin = cardPadding * -1.0; [](start = 8, length = 47)

[nit] Could you declare this right before usage?

source/uwp/Renderer/lib/XamlBuilder.cpp
anbare: The reason I had this method take the RenderContext instead of ActionInvoker is because otherwise all callers of this method have to first get the ActionInvoker, which would be 2 extra lines of redundant code assuming callers already have the RenderContext.

Right now the only caller of this method is the WrapInFullWidthTouchTarget, but in the future we can likely make the action (button) rendering code also use this method. And when we do that, the button rendering code similarly doesn't have to write those 2 redundant lines of code to obtain the ActionInvoker before calling this method.

riarenas: Fair enough. The click handler is only taking the ActionInvoker in the lambda, so I'm good with this!

anbare: Note also that I don't hold a reference to the Context. In the button click handler, I only pass along the actionInvoker, so there's no concern about a reference to the Context still being held after the method returns.

source/uwp/Renderer/lib/Util.cpp
anbare: Ah good catch. It actually works regardless since projectedAction is nullptr by default, but totally right that this line of code isn't assigning the right pointer :) Fixed and re-tested!

source/uwp/Renderer/lib/XamlBuilder.cpp
anbare: Fixed

anbare:
Also fixed an issue where the column sets were being indented by 2 pixels due to the button's border brush (it becomes quite obvious if you have an identical column set that does NOT have the selectAction, the left edge no longer is aligned).

Simple fix was to just negate the button's border that was being applied. Ideally we would be able to get the button's border thickness programmatically but when retrieving its thickness it was 0. I don't think XAML assigns the default border thickness till render time (since it pulls it from the default button style resources).

Now with my last commit, the visuals look identical regardless of whether selectAction is assigned or not (other than the container/columnset highlighting when you hover over it)

PR #754, opened 2017-10-06
UWP polish hit target visuals

anbare:
Designer feedback was that hit targets should be styled the same as list view items are. Plus the hit targets need to split the vertical spacing between rows evenly, in order to provide padding around items. Definitely looks better now :)

## Implementation details

- Created a custom `HitTarget` style for Button in `DefaultResourceDictionary.h` that styles the button similar to a ListViewItem
- Use the `spacing` on the current element to assume the same spacing on the next element and make the touch target expand its height vertically halfway between row items

(In future we should improve that behavior to actually figure out the next item's spacing, but this is good enough for now and it can be considered a bug that when different spacings are used, things get funky)

## How tested
Used `samples\v1.0\Elements\ColumnSet.SelectAction.json` and `Container.SelectAction.json` and tested/verified both of these in the UWP Visualizer, and also tested Notifications Visualizer scenarios, plus some other card payloads

## What was tested

- Keyboard accessibility
- High contrast
- When selectAction provided on Container/ColumnSet, it becomes clickable
- When not provided, still functions normally
- When clicked on, invokes the action
- Action.OpenUrl works
- Action.Submit works
- Existing buttons (actions) work
- Emphasis behavior on columns/containers
- All main scenario payloads (ActivityUpdate, CalendarReminder, FlightItinerary, etc)

## New visuals

image

image

image

### DIFF
--- a/source/uwp/Renderer/lib/DefaultResourceDictionary.h
+++ b/source/uwp/Renderer/lib/DefaultResourceDictionary.h
@@ -7,18 +7,45 @@ const PCWSTR c_defaultResourceDictionary = L"\
 \
     <ResourceDictionary.ThemeDictionaries> \
         <ResourceDictionary x:Key=\"Dark\"> \
-          <SolidColorBrush x:Key=\"TextColor.Accent\" Color=\"{ThemeResource SystemAccentColor}\"/> \
-          <SolidColorBrush x:Key=\"TextColor.Default\" Color=\"{ThemeResource SystemBaseHighColor}\"/> \
+            <SolidColorBrush x:Key=\"TextColor.Accent\" Color=\"{ThemeResource SystemAccentColor}\"/> \
+            <SolidColorBrush x:Key=\"TextColor.Default\" Color=\"{ThemeResource SystemBaseHighColor}\"/> \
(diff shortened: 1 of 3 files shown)


source/uwp/Renderer/lib/DefaultResourceDictionary.h
```diff
@@ -56,4 +56,62 @@ const PCWSTR c_defaultResourceDictionary = L"\
     <Style TargetType=\"Grid\" x:Key=\"ColumnSet\"> \
         <Setter Property=\"Margin\" Value=\"0,0,0,0\"/> \
     </Style> \
+\
+    <Style x:Key=\"HitTarget\" TargetType=\"Button\"> \
```
riarenas: >HitTarget [](start = 19, length = 9)

I wonder if we should rename this to SelectAction. In theory different hosts will want hit targets to be styled differently. We want to enable styling via resource dictionary, so we can document that to change the SelectAction behavior, they should use this key.

Alternatively for now we can ship with this, and require hosts to provide a custom renderer for all elements that support SelectAction if they want to change how it looks.

PR #757, opened 2017-10-06
Decouple XamlCardRenderer from XamlBuilder.

riarenas:
Next step in retiring the XamlBuilder class.

XamlBuilder no longer has any dependencies on the Renderer component. This enables the scenario where hosts want to modify an existing element renderer but only want to apply minor tweaks to the resulting xaml UIElement we return from the out of the box renderers.

Creating this Custom Element renderer:
```csharp
    class CustomImageRenderer : IAdaptiveElementRenderer
    {
        public UIElement Render(IAdaptiveCardElement element, AdaptiveRenderContext context, AdaptiveRenderArgs renderArgs)
        {
            AdaptiveImageRenderer imageRenderer = new AdaptiveImageRenderer();
            UIElement imageUIElement = imageRenderer.Render(element, context, renderArgs);
            Image image = imageUIElement as Image;
            if (image != null)
            {
                image.Opacity = 0.5;
            }

            return imageUIElement;
        }
    }
```
And assigining it to the AdaptiveCardRenderer
```csharp
_renderer = new AdaptiveCardRenderer();
_renderer.ElementRenderers.Set("Image", new CustomImageRenderer());

```
results in:
image

### DIFF
--- a/source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl
+++ b/source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl
@@ -1744,6 +1744,8 @@ namespace AdaptiveCards
 
             [propget] HRESULT ElementRenderers([out, retval] AdaptiveElementRendererRegistration** value);
 
+            [propget] HRESULT ResourceResolvers([out, retval] AdaptiveCardResourceResolvers** value);
+
             [propget] HRESULT ActionInvoker([out, retval] AdaptiveActionInvoker** value);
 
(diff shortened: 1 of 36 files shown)


anbare:
FYI I think this no longer compiles, I pulled these changes to test them for regressions with the new test app, and I think Becky's recent commit takes a dependency on m_hostConfig which no longer exists so the renderer no longer compiles

anbare:
Other than that fix (changing `m_hostConfig` in `XamlBuilder.cpp` for `BuildContainer` to just be `hostConfig`), all tests passed in the test app! No regressions!

riarenas:
Yep, this change took a bit to be reviewed due to its size. I was planning on getting the latest master on top of it before merging.

PR #759, opened 2017-10-09
Add Parser Registration to the shared model

RebeccaAnne:
This version passes around the parser registration rather than having it be static.


RebeccaAnne:
Pulling for now as we revisit the design.

PR #762, opened 2017-10-10
Add tapped event handler to toggle and choiceset inputs

### LINKED ISSUE #758: UWP Bug: Taps on combo boxes/check boxes propagate down to parent containers
The way XAML works is that apparently any tap events on `CheckBox` or `ComboBox` propagate down to parent controls, even though the check box is being checked or the combo box is showing the picker.

And therefore if you're displaying your card inside a button (since you want the whole card to be clickable), the parent button gets clicked when the user simply checks a check box or combo box.

In Action Center's existing code where they used combo box in toasts, they had to wire up the `Tapped` event on their `ComboBox` and set `e.Handled = true` in order to prevent the tap from propagating down further.

**Therefore fixing this in Adaptive Cards requires simply wiring up the `Tapped` event on...**

- `ComboBox`
- `CheckBox`
- `RadioButton`

and in the event handler, setting `e.Handled = true`.

Note that `TextBox` does NOT need this Tapped event wired up. TextBox correctly captures and doesn't propagate the tap event.

riarenas:
Fixes #758

### DIFF
--- a/source/uwp/Renderer/lib/XamlBuilder.cpp
+++ b/source/uwp/Renderer/lib/XamlBuilder.cpp
@@ -1760,7 +1760,12 @@ namespace AdaptiveCards { namespace Uwp
         THROW_IF_FAILED(comboBox.As(&selector));
         THROW_IF_FAILED(selector->put_SelectedIndex(selectedIndex));
 
-        THROW_IF_FAILED(comboBox.CopyTo(choiceInputSet));
+        ComPtr<IUIElement> comboBoxAsUIElement;
+        THROW_IF_FAILED(comboBox.As(&comboBoxAsUIElement));
+
+        THROW_IF_FAILED(AddHandledTappedEvent(comboBoxAsUIElement.Get()));
+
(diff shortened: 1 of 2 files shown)


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -2038,8 +2045,13 @@ namespace AdaptiveCards { namespace Uwp
...
 
-        THROW_IF_FAILED(checkBox.CopyTo(toggleInputControl));
-        renderContext->AddInputItem(adaptiveCardElement, *toggleInputControl);
+        ComPtr<IUIElement> checkboxAsUIElement;
+        THROW_IF_FAILED(checkBox.As(&checkboxAsUIElement));
+
+        THROW_IF_FAILED(AddHandledTappedEvent(checkboxAsUIElement.Get()));
+        THROW_IF_FAILED(renderContext->AddInputItem(adaptiveCardElement, checkboxAsUIElement.Get()));
+        THROW_IF_FAILED(checkboxAsUIElement.CopyTo(toggleInputControl));
+
```
riarenas: Removed blank line

PR #778, opened 2017-10-13
Implement putters for host config object model in UWP

RebeccaAnne:


anbare:
Awesome!! Would it be easy for you to also implement the putter for `AdaptiveContainer::put_SelectAction` and `AdaptiveColumnSet::put_SelectAction`?

These had a similar issue of needing to update a single instance of the shared model just like host config, so I left those as E_NOTIMPL for the time being.

source/uwp/Renderer/lib/AdaptiveActionsConfig.h
```diff
@@ -36,7 +36,13 @@ namespace AdaptiveCards { namespace Uwp
         IFACEMETHODIMP put_Spacing(ABI::AdaptiveCards::Uwp::Spacing value);
 
     private:
-        AdaptiveCards::ActionsConfig m_sharedActionsConfig;
+        ABI::AdaptiveCards::Uwp::ActionAlignment m_actionAlignment;
+        ABI::AdaptiveCards::Uwp::ActionsOrientation m_actionsOrientation;
+        UINT m_buttonSpacing;
+        UINT m_maxActions;
```
anbare: >UINT m_maxActions; [](start = 8, length = 18)

MaxActions and all these other properties aren't going to have the correct default values when you initialize these classes using the parameterless constructor, right?

That would mean hosts using the programmatic way of constructing HostConfig MUST specify every single property? Otherwise they're going to end up with values like a max of 0 actions?

I think the correct default values should be assigned when you new up each config object. So maybe that'd involve newing up a shared config object in the parameterless constructor and then calling the other constructor with that new shared object?

RebeccaAnne: Good point. I'll fix this.

---
In reply to: [145463181](https://github.com/Microsoft/AdaptiveCards/pull/778#discussion_r145463181) [](ancestors = 145463181)

RebeccaAnne:
I think that should probably be a separate change, since i want to think about whether i want to follow the same pattern for the element types as i did here.  Do we have an issue tracking?

---
In reply to: [337315291](https://github.com/Microsoft/AdaptiveCards/pull/778#issuecomment-337315291) [](ancestors = 337315291)

source/uwp/Renderer/lib/AdaptiveActionsConfig.cpp
```diff
@@ -11,94 +11,105 @@ namespace AdaptiveCards { namespace Uwp
...
         return S_OK;
     }
 
     _Use_decl_annotations_
     HRESULT AdaptiveActionsConfig::get_ShowCard(IAdaptiveShowCardActionConfig** showCardActionConfig)
     {
-        return MakeAndInitialize<AdaptiveShowCardActionConfig>(showCardActionConfig, m_sharedActionsConfig.showCard);
+        ComPtr<IAdaptiveShowCardActionConfig> localShowCardActionConfig = m_showCardActionConfig;
+        *showCardActionConfig = localShowCardActionConfig.Detach();
+        return S_OK;
```
riarenas: We have two patterns throughout the codebase for this:

I believe we should use try to standardize into using CopyTo:

return m_showCardActionConfig.CopyTo(showCardActionConfig)

thoughts?

PR #788, opened 2017-10-17
UWP renderer test app

anbare:
Adding a UWP renderer test app that renders all card payloads to images and then compares them for regressions with previous versions.

This adds a TON of new files since we need to store an image for each card payload multiplied by each host config (we render each card with each host config). And with each result we also store a JSON payload. **But this only adds 7.5 MB to the repo, and it's a very valuable addition, so I think it's worth it**. Note that I tried using JpegXr for higher compressed exported images but it made no size difference compared to PNG.

We also store the original card and host config payload so that we can show file comparisons to the user when the test fails, but these are only stored 1-to-1, so there's no multiplication factor there.

**This pull request modifies a few of the sample host configs and sample card payloads** to fix some issues that those payloads had. I also had to change the images in some of the payloads to point to picsum, for higher resolution images, otherwise the image comparison tests always kept failing.

I had to omit the WeatherLarge payload from tests since the background image always throws off the image comparison tests.

Other than those edits to sample host config/card payloads (and the UWP solution file for adding the new project), this pull request doesn't touch anything else.

**I've also added a README on how to use the test app**


anbare:
`Error CS0117: 'ContentDialog' does not contain a definition for 'CloseButtonText'`

Aw sad, I don't think SDK 14393 has that property on ContentDialog apparently. I'll make a change on Tuesday for this

anbare:
Ugh I just can't get any luck with this build server!

anbare:
OMG finally think I might have figured it out, in the solution file it accidentally updated the Visual Studio version to... `VisualStudioVersion = 15.0.26730.16`

Changing it back to `VisualStudioVersion = 15.0.26403.7` makes it build. Now to re-add the test project and hope it continues to build...

anbare:
*(phew)* finally got it building on the CI server. @riarenas @RebeccaAnne it's all good for review now

anbare:
@RebeccaAnne let's get this reviewed and merged today?

PR #789, opened 2017-10-18
Trying to see if a build of unchanged master succeeds

anbare:
I'm stumped as to why the build of the test app is failing, and it continues to fail even after I removed the app from the solution... so wondering if master even builds now


anbare:
The UWP build worked, so the master config is fine

Issue #779, opened 2017-10-14
Multi-line TextBlocks should show ellipses when truncated


tongyliu:
Multi-line TextBlocks should match the behavior of single line ones when it comes to truncation -- currently with the TypeScript renderer, no ellipsis is shown if multi-line text overflows. This will be needed for Bing API news cards to display properly.

**Current:**
![image](https://user-images.githubusercontent.com/3249032/31569542-668398ea-b030-11e7-92c8-e48751e3a3a4.png)

**Desired:**
![capture](https://user-images.githubusercontent.com/3249032/31569903-b5edacf6-b033-11e7-82af-730c6bcdae58.PNG)

This is difficult/impossible to do with pure CSS (`text-overflow: ellipsis` only works for single lines, and `-webkit-line-clamp` isn't cross browser), so we might need to pre-truncate the text in JS. Will code up a prototype early next week.

anbare:
@tongyliu are you making ellipses the default behavior in the nodejs renderer?

If so, we need to create an issue to track applying that ellipses behavior to ALL the renderers, as all renderers (iOS, Windows, Android) should behave the same

tongyliu:
@anbare yep, this will make ellipses the default behavior for truncation there.

The renderer guidelines [here](https://docs.microsoft.com/en-us/adaptive-cards/display/implementingrenderer) say that TextBlocks "**SHOULD** trim any excess text with an ellipsis," so I'd assumed that the other renderers already supported this behavior and I was just dealing with a limitation of the nodejs renderer. If that's not the case though, I agree we should have an issue to track this behavior across all official renderers.

anbare:
Awesome thx I created an uber issue to track all the renderers, I don't believe UWP does this today and we would have to confirm the other renderers do too

tongyliu:
Closing as fixed.

khouzam:
Reopenning as iOS is not respecting maxLines.

dclaux:
@khouzam why did you reopen this?

dclaux:
Sorry, I posted my question exactly at the same time as I got your response.

khouzam:
After investigations, maxLines are respected but the ellipsis don't show on iOS. Will need to fix.

PR #791, opened 2017-10-18
Enable element render callbacks & truncate multi-line TextBlocks

tongyliu:
Updates the TypeScript renderer so that multi-line `TextBlock` elements get truncated and show ellipses, instead of overflowing with no indication that there is more content. Since there's no good CSS solution for truncating multi-line text, this is accomplished with a callback function that executes after the element gets rendered (can't do it before because height and width info isn't accessible then).

This approach includes the following changes:

- Adds support for registering callbacks that get called when an element is rendered

- Changes the signature of the `AdaptiveCard.render()` function to take an `HTMLElement` argument, which the card is then rendered into (as opposed to just returning an `HTMLElement` containing the card -- we need at least some control over when the card is rendered into the DOM so we can fire the `onRender` callbacks)

    - This matches the way it's done in React with the `ReactDOM.render()` method, and abstracts the `element.appendChild` call away from users

    - No rush to merge this PR if the spec for the v1.0 SDK is already locked down; Bing API cards can work off the fork if needed
  
- Renames the old `render()` methods that would return an HTMLElement to `renderElement()`

- Updates the visualizer containers to match the new signature, and ensures that cards always get rendered into an element which is part of the DOM

This addresses #779.

### DIFF
--- a/source/nodejs/adaptivecards-visualizer/src/app.ts
+++ b/source/nodejs/adaptivecards-visualizer/src/app.ts
@@ -34,7 +34,7 @@ function setContent(element) {
     contentContainer.appendChild(element);
 }
 
-function renderCard(): HTMLElement {
+function renderCard(target: HTMLElement) {
     document.getElementById("errorContainer").hidden = true;
     lastValidationErrors = [];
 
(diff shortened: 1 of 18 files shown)


dclaux:
@tongyliu can you please setup some time with me to discuss these changes. The direction you are taking isn't one that I feel comfortable with. I need to better understand what you are trying to accomplish so we can decide what the best approach is.

dclaux:
This approach isn't generic enough IMO. The renderer supports the ability to dynamically toggle the visibility of any element in a rendered card; as done now, text truncation will only happen once as the card is rendered. If a previously invisible element becomes visible subsequently, a text block can still overflow. Similarly, a previously truncated text block might not need to be truncated anymore once a particular element of the card is made invisible. The right solution should be one that handles these case gracefully.

The existing updateLayout method exists for this purpose. I think TextBlock should simply override updateLayout and implement its truncation logic in there. CardElement.render can be updated to take an optional target: HTMLElement parameter; when that parameter is passed, the renderered element would be first inserted into the target element before updateLayout is called.

tongyliu:
See #997 for an updated approach. Closing this one.

PR #799, opened 2017-10-19
Ios custom renderer

nesalang:
Allows custom actions renderer


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOBaseActionElementPrivate.h
```diff
@@ -0,0 +1,17 @@
...
+//  Copyright © 2017 Microsoft. All rights reserved.
+//
+#import "ACOBaseActionElement.h"
+#import "BaseActionElement.h"
+
+using namespace AdaptiveCards;
+
+@interface ACOBaseActionElement()
+
+- (std::shared_ptr<BaseActionElement> const &)getElem;
```
bviglietta: > & [](start = 43, length = 2)

This should return by value rather than by reference. If you return by reference, the shared_ptr becomes invalid if the ACOBaseActionElement is destroyed.

source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ViewController.m
```diff
@@ -157,6 +158,9 @@ - (void)update:(NSString *) jsonStr
     
     if(renderResult.succeeded)
     {
+        ACRRegistration *registration = [ACRRegistration getInstance];
+        // enum will be part of API in next iterations when custom renderer extended to non-action type
```
bviglietta: >enum will be part of API in next iterations when custom renderer extended to non-action type [](start = 11, length = 92)

It would be good to include an issue # here to track the work to be done. TODOs like this can easily get old and stale if we don't keep track of them.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOBaseActionElementPrivate.h
nesalang: Thank you!

---
In reply to: [146323478](https://github.com/Microsoft/AdaptiveCards/pull/799#discussion_r146323478) [](ancestors = 146323478)

PR #810, opened 2017-10-23
Custom parsers

RebeccaAnne:
Custom parsers in the shared model and uwp.  Android updates and exposing our existing parsers from the uwp interface still to come.


source/shared/cpp/ObjectModel/ActionParserRegistration.h
```diff
@@ -0,0 +1,36 @@
...
+namespace AdaptiveCards
+{
+    class BaseActionElement;
+    class ElementParserRegistration;
+    class ActionParserRegistration;
+
+    class IActionElementParser
+    {
+    public:
+        virtual std::shared_ptr<BaseActionElement> Deserialize(
```
riarenas: >Deserialize [](start = 51, length = 11)

Discussed offline. Let's try to also add a version of these that don't require you to pass the registrations

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -82,51 +84,54 @@ HRESULT GenerateContainedElementsProjection(
...
-                std::static_pointer_cast<AdaptiveCards::TimeInput>(containedElement)));
+                std::dynamic_pointer_cast<AdaptiveCards::TimeInput>(containedElement)));
             break;
         case CardElementType::ToggleInput:
             RETURN_IF_FAILED(MakeAndInitialize<::AdaptiveCards::Uwp::AdaptiveToggleInput>(&projectedContainedElement,
-                std::static_pointer_cast<AdaptiveCards::ToggleInput>(containedElement)));
+                std::dynamic_pointer_cast<AdaptiveCards::ToggleInput>(containedElement)));
+            break;
+        case CardElementType::Custom:
+            RETURN_IF_FAILED(std::dynamic_pointer_cast<CustomElementWrapper> (containedElement)->GetWrappedElement(&projectedContainedElement));
```
riarenas: >> ( [](start = 75, length = 3)

[nit] extra space

source/uwp/Renderer/lib/Util.h
```diff
@@ -93,3 +93,26 @@ HRESULT ProjectedElementTypeToHString(ABI::AdaptiveCards::Uwp::ElementType proje
...
+};
+
+template<typename T, typename R> Microsoft::WRL::ComPtr<T> PeekInnards(R r)
+{
+    Microsoft::WRL::ComPtr<T> inner;
+    Microsoft::WRL::ComPtr<ITypePeek> peeker;
+
+    if (r && SUCCEEDED(r->QueryInterface(__uuidof(ITypePeek), &peeker)))
+    {
+        inner = reinterpret_cast<T *>(peeker->PeekAt(__uuidof(T)));
```
riarenas: >  [](start = 34, length = 1)

[nit] extra space.

source/uwp/Renderer/lib/AdaptiveCard.cpp
```diff
@@ -20,29 +22,73 @@ namespace AdaptiveCards { namespace Uwp
...
+        const std::string jsonString, 
+        IAdaptiveElementParserRegistration* elementParserRegistration,
+        IAdaptiveActionParserRegistration* actionParserRegistration, 
+        IAdaptiveCardParseResult** parseResult)
     {
-        std::shared_ptr<::AdaptiveCards::AdaptiveCard> sharedAdaptiveCard = ::AdaptiveCards::AdaptiveCard::DeserializeFromString(jsonString);
+        std::shared_ptr<ElementParserRegistration> sharedModelElementParserRegistration;
+        if (elementParserRegistration != nullptr)
+        {
+            ComPtr<AdaptiveElementParserRegistration> elementParserRegistrationImpl = PeekInnards<AdaptiveElementParserRegistration>(elementParserRegistration);
```
riarenas: >PeekInnards [](start = 86, length = 11)

useful!

source/shared/cpp/ObjectModel/ActionParserRegistration.h
RebeccaAnne: Discussed offline some more.  This interface defines the element parsers.  We don't want custom parsers to need two versions, and the only people likely to need our element parsers are those implementing custom parsers (who will have the registration handy).  The top level deserialize method takes the registration as an optional parameter.  Leaving these as is.

---
In reply to: [146913501](https://github.com/Microsoft/AdaptiveCards/pull/810#discussion_r146913501) [](ancestors = 146913501)

source/uwp/Renderer/lib/CustomElementWrapper.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    THROW_IF_FAILED(cardElement->put_Spacing(static_cast<ABI::AdaptiveCards::Uwp::Spacing>(value)));
+}
+
+std::string CustomElementWrapper::GetId() const
+{
+    ComPtr<IAdaptiveCardElement> cardElement;
+    m_cardElement.As(&cardElement);
+
+    Wrappers::HString id;
```
riarenas: >Wrappers::HString id;  [](start = 4, length = 22)

Is there a reason to use this over an HSTRING?
I ask because most of our other code uses HSTRINGs directly.

RebeccaAnne: HString will free the memory.  I will log an issue to follow up on other places we should be using it.

---
In reply to: [146940126](https://github.com/Microsoft/AdaptiveCards/pull/810#discussion_r146940126) [](ancestors = 146940126)

RebeccaAnne: #853

---
In reply to: [146953069](https://github.com/Microsoft/AdaptiveCards/pull/810#discussion_r146953069) [](ancestors = 146953069,146940126)

PR #816, opened 2017-10-24
.NET fixes

matthidinger:


matthidinger:
@anbare can you review this? I added tests for most changes. Ideally we can start reviewing the .NET stuff together

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
```diff
@@ -685,7 +688,7 @@ protected static HtmlTag ChoiceSetRender(AdaptiveChoiceSetInput adaptiveChoiceSe
                         .Attr("name", adaptiveChoiceSetInput.Id)
                         .AddClass("ac-input")
                         .AddClass("ac-multichoiceInput")
-                        .Style("width", "100%");
+                        .Style("width", "100%");
```
anbare: Did you accidently add a trailing space here? That's what GitHub seems to indicate

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
```diff
@@ -710,7 +713,11 @@ protected static HtmlTag ChoiceSetRender(AdaptiveChoiceSetInput adaptiveChoiceSe
 
                     foreach (var choice in adaptiveChoiceSetInput.Choices)
                     {
+                        // generate a random ID to link the <label> and <input> elements together
+                        var htmlLabelId = Guid.NewGuid().ToString().Substring(0, 6);
```
anbare: Please add a TODO comment and create a GitHub issue to track improving the ID generation. This has a few concerns,..

- Two inputs could get the same ID, yes it's unlikely but I believe by only using 6 characters of the GUID it could happen, so we would want to store the IDs and if we conflict generate a new one
- We could conflict with IDs in the parent DOM, since this is being inserted into one giant DOM

It's not a high priority concern but it's worthy of a TODO and an issue to track it

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
```diff
@@ -750,7 +760,11 @@ protected static HtmlTag ChoiceSetRender(AdaptiveChoiceSetInput adaptiveChoiceSe
 
                 foreach (var choice in adaptiveChoiceSetInput.Choices)
                 {
+                    // generate a random ID to link the <label> and <input> elements together
+                    var htmlLabelId = Guid.NewGuid().ToString().Substring(0, 6);
```
anbare: TODO here too

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
```diff
@@ -903,11 +919,15 @@ protected static HtmlTag TimeInputRender(AdaptiveTimeInput input, AdaptiveRender
 
         protected static HtmlTag ToggleInputRender(AdaptiveToggleInput toggleInput, AdaptiveRendererContext context)
         {
+            // generate a random ID to link the <label> and <input> elements together
+            var htmlLabelId = Guid.NewGuid().ToString().Substring(0, 6);
```
anbare: TODO here too

source/dotnet/Library/AdaptiveCards/AdaptiveCard.cs
```diff
@@ -93,14 +121,47 @@ public static AdaptiveCardParseResult FromJson(string json)
...
+            var card = (AdaptiveCard) Activator.CreateInstance(objectType);
+            // Get the version from the payload
+            card.Version = null;
+
+            serializer.Populate(jObject.CreateReader(), card);
+
+            if (card.Version == null)
+                return null;
+
+            card.Actions.RemoveAll(item => item == null);
```
anbare: Add a note here saying that unknown types come in as null? The only reason I vaguely know this is the case is because I overheard you saying that it parses unknown types as null or something... otherwise this code seems very unlogical and unexplained :)

anbare: Also do we need the same logic when parsing containers/column sets? Do you have code there that does this logic? Or should we iterate through all Body descendants here and remove all null, not just for first-level children?

source/dotnet/Test/AdaptiveCards.Test/SerializationTests.cs
```diff
@@ -48,6 +48,65 @@ public void TestCardsSerializeInTheCorrectOrder()
...
+      ""text"": ""Hello""
+    },
+    {
+      ""text"": ""Hello""
+    }
+  ]
+}";
+
+            var result = AdaptiveCard.FromJson(json);
+            Assert.AreEqual(1, result.Card.Body.Count);
```
anbare: Aren't we supposed to fail here? I thought `type` is a required property on all elements, and therefore the schema is invalid (not just warning but invalid)

matthidinger: Yea I think that's actually correct. I'll update the test and add this to our sample test scenarios

matthidinger:
Pushed updates

source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            serializer.Populate(jObject.CreateReader(), card);
+
+            // TODO: return parse result error that version must be specified
+            if (card.Version == null)
+                return null;
+
+            // When objects without a valid "type" are parsed they return null but still get added to the collections, this removes them
+            card.Actions.RemoveAll(item => item == null);
+            card.Body.RemoveAll(item => item == null);
```
anbare: Don't we need this logic on Container and Column too? So they would also need a custom converter?

anbare:
Looks all good except the last comment about needing to remove null children from Container and Column too? We need to do that right?

anbare:
Approving changes and will file issue for unknown children types within `Container` and `Column` being returned as null in the children lists

PR #854, opened 2017-10-25
Expose known element parsers from the uwp interface

RebeccaAnne:


source/uwp/Renderer/lib/AdaptiveElementParserRegistration.h
```diff
@@ -60,4 +62,39 @@ namespace AdaptiveCards { namespace Uwp
...
+            ABI::AdaptiveCards::Uwp::IAdaptiveElementParserRegistration* elementParserRegistration,
+            ABI::AdaptiveCards::Uwp::IAdaptiveActionParserRegistration* actionParserRegistration,
+            ABI::AdaptiveCards::Uwp::IAdaptiveCardElement** element)
+    {
+        std::string jsonString;
+        JsonObjectToString(jsonObject, jsonString);
+
+        std::shared_ptr<AdaptiveCards::ElementParserRegistration> sharedModelElementParserRegistration;
+        ComPtr<AdaptiveElementParserRegistration> elementParserRegistrationImpl = PeekInnards<AdaptiveElementParserRegistration>(elementParserRegistration);
+        if (elementParserRegistrationImpl != nullptr)
```
riarenas: >nullptr [](start = 45, length = 7)

What happens if either registration ends up as null? I don't believe the shared model parsers' DeserializeFromString does null checks on the registrations.

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.h
```diff
@@ -22,6 +23,13 @@ namespace AdaptiveCards { namespace Uwp
             _In_ ABI::AdaptiveCards::Uwp::IAdaptiveRenderContext* renderContext,
             _In_ ABI::AdaptiveCards::Uwp::IAdaptiveRenderArgs* renderArgs,
             _COM_Outptr_ ABI::Windows::UI::Xaml::IUIElement** result);
+
+        IFACEMETHODIMP FromJson(
+            ABI::Windows::Data::Json::IJsonObject *,
+            ABI::AdaptiveCards::Uwp::IAdaptiveElementParserRegistration* elementParsers,
+            ABI::AdaptiveCards::Uwp::IAdaptiveActionParserRegistration* actionParsers,
+            ABI::AdaptiveCards::Uwp::IAdaptiveCardElement **);
```
riarenas: [nit] spacing around *

source/uwp/Renderer/lib/AdaptiveElementParserRegistration.h
RebeccaAnne: Updated this code to create one if there isn't one, although i toyed with the idea of just returning invalid arg in this case, since we expect this to be called from a custom parser where we will have just passed them the registrations.

---
In reply to: [147445256](https://github.com/Microsoft/AdaptiveCards/pull/854#discussion_r147445256) [](ancestors = 147445256)

PR #863, opened 2017-10-27
Expose element serialization from the UWP layer.

RebeccaAnne:


source/android/build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.3'
+        classpath 'com.android.tools.build:gradle:3.0.0'
```
riarenas: >classpath 'com.android.tools.build:gradle:3.0.0' [](start = 8, length = 48)

Was this added by an Android studio update?

RebeccaAnne: Yes.  I've reverted them for now as it seems our build machine isn't ready for that version yet.

---
In reply to: [147764367](https://github.com/Microsoft/AdaptiveCards/pull/863#discussion_r147764367) [](ancestors = 147764367)

Issue #517, opened 2017-07-26
Fixed height cards should clip partial content


matthidinger:
# Implementation status
- [x] TypeScript
- [x] UWP
- [x] WPF
- [x] iOS
- [x] Android
- [ ] Documentation

# The problem
Currently fixed height cards like Timeline and Tiles will show partial text at the bottom of the card, which isn't desirable

![image](https://user-images.githubusercontent.com/1432195/28630360-f1cfdd90-71de-11e7-8738-3d0014a1fa3e.png)

dclaux:
What is the desired behavior? Is the goal to preserve the bottom padding at all cost (even at the cost of cutting a line of text in the middle horizontally) or is it that the text block in the above example is trimmed to three lines (probably even 2 actually)?

anbare:
I believe the desired behavior would be that the text block in the example becomes trimmed to 3 lines (or possibly 2 lines). That's how adaptive XML tiles function today.

So no text randomly cut off, and padding of at least the host config's padding amount is preserved (potentially even more padding would be visibly present depending on font size and where the line breaks end up)

dclaux:
Very hard to do. No automatic/CSS-based way to do it. I would have to somehow compute the remaining space, determine that this particular text block is the element that overflows and calculate/set its maxlines property. Unless someone else wants to do it, this will be a limitation of the HTML renderer.

matthidinger:
Ropening for tracking. It has to be done in uwp

khouzam:
This has been implemented

PR #866, opened 2017-10-31
Use WholeItemsPanel instead of StackPanels to avoid clipping in fixed height cards.

riarenas:
UWP implementation for #517


source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl
```diff
@@ -1940,5 +1944,24 @@ namespace AdaptiveCards
             [default] interface IAdaptiveRenderArgs;
         };
 
+        [uuid(cced6652-8ca5-4841-b650-851770ec6ecf)]
+        [exclusiveto(WholeItemsPanel)]
+        [version(NTDDI_WIN10_RS1)]
+        interface IWholeItemsPanel : IInspectable
```
RebeccaAnne: >IWholeItemsPanel [](start = 18, length = 16)

Does this need to be exposed from our API?  It seems a little out of place in the AdaptiveCards.Uwp namespace...

riarenas: I'll see if it can be removed

---
In reply to: [148412119](https://github.com/Microsoft/AdaptiveCards/pull/866#discussion_r148412119) [](ancestors = 148412119)

Issue #841, opened 2017-10-25
Spacing/Separator aren't working in WPF, not sure about HTML


matthidinger:
Currently these attributes are not being properly renderered in WPF

matthidinger:
Fixed with #870

Issue #839, opened 2017-10-25
.NET Renderer currently registers each individual Action subtype vs the base Action


matthidinger:
The AdaptiveCardRenderer (HTML and WPF) should render based on Action base type, not per Action name as designed today. Need to:

- [ ] make sure the Native Platform styling works still 
- [ ] update [docs](https://docs.microsoft.com/en-us/adaptive-cards/display/libraries/wpf) with new style names

```charp
           ElementRenderers.Set<AdaptiveSubmitAction>(XamlSubmitAction.Render);
            ElementRenderers.Set<AdaptiveOpenUrlAction>(XamlOpenUrlAction.Render);
            ElementRenderers.Set<AdaptiveShowCardAction>(XamlShowCardAction.Render);
```

matthidinger:
This is fixed with #870

Issue #837, opened 2017-10-25
AdaptiveTypedElement missing Id property


matthidinger:
Re #781 

Id is an optional property in the schema and is missing in the .NET model. We should consider if we should auto-generate an "intelligent ID" during parsing, or pre-render if one isn't provided in the card. Based

matthidinger:
Fixed in #870

PR #870, opened 2017-11-01
.NET deserialization fixes and WPF improvements

### LINKED ISSUE #850: .NET WPF e.Data is on the AdaptiveActionEventArgs instead of Action.Submit
MainWindow.xaml.cs

```csharp
     else if (e.Action is AdaptiveSubmitAction submitAction)
            {
                // TODO: Shouldn't e.Data be on the SubmitAction?
                MessageBox.Show(this, JsonConvert.SerializeObject(e.Data, Formatting.Indented), "SubmitAction");
            }
```

matthidinger:
This fixes #850 #841 #837 #839


anbare:
>         public string Title { get; set; }

This isn't new, but what is this Title property on the top level Card?

---
Refers to: source/dotnet/Library/AdaptiveCards/AdaptiveCard.cs:94 in f89b75a. [](commit_id = f89b75a4698b012d728e695c3e11fbeb1ffe3572, deletion_comment = False)

source/dotnet/Library/AdaptiveCards/AdaptiveColumn.cs
```diff
@@ -2,6 +2,7 @@
 using Newtonsoft.Json;
 
 using System;
+using System.Reflection;
```
anbare: Unnecessary addition? This was the only added line here in this file

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElement.cs
```diff
@@ -4,6 +4,12 @@
 
 namespace AdaptiveCards
 {
+    //[Serializable]
+    public delegate void TypedEventHandler<in TSender, in TEventArgs>(
```
anbare: >TypedEventHandler [](start = 25, length = 17)

Not sure what the purpose of this is, but does it have chances for conflict with Windows.Foundation.TypedEventHandler? [https://docs.microsoft.com/en-us/uwp/api/windows.foundation.typedeventhandler](https://docs.microsoft.com/en-us/uwp/api/windows.foundation.typedeventhandler)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/RenderedAdaptiveCard.cs
```diff
@@ -16,15 +18,35 @@ public RenderedAdaptiveCard(FrameworkElement frameworkElement, AdaptiveCard orig
...
         internal void InvokeOnAction(AdaptiveActionEventArgs args)
         {
             OnAction?.Invoke(this, args);
         }
+
+        /// <summary>
+        /// Gather the current values from any inputs on the card
+        /// </summary>
+        /// <returns></returns>
+        public virtual JObject GetInputData()
```
anbare: >GetInputData [](start = 31, length = 12)

This is diverged from what the UWP API is. I believe the UWP API has a UserInputs property, and then you can call AsJson() or AsValueSet().

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveElementRenderers.cs
```diff
@@ -1,16 +1,54 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
 
 namespace AdaptiveCards.Rendering
 {
+    public class AdaptiveActionHandlers
```
anbare: >AdaptiveActionHandlers [](start = 17, length = 22)

Does this deserve to be in its own file, rather than including this class in another class' file?

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveElementRenderers.cs
```diff
@@ -1,16 +1,54 @@
...
+        {
+            typeof(AdaptiveOpenUrlAction),
+            typeof(AdaptiveSubmitAction),
+            typeof(AdaptiveShowCardAction)
+        };
+
+        /// <summary>
+        /// Adds support for a given action type. Any action in a payload not specifed here will be dropped from the rendered card
+        /// </summary>
+        public void AddSupportedAction<TAction>()
```
anbare: >AddSupportedAction [](start = 20, length = 18)

Maybe instead of Add and Remove, we should use "EnableAction" and "DisableAction"?

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElement.cs
matthidinger: It's the same thing, but isn't part of net standard so I had to reuse the concept

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveElementRenderers.cs
matthidinger: Def

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveElementRenderers.cs
matthidinger: Hmm I'm on the fence on this one. Disabled to me feels like it would just gray out the action or something... Any other ideas?

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElement.cs
anbare: That's too bad, makes sense

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveElementRenderers.cs
anbare: I think part of the problem is the name of the class, `AdaptiveActionHandlers`. When I see a method called `Add` on that, I expect to be adding a handler.

Maybe change the class to `AdaptiveSupportedActions`? And then the existing names of `AddSupportedAction` and whatnot make sense. Or we could even simply make them `Add()` at that point

matthidinger: Long-term I was hoping to evolve this class to support the inline delegate so you can wire up the action and it's callback at once, but I didn't have time to fully work through that design yet

anbare: Well the other problem with supporting the callback here is that it's a global callback now, rather than a callback on the instance of the RenderedAdaptiveCard. Essentially it would be what we used to have, where on the Renderer you could wire up the action handlers, which we chose to move away from.

So I'm not sure it'd make sense to support wiring up the callback at this location.

mrbrennonwilliams: What about TryAppendAction or TryAddAction with bool result? My understanding is that an unsupported action would be dropped so we probably want to know that it failed?

matthidinger: The callback is given an instance of the RenderedAdaptiveCard which would provide some (but not all) context for every scenario to properly wire up the callback. 

@mrbrennonwilliams this is so the host can "enable" support for new types of actions specifically for them, e.g., "Action.MyThing". By default the Renderer will drop an action that it doesn't understand (which we will provide as a Warning in the RendereredAdaptiveCard object).

matthidinger:
This fixes #850 #841 #837 #839

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveElementRenderers.cs
mrbrennonwilliams: Thanks for the clarification @matthidinger 

Excuse my questions - just trying to learn more.

I guess allowing host specific actions will introduce a significant level of potential fragmentation in bot functionality?

I can see how this could benefit a host, and in particular developers working against a single host platform - is this feature intended specifically for that use case? I would assume dev's looking for the largest supported platform surface would continue to author against a "lowest common denominator" in terms of functionality.

If it is intended specifically for hosts, should it then be named AddPlatformAction or AddSupportedHostAction?

Cheers

matthidinger: Bots are actually only one use-case for Adaptive Cards. They will used in Windows for different scenarios too, like notifications and the new "Timeline" experience, as well as in Outlook for their Email Cards (again coming soon). Custom actions can of course cause fragmentation in the ecosystem, so our guidance to devs will to be cognizant of that fact, but we definitely want the schema elements and actions to be extensible for any type of scenario

matthidinger:
I think I'm going to close this PR and merge this in, with the caveat that we might change the ActionHandler signature once we've had time to iterate on it. @anbare  objections?

anbare:
Sounds fine with me

PR #871, opened 2017-11-01
Date time

nesalang:


source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +165,330 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
     m_hAlignment = value;
 }
 
+bool TextBlock::scanForISO8601(std::string::const_iterator &itr, std::string::const_iterator &end,
```
riarenas: >scanForISO8601 [](start = 16, length = 14)

Any specific reason why you take the iterators instead of the std::string?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include "CppUnitTest.h"
+#include "TextBlock.h"
+#include <time.h>
+
+using namespace Microsoft::VisualStudio::CppUnitTestFramework;
+using namespace AdaptiveCards;
+using namespace std;
+
+namespace AdaptiveCardsSharedModelUnitTest
+{
```
riarenas: >	 [](start = 1, length = 1)

[nit] replace tabs with spaces

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
```diff
@@ -0,0 +1,207 @@
...
+using namespace std;
+
+namespace AdaptiveCardsSharedModelUnitTest
+{		
+	TEST_CLASS(UnitTest1)
+	{
+	public:
+		TEST_METHOD(ISO8601TransformTest)
+		{
+			TextBlock* blck = new TextBlock();
```
riarenas: >TextBlock* blck = new TextBlock(); [](start = 3, length = 34)

Why not use stack variables?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModel.sln
```diff
@@ -0,0 +1,41 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
riarenas: >Microsoft [](start = 0, length = 9)

I'm getting several build and linker errors when trying to build this solution. Are you sure all the necessary files are in the Pull request branch?

nesalang: I will double check; thanks

---
In reply to: [148398486](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148398486) [](ancestors = 148398486)

source/shared/cpp/ObjectModel/TextBlock.cpp
riarenas: Could we use the stl regex library to avoid doing the parsing ourselves?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
```diff
@@ -0,0 +1,207 @@
...
+			Assert::IsNotNull(blck);
+			string testString = "{a{DATE(2017-02-13T20:46:30Z, Short)}}";
+			blck->SetText(testString);
+			Assert::AreEqual<string>("{a{DATE(2017-02-13T20:46:30Z, Short)}}", blck->GetText());
+			delete blck;
+		}
+
+		TEST_METHOD(ISO8601TransformBadInputTest2)
```
riarenas: Yay! Tests! could you name these in such a way that they specifically call out what is wrong with this input?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
```diff
@@ -0,0 +1,207 @@
...
+		TEST_METHOD(TransformToDateTest1)
+		{
+			TextBlock* blck = new TextBlock();
+			string testString = "{{DATE(2017-02-13T20:46:30Z, Short)}}";
+			blck->SetText(testString);
+			Assert::AreEqual<string>("02/13/2017", blck->GetText());
+			delete blck;
+		}
+
+		TEST_METHOD(TransformToDateTest2)
```
riarenas: >TransformToDateTest2 [](start = 14, length = 20)

TranformToTime?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
nesalang: good point

---
In reply to: [148399471](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148399471) [](ancestors = 148399471)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
```diff
@@ -0,0 +1,207 @@
...
+			TextBlock* blck = new TextBlock();
+			string testString = "{{DATE(2017-10-28T01:47:00Z, Short)}}";
+			blck->SetText(testString);
+			Assert::AreEqual<string>("10/27/2017", blck->GetText());
+			delete blck;
+		}
+		TEST_METHOD(TransformToTimeTest)
+		{
+			TextBlock* blck = new TextBlock();
+			string testString = "{{Time(2017-10-28T02:17:00Z, Short)}}";
```
riarenas: >Short [](start = 53, length = 5)

I was under the impression that time does not have a long/short version. The docs just say it's TIME

riarenas: Let's also make sure that it's supposed to be case insensitive.

---
In reply to: [148399685](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148399685) [](ancestors = 148399685)

nesalang: thanks for letting me know! yes, it made no sense to me, so it returns the same result regardless

---
In reply to: [148399685](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148399685) [](ancestors = 148399685)

nesalang: parsing is not case sensitive

---
In reply to: [148399905](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148399905) [](ancestors = 148399905,148399685)

riarenas: The spec for these features is here:
[https://docs.microsoft.com/en-us/adaptive-cards/create/textfeatures](https://docs.microsoft.com/en-us/adaptive-cards/create/textfeatures)


---
In reply to: [148400013](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148400013) [](ancestors = 148400013,148399685)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
nesalang: thanks

---
In reply to: [148399549](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148399549) [](ancestors = 148399549)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
riarenas: I mean, are we supposed to actually parse things if the function is in lower case? The docs say the functions are in upper case.

---
In reply to: [148400146](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148400146) [](ancestors = 148400146,148399905,148399685)

source/shared/cpp/ObjectModel/TextBlock.h
```diff
@@ -61,5 +62,12 @@ class TextBlock : public BaseCardElement
     bool m_wrap;
     unsigned int m_maxLines;
     HorizontalAlignment m_hAlignment;
+    std::string parseISO8601(std::string::const_iterator &begin, std::string::const_iterator &end) const;
+	bool TextBlock::scanForISO8601(std::string::const_iterator &itr, std::string::const_iterator &end,
+		bool &isDate, std::ostringstream &parsedostr, std::ostringstream &ostr) const;
+    bool ISO8601ToTm(std::string::const_iterator& begin, 
+            std::string::const_iterator& end, struct tm* result, std::ostringstream &ostr) const;
+    bool completeParsing(std::string::const_iterator &begin,
+            std::string::const_iterator &end, bool &isShort, std::ostringstream &ostr) const;
```
riarenas: all of these can probably be static

source/shared/cpp/ObjectModel/TextBlock.h
```diff
@@ -61,5 +62,12 @@ class TextBlock : public BaseCardElement
     bool m_wrap;
     unsigned int m_maxLines;
     HorizontalAlignment m_hAlignment;
+    std::string parseISO8601(std::string::const_iterator &begin, std::string::const_iterator &end) const;
```
riarenas: >parseISO8601 [](start = 16, length = 12)

[nit] Method names start with UpperCase ParseISO...

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: i think we can;  I thought it would be simple enough do parsing ourselves; after start working on it; i didn't get chance to think back on regex until finished.

---
In reply to: [148398764](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148398764) [](ancestors = 148398764)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
```diff
@@ -0,0 +1,207 @@
+#include "stdafx.h"
```
riarenas: >include [](start = 1, length = 7)

There's a typo in the name of this file

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +165,330 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+    };
+    const int cnts = 3;
+    int idx = 0, gateKeeper = NoneParsed;
+    const char *patterns[] = {"IME", "ATE"}; 
+    for(; itr != end && gateKeeper != DoneCheck; itr++)
+    {
+		if(*itr == '{')
+		{
+            gateKeeper <<= 1;
+			ostr << *itr;
```
RebeccaAnne: >			 [](start = 0, length = 3)

[nit] spaces instead of tabs

source/shared/cpp/ObjectModel/TextBlock.cpp
RebeccaAnne: Let's look into changing this to regex in order to simplify maintainability going forward.

---
In reply to: [148403023](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148403023) [](ancestors = 148403023,148398764)

nesalang: i think that is right approach for maintenance, can we do it after November since we have time constraint? 

---
In reply to: [148415965](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148415965) [](ancestors = 148415965,148403023,148398764)

nesalang: i wanted to keep moving forward with text process.

---
In reply to: [148396093](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148396093) [](ancestors = 148396093)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
nesalang: o.k. sounds good to me about case. I think it should be case sensitive. for time, spec doesn't have that, so we can just ignore it?

---
In reply to: [148400224](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148400224) [](ancestors = 148400224,148400013,148399685)

source/shared/cpp/ObjectModel/TextBlock.cpp
riarenas: Sorry, this comment was meant for the Parse method. You always call it with the string's begin() and end() iterators.

---
In reply to: [148613744](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148613744) [](ancestors = 148613744,148396093)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
riarenas: I would check how it works in the html visualizer.

---
In reply to: [148614497](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148614497) [](ancestors = 148614497,148400224,148400013,148399685)

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: sorry, i'm not following you; can you explain?

---
In reply to: [148614839](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148614839) [](ancestors = 148614839,148613744,148396093)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/TimeAndDateFucntionTest.cpp
nesalang: right, that's good idea. let me do that.

---
In reply to: [148615263](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148615263) [](ancestors = 148615263,148614497,148400224,148400013,148399685)

source/shared/cpp/ObjectModel/TextBlock.cpp
riarenas: The parseISO8601 method is always called with the string's begin() and end() iterators. That function should probably just take the string instead of begin and end iterators

---
In reply to: [148615412](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148615412) [](ancestors = 148615412,148614839,148613744,148396093)

nesalang: that's very good point.
yup, that's correct.
thanks


---
In reply to: [148616076](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148616076) [](ancestors = 148616076,148615412,148614839,148613744,148396093)

riarenas: I think if we don't do it now we won't do it. I believe using regex would simplify this code a lot and is worth doing now.

---
In reply to: [148613565](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148613565) [](ancestors = 148613565,148415965,148403023,148398764)

nesalang: let me work on it

---
In reply to: [148619282](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148619282) [](ancestors = 148619282,148613565,148415965,148403023,148398764)

nesalang: done

---
In reply to: [148639830](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148639830) [](ancestors = 148639830,148619282,148613565,148415965,148403023,148398764)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -86,7 +89,7 @@ Json::Value TextBlock::SerializeToJsonValue()
 
 std::string TextBlock::GetText() const
 {
-    return m_text;
+    return ParseISO8601UsingRegex();
```
riarenas: >ParseISO8601UsingRegex [](start = 11, length = 22)

UsingRegex suffix is unnecessary

nesalang: Agree

---
In reply to: [148858644](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148858644) [](ancestors = 148858644)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,126 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
     m_hAlignment = value;
 }
 
+std::string TextBlock::ParseISO8601UsingRegex() const
+{
+	std::regex pattern("\\{\\{(DATE|TIME)\\((\\d{4})-{1}(\\d{2})-{1}(\\d{2})T(\\d{2}):{1}(\\d{2}):{1}(\\d{2})(Z|([+-])(\\d{2}):{1}(\\d{2}))(\\)|, S[Hh][Oo][Rr][Tt]\\)|, L[Oo][Nn][Gg]\\))\\}\\}");
```
riarenas: > L[Oo][Nn][Gg] [](start = 165, length = 14)

I think everything should be case insensitive. This pattern wouldn't work for date(..., sHoRt). would it? A potential solution to not make the regex any more complicated is to transform the text to lowercase, and replace the regex in the original text.

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,126 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+        Format,
+    };
+    int factor = 1, hr = 0, mn = 0;
+    struct tm parsedTm = { 0 };
+    int *addrs[]  = {&parsedTm.tm_year, &parsedTm.tm_mon, 
+        &parsedTm.tm_mday, &parsedTm.tm_hour, &parsedTm.tm_min, 
+        &parsedTm.tm_sec, &hr, &mn}; 
+    std::vector<int> indexer = {YEAR, Month, Day, Hour, Min, Sec, TZHr, TZMn};
+
+    while(std::regex_search (text, matches, pattern))
```
riarenas: >  [](start = 27, length = 1)

[nit] remove space

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: good point

---
In reply to: [148859877](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148859877) [](ancestors = 148859877)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
```diff
@@ -0,0 +1,206 @@
...
+        TEST_METHOD(TwoISO8601WithText)
+        {
+            TextBlock blck;
+            // New York 
+            string testString = "You have arrived in New York on {{DATE(2017-10-27T22:27:00-04:00, SHORT)}} at {{TIME(2017-10-27T22:27:00-04:00, SHORT)}}";
+            blck.SetText(testString);
+            Assert::AreEqual<string>("You have arrived in New York on 10/27/17 at 07:27:00 PM", blck.GetText());
+        }
+
+        TEST_METHOD(LoacleDependentDateTest)
```
riarenas: >LoacleDependentDateTest [](start = 20, length = 23)

How does this test take the locale into account?

nesalang: thanks for catching this; i was going to change locale info through imbuse, but it looks like it only affect cout; i wil remove the test. 

---
In reply to: [148861012](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148861012) [](ancestors = 148861012)

riarenas: We still need to test with different locales. Can't you use something like std::setlocale?

---
In reply to: [148861694](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148861694) [](ancestors = 148861694,148861012)

nesalang: good, thanks

---
In reply to: [148862332](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148862332) [](ancestors = 148862332,148861694,148861012)

source/shared/cpp/ObjectModel/TextBlock.cpp
riarenas: [nit] space between while and () same with ifs

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,126 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+        Min,
+        Sec,
+        TimeZone,
+		TZHr = 10,
+		TZMn,
+        Format,
+    };
+    int factor = 1, hr = 0, mn = 0;
+    struct tm parsedTm = { 0 };
+    int *addrs[]  = {&parsedTm.tm_year, &parsedTm.tm_mon,
```
riarenas: >  [](start = 16, length = 1)

[nit] double space

source/shared/cpp/ObjectModel/TextBlock.cpp
RebeccaAnne: >	 [](start = 0, length = 1)

Looks like there are still tabs in this file

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,126 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
     m_hAlignment = value;
 }
 
+std::string TextBlock::ParseISO8601UsingRegex() const
```
RebeccaAnne: >ParseISO8601UsingRegex [](start = 23, length = 22)

I tried loading up the TextBlock.DateTimeFormatting.json sample in the UWP visualizer, and it looks like the date isn't working...

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
nesalang: I tested it manually, it worked as api doc specified, marking it resolved.

---
In reply to: [148862549](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148862549) [](ancestors = 148862549,148862332,148861694,148861012)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,126 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+        Day,
+        Hour,
+        Min,
+        Sec,
+        TimeZone,
+		TZHr = 10,
+		TZMn,
+        Format,
+    };
+    int factor = 1, hr = 0, mn = 0;
```
RebeccaAnne: >int factor = 1, hr = 0, mn = 0; [](start = 4, length = 31)

This code is a little tricky to read.  Could you add some more comments and maybe longer/more descriptive variable names?  (ie, what does "factor" represent?  "mn" is minutes?  Lets just call it minutes)

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: it's because spacing in regex; it will update new commits.

---
In reply to: [149149659](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149149659) [](ancestors = 149149659)

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: sure; thanks for insights


---
In reply to: [149151448](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149151448) [](ancestors = 149151448)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
riarenas: Which locales did you test? What is the output for a long DATE when locale is set to say es-mx or to ru?

---
In reply to: [149151225](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149151225) [](ancestors = 149151225,148862549,148862332,148861694,148861012)

nesalang: i tested with es_CO and jp_JP

---
In reply to: [149152236](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149152236) [](ancestors = 149152236,149151225,148862549,148862332,148861694,148861012)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
```diff
@@ -0,0 +1,206 @@
...
+            blck.SetText(testString);
+            Assert::AreEqual<string>("02/13/17", blck.GetText());
+        }
+
+        TEST_METHOD(TransformToDateTest2)
+        {
+            TextBlock blck;
+            string testString = "{{TIME(2017-10-27T18:19:09Z, SHORT)}}";
+            blck.SetText(testString);
+            Assert::AreEqual<string>("11:19:09 AM", blck.GetText());
```
riarenas: >09 [](start = 44, length = 2)

I believe we shouldn't show the seconds. The api documentation drops the seconds. Could you follow up with Matt to make sure?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+        TEST_METHOD(TransformToDateTest1)
+        {
+            TextBlock blck;
+            string testString = "{{DATE(2017-02-13T20:46:30Z, SHORT)}}";
+            blck.SetText(testString);
+            Assert::AreEqual<string>("02/13/17", blck.GetText());
+        }
+
+        TEST_METHOD(TransformToDateTest2)
```
riarenas: >TransformToDateTest2 [](start = 20, length = 20)

[nit] this is testing TIME, not DATE. Name should probably change

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
```diff
@@ -0,0 +1,206 @@
...
+        {
+            TextBlock blck;
+            string testString = "{{TIME(2017-10-27T18:19:09Z, SHORT)}}";
```
riarenas: >SHORT [](start = 62, length = 5)

TIME does not have SHORT, LONG versions

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
riarenas: [nit] typo in locale

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
nesalang: yes, that's what i thought, but html renderer shows seconds; i will double check

---
In reply to: [149153560](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149153560) [](ancestors = 149153560)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
nesalang: French example, 
Long
Expected:<Monday, February 13, 2017> 
Actual:<lundi, février 13, 2017>
Short
Expected:<02/13/17> 
Actual:<13/02/2017>


---
In reply to: [149152763](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149152763) [](ancestors = 149152763,149152236,149151225,148862549,148862332,148861694,148861012)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
riarenas: [nit] would you mind organizing these tests? TransformToTimeTest2 is placed before TransformToTimeTest1. Optionally, give them more meaningful names, and then the order wouldn't matter as much

---
In reply to: [149153835](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149153835) [](ancestors = 149153835)

source/shared/cpp/ObjectModel/TextBlock.cpp
riarenas: There's still tabs in this iteration

---
In reply to: [149141278](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149141278) [](ancestors = 149141278)

nesalang: i guess there is something wrong with my vim setting.
let me look into that; thanks

---
In reply to: [149176815](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149176815) [](ancestors = 149176815,149141278)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,141 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+	std::regex pattern("\\{\\{(DATE|Date|TIME|Time)\\((\\d{4})-{1}(\\d{2})-{1}(\\d{2})T(\\d{2}):{1}(\\d{2}):{1}(\\d{2})(Z|([+-])(\\d{2}):{1}(\\d{2}))(\\)|,Short\\)|,SHORT\\)|,Long\\)|,LONG\\))\\}\\}");
+    std::smatch matches;
+    std::string text = m_text;
+    std::ostringstream parsedostr;
+    time_t offset = 0; 
+    bool isDate = false, isShort = true;
+    enum MatchIndex
+    {
+        IsDate = 1,
+        YEAR,
```
riarenas: >YEAR [](start = 8, length = 4)

[nit] why is this the only one in All caps?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
nesalang: will do

---
In reply to: [149176496](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149176496) [](ancestors = 149176496,149153835)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,141 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+        Month,
+        Day,
+        Hour,
+        Min,
+        Sec,
+        TimeZone,
+        TZHr = 10,
```
riarenas: >TZHr [](start = 8, length = 4)

Is this for the hour offset in a timezone? maybe needs a better name?

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: good catch.

---
In reply to: [149177085](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149177085) [](ancestors = 149177085)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
riarenas: Cool! could we check with Matt what the decision was for this? should they actually get translated?

---
In reply to: [149163976](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149163976) [](ancestors = 149163976,149152763,149152236,149151225,148862549,148862332,148861694,148861012)

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: ok. TimeZoneHourOffset?

---
In reply to: [149177242](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149177242) [](ancestors = 149177242)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,141 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+        TZMn,
+        Format,
+    };
+    int factor = 1, hours = 0, minutes = 0;
+    struct tm parsedTm = { 0 };
+    int *addrs[] = {&parsedTm.tm_year, &parsedTm.tm_mon,
```
riarenas: >* [](start = 8, length = 1)

What is the reason for this being a pointer?

nesalang: it holds addresses

---
In reply to: [149178352](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149178352) [](ancestors = 149178352)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,141 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+        if(matches[Format].matched)
+        {
+           char format = matches[Format].str()[0];
+           isShort = (format != ')' && matches[Format].str()[1] == 'L')? false : true;
+        }
+
+        // converts to local time from utc
+        if(!localtime_s(&result, &utc))
+        {
+            // localtime_s double counts daylight saving time
```
RebeccaAnne: >localtime_s double counts daylight saving time [](start = 15, length = 46)

Is this a known bug in localtime_s?

nesalang: there is in localtime
it uses global variable.

---
In reply to: [149179117](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149179117) [](ancestors = 149179117)

source/shared/cpp/ObjectModel/TextBlock.cpp
riarenas: >  [](start = 56, length = 1)

[nit] Please check for trailing spaces throughout this file

nesalang: thanks; i need to import vim settings from mac.
it kept get lost after windows update

---
In reply to: [149179442](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149179442) [](ancestors = 149179442)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,141 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
     m_hAlignment = value;
 }
 
+std::string TextBlock::ParseISO8601UsingRegex() const
+{
+	std::regex pattern("\\{\\{(DATE|Date|TIME|Time)\\((\\d{4})-{1}(\\d{2})-{1}(\\d{2})T(\\d{2}):{1}(\\d{2}):{1}(\\d{2})(Z|([+-])(\\d{2}):{1}(\\d{2}))(\\)|,Short\\)|,SHORT\\)|,Long\\)|,LONG\\))\\}\\}");
```
riarenas: >(DATE|Date|TIME|Time [](start = 27, length = 20)

This will fail for {{daTE(2017-02-14T06:08:39Z,Short)}} which is valid in the typescript renderer/visualizer

nesalang: i think math said he is aiming for stricker standard.
let me check with mathh.

---
In reply to: [149181253](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149181253) [](ancestors = 149181253)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,141 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+    std::vector<int> indexer = {YEAR, Month, Day, Hour, Min, Sec, TZHr, TZMn};
+
+    while(std::regex_search(text, matches, pattern))
+    {
+        parsedostr << matches.prefix().str();
+
+        if(matches[IsDate].matched)
+        {
+            // Date is matched
+            isDate = (matches[IsDate].str()[0] == 'D')? true : false;
```
riarenas: >? true : false; [](start = 54, length = 15)

Aren't these unnecessary? matches[IsDate].str()[0] == 'D' is already a bool.

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,141 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+        }
+
+        for(unsigned int idx = 0; idx < indexer.size(); idx++)
```
riarenas: >unsigned int idx = 0; idx < indexer.size(); idx++ [](start = 12, length = 49)

[nit] consider changing this into a range-based for:

for(auto& index : indexer)
{
...
}

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,141 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+        Format,
+    };
+    int factor = 1, hours = 0, minutes = 0;
+    struct tm parsedTm = { 0 };
+    int *addrs[] = {&parsedTm.tm_year, &parsedTm.tm_mon, 
+        &parsedTm.tm_mday, &parsedTm.tm_hour, &parsedTm.tm_min, 
+        &parsedTm.tm_sec, &hours, &minutes}; 
+    std::vector<int> indexer = {YEAR, Month, Day, Hour, Min, Sec, TZHr, TZMn};
+
+    while(std::regex_search(text, matches, pattern))
```
riarenas: >while( [](start = 4, length = 6)

[nit] styling, please leave a space before the ( in if, while, for clauses to match the rest of the codebase.

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,141 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+        offset += ((nTzOffset / 100) * 3600 + (nTzOffset % 100) * 60);
+        // add offset to utc
+        utc += offset;
+        struct tm result = {0};
+
+        // match for long/short
+        if(matches[Format].matched)
+        {
+           char format = matches[Format].str()[0];
+           isShort = (format != ')' && matches[Format].str()[1] == 'L')? false : true;
```
riarenas: >? false : true; [](start = 71, length = 15)

The previous expression is already a bool, no need for this operator.

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +168,141 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+        // Z == UTC, 
+        // + == time added from UTC
+        // - == time subtracted from UTC
+        if(matches[TimeZone].matched)
+        {
+            
+            char zone = matches[TimeZone].str()[0];
+            if(zone == 'Z')
+            {
+                factor = 0;
```
RebeccaAnne: >factor [](start = 16, length = 6)

Instead of having this "factor" variable, can you just calculate the offset inside these if statements?  I think that will be clearer.

source/shared/cpp/ObjectModel/TextBlock.cpp
riarenas: Still there :)

---
In reply to: [148859669](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r148859669) [](ancestors = 148859669,148858644)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
```diff
@@ -0,0 +1,197 @@
...
+            string testString = "{{TIME(2017-10-28T11:25:00+09:00)}}";
+            blck.SetText(testString);
+            Assert::AreEqual<string>("07:25:00 PM", blck.GetText());
+        }
+
+        TEST_METHOD(TransformToDateTest6)
+        {
+            TextBlock blck;
+            // seoul 
+            string testString = "{{DATE(2017-10-28T11:25:00+09:00,LONG)}}";
```
riarenas: >+09:00 [](start = 59, length = 6)

Please check with Matt how the timezone should be specified. The Typescript renderer uses +09 and breaks on 09:00

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: Thanks~

---
In reply to: [149210701](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149210701) [](ancestors = 149210701,148859669,148858644)

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: will do; thanks

---
In reply to: [149207223](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149207223) [](ancestors = 149207223)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
nesalang: o.k.

---
In reply to: [149213501](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149213501) [](ancestors = 149213501)

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: do you mean matches[isDate] is bool? it would match both Date and Time string.


---
In reply to: [149192396](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149192396) [](ancestors = 149192396)

riarenas: I mean (matches[IsDate].str()[0] == 'D') is already a boolean expression. The conditional operator is not necessary

---
In reply to: [149215274](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149215274) [](ancestors = 149215274,149192396)

nesalang: thanks; that's good point

---
In reply to: [149217300](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149217300) [](ancestors = 149217300,149215274,149192396)

source/shared/cpp/ObjectModel/TextBlock.cpp
riarenas: Does the same thing apply for localtime_r?

---
In reply to: [149179338](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149179338) [](ancestors = 149179338,149179117)

nesalang: No, it doe not

---
In reply to: [149517346](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149517346) [](ancestors = 149517346,149179338,149179117)

nesalang: localtime_r doesn't have bug

---
In reply to: [150949291](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150949291) [](ancestors = 150949291,149517346,149179338,149179117)

riarenas: Wouldn't this mean we'd be one hour off in the platforms that use localtime_r?

---
In reply to: [150949653](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150949653) [](ancestors = 150949653,150949291,149517346,149179338,149179117)

source/shared/cpp/ObjectModel/TextBlock.h
```diff
@@ -61,5 +62,6 @@ class TextBlock : public BaseCardElement
     bool m_wrap;
     unsigned int m_maxLines;
     HorizontalAlignment m_hAlignment;
+    std::string ParseISO8601() const;
```
riarenas: >ParseISO8601 [](start = 16, length = 12)

let's call this ParseDateTime. the ISO is no longer accurate

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +174,144 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+            {
+                parsedostr << std::put_time(&result, "%I:%M %p");
+            }
+        }
+        text = matches.suffix().str();
+        parsedTm = {0};
+        offset = 0;
+        hours = minutes = 0;
+        isDate = false;
+        formatStyle = 0;
```
riarenas: Would it help to define this variables inside the loop? that way you don't have to do this awkward cleanup afer each iteration.

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +174,144 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+        // converts to local time from utc
+        if (!LOCALTIME(&result, &utc))
+        {
+            // localtime_s double counts daylight saving time
+            if (result.tm_isdst == 1)
+                result.tm_hour -= 1;
+
+            if (isDate)
+            {
+                switch(formatStyle)
```
riarenas: What would be the effect here if we add a new style that starts with one of these letters in the future?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
```diff
@@ -0,0 +1,209 @@
+#include "stdafx.h"
```
riarenas: Would it make sense to name this file DateAndTimeUnitTests.cpp instead?

riarenas: In addition to these unit tests, could you please post a screenshot of the rendered card from the payload in the date/time spec?

source/shared/cpp/ObjectModel/TextBlock.h
nesalang: thanks

---
In reply to: [150950502](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150950502) [](ancestors = 150950502)

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: yes, but this way,  we don't keep pushing and popping the variable on stack.

---
In reply to: [150951877](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150951877) [](ancestors = 150951877)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
nesalang: thanks

---
In reply to: [150952950](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150952950) [](ancestors = 150952950)

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: that's something we have to deal with later. i don't want to make it more complex than necessary.

---
In reply to: [150952262](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150952262) [](ancestors = 150952262)

source/shared/cpp/ObjectModel/TextBlock.cpp
riarenas: I will leave it to style preference I guess, but this seems like a case of over optimizing something that actually makes the code to be more bug prone.

---
In reply to: [150964853](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150964853) [](ancestors = 150964853,150951877)

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: no, that's incorrect; i will fix the comments to make it clear. thanks

---
In reply to: [150950310](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150950310) [](ancestors = 150950310,150949653,150949291,149517346,149179338,149179117)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/AdaptiveCardsSharedModelUnitTest.cpp
nesalang: i verify it; it works; but you have change ,Short -> , SHORT.
You can also try it.

---
In reply to: [150953178](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150953178) [](ancestors = 150953178)

riarenas: What do you mean? [https://github.com/Microsoft/AdaptiveCards/issues/908](https://github.com/Microsoft/AdaptiveCards/issues/908) that issue has a very lengthy example payload that should not require any changes, as it is the final spec.

---
In reply to: [150965757](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150965757) [](ancestors = 150965757,150953178)

nesalang: oh, i see; i thought you were referring to one of the sample examples. it's good suggestion

---
In reply to: [150966324](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150966324) [](ancestors = 150966324,150965757,150953178)

source/shared/cpp/ObjectModel/TextBlock.cpp
riarenas: Seems like if we're aware that it's going to be an issue, it should be addressed. Is it too big of a change to get the whole format string rather than the first character?

---
In reply to: [150965213](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150965213) [](ancestors = 150965213,150952262)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -165,3 +174,191 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+                }
+            }
+            else
+            {
+                if (parsedTm.tm_mday <= 31)
+                {
+                    isValidForm = true;
+                }
+            }
+        }
```
riarenas: [nit] consider extracting this section into its own operation. ValidateDateTime or something like it.

nesalang: o.k.


---
In reply to: [150997368](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150997368) [](ancestors = 150997368)

source/shared/cpp/ObjectModel/TextBlock.cpp
RebeccaAnne: Just the addition of minutes + seconds, right?  That's fine

---
In reply to: [149214345](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r149214345) [](ancestors = 149214345,149207223)

source/shared/cpp/ObjectModel/TextBlock.cpp
RebeccaAnne: I agree with Ricardo here.  Let's define our variables in the appropriate scope.

---
In reply to: [150965532](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r150965532) [](ancestors = 150965532,150964853,150951877)

source/uwp/AdaptiveCards.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
```
riarenas: Do not check in the changes to this file.

nesalang: Yes,

---
In reply to: [151219045](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r151219045) [](ancestors = 151219045)

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: ok

---
In reply to: [151218816](https://github.com/Microsoft/AdaptiveCards/pull/871#discussion_r151218816) [](ancestors = 151218816,150965532,150964853,150951877)

PR #880, opened 2017-11-02
[UWP] Fix Auto Image Size Rendering

riarenas:
Fix auto image behavior so the image maintains its native size, and gets scaled back in cases where it doesn't fit its container.


source/uwp/Renderer/lib/XamlBuilder.h
```diff
@@ -137,11 +137,16 @@ namespace AdaptiveCards { namespace Uwp
...
+            _In_ ABI::AdaptiveCards::Uwp::IAdaptiveRenderArgs* renderArgs);
         template<typename T>
         void SetImageSource(T* destination, ABI::Windows::UI::Xaml::Media::IImageSource* imageSource);
         template<typename T>
-        void SetImageOnUIElement(_In_ ABI::Windows::Foundation::IUriRuntimeClass* imageUrl, T* uiElement, ABI::AdaptiveCards::Uwp::IAdaptiveRenderContext* renderContext);
+        void SetImageOnUIElement(
+            _In_ ABI::Windows::Foundation::IUriRuntimeClass* imageUrl,
+            T* uiElement,
+            ABI::AdaptiveCards::Uwp::IAdaptiveCardResourceResolvers* resolvers,
+            ABI::Windows::UI::Xaml::IFrameworkElement* parentElement);
```
riarenas: >ABI::Windows::UI::Xaml::IFrameworkElement* parentElement [](start = 12, length = 56)

This was not needed. Removing

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -425,11 +434,13 @@ namespace AdaptiveCards { namespace Uwp
...
-    void XamlBuilder::SetImageOnUIElement(_In_ ABI::Windows::Foundation::IUriRuntimeClass* imageUri, T* uiElement, IAdaptiveRenderContext* renderContext)
+    void XamlBuilder::SetImageOnUIElement(
+        _In_ ABI::Windows::Foundation::IUriRuntimeClass* imageUri,
+        T* uiElement,
+        IAdaptiveCardResourceResolvers* resolvers)
     {
         // Get the resource resolvers
-        ComPtr<IAdaptiveCardResourceResolvers> resolvers;
-        THROW_IF_FAILED(renderContext->get_ResourceResolvers(&resolvers));
+        ComPtr<IAdaptiveCardResourceResolvers> localResolvers(resolvers);
```
RebeccaAnne: > ComPtr<IAdaptiveCardResourceResolvers> localResolvers(resolvers); [](start = 7, length = 66)

Looks like this isn't used?

riarenas: Yep good catch. I added it at some point as I needed to do a .As, but I removed it at later.

---
In reply to: [149127556](https://github.com/Microsoft/AdaptiveCards/pull/880#discussion_r149127556) [](ancestors = 149127556)

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1214,41 +1263,93 @@ namespace AdaptiveCards { namespace Uwp
...
+            // Set Auto, None, and Stretch to Stretch_UniformToFill. An ellipse set to Stretch_Uniform ends up with size 0.
+            if (size == ABI::AdaptiveCards::Uwp::ImageSize::None ||
+                size == ABI::AdaptiveCards::Uwp::ImageSize::Stretch ||
+                size == ABI::AdaptiveCards::Uwp::ImageSize::Auto)
             {
-                ComPtr<IShape> ellipseAsShape;
-                THROW_IF_FAILED(ellipse.As(&ellipseAsShape));
-
                 THROW_IF_FAILED(ellipseAsShape->put_Stretch(Stretch::Stretch_UniformToFill));
             }
```
riarenas: Stretch_UniformToFill is the default value. Since we're not changing this for any size, I'm removing these lines.

PR #888, opened 2017-11-03
Refactor object model to store uwp objects for properties and generat…

RebeccaAnne:
…e shared model when needed for serialization

### DIFF
--- a/source/uwp/Renderer/lib/Util.cpp
+++ b/source/uwp/Renderer/lib/Util.cpp
@@ -101,7 +101,7 @@ std::shared_ptr<TSharedBaseType> GetSharedModel(TAdaptiveBaseType * item)
     ComPtr<TAdaptiveType> adaptiveElement = PeekInnards<TAdaptiveType>(item);
 
     std::shared_ptr<TSharedType> sharedModelElement;
-    if (SUCCEEDED(adaptiveElement->GetSharedModel(sharedModelElement)))
+    if (adaptiveElement && SUCCEEDED(adaptiveElement->GetSharedModel(sharedModelElement)))
         return sharedModelElement;
     else
         return nullptr;
@@ -161,6 +161,11 @@ HRESULT GenerateSharedElements(
                 baseCardElement = std::make_shared<CustomElementWrapper>(item);
                 break;
         }
+        if (baseCardElement == nullptr)
+        {
+            return E_INVALIDARG;
+        }
+
         containedElements.push_back(baseCardElement);
 
         return S_OK;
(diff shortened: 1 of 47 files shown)


source/uwp/Renderer/lib/Util.cpp
```diff
@@ -73,6 +76,199 @@ bool Boolify(const boolean value)
...
+    return S_OK;
+}
+
+template <typename TSharedBaseType, typename TSharedType, typename TAdaptiveBaseType, typename TAdaptiveType>
+std::shared_ptr<TSharedBaseType> GetSharedModel(TAdaptiveBaseType * item)
+{
+    ComPtr<TAdaptiveType> adaptiveElement = PeekInnards<TAdaptiveType>(item);
+
+    std::shared_ptr<TSharedType> sharedModelElement;
+    if (SUCCEEDED(adaptiveElement->GetSharedModel(sharedModelElement)))
```
riarenas: >adaptiveElement [](start = 18, length = 15)

Should we check adaptiveElement for null before doing this?

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -73,6 +76,199 @@ bool Boolify(const boolean value)
...
+                baseCardElement = GetSharedModel<AdaptiveCards::BaseCardElement, AdaptiveCards::TimeInput, ABI::AdaptiveCards::Uwp::IAdaptiveCardElement, AdaptiveCards::Uwp::AdaptiveTimeInput>(item);
+                break;
+            case ABI::AdaptiveCards::Uwp::ElementType::ToggleInput:
+                baseCardElement = GetSharedModel<AdaptiveCards::BaseCardElement, AdaptiveCards::ToggleInput, ABI::AdaptiveCards::Uwp::IAdaptiveCardElement, AdaptiveCards::Uwp::AdaptiveToggleInput>(item);
+                break;
+            case ABI::AdaptiveCards::Uwp::ElementType::Custom:
+                baseCardElement = std::make_shared<CustomElementWrapper>(item);
+                break;
+        }
+        containedElements.push_back(baseCardElement);
```
riarenas: >push_back [](start = 26, length = 9)

GetSharedModel may return nullptr. Do we want to add those to the containedElements anyways?

source/uwp/Renderer/lib/Util.h
```diff
@@ -44,6 +44,34 @@ HRESULT GetSpacingSizeFromSpacing(
...
+
+HRESULT GenerateSharedAction(
+    ABI::AdaptiveCards::Uwp::IAdaptiveActionElement* action,
+    std::shared_ptr<AdaptiveCards::BaseActionElement>& sharedAction); 
+
+HRESULT GenerateSharedActions(
+    ABI::Windows::Foundation::Collections::IVector<ABI::AdaptiveCards::Uwp::IAdaptiveActionElement*>* items,
+    std::vector<std::shared_ptr<AdaptiveCards::BaseActionElement>>& containedElements);
+
+HRESULT GenerateSharedImages(
```
riarenas: >GenerateSharedImages [](start = 8, length = 20)

Do you think these specific ones should stay here, or move to the ImageSet, FactSet and ChoiceSet classes?

RebeccaAnne: I put them here because the ones that go in the other direction are already here (GenerateFactsProjection, GenerateImagesProjection, etc).  I'm going to leave as is for now unless you feel strongly.

---
In reply to: [149742053](https://github.com/Microsoft/AdaptiveCards/pull/888#discussion_r149742053) [](ancestors = 149742053)

riarenas: I do not feel strongly.

---
In reply to: [149750292](https://github.com/Microsoft/AdaptiveCards/pull/888#discussion_r149750292) [](ancestors = 149750292,149742053)

Issue #872, opened 2017-11-01
AdaptiveInput Id property should be required without hiding base Id property


matthidinger:
```csharp
        [JsonRequired]
        public new string Id { get; set; }
```

mrbrennonwilliams:
dotneterrors branch has this fix. I believe this should be closed now @matthidinger ?

matthidinger:
Fixed with #891

PR #891, opened 2017-11-03
Added .NET error/warning results to parsing/rendering

### LINKED ISSUE #835: ParseResult needs to return errors/warnings

### LINKED ISSUE #847: .NET full versioning support
- [ ] `AdaptiveCardConverter.cs` should return a parse error if the version is not specified
- [ ] Needs to check for `MinVersion` and return the `FallbackText` is greater than current version

### LINKED ISSUE #849: .NET deserialization needs to return a warning when it removes an unknown element type
AdaptiveTypedElementConverter.cs 

```csharp
     public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
        {
            var jObject = JObject.Load(reader);
            // Create target object based on JObject
            var typeName = jObject["type"]?.Value<string>() ?? jObject["@type"]?.Value<string>();
            if (typeName == null)
            {
                throw new JsonException("AdaptiveCard elements must contain a 'type' property");
            }

            if (TypedElementTypes.Value.TryGetValue(typeName, out var type))
            {
                var result = Activator.CreateInstance(type);
                serializer.Populate(jObject.CreateReader(), result);
                return result;
            }
            
            // TODO: log unknown element type 
            return null;
        }
````

### PR DESCRIPTION
**Pull request by matthidinger**
Re #674 
Fixes #849
Fixes #847 
Fixes #835

### CODE REVIEW

**Comment by matthidinger** [comment]
@riarenas / @anbare  I'm pretty happy with how this turned out. It's not the entire set of errors/warnings but it's pretty close. and feels really clean from an API perspective (in MainWindows.xaml.cs)

File: source/dotnet/Library/AdaptiveCards.Rendering.Wpf/RenderedAdaptiveCard.cs [hunk 1]
```diff
@@ -20,6 +20,10 @@ public RenderedAdaptiveCard(FrameworkElement frameworkElement, AdaptiveCard orig
 
         internal Dictionary<string, Func<object>> InputBindings { get; set; } = new Dictionary<string, Func<object>>();
 
+        /// <summary>
+        /// If any modifications were done to the rendered card they will be reported here
```
anbare: I get why you call them "modifications", but that just feels odd to me. I think "Any warnings encountered while rendering will be reported here" makes more sense. For example, I don't really see a custom renderer missing as a "modification", I see that as a "warning", as in "Hey, I'm warning you that we couldn't render an element"

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/RenderedAdaptiveCard.cs
```diff
@@ -20,6 +20,10 @@ public RenderedAdaptiveCard(FrameworkElement frameworkElement, AdaptiveCard orig
...
+        /// </summary>
+        public IList<AdaptiveViolation> Warnings { get; internal set; } = new List<AdaptiveViolation>();
```
anbare: `AdaptiveViolation` seems like a scary name. Not sure it should be a "violation" if there was simply an unknown type encountered, for example?

source/dotnet/Library/AdaptiveCards/AdaptiveSchemaVersion.cs
```diff
@@ -19,17 +19,25 @@ public AdaptiveSchemaVersion(string versionString)
...
 
-            var minor = 0;
-            if (splits.Length > 1)
+                Major = major;
+                Minor = minor;
+            }
+            catch (Exception ex)
             {
-                minor = int.Parse(splits[1]);
+                throw new JsonReaderException($"'{versionString}' is not a valid version identifier", ex);
```
anbare: Should this also be an `AdaptiveSchemaException`?

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
```diff
@@ -73,7 +85,9 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
             var typeName = jObject["type"]?.Value<string>() ?? jObject["@type"]?.Value<string>();
             if (typeName == null)
             {
-                throw new JsonException("AdaptiveCard elements must contain a 'type' property");
+                //_parseResult.Errors.Add(new AdaptiveViolation(2, "AdaptiveCard elements must contain a 'type' property"));
+                //return null;
+                throw new JsonException("Missing required 'type' property on adaptive card element");
```
anbare: Should be `AdaptiveSchemaException`?

source/dotnet/Library/AdaptiveCards/AdaptiveCardParseResult.cs
```diff
@@ -8,11 +9,27 @@ namespace AdaptiveCards
...
+    public class AdaptiveViolation
+    {
+        public AdaptiveViolation(int code, string message)
         {
-            Card = card;
+            Code = code;
+            Message = message;
         }
+
+        public int Code { get; set; }
```
anbare: We should have a `readonly` collection of known codes?

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveRenderContext.cs
```diff
@@ -120,7 +122,16 @@ public virtual dynamic MergeInputData(dynamic data)
...
         {
-            return ElementRenderers.Get(element.GetType())(element, this);
+            var renderer = ElementRenderers.Get(element.GetType());
+            if (renderer != null)
+            {
+                return renderer.Invoke(element, this);
+            }
+            else
+            {
+                Warnings.Add(new AdaptiveViolation(-1, $"No renderer for element type '{element.Type}'"));
```
anbare: The integer Code should be passed in via a `readonly` property, otherwise a value of -1, and later 5 are meaningless and prone to mistakes/conflicts

anbare:
Yeah seems good to me, I'm indifferent overall, either way works.

matthidinger:
Last commit fixes #892 and #872

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveRenderContext.cs
matthidinger: Yea those are temp, we will standardize these. I'll add TODOs

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/RenderedAdaptiveCard.cs
matthidinger: Agree, since it's just for warnings now I'll change the name

Issue #903, opened 2017-11-08
Column elements should be extensible through the element type registry


tongyliu:
Columns are currently missing from the element type registry in the TypeScript renderer, which makes it difficult for clients to implement custom functionality for columns.

Ideally, there should be a registry entry for each type of element that is explicit in the schema (i.e. even though columns aren't considered standalone elements, they are still present in the schema with `"type": "Column"`, so it should be possible to register a custom class for that type).

dclaux:
Not sure I understand what the goal is. Could you clarify?

tongyliu:
The use case we ran into for Bing API cards was creating custom column elements that would be able to automatically hide child elements that overflow the column bounds (since our cards are fixed height). Since we're adding behavior to the column elements, while not changing the way they're laid out in the ColumnSet, it's useful to be able to extend columns separately.

dclaux:
Shouldn't that simply be the behavior of Column then? But I'm confused how you are able to accomplish the desired behavior anyway - for a column to know to not render some elements if they would be outside its boundaries, the column would need to know its height at the time its render method is called, no? But it doesn't, because its height is determined by its content. Furthermore, the card is renderered as a simple HTMLElement and it's only after that element is inserted in the DOM of the host that it gets a height. I'd be curious to understand your logic.

tongyliu:
You're right that it requires the element to have been inserted into the DOM -- which is why this particular use case also depends on the change that enables element render callbacks (#791). The way I currently have it working is that the overridden Column class renders elements with `visibility: hidden` in the initial render call, and registers a callback that goes through and marks elements that fit within the container visible after the Column has been added to the DOM.

I'd definitely be open to work with you guys on making this part of the standard Column behavior -- just wasn't sure initially if it was a large enough use case that you'd want to include it, so I went through the renderer's extensibility framework first.

tongyliu:
Closing this issue -- as discussed, we'll implement the desired truncation behavior as part of the renderer, so there's no longer a need to have columns be extensible.

PR #904, opened 2017-11-08
Add Column to element type registry

tongyliu:
This addresses #903.


dclaux:
I do not thing this change can go in. I do not understand what problem it solves, however it does _introduce_ a problem, namely the ability to use a Column outside a ColumnSet which is not desirable.

tongyliu:
Added a comment to the issue. We could potentially still prevent users from using Columns outside ColumnSets by having checks inside the actual Column class though.

tongyliu:
Closing, per discussion.

PR #911, opened 2017-11-09
Fix appending choices to existing ChoiceSetInput on ReplaceElement

tony-zhu:
New choices are appended to existing choices on ReplaceElement. Clear the old ones first.


matthidinger:
@dclaux if you're good with this change you can go ahead and merge it in. Tony needs to be added to the Microsoft org and he's currently oof

PR #949, opened 2017-11-14
fix some properties in host config don't accept 0 or false as value;

QingpingMeng:
fix some properties in host config don't accept 0 or false as value;
clear the array in parse function to avoid append due to multiple parse


QingpingMeng:
@dclaux Could you review this?

PR #950, opened 2017-11-15
Add errors/warnings to UWP Rendering layer

riarenas:
Parse errors/warning changes will come in separate PR for ease of reviewing. (And because I'm having some issues with it)


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -651,6 +659,16 @@ namespace AdaptiveCards { namespace Uwp
...
             }
+            else
+            {
+                HString elementTypeWrapper;
+                elementTypeWrapper.Set(elementType);
+                std::wstring errorString = L"No Renderer found for type: ";
+                errorString += elementTypeWrapper.GetRawBuffer(nullptr);
+                renderContext->AddError(
+                    FailureStatusCode::NoRendererForType,
+                    HStringReference(errorString.c_str()).Get());
```
riarenas: Taking suggestions for a prettier way to concatenate these strings :(

source/uwp/Renderer/lib/AdaptiveFailure.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "AdaptiveFailure.h"
+
+using namespace Microsoft::WRL;
+using namespace Microsoft::WRL::Wrappers;
+using namespace ABI::AdaptiveCards::Uwp;
+
+namespace AdaptiveCards { namespace Uwp
+{
+
+    HRESULT AdaptiveFailure::RuntimeClassInitialize()
```
riarenas: >AdaptiveFailure [](start = 12, length = 15)

Still looking for a better name than this :(

source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl
```diff
@@ -1945,6 +1976,28 @@ namespace AdaptiveCards
             [default] interface IAdaptiveRenderArgs;
         };
 
+        [
+            uuid(c0222295-bbec-4ff6-bdf3-f207ce78df69),
+            version(NTDDI_WIN10_RS1)
+        ]
+        interface IAdaptiveFailure : IInspectable
```
riarenas: >IAdaptiveFailure [](start = 18, length = 16)

Still looking for a better name.

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -651,6 +659,16 @@ namespace AdaptiveCards { namespace Uwp
...
             }
+            else
+            {
+                HString elementTypeWrapper;
+                elementTypeWrapper.Set(elementType);
+                std::wstring errorString = L"No Renderer found for type: ";
+                errorString += elementTypeWrapper.GetRawBuffer(nullptr);
+                renderContext->AddFailure(
+                    FailureStatusCode::NoRendererForType,
+                    HStringReference(errorString.c_str()).Get());
```
riarenas: Any suggestions on how to make this better?

source/uwp/Visualizer/ViewModel/DocumentViewModel.cs
```diff
@@ -179,6 +195,7 @@ public static void InitializeRenderer(AdaptiveHostConfig hostConfig)
 
                 // Custom resource resolvers
                 _renderer.ResourceResolvers.Set("symbol", new MySymbolResourceResolver());
+                _renderer.ElementRenderers.Remove("FactSet");
```
riarenas: >_renderer.ElementRenderers.Remove("FactSet"); [](start = 16, length = 45)

Revert. Had this to test the "No renderer for element" warning

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -699,4 +699,10 @@ HRESULT ProjectedElementTypeToHString(ABI::AdaptiveCards::Uwp::ElementType proje
 {
     CardElementType sharedElementType = static_cast<CardElementType>(projectedElementType);
     return UTF8ToHString(CardElementTypeToString(sharedElementType), result);
-}
\ No newline at end of file
+}
+
+bool IsWarning(int FailureStatusCode)
+{
+    int mask = 0xFF000000;
```
riarenas: >0xFF000000 [](start = 15, length = 10)

This should be 0xF0000000

source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl
```diff
@@ -1945,6 +1976,28 @@ namespace AdaptiveCards
...
+        {
+            [propget] HRESULT FailureStatusCode([out, retval] FailureStatusCode* value);
+            [propput] HRESULT FailureStatusCode([in] FailureStatusCode value);
```
riarenas: >propput [](start = 13, length = 7)

I think the putters are not needed. I'll see if I need them for once I finish integrating the shared model warnings/errors.

source/uwp/Renderer/idl/AdaptiveCards.Uwp.idl
RebeccaAnne: What if separated this out into warnings vs. errors everywhere?  We could make two separate interfaces/enums/add methods/etc. one for warnings and one for errors.  

---
In reply to: [151170456](https://github.com/Microsoft/AdaptiveCards/pull/950#discussion_r151170456) [](ancestors = 151170456)

source/uwp/Renderer/lib/XamlBuilder.cpp
RebeccaAnne: Not really, but elementType should be an HString to begin with, then you don't need the extra wrapper variable.

---
In reply to: [151178751](https://github.com/Microsoft/AdaptiveCards/pull/950#discussion_r151178751) [](ancestors = 151178751)

riarenas: Yep, good call.

---
In reply to: [151264474](https://github.com/Microsoft/AdaptiveCards/pull/950#discussion_r151264474) [](ancestors = 151264474,151178751)

source/uwp/Renderer/lib/Util.cpp
riarenas: Function is no longer needed

---
In reply to: [151202047](https://github.com/Microsoft/AdaptiveCards/pull/950#discussion_r151202047) [](ancestors = 151202047)

source/uwp/Renderer/lib/AdaptiveRenderContext.cpp
```diff
@@ -87,4 +110,6 @@ namespace AdaptiveCards { namespace Uwp
 
         return S_OK;
     }
+
+
```
RebeccaAnne: [nit] extra lines

PR #951, opened 2017-11-16
Update tools to use latest schema

danmarshall:
There is not any new code here, just the sample objects have been updated.


matthidinger:
This was merged in with #984

PR #952, opened 2017-11-16
Add parser errors to shared model

riarenas:
Wires up parser errors from the shared model to the UWP renderer. Warnings pending.

### DIFF
--- a/source/shared/cpp/ObjectModel/ParseUtil.h
+++ b/source/shared/cpp/ObjectModel/ParseUtil.h
@@ -138,6 +138,7 @@ T ParseUtil::GetEnumValue(const Json::Value& json, AdaptiveCardSchemaKey key, T 
     }
     catch (const std::out_of_range&)
     {
+        // TODO: Uncomment and add to warnings instead of throwing.
         // throw AdaptiveCardParseException("Enum type was out of range. Actual: " + propertyValueStr);
         return defaultEnumValue;
     }
--- a/source/shared/cpp/ObjectModel/ParseUtil.h
+++ b/source/shared/cpp/ObjectModel/ParseUtil.h
@@ -120,7 +120,7 @@ T ParseUtil::GetEnumValue(const Json::Value& json, AdaptiveCardSchemaKey key, T 
         {
             if (isRequired)
             {
-                throw AdaptiveCardParseException("Property is required but was found empty: " + propertyName);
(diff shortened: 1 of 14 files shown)


source/shared/cpp/ObjectModel/ParseUtil.h
```diff
@@ -130,15 +130,16 @@ T ParseUtil::GetEnumValue(const Json::Value& json, AdaptiveCardSchemaKey key, T
...
+            throw AdaptiveCardParseException(ErrorStatusCode::InvalidPropertyValue, "Enum type was invalid. Expected type string.");
         }
 
         propertyValueStr = propertyValue.asString();
         return enumConverter(propertyValueStr);
     }
     catch (const std::out_of_range&)
     {
-        throw AdaptiveCardParseException("Enum type was out of range. Actual: " + propertyValueStr);
+        // throw AdaptiveCardParseException("Enum type was out of range. Actual: " + propertyValueStr);
```
riarenas: This will now add a warning instead of throwing an error.

RebeccaAnne: Add a comment here saying that just in case.

---
In reply to: [151812267](https://github.com/Microsoft/AdaptiveCards/pull/952#discussion_r151812267) [](ancestors = 151812267)

source/uwp/Renderer/lib/AdaptiveCardParseResult.cpp
```diff
@@ -55,4 +55,11 @@ namespace AdaptiveCards { namespace Uwp
     {
         return m_warnings.CopyTo(value);
     }
+
+        _Use_decl_annotations_
+        HRESULT AdaptiveCardParseResult::put_AdaptiveCard(_In_ IAdaptiveCard* value)
+        {
+                m_adaptiveCard = value;
+                return S_OK;
+        }
```
RebeccaAnne: [nit] extra indenting

source/shared/cpp/ObjectModel/ParseUtil.h
riarenas: Yeah, I just remembered I forgot to add the comment so I added it to the PR so you wouldn't be surprised by it 

---
In reply to: [151813511](https://github.com/Microsoft/AdaptiveCards/pull/952#discussion_r151813511) [](ancestors = 151813511,151812267)

Issue #833, opened 2017-10-25
[UWP] Change namespace to AdaptiveCards.Rendering.Uwp


riarenas:
UWP implementation of #653 

Consider making the namespace a #define so that we can have a different namespace when making renderer drops into Windows.

matthidinger:
@riarenas this is done but so you want to use it to track the #define work? if not, feel free to close

PR #974, opened 2017-11-21
Change UWP renderer namespace to AdaptiveCards.Rendering.Uwp

riarenas:


riarenas:
Fixes #833

riarenas:
Shared model build is broke due to the solution missing some files. Not related with these changes. I address that break in my pull request to fix date / time

PR #990, opened 2017-11-27
Convert HTML in TextBlocks to Xaml runs

RebeccaAnne:
This handles processing html into XAML runs. I won't check this in until I've had a chance to validate against Joseph's work once that's available. In the meantime I included a test file which has the HTML I'm expecting in the pull request for reference. I'll remove that before checking in.


RebeccaAnne:
>     "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",

Included test file for reference.  Will remove before checking in.

---
Refers to: samples/v1.0/Elements/TextBlock.Markdown.Html.json:2 in 8e6f91c. [](commit_id = 8e6f91c91cc209e9d27b3a7dc31470063a6ef62e, deletion_comment = False)

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1123,6 +1123,243 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
         THROW_IF_FAILED(textConfig->get_MaxWidth(&maxWidth));
         StyleXamlTextBlock(textSize, textColor, containerStyle, Boolify(isSubtle), wrap, maxWidth, textWeight, xamlTextBlock, hostConfig);
     }
+
```
riarenas: [nit] extra whitespace

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1123,6 +1123,243 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
...
+        RETURN_IF_FAILED(GetTextFromXmlNode(node, text.GetAddressOf()));
+
+        ComPtr<ABI::Windows::UI::Xaml::Documents::IRun> localRun = XamlHelpers::CreateXamlClass<ABI::Windows::UI::Xaml::Documents::IRun>(HStringReference(RuntimeClass_Windows_UI_Xaml_Documents_Run));
+        RETURN_IF_FAILED(localRun->put_Text(text.Get()));
+
+        *run = localRun.Detach();
+        return S_OK;
+    }
+
+    HRESULT AddListInlines(
```
riarenas: >AddListInlines [](start = 12, length = 14)

[nit] Inline?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1123,6 +1123,243 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
...
+                wchar_t buffer[10];
+                swprintf_s(buffer, ARRAYSIZE(buffer), L"%d. ", iteration);
+
+                std::wstring numberElementString(buffer);
+                listElementString += numberElementString;
+            }
+            listElementString += text.GetRawBuffer(nullptr);
+
+            HString listElementHString;
+            RETURN_IF_FAILED(WindowsCreateString(listElementString.c_str(), listElementString.length(), listElementHString.GetAddressOf()));
```
riarenas: >WindowsCreateString [](start = 29, length = 19)

Would listElementHString.Set(listElementString.c_str()) work instead of calling WIndowsCreateString?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1123,6 +1123,243 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
...
+        RETURN_IF_FAILED(run.As(&runAsInline));
+
+        RETURN_IF_FAILED(inlines->Append(runAsInline.Get()));
+        return S_OK;
+    }
+
+    HRESULT SetTextOnXamlTextBlock(
+        IAdaptiveRenderContext* renderContext,
+        HSTRING textIn,
+        ITextBlock * textBlock)
```
riarenas: > ITextBlock * [](start = 7, length = 13)

[nit] spacing

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1123,6 +1123,243 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
...
+                if (isOrderedListResult == 0 || isUnorderedListResult == 0)
+                {
+                    nodeHandled = SUCCEEDED(AddListInlines(childNode.Get(), (isOrderedListResult == 0), inlines.Get()));
+                }
+                else if (isLinkResult == 0)
+                {
+                    nodeHandled = SUCCEEDED(AddLinkInline(childNode.Get(), inlines.Get()));
+                }
+                
+                if (!nodeHandled)
```
riarenas: >if (!nodeHandled) [](start = 16, length = 17)

What do you think of using an else if for this one as well? we can keep the !nodeHandled as an error case when one of the AddXXInline functions report failure. In this version if AddList or AddLink fail, it will still try to AddText

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1123,6 +1123,243 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
...
+                bool nodeHandled = false;
+                if (isOrderedListResult == 0 || isUnorderedListResult == 0)
+                {
+                    nodeHandled = SUCCEEDED(AddListInlines(childNode.Get(), (isOrderedListResult == 0), inlines.Get()));
+                }
+                else if (isLinkResult == 0)
+                {
+                    nodeHandled = SUCCEEDED(AddLinkInline(childNode.Get(), inlines.Get()));
+                }
+
```
riarenas: [nit] whitespace

source/uwp/Renderer/lib/XamlBuilder.cpp
riarenas: yep, makes sense.

source/uwp/Renderer/lib/XamlBuilder.cpp
RebeccaAnne: Yeah, that was my intent.  The scenario i ran into was that the link was missing it's href (i had manually added the HTML incorrectly).  At the time i was thinking that it would be better to display the malformed html as text than fail to render the card.  That said, now that i think about it the html is something we're generating (not a card author's mistake) so maybe it's fine to fail out here if it's bogus for some reason.

---
In reply to: [153663351](https://github.com/Microsoft/AdaptiveCards/pull/990#discussion_r153663351) [](ancestors = 153663351)

RebeccaAnne:
Closing this old PR out.  Will send an updated one soon.

PR #993, opened 2017-11-28
Added Javascript facade API (see README), cleaned up packaging for npm

matthidinger:


source/nodejs/adaptivecards/src/rendercard.ts
```diff
@@ -13,15 +13,11 @@ export function renderCard(card: IAdaptiveCard | string, options?: RenderOptions
     if (typeof card === "string") {
         card = <IAdaptiveCard>JSON.parse(card);
```
dclaux: What happens if this fails?

source/nodejs/adaptivecards/src/rendercard.ts
```diff
@@ -13,15 +13,11 @@ export function renderCard(card: IAdaptiveCard | string, options?: RenderOptions
...
 
-    let adaptiveCard = new AdaptiveCard();
-    adaptiveCard.parse(card);
-    adaptiveCard.onAction = options.onAction;
+    options = options || {};
 
+    // Parse the host config
     let hostConfig: HostConfig;
-
     if (isHostConfig(options.hostConfig)) {
```
dclaux: This seems weak. I don't think the isHostConfig method is necessary. Should be something like this:
```
if (typeof hostConfig === "string") {
    hostConfig = new HostConfig(JSON.parse(hostConfig));
}
else if (typeof hostConfig === "object" && !(hostConfig instanceof HostConfig)) {
    hostConfig = new HostConfig(hostConfig);
}
```

source/nodejs/adaptivecards/src/rendercard.ts
```diff
@@ -31,16 +27,22 @@ export function renderCard(card: IAdaptiveCard | string, options?: RenderOptions
...
         AdaptiveCard.processMarkdown = options.processMarkdown;
     }
-    
-    adaptiveCard.hostConfig = hostConfig;
+
+    // Render the card
     return adaptiveCard.render();
 }
 
 function isHostConfig(hostConfig: HostConfig | string | any): hostConfig is HostConfig {
```
dclaux: Should be removed

source/nodejs/adaptivecards/src/schema.ts
```diff
@@ -27,17 +27,22 @@ export interface IBackgroundImage {
     url: string;    
 }
 export interface IAdaptiveCard extends ITypedElement {
-    version?: IVersion | string,
-    actions?: (IActionBase)[];
+    type: string;
```
dclaux: ITypedElement already has a type property. Either remove ITypedElement or remove type for IAdapativeCard

source/nodejs/adaptivecards/src/schema.ts
```diff
@@ -27,17 +27,22 @@ export interface IBackgroundImage {
...
+    version?: IVersion | string;
     backgroundImage?: IBackgroundImage | string;
-    body?: (ITextBlock | IImage | IImageSet | IFactSet | IContainer | IColumnSet | IInputDate | IInputNumber | IInputText | IInputTime | IInputToggle)[];
+    body?: (any)[];
+    actions?: (IActionBase)[];    
     speak?: string;
-    title?: string;
+    [propName: string]: any;
```
dclaux: Why this?

source/nodejs/adaptivecards/src/schema.ts
```diff
@@ -27,17 +27,22 @@ export interface IBackgroundImage {
...
 }
 export interface ICardElement extends ITypedElement {
-    speak?: string;
+    type: string,
+    id?: string;
     horizontalAlignment?: HorizontalAlignment;
-    separation?: Separation;
+    spacing?: Separation;
```
dclaux: But that's not correct... Spacing is of type Spacing, not Separation (the type Separation is gone)

source/nodejs/adaptivecards/src/schema.ts
```diff
@@ -27,17 +27,22 @@ export interface IBackgroundImage {
...
+    separator?: boolean;
+    [propName: string]: any;
```
dclaux: Why this?

matthidinger: It signifies that extra properties are allowed on the interface. Per [TypeScript docs](https://www.typescriptlang.org/docs/handbook/interfaces.html)

"However, a better approach might be to add a string index signature if you’re sure that the object can have some extra properties that are used in some special way."

source/nodejs/adaptivecards/src/schema.ts
matthidinger: See below comment

source/nodejs/adaptivecards/src/schema.ts
dclaux: I get that, but why do we want this?

source/nodejs/adaptivecards/src/schema.ts
matthidinger: Good catch, fixed in latest commit

source/nodejs/adaptivecards/src/schema.ts
matthidinger: Because adaptive cards/elements allow extra properties to be included

dclaux: Oh, I didn't realize we officially allowed that.

matthidinger: Indeed :) they should be used very sparingly of course, but it's a supported construct. Let me know if you can approve the review

source/nodejs/adaptivecards/src/schema.ts
```diff
@@ -134,7 +133,7 @@ export type TextColor = "default" | "dark" | "light" | "accent" | "good" | "warn
 export type TextInputStyle = "text" | "tel" | "url" | "email";
 export type TextSize = "small" | "normal" | "medium" | "large" | "extraLarge";
 export type TextWeight = "lighter" | "normal" | "bolder";
-export type Separation = "none" | "default" | "strong";
+export type Spacing = "none" | "default" | "small" | "medium" | "large" | "extraLarge";
```
dclaux: All these enums are already declared in Enums, why do we need to re-declare them here, and in a different way for that matter? Let's use the real enums instead of duplicating. That applies to Spacing and all the others.

dclaux: And remove the locally defined enums from schema.ts

source/nodejs/adaptivecards/src/adaptivecards.ts
```diff
@@ -1,5 +1,5 @@
 export * from "./card-elements";
 export * from "./enums";
 export * from "./host-config";
-export { IAdaptiveCard, ICardElement } from "./schema";
+export { IAdaptiveCard, ICardElement, Spacing, HorizontalAlignment } from "./schema";
```
dclaux: There is no need to export these two enums given they already exist in enums.ts

matthidinger: Fixed

source/nodejs/adaptivecards/src/schema.ts
matthidinger: Fixed

PR #997, opened 2017-11-29
Show ellipses when truncating TextBlocks due to maxLines

tongyliu:
*This is an updated approach to PR #791 -- creating a new PR since the render callback behavior is no longer desired.*

Updates the TypeScript renderer so that multi-line `TextBlock` elements get truncated and show ellipses, instead of overflowing with no indication that there is more content. Uses the `updateLayout()` function to trigger truncation.

This approach includes the following changes:

- Changes the signature of the `AdaptiveCard.render()` function to take an optional `HTMLElement` argument, which the card is then rendered into (as opposed to just returning an `HTMLElement` containing the card -- we need at least some control over when the card is rendered into the DOM so we can execute JS afterwards)

  - This matches the way it's done in React with the `ReactDOM.render()` method, and abstracts the `element.appendChild` call away from users
 &nbsp;

- Updates the visualizer containers to match the new signature, and ensures that cards always get rendered into an element which is part of the DOM

- Adds a toggle so the JS truncation can be turned off if desired (e.g. in performance critical situations)

This addresses #779.

### DIFF
--- a/source/nodejs/adaptivecards-visualizer/src/containers/host-container.ts
+++ b/source/nodejs/adaptivecards-visualizer/src/containers/host-container.ts
@@ -237,7 +237,7 @@ export abstract class HostContainer {
         });
     }
 
-    protected renderContainer(adaptiveCard: AdaptiveCard, target:HTMLElement): HTMLElement {
+    protected renderContainer(adaptiveCard: AdaptiveCard, target: HTMLElement): HTMLElement {
         return null;
     }
 
--- a/source/nodejs/adaptivecards-visualizer/src/containers/host-container.ts
+++ b/source/nodejs/adaptivecards-visualizer/src/containers/host-container.ts
@@ -237,7 +237,7 @@ export abstract class HostContainer {
         });
     }
 
-    protected renderContainer(renderedCard: HTMLElement): HTMLElement {
+    protected renderContainer(adaptiveCard: AdaptiveCard, target:HTMLElement): HTMLElement {
         return null;
     }
 
@@ -304,22 +304,20 @@ export abstract class HostContainer {
         this.styleSheet = styleSheet;
     }
 
-    render(renderedCard: HTMLElement, speechString: string, showSpeechXml: boolean = false): HTMLElement {
+    render(adaptiveCard: AdaptiveCard, target: HTMLElement, showSpeechXml: boolean = false): HTMLElement {
         var element = document.createElement("div");
+        target.appendChild(element);
 
-        if (renderedCard) {
-            var renderedContainer = this.renderContainer(renderedCard);
+        if (adaptiveCard) {
+            var renderedContainer = this.renderContainer(adaptiveCard, element);
 
             if (renderedContainer) {
-                element.appendChild(renderedContainer);
-
                 var separator = document.createElement("div");
                 separator.style.height = "20px";
-
                 element.appendChild(separator);
             }
 
-            var renderedSpeech = this.renderSpeech(speechString);
+            var renderedSpeech = this.renderSpeech(adaptiveCard.renderSpeech());
 
             if (renderedSpeech) {
                 element.appendChild(renderedSpeech);
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -394,7 +394,7 @@ export class TextBlock extends CardElement {
     maxLines: number;
 
     private _computedLineHeight: number;
-    private _innerHtml: string;
+    private _originalInnerHtml: string;
 
     protected internalRender(): HTMLElement {
         if (!Utils.isNullOrEmpty(this.text)) {
@@ -532,7 +532,7 @@ export class TextBlock extends CardElement {
             }
 
             if (AdaptiveCard.useAdvancedTextBlockTruncation) {
-                this._innerHtml = element.innerHTML;
+                this._originalInnerHtml = element.innerHTML;
             }
 
             return element;
@@ -601,7 +601,7 @@ export class TextBlock extends CardElement {
         if (AdaptiveCard.useAdvancedTextBlockTruncation) {
             // Reset the element's innerHTML in case the available room for
             // content has increased
-            this.renderedElement.innerHTML = this._innerHtml;
+            this.renderedElement.innerHTML = this._originalInnerHtml;
             this.truncateIfSupported();
         }
     }
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -666,7 +666,8 @@ export class TextBlock extends CardElement {
                 if (fits()) {
                     bestBreakIdx = idx;
                     idx = TextBlock.findNextCharacter(fullText, idx);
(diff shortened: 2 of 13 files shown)


dclaux:
Looking pretty good at first glance. Will do a full review tomorrow. Thanks for implementing this new approach!

source/nodejs/adaptivecards-visualizer/src/containers/host-container.ts
```diff
@@ -237,7 +237,7 @@ export abstract class HostContainer {
         });
     }
 
-    protected renderContainer(renderedCard: HTMLElement): HTMLElement {
+    protected renderContainer(adaptiveCard: AdaptiveCard, target:HTMLElement): HTMLElement {
```
dclaux: Space missing after target:

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -520,14 +523,15 @@ export class TextBlock extends CardElement {
...
-                    element.style.maxHeight = (computedLineHeight * this.maxLines) + "px";
+                    element.style.maxHeight = (this.computedLineHeight * this.maxLines) + "px";
                     element.style.overflow = "hidden";
                 }
             }
             else {
                 element.style.whiteSpace = "nowrap";
             }
 
+            this.innerHtml = element.innerHTML;
```
dclaux: This shouldn't be necessary. Once a card element is rendered, its renderedElement property is set to the element returned by internalRender. So in updateLayout, you should simply be able to refer to this.renderedElement ansd you shouldn't need this private innerHtml member.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -589,6 +593,107 @@ export class TextBlock extends CardElement {
 
         return null;
     }
+
+    updateLayout(processChildren: boolean = false) {
+        // Reset the element's innerHTML in case the available room for content
+        // has increased
+        this.renderedElement.innerHTML = this.innerHtml;
```
dclaux: You should be able to use this.renderedElement.innerHtml instead.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -589,6 +593,107 @@ export class TextBlock extends CardElement {
...
+
+            if (truncationSupported) {
+                var element = <HTMLElement>children[0];
+                var maxHeight = this.computedLineHeight * this.maxLines;
+                this.truncate(element, maxHeight);
+            }
+        }
+    }
+
+    private truncate(element: HTMLElement, maxHeight: number) {
```
dclaux: Can we move this function to utils.ts?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -589,6 +593,107 @@ export class TextBlock extends CardElement {
...
+        var breakableIndices = TextBlock.findBreakableIndices(fullText);
+        var lo = 0;
+        var hi = breakableIndices.length;
+        var bestBreakIdx = 0;
+
+        // Do a binary search for the longest string that fits
+        while (lo < hi) {
+            var mid = Math.floor((lo + hi) / 2);
+            truncateAt(breakableIndices[mid]);
+            if (fits()) {
```
dclaux: Blank line above if

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -589,6 +593,107 @@ export class TextBlock extends CardElement {
...
+            }
+        }
+
+        truncateAt(bestBreakIdx);
+
+        // If we have extra room, try to expand the string letter by letter
+        // (covers the case where we have to break in the middle of a long word)
+        if (maxHeight - element.scrollHeight >= this.computedLineHeight - 1.0) {
+            let idx = TextBlock.findNextCharacter(fullText, bestBreakIdx);
+            while (idx < fullText.length) {
```
dclaux: Blank line above while

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -589,6 +593,107 @@ export class TextBlock extends CardElement {
...
+                truncateAt(idx);
+                if (fits()) {
+                    bestBreakIdx = idx;
+                    idx = TextBlock.findNextCharacter(fullText, idx);
+                } else {
```
dclaux: else on new line

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -589,6 +593,107 @@ export class TextBlock extends CardElement {
...
+                    break;
+                }
+            }
+            truncateAt(bestBreakIdx);
```
dclaux: Blank line before truncateAt

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -589,6 +593,107 @@ export class TextBlock extends CardElement {
...
+        }
+    }
+
+    private static findBreakableIndices(html: string): Array<number> {
```
dclaux: Can we move this to utils.ts?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -589,6 +593,107 @@ export class TextBlock extends CardElement {
...
+        var results: Array<number> = [];
+
+        var idx = TextBlock.findNextCharacter(html, -1);
+        while (idx < html.length) {
```
dclaux: Blank line above while

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -589,6 +593,107 @@ export class TextBlock extends CardElement {
...
+            if (html[idx] == ' ') {
+                results.push(idx);
+            }
+            idx = TextBlock.findNextCharacter(html, idx);
```
dclaux: Blank line above idx =

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -589,6 +593,107 @@ export class TextBlock extends CardElement {
...
+        }
+
+        return results;
+    }
+
+    private static findNextCharacter(html: string, currIdx: number): number {
```
dclaux: Can we move this to utils.ts?

source/nodejs/adaptivecards/src/card-elements.ts
tongyliu: Since `this.renderedElement` is a reference to the element, after the truncation code executes, `this.renderedElement.innerHTML` will reflect the truncated content. If the card layout then changes such that the available room for the TextBlock has increased, we need a way to bring the original content back.

dclaux: Makes sense. Maybe a better name would be _nonTruncatedInnerHtml or _originalInnerHtml or something of the sort, but that's a nit pick.

tongyliu:
Don't think I have write access to the repo. If someone who does could merge that'd be great 👍

dclaux:
Ah yes, sorry :) I'll merge.

PR #1003, opened 2017-11-29
Jwoo/markdown

nesalang:
Not everything flushed out, but functionality is there


source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/MarkDownUnitTest.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+using namespace Microsoft::VisualStudio::CppUnitTestFramework;
+using namespace AdaptiveCards;
+using namespace std;
+
+namespace AdaptiveCardsSharedModelUnitTest
+{
+    TEST_CLASS(EmphasisLeftDelimterTest)
+    {
+        TEST_METHOD(LefDelimiterTest)
```
riarenas: >LefDelimiterTest [](start = 20, length = 16)

Is lef shorthand for left? let's use the complete word.

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/MarkDownUnitTest.cpp
```diff
@@ -0,0 +1,318 @@
...
+        {
+            MarkDownParser blck("*foo bar*");
```
riarenas: >blck [](start = 27, length = 4)

Whis is this variable called blck? Let's try to use more meaningful names for variables.

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,430 @@
...
+#include <iostream>
+#include "MarkDownParser.h"
+
+using namespace AdaptiveCards;
+
+MarkDownParser::MarkDownParser (const std::string &txt) : m_currentDelimiterType (Init), 
+    m_currentWordIndex(0), m_lookBehind(Init), m_delimiterCnts(0), m_EmphasisState (EmphasisNotDetected), 
+    m_checkLookAhead(false), m_checkIntraWord(false), m_isRightEmphasisDetecting(false), m_text(txt), 
+    m_textBegin(m_text.begin()), m_curPos(m_text.begin()), m_textEnd(m_text.end()), m_LeftEmphasisDetecting(0),
+    m_leftLookUpTable (), m_rightLookUpTable (), m_tokenizedString(1, "")
```
riarenas: Consider initializing all these extra members in the header file.

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/MarkDownUnitTest.cpp
nesalang: thanks

---
In reply to: [154182001](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154182001) [](ancestors = 154182001)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/DateAndTimeUnitTest.cpp
```diff
@@ -1,6 +1,7 @@
 #include "stdafx.h"
 #include "CppUnitTest.h"
 #include "TextBlock.h"
+#include "MarkDownParser.h"
```
riarenas: This include doesn't seem necessary

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/DateAndTimeUnitTest.cpp
```diff
@@ -265,4 +266,5 @@ namespace AdaptiveCardsSharedModelUnitTest
 
         }
     };
```
riarenas: >}; [](start = 4, length = 2)

[nit] extra line

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -1,4 +1,4 @@
-#if defined(__ANDROID__) || (__APPLE__)
```
riarenas: >if [](start = 1, length = 2)

Am I missing where we use the markdown parser on a textblock? seems like it should be somewhere in this file.

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,430 @@
...
+        {
+            stk.push_back(top_left);
+            lookBehind |= (m_tokenizedString[top_left->m_idx][0] == '*')? Asterisk : Underscore; 
+            ++top_left;
+        }
+        else
+        {
+            curr_left_delim = stk.back();
+            // because of rule #9 & #10 and multiple of 3 rule, left delim jump ahead of right delim,
+            // so need to check this condition.
```
riarenas: There should be a comment at some point pointing to the rules you are referring to.

source/shared/cpp/ObjectModel/TextBlock.cpp
nesalang: I think Becky's change would address that.

---
In reply to: [154206301](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154206301) [](ancestors = 154206301)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/DateAndTimeUnitTest.cpp
nesalang: thanks!

---
In reply to: [154205464](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154205464) [](ancestors = 154205464)

source/shared/cpp/ObjectModel/TextBlock.cpp
riarenas: Becky's change addresses usage in the UWP layer. This change should probably include it for all shared model consumers.

---
In reply to: [154207709](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154207709) [](ancestors = 154207709,154206301)

nesalang: I agree, before I make changes here, let me check with Becky how she is using it, so I wouldn't interfere with her work.

---
In reply to: [154208020](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154208020) [](ancestors = 154208020,154207709,154206301)

source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
```diff
@@ -114,6 +114,7 @@
     </ClCompile>
   </ItemDefinitionGroup>
   <ItemGroup>
+    <ClInclude Include="..\..\shared\cpp\ObjectModel\MarkDownParser.h" />
```
riarenas: ><ClInclude Include="..\..\shared\cpp\ObjectModel\MarkDownParser.h" /> [](start = 4, length = 69)

This project does not include any shared object model files. Please add them to the ObjectModelProjection project instead.

nesalang: Thanks

---
In reply to: [154217414](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154217414) [](ancestors = 154217414)

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,84 @@
...
+    struct Emphasis
+    {
+        int m_emphCnts;
+        unsigned m_idx;
+        std::vector<int> m_tags;
+        Emphasis(unsigned int empCnts, unsigned idx) : 
+            m_emphCnts(empCnts), m_idx(idx), m_tags(){};
+    };
+
+    bool IsLeftEmphasisDelimiter(void);
```
riarenas: >void [](start = 33, length = 4)

Avoid using void inside parameter list.

bool function() is the preferred way.

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,84 @@
...
+    bool IsRightEmphasisDelimiter(void);
+    bool IsMarkDownDelimiter(char ch);
```
riarenas: These look like they could be private.

nesalang: that's right, thanks


---
In reply to: [154243800](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154243800) [](ancestors = 154243800)

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+MarkDownParser::MarkDownParser (const std::string &txt) : m_currentDelimiterType (Init), 
+    m_currentWordIndex(0), m_lookBehind(Init), m_delimiterCnts(0), m_EmphasisState (EmphasisNotDetected), 
+    m_checkLookAhead(false), m_checkIntraWord(false), m_isRightEmphasisDetecting(false), m_text(txt), 
+    m_textBegin(m_text.begin()), m_curPos(m_text.begin()), m_textEnd(m_text.end()), m_LeftEmphasisDetecting(0),
+    m_leftLookUpTable (), m_rightLookUpTable (), m_tokenizedString(1, "")
+{
+}
+
+int MarkDownParser::adjustDelimsCntsAndMetaData(int leftOver, std::list<Emphasis>::iterator left_itr,
```
RebeccaAnne: >Cnts [](start = 32, length = 4)

Please don't use abbreviations in function or variable names.  It makes the code harder to read.

RebeccaAnne: >adjustDelimsCntsAndMetaData [](start = 20, length = 27)

Could you please add a function comment explaining this method?

nesalang: sure

---
In reply to: [154402126](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154402126) [](ancestors = 154402126)

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,84 @@
...
+    enum EmphasisType
+    {
+        EmphasisItalic = 1,
+        EmphasisBold = 2,
+    };
+
+    enum EmphasisState
+    {
+        EmphasisDetected = 0x1,
+        EmphasisDetecting = 0x2,
```
RebeccaAnne: Can you add some comments here?  What is the difference between EmphasisDetected and EmphasisDetecting?

source/shared/cpp/ObjectModel/TextBlock.cpp
RebeccaAnne: Ricardo and I talked offline.  It does seems a little strange at first to not have it consumed in the text block, but since converting to HTML is more of an implementation detail than it is a functionality we want to expose on the textblock i'm fine with leaving this as is.

---
In reply to: [154208465](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154208465) [](ancestors = 154208465,154208020,154207709,154206301)

RebeccaAnne: For what it's worth, my consumption looks like this:

        MarkDownParser markdownParser(HStringToUTF8(textIn));
        auto htmlString = markdownParser.TransformToHtml();

        HString htmlHString;
        UTF8ToHString(htmlString, htmlHString.GetAddressOf());

        HRESULT hr = xmlDocumentIO->LoadXml(htmlHString.Get());

---
In reply to: [154409975](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154409975) [](ancestors = 154409975,154208465,154208020,154207709,154206301)

nesalang: understood & thanks

---
In reply to: [154409975](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154409975) [](ancestors = 154409975,154208465,154208020,154207709,154206301)

source/shared/cpp/ObjectModel/MarkDownParser.cpp
nesalang: added the additional comment at line 28-29

---
In reply to: [154207523](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154207523) [](ancestors = 154207523)

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,92 @@
...
+namespace AdaptiveCards
+{
+class MarkDownParser
+{
+public:
+    MarkDownParser (const std::string &txt) : m_currentDelimiterType (Init), 
+        m_currentWordIndex(0), m_lookBehind(Init), m_delimiterCnts(0), m_EmphasisState (EmphasisNotDetected), 
+        m_checkLookAhead(false), m_checkIntraWord(false), m_isRightEmphasisDetecting(false), m_text(txt), 
+        m_textBegin(m_text.begin()), m_curPos(m_text.begin()), m_textEnd(m_text.end()), m_LeftEmphasisDetecting(0),
+        m_leftLookUpTable (), m_rightLookUpTable (), m_tokenizedString(1, "") {};
```
riarenas: Sorry if there was some confusion in my previous comment.

Any members that are always initialized with the same values, should be initialized in the header during declaration:
m_checkLookAhead = false

Keep the members that depend on the passed in string, like m_text and m_textBegin in the cpp file.

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,92 @@
...
+    };
+
+    enum EmphasisState
+    {   
+        // state for the first encounter of emphasis delimiter 
+        EmphasisDetected = 0x1,  
+        // state for a emphasis delimiter run 
+        EmphasisDetecting = 0x2,   
+        EmphasisOn = EmphasisDetected | EmphasisDetecting,
+        // state for when a empahsis delimiter run is ended
```
riarenas: >empahsis [](start = 28, length = 8)

[nit] typo

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+        if (m_EmphasisState & EmphasisOn)
+        {
+            DelimiterType delimiterType = (*m_curPos == '*')? Asterisk : Underscore;
+
+            if (m_curPos != m_textBegin && (m_EmphasisState & EmphasisDetected))
+            {
+                m_tokenizedString.push_back("");
+                ++m_currentWordIndex;
+            }
```
RebeccaAnne: I realize that there are a lot of edge cases and complexities to this markdown parsing, but I'm finding this code generally difficult to read.  It would help both me as a code reviewer and any future developers who need to read this code if we could spend some time to make this code a little easier to understand.

More comments would help a lot:
- Since this is fairly non-intuitive code, a description at the top describing the overall functioning of the state machine at a high level might be helpful to give some context.
- Function level comments.  What's a symbol table and what is entailed in generating it?  
- Comments on blocks of code.  I can see that when we're in the EmphasisDetected state, but not at the beginning of the text, we're adding an empty string to the tokenized string and incrementing the word index.  I don't immediately know why.  What does adding an empty string to the tokenized string represent?  Why do we not need to do this if we're at the beginning of the string?  Why does this state indicate a new "Word"?

In addition to comments, think about how you're naming variables and functions.  In general, more descriptive is almost always better.  Also think about the meaning of the names you're giving them. 
This function is doing a lot of things, but it's not obviously updating or creating any variables with names like "symbol table".  A function comment will help, but making the code itself more obvious is even better.  Should this actually be GenerateTokenizedString?  Should some of the things being updated be more obviously named as symbol tables?  Is m_currentWordIndex actually about "words"?

Lastly, if you can thing of any ways that the structure of the code itself can be simplified or clarified that is the most helpful thing of all.  It may be the nature of the beast that this is complex code, but it's something to think about.  For example.  
-There are a lot of member variables representing state.  Do they all need to be member variables, or could they be local to the part of the code that needs them?  Ultimately there's only one function on the outside of this thing, so theoretically we shouldn't need to maintain a lot of state.  Declaring the variables close to where they're used makes the code easier to understand since we can get context on how they're used near the place where they're defined.  (P.S., what happens if you call TransformToHtml more than once?  does the state get reset?)
- Breaking out things like this into separate functions from TransformToHtml like you've done here helps, so this is great.  Are there any more parts of TransformToHtml that would make sense as helper functions?

I love that you've written so many tests for this thing so we have confidence that it works well, but code that's tricky makes me nervous even when we know it works, so the clearer we can be in the code itself the better.

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,425 @@
...
+bool MarkDownParser::IsLeftEmphasisDelimiter () 
+{
+    bool isLeftEmphasisDetecting = ((m_EmphasisState & ~EmphasisOn) && 
+                          m_delimiterCnts && 
+                          !isspace (*m_curPos) && 
+                          !(m_lookBehind == Alphanumeric && ispunct(*m_curPos)) && 
+                          !(m_lookBehind == Alphanumeric && m_currentDelimiterType == Underscore));
+    if (isLeftEmphasisDetecting)
+    {
+        m_leftLookUpTable.push_back(
```
RebeccaAnne: >m_leftLookUpTable [](start = 8, length = 17)

Functions called Is* that return a bool shouldn't have side effects.  Lets either rename this function to describe what it's actually doing (and make isLeftEmphasisDelimiter an out parameter) or move the part that's updating the member variables someplace else.

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,425 @@
...
+// It generally supports more stricker version of the rules
+std::string MarkDownParser::TransformToHtml()
+{
+    GenSymbolTable ();
+    if (!m_leftLookUpTable.size () || !m_rightLookUpTable.size ())
+    {
+        return "<p>" + m_text + "</p>";
+    }
+
+    std::vector<std::list<Emphasis>::iterator> stk;
```
RebeccaAnne: >stk [](start = 47, length = 3)

Give this a more descriptive name please.

source/shared/cpp/ObjectModel/MarkDownParser.cpp
nesalang: Thanks for the great advice!

---
In reply to: [154425566](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154425566) [](ancestors = 154425566)

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,425 @@
...
+
+            // emphasis found
+            if (delimCnts % 2)
+            {
+                curr_left_delim->m_tags.push_back(EmphasisItalic);
+                top_right->m_tags.push_back(EmphasisItalic);
+            }
+
+            // strong emphasis found
+            for (int rpts = 0; rpts < delimCnts / 2; rpts++)
```
RebeccaAnne: >rpts [](start = 21, length = 4)

rpts is repetitions?  Either give this a more descriptive name, or, alternatively, in the case of a for loop like this where the loop variable itself isn't too important it's ok to just call it "i"

source/uwp/AdaptiveCards.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
```
riarenas: Let's make sure that the changes to this file are not checked in.

source/shared/cpp/ObjectModel/MarkDownParser.cpp
nesalang: i'm working on init method, and will remove unnecessary member variables and use locals instead to increase readability. 

meanwhile, i will refactor to make it more readable. rules themselves were very complex, so it was difficult to come up with abstraction that simplifies the rules, but i do understand the rules better, so I will attempt to make it clearer . 

---
In reply to: [154427017](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154427017) [](ancestors = 154427017,154425566)

source/shared/cpp/ObjectModel/MarkDownParser.h
RebeccaAnne: Thanks!  Does this mean that if the state is EmphasisDetected that m_curPos is pointing at a either a * or a _?  And that EmphasisDetecting is in-between the start and end of an emphasis run?  Could we call them something EmphasisStart, Emphasis (InsideEmphasis?), EmphasisEnd, OutsideEmphasis?  The detected and detecting language is confusing to me, let me know if i'm missing a subtlety here.

---
In reply to: [154409330](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154409330) [](ancestors = 154409330)

source/uwp/AdaptiveCards.sln
nesalang: absolutely!

---
In reply to: [154430317](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154430317) [](ancestors = 154430317)

source/shared/cpp/ObjectModel/MarkDownParser.h
nesalang: you are absolutely correct. i'll make change; thank you!

---
In reply to: [154430904](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154430904) [](ancestors = 154430904,154409330)

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,96 @@
...
+    void GenSymbolTable();
+    void UpdateState();
+    int AdjustEmphasisCounts(int leftOver, std::list<Emphasis>::iterator left_itr, 
+            std::list<Emphasis>::iterator right_itr);
+
+    DelimiterType m_currentDelimiterType = Init;
+    unsigned int m_currentWordIndex = 0;
+    int m_lookBehind = Init;
+    int m_delimiterCnts = 0;
+    int m_EmphasisState = EmphasisNone;
```
riarenas: >m_EmphasisState [](start = 8, length = 15)

[nit] members should start with lowercase. m_emphasisState. Same for m_leftEmphasisDetecting

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,96 @@
...
+    bool m_checkLookAhead = false;
+    bool m_checkIntraWord = false;
+    int m_LeftEmphasisDetecting = false;
```
riarenas: >false [](start = 34, length = 5)

0?

nesalang: thanks

---
In reply to: [154471694](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154471694) [](ancestors = 154471694)

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,96 @@
...
+
+    std::string m_text;
+    std::string::iterator m_textBegin;
```
riarenas: >m_textBegin [](start = 26, length = 11)

Please correct me if I'm missing something, but it seems like you are only using m_textBegin inside the  GenSymbolTable method. You can just declare a local variable in that method instead of keeping it as a member of the class.

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,96 @@
...
+    std::string::iterator m_curPos;
+    std::string::iterator m_textEnd;
```
riarenas: >m_textEnd [](start = 26, length = 9)

You can remove this member, and always refer to m_text::end(), since I don't see any places where you are modifying the text.

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,478 @@
...
+    return delimCnts;
+}
+
+// Following the rules speicified in CommonMark (http://spec.commonmark.org/0.27/)
+// It generally supports more stricker version of the rules
+// push left delims to stack, until matching right delim is found,
+// update emphasis counts and type to build string after search is complete
+std::string MarkDownParser::TransformToHtml()
+{
+    GenSymbolTable ();
```
riarenas: >  [](start = 18, length = 1)

[nit] there shouldn't be a space between the function name and the args, I see that in multiple places.

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,477 @@
+#include <vector>
+#include <iomanip>
+#include <iostream>
+#include "MarkDownParser.h"
+
+using namespace AdaptiveCards;
+
+MarkDownParser::MarkDownParser(const std::string &txt) : m_text(txt), m_curPos(m_text.begin())
```
riarenas: >m_text.begin() [](start = 79, length = 14)

Since you only initialize the current position here, this means that, unless I'm missing something here, there will be problems when doing something like this:

std::string firstString = "*hello, world*;
auto parser = MarkdownParser(firstString);
std::string parsedString = parser.TransformToHtml();
std::string wrongString = parser.TransformToHtml();

Additionally, since you provide a SetText operation, it implies you can reuse the same parser for multiple strings, but that wouldn't work.

Something that could help to solve this would be to remove the m_curPos member altogether, and make the private helpers that need it receive it as an argument.

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,94 @@
...
+class MarkDownParser
+{
+public:
+    MarkDownParser(const std::string &txt); 
+
+    enum DelimiterType
+    {
+        Init,
+        Alphanumeric,
+        Puntuation,
```
riarenas: [nit] typo: punctuation

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,477 @@
...
+}
+
+// Following the rules speicified in CommonMark (http://spec.commonmark.org/0.27/)
+// It generally supports more stricker version of the rules
+// push left delims to stack, until matching right delim is found,
+// update emphasis counts and type to build string after search is complete
+std::string MarkDownParser::TransformToHtml()
+{
+    GenSymbolTable();
+    if (!m_leftLookUpTable.size() || !m_rightLookUpTable.size())
```
riarenas: .Size() returns an int. It's usually more readable to compare against 0 than to treat the result as a bool.

Since you don't really care about the actual size, but just want to check if the list is empty, you can use m_leftLookupTable.Empty()

source/shared/cpp/ObjectModel/MarkDownParser.cpp
riarenas: This comment was marked as resolved but there were no changes.

---
In reply to: [154426600](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154426600) [](ancestors = 154426600)

nesalang: it was changed at iteration 11.

---
In reply to: [154735329](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154735329) [](ancestors = 154735329,154426600)

riarenas: Sorry, missed it.

---
In reply to: [154736117](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154736117) [](ancestors = 154736117,154735329,154426600)

source/shared/cpp/ObjectModel/MarkDownParser.cpp
nesalang: i'm going to add dirty bit and init method. when once used, next time it get called, it will be initialized.

---
In reply to: [154718912](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154718912) [](ancestors = 154718912)

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,94 @@
...
+    unsigned int m_currentWordIndex = 0;
+    int m_lookBehind = Init;
+    int m_delimiterCnts = 0;
+    int m_emphasisState = EmphasisNone;
+    bool m_checkLookAhead = false;
+    bool m_checkIntraWord = false;
+    int m_LeftEmphasisDetecting = 0;
+
+    std::string m_text;
+    std::string::iterator m_curPos;
```
riarenas: >m_curPos [](start = 26, length = 8)

This class seems to have two different types of member variables. 

1. Members needed to generate the html, such as the lookup tables. These members deal directly with the main responsibility of this class, which is to parse the markdown into html.

2. Members that are only used during GenSymbolTable() and the helpers that are called from there. These members do not deal with the class' responsibility, and should be removed in favor of local variables declared and initialized during the symbol table generation. That means the class needs to hold less state, which will most likely make the code more readable and maintainable, as you will need to change most of the helpers to take whatever information they need to work as arguments.

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,96 @@
...
+    bool PushRightEmphasisToLookUpTableIfValid();
+    bool IsMarkDownDelimiter(char ch);
+    DelimiterType GetDelimiterTypeForCharAtCurrentPosition();
+    bool IsEmphasisDelimiterRun(MarkDownParser::DelimiterType delimiterRunType);
+    void UpdateCurrentEmphasisRunState(MarkDownParser::DelimiterType emphasisType);
+    void ResetCurrentEmphasisRunState(MarkDownParser::DelimiterType emphasisType);
+    void StartNewTokenCapture();
+    void GetCh(char ch);
+    void PutBackCh();
+    void GenSymbolTable();
```
riarenas: >GenSymbolTable [](start = 9, length = 14)

Let's use the full name for this. GenerateSymbolTable

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,94 @@
...
+
+    enum DelimiterType
+    {
+        Init,
+        Alphanumeric,
+        Puntuation,
+        Escape,
+        WhiteSpace,
+        Underscore = 0x8,
+        Asterisk   = 0x10,
```
RebeccaAnne: Why are these set to these values?  It looks like maybe you want a bitmask here, but i didn't notice a place where this value was being used as a bitmask.  If you need bitmask values, lets have all the values be bitmask values, otherwise just let these be sequential.

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,94 @@
...
+        // init state for emphasis detection 
+        EmphasisNone = 0x8,
+        OutsideEmphasis = EmphasisEnd | EmphasisNone,
+        EndOfWords = 0x10,
+        BeginOfWords = 0x20,
+    };
+
+    struct Emphasis
+    {
+        int m_emphCnts;
```
RebeccaAnne: >m_emphCnts [](start = 12, length = 10)

[nit] Please expand the variable names (m_emphasisCount?)

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,96 @@
...
```
riarenas: [nit] use the full name here.

source/shared/cpp/ObjectModel/MarkDownParser.h
nesalang: sure

---
In reply to: [154741027](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154741027) [](ancestors = 154741027)

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,477 @@
...
+}
+
+MarkDownParser::DelimiterType MarkDownParser::GetDelimiterTypeForCharAtCurrentPosition()
+{ 
+    return (*m_curPos == '*')? Asterisk : Underscore;
+}
+
+void MarkDownParser::StartNewTokenCapture()
+{
+    m_tokenizedString.push_back("");
```
RebeccaAnne: What does m_tokenizedString represent?  What does it mean to add an empty string to this vector?

nesalang: sorry, i will add comments. it contains tokens. adding empty string to it begins capturing new token. 

---
In reply to: [154743444](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154743444) [](ancestors = 154743444)

source/shared/cpp/ObjectModel/MarkDownParser.h
riarenas: >m_emphCnts [](start = 12, length = 10)

Since member visibility defaults to public for structs, struct members usually don't have an m_ prefix. Also, since the struct is called Emphasis, having the emph prefix in this particular member is not necessary. (and let's use the full name instead of cnts)

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,96 @@
...
+        unsigned m_idx;
```
riarenas: >m_idx [](start = 17, length = 5)

Would this make sense if renamed to position? if it doesn't, let's use the whole word: index

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -0,0 +1,96 @@
...
+    bool m_checkIntraWord = false;
+    int m_LeftEmphasisDetecting = false;
+
+    std::string m_text;
+    std::string::iterator m_textBegin;
+    std::string::iterator m_curPos;
+    std::string::iterator m_textEnd;
+    std::list<Emphasis> m_leftLookUpTable;
+    std::list<Emphasis> m_rightLookUpTable;
+    std::vector<std::string> m_tokenizedString = std::vector<std::string>(1, "");
```
riarenas: >m_tokenizedString [](start = 29, length = 17)

This is a collection of strings, therefore its name should be plural, and consider giving it a more representative name.

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/MarkDownUnitTest.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+namespace AdaptiveCardsSharedModelUnitTest
+{
+    TEST_CLASS(EmphasisLeftDelimterTest)
+    {
+        TEST_METHOD(LeftDelimiterTest)
+        {
+            MarkDownParser parser("*foo bar*");
+            Assert::AreEqual<string>("<p><em>foo bar</em></p>", parser.TransformToHtml());
+        }
```
riarenas: >} [](start = 8, length = 1)

Formatting: let's leave a blank line between test methods for easier readibility

source/shared/cpp/ObjectModel/MarkDownParser.h
nesalang: it stores index of token table, so it makes more sense to keep it index; i will change it to whole word.

---
In reply to: [154744159](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154744159) [](ancestors = 154744159)

source/shared/cpp/ObjectModel/MarkDownParser.h
nesalang: sounds good. thanks

---
In reply to: [154744055](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154744055) [](ancestors = 154744055)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/MarkDownUnitTest.cpp
```diff
@@ -0,0 +1,314 @@
...
+            MarkDownParser parser("5_6_78");
+            Assert::AreEqual<string>("<p>5_6_78</p>", parser.TransformToHtml());
+        }
+        TEST_METHOD(UnderscoreLeftDelimiterCanBeProceededAndFollowedByPunct)
+        {
+            MarkDownParser parser("foo-_(bar)_");
+            Assert::AreEqual<string>("<p>foo-<em>(bar)</em></p>", parser.TransformToHtml());
+        }
+    };
+    TEST_CLASS(EmphasisDelimterTest)
```
riarenas: [nit] typo: delimiter

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,477 @@
...
+    }
+    return false;
+}
+
+bool MarkDownParser::IsRightEmphasisDelimiter()
+{
+    if (isspace(*m_curPos) && 
+       (m_lookBehind != WhiteSpace) && 
+       (m_checkLookAhead || m_checkIntraWord || m_currentDelimiterType == Asterisk))
+    {
```
riarenas: >         [](start = 5, length = 8)

[nit] trailing whitespace

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -0,0 +1,477 @@
...
+
+            // found the left and right delimiter, gen string
+            if (curr_left_delim_type == curr_right_delim_type)
+            {
+                // rule #9 & #10, sum of delim cnts can't be multipe of 3 
+                if (!((curr_left_delim->m_emphCnts + top_right->m_emphCnts) % 3))
+                {
+                      if (m_tokenizedString.size() != top_right->m_idx + 1)
+                      {
+                          // check adjcent char is space
```
khouzam: >adjcent [](start = 35, length = 7)

Adjacent

nesalang: thanks!

---
In reply to: [154774189](https://github.com/Microsoft/AdaptiveCards/pull/1003#discussion_r154774189) [](ancestors = 154774189)

PR #1014, opened 2017-12-01
Use shared xamlBuilder in image renderer

### LINKED ISSUE #991: [UWP] Image renderer should not instantiate its own XamlBuilder.
Image renderer instantiates its own XamlBuilder member. This causes image tracking to not work correctly for the asynchronous versions of render, as well as causing the Synchronous versions to use the image downloader instead of letting xaml handle the image loading.

riarenas:
Image renderer needs to share the same XamlBuilder as the AdaptiveCardRenderer, otherwise the XamlBuilder's image tracking capabilities and the settings to allow xaml to load the images do not sync correctly.

Fixes #991

### DIFF
--- a/source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
+++ b/source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
@@ -920,9 +920,10 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
 
         HRESULT RenderAdaptiveCard([in] AdaptiveCard* adaptiveCard, [out, retval] RenderedAdaptiveCard** result);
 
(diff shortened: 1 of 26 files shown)


source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -920,9 +920,10 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
 
         HRESULT RenderAdaptiveCard([in] AdaptiveCard* adaptiveCard, [out, retval] RenderedAdaptiveCard** result);
 
+        // HRESULT RenderCardAsXamlAsync([in] AdaptiveCard* adaptiveCard, [out, retval] Windows.Foundation.IAsyncOperation<RenderedAdaptiveCard*>** result);
```
RebeccaAnne: >// HRESULT RenderCardAsXamlAsync([in] AdaptiveCard* adaptiveCard, [out, retval] Windows.Foundation.IAsyncOperation<RenderedAdaptiveCard*>** result); [](start = 8, length = 148)

Can we remove these commented out lines?

PR #1028, opened 2017-12-04
Truncate bottom overflows for fixed height cards

tongyliu:
Updates the TypeScript renderer to enable automatically truncating content that overflows the bottom of a fixed height card. The truncation logic is triggered in the `updateLayout()` method of the `AdaptiveCard` class, though it could be easily moved to the base container class if we ever want to be able to handle content overflows on any container, not just the bottom of the card. (This isn't really necessary at the moment since there's no notion of fixed height containers yet).

The default behavior for when an element overflows the bottom of the card is to hide the entire element. Two new methods are added to the `CardElement` class (`truncateOverflows()` and `undoOverflowTruncation()`) for elements that want to override this behavior (e.g. for TextBlocks, these methods are overridden so that text content can be partially clipped).

**Example:**

capture

Everything is hidden behind the `AdaptiveCard.useAdvancedCardBottomTruncation` flag (defaults to false, since this should only be enabled for fixed height cards).

This addresses #517.

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -443,7 +443,7 @@ export abstract class CardElement {
         // If the element is going to be hidden, reset any changes that were due
         // to overflow truncation (this ensures that if the element is later
         // un-hidden it has the right content)
-        if (!value) {
+        if (AdaptiveCard.useAdvancedCardBottomTruncation && !value) {
             this.undoOverflowTruncation();
         }
 
@@ -620,7 +620,8 @@ export class TextBlock extends CardElement {
                 element.style.whiteSpace = "nowrap";
             }
 
-            if (AdaptiveCard.useAdvancedTextBlockTruncation) {
+            if (AdaptiveCard.useAdvancedTextBlockTruncation
+                || AdaptiveCard.useAdvancedCardBottomTruncation) {
                 this._originalInnerHtml = element.innerHTML;
             }
 
@@ -3575,6 +3576,7 @@ export class AdaptiveCard extends ContainerWithActions {
 
     static preExpandSingleShowCardAction: boolean = false;
     static useAdvancedTextBlockTruncation: boolean = true;
+    static useAdvancedCardBottomTruncation: boolean = false;
 
     static readonly elementTypeRegistry = new ElementTypeRegistry();
     static readonly actionTypeRegistry = new ActionTypeRegistry();
@@ -3747,7 +3749,10 @@ export class AdaptiveCard extends ContainerWithActions {
 
     updateLayout(processChildren: boolean = true) {
         super.updateLayout(processChildren);
-        this.handleBottomOverflows();
+
+        if (AdaptiveCard.useAdvancedCardBottomTruncation) {
+            this.handleBottomOverflows();
+        }
     }
 
     canContentBleed(): boolean {
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -291,14 +291,13 @@ export abstract class CardElement {
         // Does nothing in base implementation
     }
 
-    handleOverflow(maxHeight: number) {
+    private handleOverflow(maxHeight: number) {
         if (this.isVisible || this.isHiddenDueToOverflow()) {
             var handled = this.truncateOverflows(maxHeight);
 
             // Even if we were unable to truncate the element to fit this time,
             // it still could have been previously truncated
-            this._truncatedDueToOverflow
-                = handled || this._truncatedDueToOverflow;
+            this._truncatedDueToOverflow = handled || this._truncatedDueToOverflow;
 
             if (!handled) {
                 this.hideElementDueToOverflow();
@@ -309,7 +308,7 @@ export abstract class CardElement {
         }
     }
 
-    resetOverflow(): boolean {
+    private resetOverflow(): boolean {
         var sizeChanged = false;
 
         if (this._truncatedDueToOverflow) {
@@ -325,8 +324,8 @@ export abstract class CardElement {
         return sizeChanged;
     }
 
-    isOnPage(): boolean {
-        return !!(this._renderedElement && this._renderedElement.offsetHeight);
+    isRendered(): boolean {
+        return this._renderedElement && this._renderedElement.offsetHeight > 0;
     }
 
     isAtTheVeryTop(): boolean {
@@ -688,8 +687,7 @@ export class TextBlock extends CardElement {
     }
 
     updateLayout(processChildren: boolean = false) {
-        if (AdaptiveCard.useAdvancedTextBlockTruncation
-            && this.maxLines && this.isOnPage()) {
+        if (AdaptiveCard.useAdvancedTextBlockTruncation && this.maxLines && this.isRendered()) {
             // Reset the element's innerHTML in case the available room for
             // content has increased
             this.restoreOriginalContent();
@@ -3377,12 +3375,13 @@ function raiseInlineCardExpandedEvent(action: ShowCardAction, isExpanded: boolea
 
 function raiseElementVisibilityChangedEvent(element: CardElement,
                                             shouldUpdateLayout: boolean = true) {
-    let card = element.getRootElement() as AdaptiveCard;
+    let rootElement = element.getRootElement();
 
     if (shouldUpdateLayout) {
-        card.updateLayout();
+        rootElement.updateLayout();
     }
 
+    let card = rootElement as AdaptiveCard;
     let onElementVisibilityChangedHandler = (card && card.onElementVisibilityChanged) ? card.onElementVisibilityChanged : AdaptiveCard.onElementVisibilityChanged;
 
     if (onElementVisibilityChangedHandler != null) {
@@ -3794,7 +3793,7 @@ export class AdaptiveCard extends ContainerWithActions {
     }
 
     private handleBottomOverflows() {
-        if (!this.isOnPage()) {
+        if (!this.isRendered()) {
             return;
         }
 
@@ -3810,7 +3809,7 @@ export class AdaptiveCard extends ContainerWithActions {
             if (elt) {
                 switch (Utils.getFitStatus(elt, boundary)) {
                     case Enums.ContainerFitStatus.FullyInContainer:
-                        let sizeChanged = cardElement.resetOverflow();
+                        let sizeChanged = cardElement['resetOverflow']();
                         // If the element's size changed after resetting content,
                         // we have to check if it still fits fully in the card
                         if (sizeChanged) {
@@ -3819,10 +3818,10 @@ export class AdaptiveCard extends ContainerWithActions {
                         break;
                     case Enums.ContainerFitStatus.Overflowing:
                         let maxHeight = boundary - elt.offsetTop;
-                        cardElement.handleOverflow(maxHeight);
+                        cardElement['handleOverflow'](maxHeight);
                         break;
                     case Enums.ContainerFitStatus.FullyOutOfContainer:
-                        cardElement.handleOverflow(0);
+                        cardElement['handleOverflow'](0);
                         break;
                 }
             }
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -3377,11 +3377,12 @@ function raiseInlineCardExpandedEvent(action: ShowCardAction, isExpanded: boolea
 
 function raiseElementVisibilityChangedEvent(element: CardElement,
                                             shouldUpdateLayout: boolean = true) {
+    let card = element.getRootElement() as AdaptiveCard;
+
     if (shouldUpdateLayout) {
-        element.getRootElement().updateLayout();
+        card.updateLayout();
     }
 
-    let card = element.getRootElement() as AdaptiveCard;
     let onElementVisibilityChangedHandler = (card && card.onElementVisibilityChanged) ? card.onElementVisibilityChanged : AdaptiveCard.onElementVisibilityChanged;
 
     if (onElementVisibilityChangedHandler != null) {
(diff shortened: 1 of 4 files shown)


matthidinger:
Nice!

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3377,11 +3377,12 @@ function raiseInlineCardExpandedEvent(action: ShowCardAction, isExpanded: boolea
 
 function raiseElementVisibilityChangedEvent(element: CardElement,
                                             shouldUpdateLayout: boolean = true) {
+    let card = element.getRootElement() as AdaptiveCard;
```
dclaux: Technically, the root element does not have to be a card. Your code shouldn't make this assumption.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -360,12 +436,19 @@ export abstract class CardElement {
...
+        return this._isVisible;
     }
 
     set isVisible(value: boolean) {
-        if (this._isVisibile != value) {
-            this._isVisibile = value;
+        // If the element is going to be hidden, reset any changes that were due
+        // to overflow truncation (this ensures that if the element is later
+        // un-hidden it has the right content)
+        if (!value) {
```
dclaux: Shouldn't you only do this if ```value != this._isVisible```?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -604,29 +687,52 @@ export class TextBlock extends CardElement {
     }
 
     updateLayout(processChildren: boolean = false) {
-        if (AdaptiveCard.useAdvancedTextBlockTruncation) {
+        if (AdaptiveCard.useAdvancedTextBlockTruncation
```
dclaux: The full if statement would easily fit on one line

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -257,6 +291,44 @@ export abstract class CardElement {
         // Does nothing in base implementation
     }
 
+    handleOverflow(maxHeight: number) {
+        if (this.isVisible || this.isHiddenDueToOverflow()) {
+            var handled = this.truncateOverflows(maxHeight);
+
+            // Even if we were unable to truncate the element to fit this time,
+            // it still could have been previously truncated
+            this._truncatedDueToOverflow
```
dclaux: Assignment should be on single line

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -257,6 +291,44 @@ export abstract class CardElement {
         // Does nothing in base implementation
     }
 
+    handleOverflow(maxHeight: number) {
```
dclaux: Too bad this is public... I know why, but it isn't great. When I hit those situations, I made the methods protected or private and I call them using the object["methodName"] pattern, which isn't great but a good approximation of what the "internal" modifier would be.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -257,6 +291,44 @@ export abstract class CardElement {
...
+            if (!handled) {
+                this.hideElementDueToOverflow();
+            }
+            else if (handled && !this._isVisible) {
+                this.showElementHiddenDueToOverflow();
+            }
+        }
+    }
+
+    resetOverflow(): boolean {
```
dclaux: Same here, ideally shouldn't be public.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -257,6 +291,44 @@ export abstract class CardElement {
...
+        }
+
+        if (this.isHiddenDueToOverflow) {
+            this.showElementHiddenDueToOverflow();
+        }
+
+        return sizeChanged;
+    }
+
+    isOnPage(): boolean {
```
dclaux: Should that be renamed "overflows" or "isOverflowing"?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -257,6 +291,44 @@ export abstract class CardElement {
...
+        return !!(this._renderedElement && this._renderedElement.offsetHeight);
```
dclaux: Is the "!!" a typo? If not, I am not familiar with what it means, if you don't mind educating me :)

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2860,6 +2966,10 @@ export class Container extends CardElement {
             invokeSetParent(this._selectAction, this);
         }
     }
+
+    get items(): Array<CardElement> {
```
dclaux: This can't exist. If we introduce this, developers will use it to add new elements, which they can't do because that doesn't properly set the added element's parent. Developers are supposed to use addItem to add elements. Please remove.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3171,6 +3281,10 @@ export class ColumnSet extends CardElement {
             invokeSetParent(this._selectAction, this);
         }
     }
+
+    get columns(): Array<Column> {
```
dclaux: Needs to be removed. Developers are supposed to call addColumn that properly sets the parent.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3628,9 +3745,75 @@ export class AdaptiveCard extends ContainerWithActions {
...
+    updateLayout(processChildren: boolean = true) {
+        super.updateLayout(processChildren);
+        this.handleBottomOverflows();
+    }
+
     canContentBleed(): boolean {
         return true;
     }
+
+    private getChildElements(): Array<CardElement> {
```
dclaux: There shouldn't be a method that needs to understand the internals of various element types. This doesn't work with extensibility; a developer won't be able to add a new element type that itself has children, because this method won't know how to gather its children.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3628,9 +3745,75 @@ export class AdaptiveCard extends ContainerWithActions {
...
+                    children.push(item);
+                }
+            }
+        };
+
+        addChildren(this);
+        return children;
+    }
+
+    private handleBottomOverflows() {
```
dclaux: This needs to be updated so that each element triggers the same logic in its own children instead of relaying on collecting all the elements beforehand. What you have now doesn't work with extensibility.

source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -238,3 +238,19 @@ function findNextCharacter(html: string, currIdx: number): number {
 
     return currIdx;
 }
+
+export function getFitStatus(element: HTMLElement,
```
dclaux: Fits on one line

source/nodejs/adaptivecards/src/card-elements.ts
tongyliu: This method actually doesn't check for overflow, it's just a check for whether the element has been added to the DOM yet.

source/nodejs/adaptivecards/src/card-elements.ts
tongyliu: It coerces `this._renderedElement && this._renderedElement.offsetHeight` into a boolean -- otherwise the result of evaluating that expression would be either false or a number.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Ah got it. How about "isRendered" then?

source/nodejs/adaptivecards/src/card-elements.ts
tongyliu: That's a good point. As it is now, new element types would be entirely hidden if they overflow the card, so I was originally thinking developers would override `truncateOverflows()` and `resetOverflowTruncation()` methods if their custom element has children -- but then I realized we don't need to treat that as a special case. I'll work on moving the logic into the `Container` and `ColumnSet` classes.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: So is it equivalent to ```this._renderedElement && this._renderedElement.offsetHeight > 0```? If so, can we use that syntax instead? I find the these kinds of syntax shortcuts difficult to read.
Thanks for the explanation, I leanred something new today.

tongyliu: Yep, that would work as well. Will update.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -325,8 +324,8 @@ export abstract class CardElement {
         return sizeChanged;
     }
 
-    isOnPage(): boolean {
-        return !!(this._renderedElement && this._renderedElement.offsetHeight);
+    isRendered(): boolean {
```
dclaux: Ah - now it looks like  this should be named isOverflowing, no?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3144,6 +3181,20 @@ export class ColumnSet extends CardElement {
         }
     }
 
+    protected truncateOverflow(maxHeight: number): boolean {
+        for (let column of this._columns) {
+            column['handleOverflow'](maxHeight);
+        }
+
+        return true;
```
dclaux: Is that right? As is, ColumnSet always reports it truncated overflow, which will not be true in many cases. Shouldn't it be true only if at least one column does truncate overflow?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2726,6 +2724,49 @@ export class Container extends CardElement {
         return element;
     }
 
+    protected truncateOverflow(maxHeight: number): boolean {
+        // Add 1 to account for rounding differences between browsers
+        var boundary = this.renderedElement.offsetTop + maxHeight + 1;
+        this.handleBottomOverflow(boundary);
```
dclaux: It doesn't look like handleBottomOverflow is called from any other place. As I was reading your changes, this indirection somewhat confused me. If this is indeed the only call site, let's move the code in handleBottomOverflow into truncateOverflow.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2726,6 +2724,49 @@ export class Container extends CardElement {
...
+        return true;
```
dclaux: Should be true only if at least one element in the container did get truncated, no?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3761,77 +3809,18 @@ export class AdaptiveCard extends ContainerWithActions {
...
         super.updateLayout(processChildren);
 
-        if (AdaptiveCard.useAdvancedCardBottomTruncation) {
-            this.handleBottomOverflows();
+        if (AdaptiveCard.useAdvancedCardBottomTruncation && this.isRendered()) {
+            var card = this.renderedElement;
+            var padding = this.hostConfig.getEffectivePadding(Enums.Padding.Default);
+            var boundary = card.offsetTop + card.offsetHeight - padding + 1;
+
+            this.handleBottomOverflow(boundary);
```
dclaux: Why not call this.truncateOverflow instead?

source/nodejs/adaptivecards/src/card-elements.ts
tongyliu: Still just a check that the card is in the DOM -- when `updateLayout()` gets called, we need to trigger the overflow handling logic regardless of whether or not the card actually does overflow, because it could be the case that there is now more space available for a previously truncated card.

source/nodejs/adaptivecards/src/card-elements.ts
tongyliu: With this implementation, Containers will always return true because they can always be truncated to fit any height (in the worst case, all their children get hidden). I'd prefer not to hide Containers as a whole, because then we have to deal with managing inherited visibility properties (tried this earlier, there's some cross-browser differences that are a pain).

source/nodejs/adaptivecards/src/card-elements.ts
tongyliu: As it is now, since Columns are Containers, which can always get truncated, this will always return true, even if we look at the truncation status for each column.

source/nodejs/adaptivecards/src/card-elements.ts
tongyliu: The thought was that calling `truncateOverflow` would be a bit of a misnomer because it also handles the case where content is expanded due to more room becoming available.

Containers are a bit different in that the code in `handleBottomOverflow` is optimized so that we don't actually need to call `resetOverflow` for them -- `handleBottomOverflow` expands the content if needed. This lets us avoid triggering multiple element visibility changed events in succession, where an element might become visible due to resetting overflow, but then immediately be hidden again.

I'll look into moving some stuff around to see if we can make this clearer.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: I'm not getting this. Your own comment reads "Child implementations should return true if the element was truncated to fit within maxHeight, false otherwise." Container always returning true means that the container will appear as always being truncated, when it should only appear as truncated if it truncated (or even fully hid) at least one of its children, no?

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Same question a for Container.

source/nodejs/adaptivecards/src/card-elements.ts
tongyliu: The key here is that `truncateOverflow` will only get called when there is a need for the container to truncate/hide at least one of its children, so keeping track of whether or not any children were actually hidden or truncated and then returning that value would be functionally equivalent to always returning true.

In any case, I think we can get around having to keep track of whether the element was *actually* truncated -- I'm using the return value from this method more as an indicator of whether the element should be hidden or not. We'll see in the next update.

dclaux:
Approved - let me know if it's ok to merge at this time or you plan other minor changes.

tongyliu:
Feel free to merge 👍

PR #1036, opened 2017-12-08
Initialize enum mappings on first call instead of using static initia…

riarenas:
…lizers


source/shared/cpp/ObjectModel/Enums.h
```diff
@@ -357,4 +357,38 @@ GenerateStringToEnumMap(const std::unordered_map<T, std::string, EnumHash>& keyT
     }
     return result;
 }
+
+void InitializeAdaptiveCardSchemaKeyEnumMappings();
```
RebeccaAnne: >InitializeAdaptiveCardSchemaKeyEnumMappings [](start = 5, length = 43)

We probably don't need these here in the header as they're not called outside of Enums.cpp

Issue #862, opened 2017-10-27
Make Container styles extensible


dclaux:
# Problem
Container styles are designed to solve one problem: make it possible for a container to have a background color while at the same time enforcing that text rendered on that container meets minimum contrast requirements for accessibility. Currently, Adaptive only supports two container styles, namely **default** and **emphasis**. It was always intended that this list would grow, the question is what is the right way to make it grow.

# Solution
There are two main approaches:
1. The Adaptive spec introduces additional container styles and all renderers/hosts support them
2. Or it is possible for a host to declare its own custom container styles in addition to the built-in ones

I propose we adopt **option 2** as it adds a lot of flexibility. Some hosts, like Outlook, will likely make extensive use of container styles and will want to have a somewhat wide selection available to card authors. These hosts shouldn't force other hosts to support all these styles, so an extensibility model seems like the right way to go. Note that this doesn't mean Adaptive can't add new built-in styles in the future.

This solution however leads to another question: should it be up to hosts to handle their custom Container styles whichever way they want, or should Adaptive provide the extensibility as a built-in feature. I propose that Adaptive exposes this as a built-in feature, via HostConfig, by adding a **customStyles** property to the containerStyles section:

```
containerStyles: {
    default: {
        backgroundColor: "#FFFFFF",
        fontColors: {
            default: { normal: "#333333", subtle: "#EE333333" },
            accent: { ... },
            ... abbreviated
        }
    },
    emphasis: {
        backgroundColor: "#08000000",
        fontColors: {
            default: { normal: "#333333", subtle: "#EE333333" },
            accent: { ... },
            ... abbreviated
        }
    },
    customStyles: [
        {
            name: "myCustomContainerStyle",
            style: {
                backgroundColor: "#FF0000",
                fontColors: {
                    default: { normal: "#333333", subtle: "#EE333333" },
                    accent: { ... },
                    ... abbreviated
                }
            }
        }
    ]
}
```

When applying a style to a container, a renderer would always prefer built-in styles (if a custom style is defined with the name **emphasis**, it will never be used because the emphasis style is explicitly defined) and would look up a style's definition by name in the **customStyles** collection if the style isn't a built-in one.

This change also requires that the Container.style property be of type string rather than the ContainerStyle enumeration.

PR #1046, opened 2017-12-14
Add support for custom Container styles in HostConfig

dclaux:
This change addresses https://github.com/Microsoft/AdaptiveCards/issues/862.

This is a capability of the TypeScript renderer and is independent of the Adaptive v1.0 schema. This change is required by Outlook.


source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -230,17 +195,140 @@ export class ActionsConfig {
...
+        }
+    }
+}
+
 export class ContainerStyleSet {
-    readonly default: ContainerStyleDefinition = new ContainerStyleDefinition();
-    readonly emphasis: ContainerStyleDefinition = new ContainerStyleDefinition();
+    private _default = new NamedContainerStyleDefinition(Enums.ContainerStyle.Default, true);
+    private _emphasis = new NamedContainerStyleDefinition(Enums.ContainerStyle.Emphasis, true);
+    private _allStyles: Array<NamedContainerStyleDefinition> = [];
```
QingpingMeng: Why not make this a dictionary?

dclaux: I have updated the code to use Map instead. Thanks!

dclaux:
I'll leave the "var"s for now because there are many in the whole code base, but agreed we should use "let" pretty much everywhere instead. I'll open an issue to change all var into lets.

Issue #1050, opened 2017-12-19
Text needs to be html escaped before being processed by the MarkDownParser


eschavez:
Example text: `3<5`
This needs to be escaped to: `3&lt;5`
Currently MarkdownParser takes text as is and would produce: `<p>3<5</p>` which is invalid html. Correct output should be: `<p>3&lt;5</p>`

eschavez:
Fixed and pushed to master

PR #1051, opened 2017-12-19
Escape text in MarkdownParser

eschavez:
Resolving #1050


source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -37,3 +37,32 @@ void MarkDownParser::ParseBlock()
...
+    {
+        switch (m_text[i])
+        {
+        case '<':
+            escaped += "&lt;";
+            break;
+        case '>':
+            escaped += "&gt;";
+            break;
+        case '"':
```
riarenas: >case [](start = 8, length = 4)

Do we need to escape single quotes ' at all?

PR #1063, opened 2018-01-04
Remove renderCard API

dclaux:
We have agreed that providing two ways of doing the same thing (e.g. using renderCard vs using the object model to render a card) was not desirable. Thus we are removing renderCard.


matthidinger:
Thanks David, are you planning to add the other enhancements we talked about to this PR? AdaptiveCard ctor to take a string or JSON object, and same with HostConfig ctor?

dclaux:
I'll try.

PR #1064, opened 2018-01-05
Switch visualizer editor from Ace to Monaco

dclaux:
Monaco is the editor used in VS Code. Among other things, it supports IntelliSense which will make it easier to author cards.

https://microsoft.github.io/monaco-editor/index.html


QingpingMeng:
Going back/forward button doesn't re-render the card using correct theme

dclaux:
That likely was already the case before, but I'll take a look.

QingpingMeng:
otherwise it looks good to me

dclaux:
Navigation is fixed

dclaux:
The code hasn't changed, it's the exact same but now hooked up to the Monaco editor. sessionStorage.getItem/setItem was throwing an exception for a reason I don't understand.

The difference is Monaco load asynchronously (to be honest it's real not that easy to code against, Ace was a lot better) when Ace initializes synchronously. The code to re-load the previous payload was in the callback that is invoked after Monaco has finished initializing; I moved it out, and it now works.

I do not know why the schema won't load. I will look into this next week.

dclaux:
It's very possible that this is simply a cross domain request problem, since the schema is hosted on adaptivecards.io

QingpingMeng:
it looks good to me

matthidinger:
The visualizer looks good. Seems some parts of the schema aren't compatible with their IntelliSense provider though -- everything in body isn't registering for me. the `actions` array works as expected though.  Feel free to merge, will have to investigate that stuff in a later PR

dclaux:
Correct - the schema doesn't seem to be up to date/finalized. I haven't touched it in this PR. I will merge these changes.

PR #1074, opened 2018-01-12
Jwoo/ios parsing rendering results

nesalang:


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
```diff
@@ -29,8 +31,17 @@ + (ACOAdaptiveCardParseResult *)fromJson:(NSString *)payload;
...
+        catch(const AdaptiveCardParseException& e)
         {
+            // converts AdaptiveCardParseException to NSError
+            ErrorStatusCode errorStatusCode = e.GetStatusCode();
+            NSInteger errorCode = (long)errorStatusCode;
+            NSString *errorMessage= [NSString stringWithCString:e.GetMessage().c_str()
+                                                  encoding:[NSString defaultCStringEncoding]];
+            NSError *parseError = [NSError errorWithDomain:ACRParseErrorDomain
+                                                      code:errorCode
+                                                  userInfo:@{NSLocalizedFailureReasonErrorKey:errorMessage}];
```
bviglietta: >NSLocalizedFailureReasonErrorKey:errorMessage [](start = 61, length = 45)

This error message isn't actually going to be localized.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCardParseResult.h
```diff
@@ -5,12 +5,14 @@
...
-#import "ACOParseResult.h"
+#import <Foundation/Foundation.h>
 
 @class ACOAdaptiveCard;
 
-@interface ACOAdaptiveCardParseResult:ACOParseResult
+@interface ACOAdaptiveCardParseResult:NSObject
 
 @property ACOAdaptiveCard *card;
+@property BOOL IsValid;
```
bviglietta: >IsValid [](start = 15, length = 7)

`isValid`, with a lowecase i, to conform to normal naming conventions.

You might also consider making all these fields readonly (and immutable in the case of parseErrors), since there's no reason for callers to modify these objects.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
bviglietta: The error string will still not be localized. Just calling NSLocalizedString doesn't localize the string for you.

Assuming that the Adaptive Cards team is not planning to localize its error messages any time soon, I would recommend not putting the error message into the localizedFailureReason property.

---
In reply to: [161899565](https://github.com/Microsoft/AdaptiveCards/pull/1074#discussion_r161899565) [](ancestors = 161899565)

nesalang: Understood, thanks Ben

---
In reply to: [161927305](https://github.com/Microsoft/AdaptiveCards/pull/1074#discussion_r161927305) [](ancestors = 161927305,161899565)

Issue #908, opened 2017-11-08
Date Time formatting final spec


matthidinger:
<table>
<tr>
<th>Release</th>
<th>Renderer status</th>
<th>Tasks status</th>
</tr>
<tr>
<td>Not scheduled
<td>✔️ .NET (#959)</td>
<td></td>
</tr>
</table>

# Implementation status

- [X] TypeScript (https://github.com/Microsoft/AdaptiveCards/pull/966)
- [x] UWP
- [ ] .NET (.NET uses DateTime.Parse and is therefore not strict enough)
- [x] iOS
- [x] Android
- [ ] Documentation

# Problem

Given that ISO8601 has many accepted string formats (with and without hyphens, colons, timezone offset vs UTC, etc) it makes sense for us to clearly define a strict `DATE`/`TIME` function format and ensure all renderers adhere to it. 

This will improve reliability for card authors and prevent unfortunate situations where a developer tests on one renderer and the date shows properly, but then the same payload goes to a different renderer and the user sees a weird {{DATE(....)}} string in the middle of the card.

## Language support

Additionally, we need to know the language of the card to properly format the LONG DATE. 

If the card is sent in English, but the user's device is in Spanish then the date formatting will take the device culture and format the long date into spanish, which will look really weird in the middle of english text.

Eg., this will look super weird to a user:

Payload | Output
---|---
|`"Your order was shipped {{DATE(2017-02-13T20:46:30Z, LONG)}}"` | "Your order was shipped Lunes, 13 de Febrero, 2017" 
 

# Specification

## AdaptiveCard `lang` attribute

- [ ]  Add a `lang` attribute to `AdaptiveCard`

Consensus is to add a top-level property to AdaptiveCard to support the language. This aligns with [HTML's design](https://www.w3.org/International/tutorials/language-decl/) of `<html lang="en">`

## Date/Time parsing spec 

1. **CASE SENSITIVE** (must be all caps)
1. **NO SPACES** between the `{{`, `}}`, or parentheses
1. **STRICT [RFC 3389](https://tools.ietf.org/html/rfc3339) FORMATTING** (See examples below)
1. **MUST BE** a valid date and time
1. **ALL FAILURES MUST** display the raw string in the card. No friendly message attempted. (The goal being to make the developer immediately aware there is a problem)

### Valid formats

The only accepted format follows [RFC 3389](https://tools.ietf.org/html/rfc3339)

[...]

PR #1102, opened 2018-02-05
Add language for Date formatting

almedina-ms:
Add lang tag for parsing date in different languages. The languages follow the names defined in the ISO 639 codes (shortest code)

Work for issue: https://github.com/Microsoft/AdaptiveCards/issues/908


dclaux:
As far as I know we haven't arrested the spec on the lang tag. @matthidinger what's the status of this? I believe you were on point to write the spec.

source/shared/cpp/ObjectModel/ParseUtil.cpp
```diff
@@ -334,6 +337,23 @@ std::vector<std::shared_ptr<BaseCardElement>> ParseUtil::GetElementCollection(
         {
             // Use the parser that maps to the type
             elements.push_back(parser->Deserialize(elementParserRegistration, actionParserRegistration, curJsonValue));
+            if (elements.back()->GetElementType() == CardElementType::TextBlock)
+            {
+                auto textBlock = std::dynamic_pointer_cast<TextBlock>(elements.back());
```
nesalang: i learned we can't use dynamic cast for codes bound for windows integration

almedina-ms: there are multiple uses of dynamic_pointer_cast in AdaptiveColumn, AdaptiveContainer, Util.cpp, etc. should we open an issue to fix those casts too?

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
```diff
@@ -85,6 +89,17 @@ std::shared_ptr<AdaptiveCard> AdaptiveCard::Deserialize(
         ParseUtil::GetString(json, AdaptiveCardSchemaKey::BackgroundImage);
     std::string speak = ParseUtil::GetString(json, AdaptiveCardSchemaKey::Speak);
     ContainerStyle style = ParseUtil::GetEnumValue<ContainerStyle>(json, AdaptiveCardSchemaKey::Style, ContainerStyle::None, ContainerStyleFromString);
+    std::string language = ParseUtil::GetString(json, AdaptiveCardSchemaKey::Language, true /* isRequired */);
+    
+    std::locale locale;
+    try
+    {
+        locale = std::locale(language);
```
nesalang: what happens when language string is not set? i think we should continue rendering using device's system setting. i think we have facility to propagate warnings for case like this and we could it.
Have you tested the changes on other platforms? language string is not standardized, so "en-US" may work with windows but may not work with iOS.  Correct string could be "en_US" in iOS.

source/shared/cpp/ObjectModel/ParseUtil.cpp
nesalang: if they were there before, dynamic_pointer_cast is probably o.k. to use. i meant it for dynamic_cast, you probably want to check it with Becky if you wanted to be sure.

---
In reply to: [166155402](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r166155402) [](ancestors = 166155402)

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
almedina-ms: in fact this works If the string passed is "en" (english) or "es" (spanish), I'm not sure if the same behavior works on Linux

almedina-ms: but that does remind me that I forgot to update the unit tests

dclaux: We have not fully speced that lang tag support. Namely, one of the things that remains to be speced is the format of the tag. What we do not want is to be in the business of mapping a set of language strings to platform-specific languages, on each platform. We should wait until the spec is ready before we implement anything IMO.

nesalang: what other platforms did you test beside windows? can locale object be instantiated using language tag alone?

---
In reply to: [166160065](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r166160065) [](ancestors = 166160065)

almedina-ms: @matthidinger

source/shared/cpp/ObjectModel/ParseUtil.cpp
RebeccaAnne: I believe we can't use dyanmic_pointer_cast in windows.  There are places where i used them before, but when Ricardo ported this code to windows he removed them from the windows version. As of right now i believe that the windows version of the code is different than the github version of the code, which is why you still see them in github.

---
In reply to: [166159978](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r166159978) [](ancestors = 166159978,166155402)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -145,6 +149,41 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+}
+
+std::wstring TextBlock::StringToWstring(const std::string& in) const
+{
+#ifdef _WIN32
+    std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>, wchar_t> utfConverter;
+    return utfConverter.from_bytes(in);
+#else
+    std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>, wchar_t> utfConverter;
+    return utfConverter.from_bytes(input);
```
RebeccaAnne: >input [](start = 35, length = 5)

in?  Can you give me the background on why we need these ifdef'd versions?

source/shared/cpp/ObjectModel/Container.cpp
```diff
@@ -56,6 +57,26 @@ void Container::SetSelectAction(const std::shared_ptr<BaseActionElement> action)
...
+    {
+        CardElementType elementType = item->GetElementType();
+
+        if (elementType == CardElementType::Container)
+        {
+            auto element = std::dynamic_pointer_cast<Container>(item);
+            element->SetLanguage(value);
+        }
+
+        if (elementType == CardElementType::TextBlock)
```
RebeccaAnne: >if [](start = 8, length = 2)

[nit] could be else if

source/shared/cpp/ObjectModel/TextBlock.cpp
almedina-ms: The wstrings differ by size in different platforms, in ios, Linux and android the wstring characters are 4 bytes in length, while the windows ones are 2 bytes long. 
https://stackoverflow.com/questions/7232710/convert-between-string-u16string-u32string

RebeccaAnne: In this function these look the same except for the parameter name...

---
In reply to: [166771554](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r166771554) [](ancestors = 166771554)

source/shared/cpp/ObjectModel/TextBlock.cpp
```diff
@@ -145,6 +149,41 @@ void TextBlock::SetHorizontalAlignment(const HorizontalAlignment value)
...
+#endif
+}
+
+std::string TextBlock::WstringToString(const std::wstring& input) const
+{
+#ifdef _WIN32
+    if (sizeof(wchar_t) == 2)
```
RebeccaAnne: >if (sizeof(wchar_t) == 2) [](start = 4, length = 25)

If you're checking the size manually anyway, do you still need the #ifdef?

source/shared/cpp/ObjectModel/TextBlock.h
```diff
@@ -50,6 +51,11 @@ class TextBlock : public BaseCardElement
     HorizontalAlignment GetHorizontalAlignment() const;
     void SetHorizontalAlignment(const HorizontalAlignment value);
 
+    void SetLanguage(const std::locale& value);
+
+    std::wstring StringToWstring(const std::string& in) const;
+    std::string WstringToString(const std::wstring& in) const;
```
RebeccaAnne: Do these need to be public? Also, since these are generic utility functions and not really part of TextBlock consider moving them to ParseUtil.

source/shared/cpp/ObjectModel/TextBlockText.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+#include <string>
+#include <vector>
+#include "Enums.h"
+#include "TextSection.h"
+#include "TextSection.h"
+
+namespace AdaptiveCards
+{   
+    class TextBlockText
```
nesalang: name is confusing.
DateTimeFormatter may be?

source/shared/cpp/ObjectModel/TextBlockText.cpp
```diff
@@ -0,0 +1,243 @@
...
+            // measured from year 1900
+            parsedTm.tm_year -= 1900;
+            parsedTm.tm_mon -= 1;
+
+            time_t utc{};
+            // converts to ticks in UTC
+            utc = mktime(&parsedTm);
+            if (utc == -1)
+            {
+                AddTextSection(matches[0], TextSectionFormat::RegularString);
```
nesalang: Is this needed to be saved?

source/shared/cpp/ObjectModel/TextBlockText.cpp
```diff
@@ -0,0 +1,243 @@
...
+            bool formatHasSpace = matches[Format].str()[1] == ' ';
+            int formatStartIndex = formatHasSpace ? 2 : 1;
+            formatStyle = matches[Format].str()[formatStartIndex];
+        }
+
+        AddTextSection(matches.prefix().str(), TextSectionFormat::RegularString);
+
+        if (!isDate && formatStyle)
+        {
+            AddTextSection(matches[0].str(), TextSectionFormat::RegularString);
```
nesalang: Is this needed to be saved?

source/shared/cpp/ObjectModel/TextSection.h
```diff
@@ -0,0 +1,32 @@
...
+{
+    class TextSection
+    {
+    public:
+        TextSection();
+        TextSection(std::string text, TextSectionFormat format);
+        TextSection(std::string text, std::string originalText, TextSectionFormat format);
+
+        std::string GetText() const;
+        std::string GetOriginalText() const;
```
nesalang: do we need to make distinction between original text and text?
get text will return original text even if parsing validation fails, right?

source/shared/cpp/ObjectModel/TextBlockText.h
almedina-ms: I agree that the name should be changed for TextBlockText and TextSection but I think DateTimeFormatter is not a good name as this class doesn't format the date and time, it just splits the text and finds possible matches for a date or time

source/shared/cpp/ObjectModel/TextSection.h
```diff
@@ -0,0 +1,32 @@
...
+        TextSectionFormat GetFormat() const;
+        // Date format should be: MM/DD/YYYY
+        int GetDay() const;
```
nesalang: why do we need day/month/year in integer?

source/uwp/Renderer/lib/DateTimeParser.cpp
```diff
@@ -0,0 +1,74 @@
...
+    {
+        struct tm result{};
+        result.tm_mday = textSection->GetDay();
+        result.tm_mon = textSection->GetMonth();
+        result.tm_year = textSection->GetYear() >= 1900 ? textSection->GetYear() - 1900 : 0;
+
+        switch (textSection->GetFormat())
+        {
+            case TextSectionFormat::DateCompact:
+
```
nesalang: nit: extra space

source/uwp/Renderer/lib/DateTimeParser.cpp
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+std::string DateTimeParser::GenerateString(TextBlockText text)
+{
+    std::wostringstream parsedostr;
+
+    for (const auto& textSection : text.GetString())
+    {
+        struct tm result{};
```
nesalang: UWP doesn't have high level date and time formatter that use string format?

source/shared/cpp/ObjectModel/TextSection.h
almedina-ms: it's easier to retrieve them if the format is not as expected, also, in the case of c++ the date formatting needs a tm struct that needs to be filled by hand so having this deal with the day month year splitting is easier

source/shared/cpp/ObjectModel/TextBlockText.cpp
almedina-ms: This is saved if the date complies with a correct formatting (regex) but the make time function fails to generate the date as expected, so the original string is saved to be printed

nesalang: i think localtime function will fail? if utc == -1, and it can be caught in the else statement?

---
In reply to: [168269856](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r168269856) [](ancestors = 168269856)

source/shared/cpp/ObjectModel/TextBlockText.cpp
```diff
@@ -0,0 +1,243 @@
...
+                // localtime() set dst, put_time adjusts time accordingly which is not what we want since 
+                // we have already taken cared of it in our calculation
+                if (result.tm_isdst == 1)
+                {
+                    result.tm_hour -= 1;
+                }
+
+                if (isDate)
+                {
+                    std::string plainDate = std::to_string(result.tm_mon) + dateDelimiter +
```
nesalang: consider using std::put_time(&result, "%Ex").

source/shared/cpp/ObjectModel/TextSection.h
nesalang: if uwp doesn't have a date and time formatter, you can consider saving tm struct saved from textblocktext.


---
In reply to: [168269415](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r168269415) [](ancestors = 168269415)

almedina-ms: the only date time formatter I found for uwp was the COleDateTime but what I get from the format functions is that it's either language or short and long format

 [https://docs.microsoft.com/en-us/cpp/atl-mfc-shared/reference/coledatetime-class#format](https://docs.microsoft.com/en-us/cpp/atl-mfc-shared/reference/coledatetime-class#format)

I think I'll end up saving the tm struct in the TextSection class then

---
In reply to: [168272931](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r168272931) [](ancestors = 168272931,168269415)

source/shared/cpp/ObjectModel/TextBlockText.cpp
almedina-ms: in the case of c++ the compact date can be localized (I guess it should change the month/day/year order) but I will test it

---
In reply to: [168270900](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r168270900) [](ancestors = 168270900)

almedina-ms: Yes, the %Ex can be localized, for example, in English the date February 14th, 2018 is represented as 02/14/2018 in Spanish it's represented as 14/02/2018

source/shared/cpp/ObjectModel/TextBlockText.h
almedina-ms: We have the names DateTimePreparser and DateTimePreparserToken right now, if someone has a better name I can change them

source/shared/cpp/ObjectModel/TextBlockText.cpp
almedina-ms: we used to append it to the output, so I guess it must be saved

source/shared/cpp/ObjectModel/Container.cpp
```diff
@@ -56,6 +57,28 @@ void Container::SetSelectAction(const std::shared_ptr<BaseActionElement> action)
     m_selectAction = action;
 }
 
+void Container::SetLanguage(const std::string& value)
+{
+    for (auto& item : m_items)
+    {
+        CardElementType elementType = item->GetElementType();
+
+        if (elementType == CardElementType::Container)
```
RebeccaAnne: >Container [](start = 44, length = 9)

Do we also need to handle ColumnSet?

source/shared/cpp/ObjectModel/ParseUtil.cpp
```diff
@@ -334,6 +338,25 @@ std::vector<std::shared_ptr<BaseCardElement>> ParseUtil::GetElementCollection(
...
+            {
+                if (elements.back()->GetElementType() == CardElementType::Container)
+                {
+                    auto container = std::static_pointer_cast<Container>(elements.back());
+                    if (container != nullptr)
+                    {
+                        container->SetLanguage(dateLanguage);
+                    }
+                }
+            }
```
RebeccaAnne: This is the same as the code in Container.cpp, can we make this chunk of code a helper function?

RebeccaAnne: Alternatively, instead of doing this here, how about just setting the language on the card at the end, and letting the SetLanguage() methods handle propagating to the elements?

---
In reply to: [170750454](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r170750454) [](ancestors = 170750454)

source/uwp/Renderer/lib/DateTimeParser.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>, wchar_t> utfConverter;
+        return utfConverter.to_bytes(input);
+    }
+    else
+    {
+        std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> utfConverter;
+        return utfConverter.to_bytes(input);
+    }
+}
```
RebeccaAnne: I'd put these methods in util.cpp with the other string conversion functions.

source/shared/cpp/ObjectModel/Container.cpp
```diff
@@ -56,6 +57,28 @@ void Container::SetSelectAction(const std::shared_ptr<BaseActionElement> action)
...
+        {
+            auto element = std::static_pointer_cast<Container>(item);
+            element->SetLanguage(value);
+        }
+        else
+        {
+            if (elementType == CardElementType::TextBlock)
```
RebeccaAnne: just make this:
else if (elementType == CardElementType::TextBlock)

source/uwp/Renderer/lib/AdaptiveTextBlock.cpp
```diff
@@ -33,7 +34,8 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
     _Use_decl_annotations_
     HRESULT AdaptiveTextBlock::get_Text(HSTRING* text)
     {
-        return UTF8ToHString(m_sharedTextBlock->GetText(), text);
+        DateTimeParser parser(m_sharedTextBlock->GetLanguage());
+        return UTF8ToHString(parser.GenerateString(m_sharedTextBlock->GetTextForDateParsing()), text);
```
RebeccaAnne: >return UTF8ToHString(parser.GenerateString(m_sharedTextBlock->GetTextForDateParsing()), text); [](start = 8, length = 94)

Are we sure we want to do this here and not as part of rendering? I'm worried about whether we want someone who calls get_Text on the API to get the raw text or the already processed date. I can't actually think of a scenario where you want to read out the raw text, but it makes me a bit nervous that it's not possible to do so. Maybe double check with Matt if you haven't already just in case.

source/shared/cpp/ObjectModel/Container.cpp
almedina-ms: aside from containers, show card actions and column sets, are there some other elements that can have a Text block inside of it?

---
In reply to: [170754476](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r170754476) [](ancestors = 170754476)

source/shared/cpp/ObjectModel/Container.cpp
almedina-ms: aside from Container, Show Card Actions and Column Sets, are there other places where a TextBlock could be inserted?

---
In reply to: [170748153](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r170748153) [](ancestors = 170748153)

RebeccaAnne: I think that's it

---
In reply to: [170760621](https://github.com/Microsoft/AdaptiveCards/pull/1102#discussion_r170760621) [](ancestors = 170760621,170748153)

source/shared/cpp/ObjectModel/Column.cpp
```diff
@@ -133,4 +136,28 @@ std::shared_ptr<BaseActionElement> Column::GetSelectAction() const
...
+            auto element = std::static_pointer_cast<ColumnSet>(item);
+            element->SetLanguage(language);
+        }
+        else if (elementType == CardElementType::TextBlock)
+        {
+            auto element = std::static_pointer_cast<TextBlock>(item);
+            element->SetLanguage(language);
+        }
+    }
 }
```
RebeccaAnne: We have this same chunk of code in three places (here, container, and card).  Can we make this a helper function?

source/uwp/Renderer/lib/AdaptiveTextBlock.cpp
```diff
@@ -143,6 +144,21 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
         return S_OK;
     }
 
+    _Use_decl_annotations_
+        HRESULT AdaptiveTextBlock::get_Language(HSTRING* language)
```
RebeccaAnne: >         [](start = 0, length = 8)

[nit] Extra space here and on 154

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -740,4 +749,24 @@ HRESULT ProjectedElementTypeToHString(ABI::AdaptiveCards::Rendering::Uwp::Elemen
...
+
+std::wstring StringToWstring(const std::string& in) 
+{
+    std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>, wchar_t> utfConverter;
+    return utfConverter.from_bytes(in);
+}
+
+std::string WstringToString(const std::wstring& input)
+{
+    if (sizeof(wchar_t) == 2)
```
RebeccaAnne: Can you add a comment here so it's clear to future developers why we need this check?

PR #1113, opened 2018-02-07
Convert HTML in TextBlocks to Xaml runs

RebeccaAnne:
Reviving this pull request to consume the markdown parser in UWP.


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1126,6 +1127,253 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
...
+                    {
+                        RETURN_IF_FAILED(AddLinkInline(childNode.Get(), inlines.Get()));
+                    }
+                    else
+                    {
+                        RETURN_IF_FAILED(AddTextInline(renderContext, childNode.Get(), (isBoldResult == 0), (isItalicResult == 0), inlines.Get()));
+                    }
+
+                    ComPtr<ABI::Windows::Data::Xml::Dom::IXmlNode> nextChildNode;
+                    RETURN_IF_FAILED(childNode->get_NextSibling(&nextChildNode));
```
nesalang: i don't know how exactly xml parses html... what happens if html tag has child tag for example <link tag "<- i just made it up" > <bold> "hello wordl" </bold> </link tag> will get_NextSibling returns correct node?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1126,6 +1127,253 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
...
+            ComPtr<IAdaptiveHostConfig> hostConfig;
+            RETURN_IF_FAILED(renderContext->get_HostConfig(&hostConfig));
+
+            ComPtr<IAdaptiveFontWeightsConfig> fontWeightsConfig;
+            RETURN_IF_FAILED(hostConfig->get_FontWeights(&fontWeightsConfig));
+
+            ABI::Windows::UI::Text::FontWeight boldFontWeight;
+            RETURN_IF_FAILED(fontWeightsConfig->get_Bolder(&boldFontWeight.Weight));
+
+            RETURN_IF_FAILED(runAsTextElement->put_FontWeight(boldFontWeight));
```
nesalang: just to be sure, <bold><bold>"hello"</bold></bold> will work, right? adding two fontweight won't add the weight, right? thanks for entertaining silly question.

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1126,6 +1127,253 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
...
+
+            std::wstring listElementString = L"\n";
+            if (!isListOrdered)
+            {
+                listElementString += L"● ";
+            }
+            else
+            {
+                wchar_t buffer[10];
+                swprintf_s(buffer, ARRAYSIZE(buffer), L"%d. ", iteration);
```
nesalang: iteration is user defined. ordered list can have user defined starting number

source/uwp/Renderer/lib/XamlBuilder.cpp
RebeccaAnne: Assuming you mean "strong", that example makes "hello" be in bold, which i think is what's supposed to happen.

What doesn't work though is <strong> <em> "hello" </em></strong> which should be both bold and italic and only turns out bold...

---
In reply to: [167377400](https://github.com/Microsoft/AdaptiveCards/pull/1113#discussion_r167377400) [](ancestors = 167377400)

RebeccaAnne: Reworked this a bit to make sure that the recursion works properly and my second example now works.

---
In reply to: [167381058](https://github.com/Microsoft/AdaptiveCards/pull/1113#discussion_r167381058) [](ancestors = 167381058,167377400)

source/uwp/Renderer/lib/XamlBuilder.cpp
RebeccaAnne: Fixed this! Do we support the "type" attribute at all?

---
In reply to: [167377890](https://github.com/Microsoft/AdaptiveCards/pull/1113#discussion_r167377890) [](ancestors = 167377890)

source/uwp/Renderer/lib/XamlBuilder.cpp
RebeccaAnne: I think your concern is fixed by my rework. Let me know if you still have concerns on the new iteration.

---
In reply to: [167377042](https://github.com/Microsoft/AdaptiveCards/pull/1113#discussion_r167377042) [](ancestors = 167377042)

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1126,6 +1127,300 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
         THROW_IF_FAILED(textConfig->get_MaxWidth(&maxWidth));
         StyleXamlTextBlock(textSize, textColor, containerStyle, Boolify(isSubtle), wrap, maxWidth, textWeight, xamlTextBlock, hostConfig);
     }
+    
+    HRESULT AddTextInlines(
```
RebeccaAnne: >AddTextInlines [](start = 12, length = 14)

This turned out to be a lot of code, so I'm going to move it all into a helper file. Coming in a future commit (so the diffs are easier to see on this one).

RebeccaAnne:
The markdown parser is currently throwing on some of the cards. Holding this until that issue is resolved.

source/uwp/Renderer/lib/XamlBuilder.cpp
nesalang: i'm not sure what you mean by "type"

---
In reply to: [167641053](https://github.com/Microsoft/AdaptiveCards/pull/1113#discussion_r167641053) [](ancestors = 167641053,167377890)

RebeccaAnne:
No longer throwing on the latest main.

---
In reply to: [365036367](https://github.com/Microsoft/AdaptiveCards/pull/1113#issuecomment-365036367) [](ancestors = 365036367)

source/uwp/Renderer/lib/XamlBuilder.cpp
RebeccaAnne: It's an attribute on an ordered list that says whether the list should be, for example, 1,2,3,4 or a,b,c,d or I,II,III,IV. If you're not aware of it i'm guessing that means our markdown parser doesn't emit it :)

---
In reply to: [167732578](https://github.com/Microsoft/AdaptiveCards/pull/1113#discussion_r167732578) [](ancestors = 167732578,167641053,167377890)

Issue #1111, opened 2018-02-07
Adaptive Card Perf Issue


nesalang:
iOS Adaptive Cards have following performance issues:

- NSAttributedText initialization is very slow when it is initialized with HTML rendering enabled taking about 300 ms. It appears to be one time cost.

- Image loading and scaling performance can be improved. It is not as bad as NSAttributedText, but images load and subsequent processing are serialized.

PR #1124, opened 2018-02-13
Jwoo/ios perf experimental

nesalang:
fixes perf issues #1111

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
@@ -251,17 +251,18 @@ using namespace AdaptiveCards;
                 std::shared_ptr<ImageSet>imgSetElem = std::static_pointer_cast<ImageSet>(elem);
                 for(auto img :imgSetElem->GetImages()) { // loops through images in image set
                     std::shared_ptr<BaseCardElement> baseImgElem = std::static_pointer_cast<BaseCardElement>(img);
+                    /// tag a base card element with unique key
                     [self tagBaseCardElement:baseImgElem];
                     img->SetImageSize(imgSetElem->GetImageSize());
                     [self processImageConcurrently:img];
                 }
                 break;
             }
+            // continue on search
             case CardElementType::Container:
             {
                 std::shared_ptr<Container> container = std::static_pointer_cast<Container>(elem);
                 std::vector<std::shared_ptr<BaseCardElement>> &new_body = container->GetItems();
-                // update serial number that is used for generating unique key for image_map
                 [self addTasksToConcurrentQueue: new_body];
                 break;
             }
@@ -270,7 +271,6 @@ using namespace AdaptiveCards;
             {
                 std::shared_ptr<Column> colum = std::static_pointer_cast<Column>(elem);
                 std::vector<std::shared_ptr<BaseCardElement>> &new_body = colum->GetItems();
-                // update serial number that is used for generating unique key for image_map
                 [self addTasksToConcurrentQueue: new_body];
                 break;
             }
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
@@ -350,12 +350,8 @@ using namespace AdaptiveCards;
 -(void) tagBaseCardElement:(std::shared_ptr<BaseCardElement> const &) elem
 {
     std::string serial_number_as_string = std::to_string(_serialNumber);
-    // ID field is optional, if empty, add new one
-    if("" == elem->GetId()) {
-        elem->SetId("_" + serial_number_as_string);
-    } else { // concat a newly generated key to a existing id, the key will be removed after use
-        elem->SetId(elem->GetId() + "_" + serial_number_as_string);
-    }
+    // concat a newly generated key to a existing id, the key will be removed after use
+    elem->SetId(elem->GetId() + "_" + serial_number_as_string);
     ++_serialNumber;
 }
 
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
@@ -64,11 +64,6 @@ using namespace AdaptiveCards;
             _hostConfig = [config getHostConfig];
         }
         _guideFrame = frame;
-        _imageViewMap = [[NSMutableDictionary alloc] init];
-        _textMap = [[NSMutableDictionary alloc] init];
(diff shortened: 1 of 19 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
```diff
@@ -45,6 +64,11 @@ - (instancetype)init:(ACOAdaptiveCard *)card
             _hostConfig = [config getHostConfig];
         }
         _guideFrame = frame;
+        _imageViewMap = [[NSMutableDictionary alloc] init];
+        _textMap = [[NSMutableDictionary alloc] init];
+        _serial_queue = dispatch_queue_create("io.adaptiveCards.serial_queue", NULL);
+        _serial_text_queue = dispatch_queue_create("io.adaptiveCards.serial_text_queue", DISPATCH_QUEUE_SERIAL);
+        _serialNumber = 0;
```
bviglietta: You don't have to initialize these here. They've already been initialized in `initWithNibName:bundle:`

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
```diff
@@ -149,4 +175,212 @@ - (void)render
...
+                        parsedString = [parsedString stringByAppendingString:[NSString stringWithFormat:@"<style>body{font-family: '%@'; font-size:%dpx;}</style>",
+                                                                              @"verdana", [ACRTextBlockRenderer getTextBlockTextSize:txtElem->GetTextSize() withHostConfig:_hostConfig]]];
+                        // Convert html string to NSMutableAttributedString, NSAttributedString knows how to apply html tags
+                        NSData *htmlData = [parsedString dataUsingEncoding:NSUTF16StringEncoding];
+                        NSDictionary *options = @{NSDocumentTypeDocumentAttribute: NSHTMLTextDocumentType};
+                        // Initializing NSMutableAttributedString for HTML rendering is very slow
+                        NSMutableAttributedString *content = [[NSMutableAttributedString alloc] initWithData:htmlData options:options documentAttributes:nil error:nil];
+
+                         dispatch_async(dispatch_get_main_queue(),
+                             ^{ __block UILabel *lab = nil; // generate key for imageMap from image element's id
```
bviglietta: >imageMap from image element [](start = 80, length = 27)

Not an image

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
```diff
@@ -149,4 +175,212 @@ - (void)render
...
+                                  NSString *key = [NSString stringWithCString:txtElem->GetId().c_str() encoding:[NSString defaultCStringEncoding]];
+                                  // syncronize access to image map
```
bviglietta: >image map [](start = 58, length = 9)

Not an image

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
```diff
@@ -149,4 +175,212 @@ - (void)render
...
+                                  dispatch_sync(_serial_text_queue,
+                                      ^{
+                                           // UIImageView is not ready, cashe UIImage
```
bviglietta: >UIImageView is not ready, cashe UIImage [](start = 46, length = 39)

Not an image

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
nesalang: thanks

---
In reply to: [167995671](https://github.com/Microsoft/AdaptiveCards/pull/1124#discussion_r167995671) [](ancestors = 167995671)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
```diff
@@ -149,4 +170,212 @@ - (void)render
...
+}
+// add postfix to existing BaseCardElement ID to be used as key
+-(void) tagBaseCardElement:(std::shared_ptr<BaseCardElement> const &) elem
+{
+    std::string serial_number_as_string = std::to_string(_serialNumber);
+    // ID field is optional, if empty, add new one
+    if("" == elem->GetId()) {
+        elem->SetId("_" + serial_number_as_string);
+    } else { // concat a newly generated key to a existing id, the key will be removed after use
+        elem->SetId(elem->GetId() + "_" + serial_number_as_string);
```
bviglietta: You don't need the if statement. Just use this line all the time.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
```diff
@@ -149,4 +170,212 @@ - (void)render
...
+                                  // syncronize access to text map
+                                  dispatch_sync(_serial_text_queue,
+                                      ^{
+                                           // UILabel is not ready, cashe UILabel
+                                           if(!_textMap[key]) {
+                                               _textMap[key] = content;
+                                           } // UILable is ready, get label
+                                           else {
+                                               lab = _textMap[key];
+                                           }
```
bviglietta: I think it's confusing the way this map is used, mapping keys sometimes to strings and sometimes to UI elements. I also think the business of temporarily modifying the card's ID (to disambiguate between cards with the same ID?) seems fishy. But I don't understand this code base well enough to suggest any alternatives, so I'm letting those things pass.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
nesalang: thanks

---
In reply to: [168014387](https://github.com/Microsoft/AdaptiveCards/pull/1124#discussion_r168014387) [](ancestors = 168014387)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
nesalang: thanks, the code is initializing NSAttributedString's html rendering in the background while UI elements tree is being rendered. Either two can complete in any order, so I used NSDictionary that can get hold of any object types. 
Because we are using unique ID, there is no ambiguity of which type of the two will be stored in the dictionary. 

I wanted to use address of base elements as key, but it was unstable, and they were keep changing. so i had to create new id, and i piggyback on the existing ID. elements are not accessed again until UI elements complete rendering, so this shouldn't affect any pre-existing assumptions 

---
In reply to: [168017326](https://github.com/Microsoft/AdaptiveCards/pull/1124#discussion_r168017326) [](ancestors = 168017326)

PR #1129, opened 2018-02-14
Bug: ImageSetConfig - max image height

EduardTruuvaart:
Fixes an issue with TypeScript ImageSetConfig json parsing logic where maxImageHeight property wouldn't parse correctly because of typo. 

## Current behaviour
```javascript
    var imageSizeJson = {
        imageSize: 10,
        maxImageHeight: 50
    }
    var imageSetConfig = new ImageSetConfig(imageSizeJson);
    // imageSetConfig.maxImageHeight === "maxImageHeight" - it should be 50 instead
```
## Behaviour after fix
```javascript
    var imageSizeJson = {
        imageSize: 10,
        maxImageHeight: 50
    }
    var imageSetConfig = new ImageSetConfig(imageSizeJson);
    // imageSetConfig.maxImageHeight === 50
```
#1130


dclaux:
Sorry for the delay, thanks for the fix!

PR #1135, opened 2018-02-14
Move maps to function static rather than global

RebeccaAnne:


source/shared/cpp/ObjectModel/BaseCardElement.h
```diff
@@ -49,7 +49,6 @@ class BaseCardElement
     static Json::Value SerializeSelectAction(const std::shared_ptr<BaseActionElement> selectAction);
 
 private:
-    static const std::unordered_map<ActionType, std::function<std::shared_ptr<BaseActionElement>(const Json::Value&)>, EnumHash> ActionParsers;
     CardElementType m_type;
```
khouzam: Who was using ActionParsers?

RebeccaAnne: It's a remnant from how we defined parsers before we had custom parsers, and is no longer in use.

PR #1147, opened 2018-02-21
Fix to respect font-family in host config

yeohan:
## Defect
We had a wrong condition check on when to fall back to default value for font-family. As a result we ignore the font family given in the host config.

## Fix
Corrected the condition check to respect host config or fall back to default.

## Tests
Using iOS Visualizer, verify that font changes per host config.


yeohan:
Thanks!

PR #1150, opened 2018-02-21
iOS: Support borderColor/borderThickness for Container

yeohan:
## Defect
NYI. We have schema for borderColor/borderThickness, but have not made it on iOS.

## Fix
Parse from JSON and apply when rendering Container elements.

## Tests
Using iOS Visualizer, verify:
- The default style (without borderColor/Thickness) has no border
- The emphasis style (with borderColor/Thickness) has border.

### DIFF
--- a/source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
+++ b/source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
@@ -29,6 +29,7 @@
     "default": {
       "backgroundColor": "#00000000",
       "borderColor": "#FF232323",
+      "borderThickness": "1",
       "fontColors": {
         "default": {
           "normal": "#333333",
--- a/source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
+++ b/source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
@@ -28,8 +28,6 @@
   "containerStyles": {
     "default": {
       "backgroundColor": "#00000000",
-      "borderColor": "#FF232323",
-      "borderThickness": "1",
       "fontColors": {
         "default": {
           "normal": "#333333",
@@ -55,6 +53,8 @@
     },
     "emphasis": {
       "backgroundColor": "#08000000",
+      "borderColor": "#FF232323",
+      "borderThickness": "1",
       "fontColors": {
         "default": {
           "normal": "#333333",
--- a/source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
+++ b/source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
@@ -69,7 +69,7 @@
             },
             "backgroundColor": "#08000000",
             "borderColor": "#FF000000",
-            "borderThickness": "2"
+            "borderThickness": 2
               }
     },
     "spacing": {
--- a/source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
+++ b/source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
@@ -28,7 +28,7 @@
   "containerStyles": {
     "default": {
       "backgroundColor": "#00000000",
-      "borderColor": "#00232323",
+      "borderColor": "#FF232323",
       "fontColors": {
         "default": {
           "normal": "#333333",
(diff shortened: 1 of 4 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRContentStackView.mm
```diff
@@ -56,25 +59,47 @@ - (ContainerStyle)getStyle
...
+- (ContainerStyleDefinition&)paletteForHostConfig:(std::shared_ptr<HostConfig> const &)config
+{
+    return (_style == ContainerStyle::Emphasis)
+        ? config->containerStyles.emphasisPalette
+        : config->containerStyles.defaultPalette;
+}
+
+- (void)setBackgroundColorWithHostConfig:(std::shared_ptr<HostConfig> const &)config
+{
+    UIColor *color = [[self class] colorFromString:[self paletteForHostConfig:config].backgroundColor];
```
nesalang: what is advantage of using [self class] vs ACRContentStackView? 
if compiler optimizes the code, wouldn't it replace the above call with ACRContentStackView?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRContentStackView.mm
```diff
@@ -56,25 +59,47 @@ - (ContainerStyle)getStyle
...
+- (void)setBorderColorWithHostConfig:(std::shared_ptr<HostConfig> const &)config
+{
+    UIColor *color = [[self class] colorFromString:[self paletteForHostConfig:config].borderColor];
+
+    [[self layer] setBorderColor:[color CGColor]];
+}
+
+- (void)setBorderThicknessWithHostConfig:(std::shared_ptr<HostConfig> const &)config
+{
+    const CGFloat borderWidth = [[self class] borderWidthFromString:[self paletteForHostConfig:config].borderThickness];
```
nesalang: nit: do we really need to make it as method call?
borderWidthFromString seems to be simple and used only once  so it seems to make it inline.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRContentStackView.mm
yeohan: Advantage is that you don't have to spell out the exact class name ;)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRContentStackView.mm
yeohan: Yes, I believe this makes our code more readable. We have all these functions in method calls. If we want to end up with static methods, then we would need to pass _style to all of these functions. Please note that we need to get the palette corresponding to the given style.
- setBackgroundColorWithHostConfig
- setBorderColorWithHostConfig
- setBorderThicknessWithHostConfig

nesalang: I agree. I meant for borderWidthFromString method.

---
In reply to: [169716955](https://github.com/Microsoft/AdaptiveCards/pull/1150#discussion_r169716955) [](ancestors = 169716955)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRContentStackView.mm
nesalang: o.k. i was concerned more about having an additional call. but i guess it's o.k. since I assume it will be optimized by compiler.

---
In reply to: [169714151](https://github.com/Microsoft/AdaptiveCards/pull/1150#discussion_r169714151) [](ancestors = 169714151)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRContentStackView.mm
yeohan: I think I just followed the same pattern as in setBackgroundColorWithHostConfig. I'm fine to remove the helper function if it harms the readability. We shouldn't care too much about perf - this call isn't called too frequently. Readability is more important as far as perf is no concern.

Assaya:
Lovely CSS 

x

GLOBALCHERRY

source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
```diff
@@ -100,7 +102,7 @@
     },
     "imageSet": {
         "imageSize": "Medium",
-        "maxImageHeight": "maxImageHeight"
```
yeohan: This is a test break introduced from outside. Giving an explicit int number to work around.

Issue #914, opened 2017-11-09
[UWP] UWP renderer is missing mechanism to get current values from a custom input.


riarenas:

riarenas:
This is more of a feature we missed when writing the renderer and parser extensibility. Converting to Task and adding estimate.

PR #1174, opened 2018-02-27
Add handling of custom inputs

RebeccaAnne:
Fix for #914


source/uwp/Renderer/lib/InputItem.h
```diff
@@ -4,20 +4,24 @@
 
 namespace AdaptiveCards { namespace Rendering { namespace Uwp
 {
-    class InputItem
+    class DECLSPEC_UUID("BB1D1269-2243-4F34-B4EC-5216296EBBA0") InputValue :
```
eschavez: >InputValue [](start = 64, length = 10)

File should be changed to InputValue to match class

PR #1176, opened 2018-02-27
iOS: Notify rendering completion

yeohan:
This is to notify app of the rendering completion of an adaptive card. Using this, app can show the card only after it gets fully available.

## Issue
TextBlock rendering is a heavy task in iOS. It gives noticeable delay for users in hundreds milliseconds. We do render TextBlock elements async (which alone is good), but it is slow enough for users to notice they are being rendered. It looks messy so you may want to hide it while rendering is in progress. 

## Change
Provided didLoadElements, a new optional method in ACRActionDelegate.

When app implements this, we notify rendering complete by calling this back to app. Then app can handle to show/hide the adaptive card which has just been rendered completely.

The new method applies to not only TextBlock/Image/ImageSet but all other elements (which are rendered synchronously). We call didLoadElements at viewDidLoad when there are no TextBlock elements which are being rendered async. Otherwise, we let all TextBlock/Image/ImageSet elements render and call didLoadElements at the last element which completed rendering.

Updated the Visualizer code so the method is called with a NSLog call.

## Tests
Using iOS Visualizer, verify didLoadElements is called - log line will be added on debug console.
Tested all existing JSON in iOS Visualizer and confirmed didLoadElements is called in all the cases.

### DIFF
--- a/source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ViewController.m
+++ b/source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ViewController.m
@@ -233,9 +233,9 @@
     NSLog(@"Http Request fetched: %@", request);    
 }
 
-- (void)didCompleteRendering
+- (void)didLoadElements
 {
-    NSLog(@"didCompleteRendering");
+    NSLog(@"didLoadElements");
 }
 
 - (void)registerForKeyboardNotifications
--- a/source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ViewController.m
+++ b/source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ViewController.m
@@ -233,6 +233,11 @@
     NSLog(@"Http Request fetched: %@", request);    
 }
 
+- (void)didCompleteRendering
+{
+    NSLog(@"didCompleteRendering");
+}
+
 - (void)registerForKeyboardNotifications
 {
     [[NSNotificationCenter defaultCenter] addObserver:self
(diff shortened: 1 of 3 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
```diff
@@ -171,15 +178,36 @@ - (void)render
...
 
+auto _asyncRenderedElements = std::list<const BaseCardElement*>();
+
+- (void)callDidCompleteRenderingIfNeeded
+{
+#if DEBUG
+    NSLog(@"callDidCompleteRenderingIfNeeded");
+#endif
+
+    if (_asyncRenderedElements.size() == 0)
```
nesalang: consider this case, textblock is added, and completed. since _asyncRenderedElements.size() is zero, didCompleteRendering() will get called. but there were few more textblocks that were not put into dispatch queue because Image elements are being processed before it. If we have infinite queues, this won't be an issue, but I think there can be cases where this logic can fail since we don't have infinite queue.  I think you can add a additional check that makes sure we have seen all elements in the tree.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
```diff
@@ -171,15 +178,36 @@ - (void)render
        [newView.topAnchor constraintEqualToAnchor:view.topAnchor]]];
 }
 
+auto _asyncRenderedElements = std::list<const BaseCardElement*>();
```
nesalang: why make it global variable? It looks wired.

source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ViewController.m
```diff
@@ -233,6 +233,11 @@ - (void)didFetchHttpRequest:(NSURLRequest *)request
     NSLog(@"Http Request fetched: %@", request);    
 }
 
+- (void)didLoadElements
```
yeohan: Renamed to didLoadElements to better represent its functionality. The previous name didCompleteRendering sounded overpromising.

PR #1177, opened 2018-02-27
[TypeScript] Add lang support

dclaux:
Adds lang property to all Adaptive elements
Only parsed for AdaptiveCard, but technically lang can be overridden on a per-element basis in code
Valid lang format is 2 or 3 alpha characters, case insensitive


dclaux:
This PR finalizes support for https://github.com/Microsoft/AdaptiveCards/issues/908 now that we have agreed on the lang property format.

PR #1180, opened 2018-02-28
Fix Xamarin Android build break caused by changes in Android ActionHandler

khouzam:
Consolidate the actions in to ICardActionHandler and handle method signature changes.

Addresses issue #1181


eschavez:
Yes! Great to see all checks passed!

Issue #563, opened 2017-08-17
UWP renderer returns UIElement instead of FrameworkElement


anbare:
The renderer API's method signature returns a UIElement, but the actual item we're returning is a FrameworkElement...

Might as well change the method signature to return the more-specific FrameworkElement (which extends UIElement).

Natasha from Skype UWP had to actually cast the returned result to FrameworkElement since they need to set it as the content of a ContentPresenter, and ContentPresenter requires a FrameworkElement.

riarenas:
Seems completely reasonable.

PR #1183, opened 2018-02-28
Change to return FrameworkElement instead of UIElement

RebeccaAnne:
Fix for #563

### DIFF
--- a/source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
+++ b/source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
@@ -452,6 +452,7 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
     {
         [propget] HRESULT AdaptiveCard([out, retval] AdaptiveCard** value);
         [propget] HRESULT Errors([out, retval] Windows.Foundation.Collections.IVector<IAdaptiveError*>** value);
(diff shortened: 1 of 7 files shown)


source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -452,7 +452,6 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
     {
         [propget] HRESULT AdaptiveCard([out, retval] AdaptiveCard** value);
         [propget] HRESULT Errors([out, retval] Windows.Foundation.Collections.IVector<IAdaptiveError*>** value);
-        [propget] HRESULT Warnings([out, retval] Windows.Foundation.Collections.IVector<IAdaptiveWarning*>** value);
     };
```
khouzam: Seems unrelated. Also, should this stay in the interface even if we don't populate it yet?

RebeccaAnne: I'll move this to a separate change.

Per the discussion in the meeting the other day, I thought we had decided that we didn't want to have this. I'll double check with Matt before making this change. In any case, we can always add it back later if we change our mind.

---
In reply to: [171363004](https://github.com/Microsoft/AdaptiveCards/pull/1183#discussion_r171363004) [](ancestors = 171363004)

PR #1196, opened 2018-03-01
iOS: Fix crash on TextBlock rendering

yeohan:
## Defect
We make NSMutableAttributedString call in background thread. It may cause crash because NSMutableAttributedString is not thread-safe.

https://stackoverflow.com/questions/37029891/ios-crash-nsmutableattributedstring
https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/AttributedStrings/Tasks/CreatingAttributedStrings.html

Actually, we see a crash frequently on an iPhone X device.

## Fix
Moved the NSMutableAttributedString to main thread.

## Tests
iOS Visualizer works fine, render the test templates.

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
@@ -241,11 +241,13 @@ using namespace AdaptiveCards;
                         // Convert html string to NSMutableAttributedString, NSAttributedString knows how to apply html tags
                         NSData *htmlData = [parsedString dataUsingEncoding:NSUTF16StringEncoding];
                         NSDictionary *options = @{NSDocumentTypeDocumentAttribute: NSHTMLTextDocumentType};
-                        // Initializing NSMutableAttributedString for HTML rendering is very slow
-                        NSMutableAttributedString *content = [[NSMutableAttributedString alloc] initWithData:htmlData options:options documentAttributes:nil error:nil];
 
-                         dispatch_async(dispatch_get_main_queue(),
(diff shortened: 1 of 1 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRViewController.mm
```diff
@@ -241,11 +241,13 @@ - (void) addTasksToConcurrentQueue:(std::vector<std::shared_ptr<BaseCardElement>
                         // Convert html string to NSMutableAttributedString, NSAttributedString knows how to apply html tags
                         NSData *htmlData = [parsedString dataUsingEncoding:NSUTF16StringEncoding];
                         NSDictionary *options = @{NSDocumentTypeDocumentAttribute: NSHTMLTextDocumentType};
-                        // Initializing NSMutableAttributedString for HTML rendering is very slow
-                        NSMutableAttributedString *content = [[NSMutableAttributedString alloc] initWithData:htmlData options:options documentAttributes:nil error:nil];
 
-                         dispatch_async(dispatch_get_main_queue(),
```
nesalang: I have questions, Are there any impacts on performance by doing this? and is there any co-relation with rendering completion notification changes? 
Thanks

yeohan: A few from us tested the change and did not see any noticeable degradation.

There is no co-relation with rendering completion notification. We got call stack at the crash and see it's from HTML rendering.

nesalang: great;

thanks!

---
In reply to: [171457808](https://github.com/Microsoft/AdaptiveCards/pull/1196#discussion_r171457808) [](ancestors = 171457808)

PR #1208, opened 2018-03-03
Jwoo/shared custom parser for ios

nesalang:
- added custom element and parser for ios custom parsing
- ios custom pluggin stores json payload to pass it up to ios renderer at render time.


source/shared/cpp/ObjectModel/UnknownElement.h
```diff
@@ -0,0 +1,36 @@
...
+{
+class UnknownElement: public BaseCardElement
+{
+public:
+    UnknownElement();
+    UnknownElement(Json::Value const &paylaod);
+        
+    virtual Json::Value SerializeToJsonValue() override;
+    Json::Value GetJsonPayload();
+    void SetJsonPayload(const Json::Value& payload);
```
RebeccaAnne: Do you need this, or does AdditionalProperties work for this?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/CustomParsingForIOSTest.cpp
```diff
@@ -0,0 +1,116 @@
...
+        {
+            std::string testJsonString = 
+            "{\
+                \"$schema\":\"http://adaptivecards.io/schemas/adaptive-card.json\",\
+                \"type\": \"AdaptiveCard\",\
+                \"version\": \"1.0\",\
+                \"body\": [\
+                    {\
+                        \"type\": \"Random\",\
+                        \"payload\": \"You can even draw attention to certain text with color\"\
```
RebeccaAnne: >\"payload\": \"You can even draw attention to certain text with color\"\ [](start = 24, length = 72)

The card author shouldn't have to use "payload", they should be able to use whatever properties they want

source/shared/cpp/ObjectModel/UnknownElement.h
RebeccaAnne: I see, this is just calling AdditionalProperties. You don't need to expose this, the caller can just call AdditionalProperties.

---
In reply to: [172369640](https://github.com/Microsoft/AdaptiveCards/pull/1208#discussion_r172369640) [](ancestors = 172369640)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/CustomParsingForIOSTest.cpp
nesalang: yes, that's correct; i just leaved it as it is; as it works.

---
In reply to: [172369858](https://github.com/Microsoft/AdaptiveCards/pull/1208#discussion_r172369858) [](ancestors = 172369858)

source/shared/cpp/ObjectModel/UnknownElement.h
nesalang: yes, we don't need this; thanks


---
In reply to: [172369640](https://github.com/Microsoft/AdaptiveCards/pull/1208#discussion_r172369640) [](ancestors = 172369640)

RebeccaAnne: You don't need the getter either.  The caller can call GetAdditionalProperties.

---
In reply to: [172370738](https://github.com/Microsoft/AdaptiveCards/pull/1208#discussion_r172370738) [](ancestors = 172370738,172369640)

nesalang: doesn't sound weird to call getAdditonalProperty on unknown object?

---
In reply to: [172372405](https://github.com/Microsoft/AdaptiveCards/pull/1208#discussion_r172372405) [](ancestors = 172372405,172370738,172369640)

source/shared/cpp/ObjectModel/UnknownElement.cpp
```diff
@@ -0,0 +1,49 @@
...
+UnknownElement::UnknownElement(
+    Json::Value const &paylaod) :
+    BaseCardElement(CardElementType::Unknown)
+{
+}
+
+Json::Value UnknownElement::SerializeToJsonValue()
+{
+    Json::Value root = GetJsonPayload();
+    root[AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Type)] = CardElementTypeToString(CardElementType::Unknown);
```
RebeccaAnne: >root[AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Type)] = CardElementTypeToString(CardElementType::Unknown);  [](start = 4, length = 118)

You need the other base card element properties as well. Call BaseCardElement::SerializeToJsonValue

nesalang: Is there a way to append two Json::Values, otherwise i will just add those properties by name.

---
In reply to: [172373459](https://github.com/Microsoft/AdaptiveCards/pull/1208#discussion_r172373459) [](ancestors = 172373459)

PR #1216, opened 2018-03-05
Add Version handling to shared model

### LINKED ISSUE #860: [UWP] Return FallbackText when the card is not rendered.

RebeccaAnne:
Remove min version and handle switching to fallback text if the version is less than the passed in renderer version.

Fixes #860


source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
```diff
@@ -49,11 +46,13 @@ AdaptiveCard::AdaptiveCard(std::string version,
 #ifdef __ANDROID__
 std::shared_ptr<AdaptiveCard> AdaptiveCard::DeserializeFromFile(
     const std::string& jsonFile,
+    const std::string& rendererVersion,
     std::shared_ptr<ElementParserRegistration> elementParserRegistration,
     std::shared_ptr<ActionParserRegistration> actionParserRegistration) throw(AdaptiveCards::AdaptiveCardParseException)
 #else
 std::shared_ptr<AdaptiveCard> AdaptiveCard::DeserializeFromFile(
     const std::string& jsonFile,
+    double rendererVersion,
```
nesalang: this is a little bit confusing to me. who set this version?

source/shared/cpp/ObjectModel/ShowCardAction.cpp
```diff
@@ -45,7 +45,7 @@ std::shared_ptr<BaseActionElement> ShowCardActionParser::Deserialize(
     std::shared_ptr<ShowCardAction> showCardAction = BaseActionElement::Deserialize<ShowCardAction>(json);
 
     std::string propertyName = AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Card);
-    showCardAction->SetCard(AdaptiveCard::Deserialize(json.get(propertyName, Json::Value()), elementParserRegistration, actionParserRegistration));
+    showCardAction->SetCard(AdaptiveCard::Deserialize(json.get(propertyName, Json::Value()), std::numeric_limits<double>::max(), elementParserRegistration, actionParserRegistration));
```
nesalang: just curious, why max?

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
```diff
@@ -84,14 +85,20 @@ std::shared_ptr<AdaptiveCard> AdaptiveCard::Deserialize(
     ParseUtil::ExpectTypeString(json, CardElementType::AdaptiveCard);
 
     std::string version = ParseUtil::GetString(json, AdaptiveCardSchemaKey::Version);
-    std::string minVersion = ParseUtil::GetString(json, AdaptiveCardSchemaKey::MinVersion);
     std::string fallbackText = ParseUtil::GetString(json, AdaptiveCardSchemaKey::FallbackText);
+    std::string language = ParseUtil::GetString(json, AdaptiveCardSchemaKey::Language);
+
+    double versionAsDouble = ::atof(version.c_str());
```
eschavez: >versionAsDouble [](start = 11, length = 15)

Is version required to be a float?

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
RebeccaAnne: The renderer passes it in. This is the version of the renderer. Currently this is 1.0, and will be updated as we update our renderers.

---
In reply to: [172352845](https://github.com/Microsoft/AdaptiveCards/pull/1216#discussion_r172352845) [](ancestors = 172352845)

source/shared/cpp/ObjectModel/ShowCardAction.cpp
RebeccaAnne: Per discussion with Matt, show cards don't have a separate version, so we don't need to enforce version here. Pass max to effectively say we can handle anything. (because we've already checked version at the top of the main card)

---
In reply to: [172352922](https://github.com/Microsoft/AdaptiveCards/pull/1216#discussion_r172352922) [](ancestors = 172352922)

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
RebeccaAnne: As opposed to an int? I believe we need to support dot versions, so like, version 1.1.

---
In reply to: [172354938](https://github.com/Microsoft/AdaptiveCards/pull/1216#discussion_r172354938) [](ancestors = 172354938)

eschavez: no. Float makes sense. I meant that we read this as a string from the card payload but are we enforcing that it has to be a float representation? What happens for "version": "blah"?

---
In reply to: [172357010](https://github.com/Microsoft/AdaptiveCards/pull/1216#discussion_r172357010) [](ancestors = 172357010,172354938)

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
nesalang: thanks; that makes sense

---
In reply to: [172356287](https://github.com/Microsoft/AdaptiveCards/pull/1216#discussion_r172356287) [](ancestors = 172356287,172352845)

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
RebeccaAnne: Hmm... well atof will return 0, and renderer version will not be smaller than that, so we'll go along our merry way. I think that's probably reasonable, but I'll double check with Matt.

---
In reply to: [172358721](https://github.com/Microsoft/AdaptiveCards/pull/1216#discussion_r172358721) [](ancestors = 172358721,172357010,172354938)

RebeccaAnne: Matt says we should throw, will update

---
In reply to: [172362617](https://github.com/Microsoft/AdaptiveCards/pull/1216#discussion_r172362617) [](ancestors = 172362617,172358721,172357010,172354938)

nesalang:
Added my changes;

RebeccaAnne:
Thank you!

---
In reply to: [370613227](https://github.com/Microsoft/AdaptiveCards/pull/1216#issuecomment-370613227) [](ancestors = 370613227)

PR #1222, opened 2018-03-06
Add parser warnings to the shared model and UWP

RebeccaAnne:


khouzam:
Still need to fix Android for this change

eschavez:
Fixing android now

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -289,7 +289,6 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
     [version(NTDDI_WIN10_RS1)]
     typedef [v1_enum] enum ErrorStatusCode {
         InvalidJson = 0,
-        UnsupportedSchemaVersion,
```
bviglietta: >UnsupportedSchemaVersion, [](start = 8, length = 25)

Do you have any existing clients that might be affected by this breaking change?

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -455,6 +455,7 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
     {
         [propget] HRESULT AdaptiveCard([out, retval] AdaptiveCard** value);
         [propget] HRESULT Errors([out, retval] Windows.Foundation.Collections.IVector<IAdaptiveError*>** value);
+        [propget] HRESULT Warnings([out, retval] Windows.Foundation.Collections.IVector<IAdaptiveWarning*>** value);
```
bviglietta: >IVector [](start = 80, length = 7)

Should this be IVectorView? It doesn't seem like you would want callers to modify the vector.

khouzam: Good point, but we use the interface internally also to populate the vector. The changes to fix this would be out of scope of this change, errors is the same way.

source/uwp/Renderer/lib/AdaptiveCard.cpp
```diff
@@ -95,10 +96,27 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
...
+                        std::shared_ptr<::AdaptiveCards::ParseResult> sharedParseResult = ::AdaptiveCards::AdaptiveCard::DeserializeFromString(jsonString, c_rendererVersion, sharedModelElementParserRegistration, sharedModelActionParserRegistration);
                         ComPtr<IAdaptiveCard> adaptiveCard;
-                        RETURN_IF_FAILED(MakeAndInitialize<AdaptiveCard>(&adaptiveCard, sharedAdaptiveCard));
+                        RETURN_IF_FAILED(MakeAndInitialize<AdaptiveCard>(&adaptiveCard, sharedParseResult->GetAdaptiveCard()));
                         RETURN_IF_FAILED(adaptiveParseResult->put_AdaptiveCard(adaptiveCard.Get()));
+
+                        for (auto sharedWarning : sharedParseResult->GetWarnings())
+                        {
+                            ComPtr<IVector<IAdaptiveWarning*>> warnings;
+                            RETURN_IF_FAILED(adaptiveParseResult->get_Warnings(&warnings));
```
bviglietta: You can pull this out of the loop.

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
khouzam: Yes, and this is why we're doing this breaking change now.

khouzam: Yes, this is why we're doing this now, before we release.


---
In reply to: [172656255](https://github.com/Microsoft/AdaptiveCards/pull/1222#discussion_r172656255) [](ancestors = 172656255)

source/uwp/Renderer/lib/AdaptiveCard.cpp
khouzam: Good point

---
In reply to: [172677757](https://github.com/Microsoft/AdaptiveCards/pull/1222#discussion_r172677757) [](ancestors = 172677757)

PR #1247, opened 2018-03-09
Introduce base classes for elements, inputs, and actions to reduce du…

RebeccaAnne:
…plicated code.

### DIFF
--- a/source/uwp/Renderer/lib/AdaptiveColumnSet.cpp
+++ b/source/uwp/Renderer/lib/AdaptiveColumnSet.cpp
@@ -84,11 +84,18 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
 
         XamlHelpers::IterateOverVector<IAdaptiveColumn>(m_columns.Get(), [&](IAdaptiveColumn* column)
         {
-            std::shared_ptr<BaseCardElement> sharedColumn = std::make_shared<Column>();
             ComPtr<AdaptiveCards::Rendering::Uwp::AdaptiveColumn> columnImpl = PeekInnards<AdaptiveCards::Rendering::Uwp::AdaptiveColumn>(column);
-            RETURN_IF_FAILED(columnImpl->GetSharedModel(sharedColumn));
 
-            columnSet->GetColumns().push_back(std::dynamic_pointer_cast<AdaptiveCards::Column>(sharedColumn));
+            std::shared_ptr<BaseCardElement> sharedColumnBaseElement;
+            RETURN_IF_FAILED(columnImpl->GetSharedModel(sharedColumnBaseElement));
+
+            std::shared_ptr<AdaptiveCards::Column> sharedColumn = std::dynamic_pointer_cast<AdaptiveCards::Column>(sharedColumnBaseElement);
+            if (sharedColumn == nullptr)
+            {
+                return E_UNEXPECTED;
+            }
+
+            columnSet->GetColumns().push_back(sharedColumn);
 
             return S_OK;
         });
--- a/source/uwp/Renderer/lib/AdaptiveColumnSet.cpp
+++ b/source/uwp/Renderer/lib/AdaptiveColumnSet.cpp
@@ -28,7 +28,7 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
     } CATCH_RETURN;
 
     _Use_decl_annotations_
-    HRESULT AdaptiveColumnSet::RuntimeClassInitialize(const std::shared_ptr<AdaptiveCards::ColumnSet>& sharedColumnSet)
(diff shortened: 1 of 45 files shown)


source/uwp/Renderer/lib/AdaptiveColumnSet.cpp
```diff
@@ -158,11 +84,11 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
 
         XamlHelpers::IterateOverVector<IAdaptiveColumn>(m_columns.Get(), [&](IAdaptiveColumn* column)
         {
-            std::shared_ptr<Column> sharedColumn = std::make_shared<Column>();
+            std::shared_ptr<BaseCardElement> sharedColumn = std::make_shared<Column>();
             ComPtr<AdaptiveCards::Rendering::Uwp::AdaptiveColumn> columnImpl = PeekInnards<AdaptiveCards::Rendering::Uwp::AdaptiveColumn>(column);
             RETURN_IF_FAILED(columnImpl->GetSharedModel(sharedColumn));
 
-            columnSet->GetColumns().push_back(sharedColumn);
+            columnSet->GetColumns().push_back(std::dynamic_pointer_cast<AdaptiveCards::Column>(sharedColumn));
```
almedina-ms: >dynamic_pointer_cast [](start = 51, length = 20)

Should this one be a static_pointer_cast like the other ones being changed?

source/uwp/Renderer/lib/AdaptiveToggleInput.h
```diff
@@ -34,29 +36,29 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
...
-        IFACEMETHODIMP get_AdditionalProperties(_Out_ ABI::Windows::Data::Json::IJsonObject** result);
-        IFACEMETHODIMP put_AdditionalProperties(_In_ ABI::Windows::Data::Json::IJsonObject* value);
+        IFACEMETHODIMP get_AdditionalProperties(_Out_ ABI::Windows::Data::Json::IJsonObject** result) { return AdaptiveCardElementBase::get_AdditionalProperties(result); }
+        IFACEMETHODIMP put_AdditionalProperties(_In_ ABI::Windows::Data::Json::IJsonObject* value) { return AdaptiveCardElementBase::put_AdditionalProperties(value); }
 
-        IFACEMETHODIMP ToJson(_Out_ ABI::Windows::Data::Json::IJsonObject** result);
+        IFACEMETHODIMP ToJson(_Out_ ABI::Windows::Data::Json::IJsonObject** result) { return AdaptiveCardElementBase::ToJson(result); }
 
-        HRESULT GetSharedModel(_In_ std::shared_ptr<AdaptiveCards::ToggleInput>& sharedModel);
+        virtual HRESULT GetSharedModel(_In_ std::shared_ptr<AdaptiveCards::BaseCardElement>& sharedModel) override;
```
eschavez: >_In_ [](start = 39, length = 4)

_Out_ please update everywhere

source/uwp/Renderer/lib/AdaptiveColumnSet.cpp
RebeccaAnne: I made this one dynamic cast since we're down casting so theoretically we can't be sure that the basecardelement is a column. 

On the flip side, we know the implementation of GetSharedModel so we know it would be safe to static cast, and the build system in windows doesn't like dynamic casts.

I'm going to leave as dynamic for now, but i am going to revisit the Windows guidance on this and see whether we should be making a change all up.

---
In reply to: [173899229](https://github.com/Microsoft/AdaptiveCards/pull/1247#discussion_r173899229) [](ancestors = 173899229)

source/uwp/Renderer/lib/AdaptiveToggleInput.h
RebeccaAnne: I'll remove the sal. Current guidance is to reduce usage of sal when not needed. In the case, the fact that it's a non-const reference already tells us that this is writable. 

I'm looking at SAL.2 at:
[https://osgwiki.com/wiki/Cpp_Core_Guidelines#SAL:_Source_Annotation_Language](https://osgwiki.com/wiki/Cpp_Core_Guidelines#SAL:_Source_Annotation_Language)

---
In reply to: [173976055](https://github.com/Microsoft/AdaptiveCards/pull/1247#discussion_r173976055) [](ancestors = 173976055)

source/uwp/Renderer/lib/AdaptiveColumnSet.cpp
RebeccaAnne: (I will add a check on the return value though)

---
In reply to: [174513509](https://github.com/Microsoft/AdaptiveCards/pull/1247#discussion_r174513509) [](ancestors = 174513509,173899229)

PR #1250, opened 2018-03-12
Check the supporstInteractivity option in the host config before rendering input/actions

eschavez:
Also fixed a visual bug where the space between the body and the actions has no background color

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
@@ -190,6 +190,7 @@ public class AdaptiveCardRenderer
         if (baseActionElementList != null && baseActionElementList.size() > 0)
         {
             LinearLayout showCardsLayout = new LinearLayout(context);
+            showCardsLayout.setBackgroundColor(Color.parseColor(color));
             showCardsLayout.setLayoutParams(new LinearLayout.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.MATCH_PARENT));
             rootLayout.addView(showCardsLayout);
             
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
@@ -185,16 +185,23 @@ public class AdaptiveCardRenderer
 
         CardRendererRegistration.getInstance().render(renderedCard, context, fragmentManager, layout, adaptiveCard, baseCardElementList, cardActionHandler, hostConfig, style);
 
-        // Actions are optional
-        BaseActionElementVector baseActionElementList = adaptiveCard.GetActions();
-        if (baseActionElementList != null && baseActionElementList.size() > 0)
+        if (hostConfig.getSupportsInteractivity())
         {
-            LinearLayout showCardsLayout = new LinearLayout(context);
-            showCardsLayout.setBackgroundColor(Color.parseColor(color));
-            showCardsLayout.setLayoutParams(new LinearLayout.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.MATCH_PARENT));
-            rootLayout.addView(showCardsLayout);
-            
-            renderActions(renderedCard, context, fragmentManager, layout, baseActionElementList, cardActionHandler, hostConfig);
+            // Actions are optional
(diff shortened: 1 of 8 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
```diff
@@ -239,11 +247,18 @@ else if (alignment == ActionAlignment.Center.swigValue())
             viewGroup.addView(actionButtonsLayout);
         }
 
-        for (int i = 0; i < size && i < hostConfig.getActions().getMaxActions(); i++)
+        int i = 0;
+        int maxActions = (int) hostConfig.getActions().getMaxActions();
```
nesalang: if getMaxActions() return unsigned, would be better to have i unsigned?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
```diff
@@ -239,11 +247,18 @@ else if (alignment == ActionAlignment.Center.swigValue())
...
+        for (; i < size && i < maxActions; i++)
         {
             BaseActionElement actionElement = baseActionElementList.get(i);
             ActionElementRenderer.getInstance().render(renderedCard, context, fragmentManager, actionButtonsLayout, actionElement, cardActionHandler, hostConfig);
         }
+
+        if (i >= maxActions)
```
nesalang: should be i > maxActions

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveWarning.java
```diff
@@ -3,6 +3,9 @@
 public class AdaptiveWarning {
     public static final int UNKNOWN_ELEMENT_TYPE = 1;
     public static final int UNABLE_TO_LOAD_IMAGE = 2;
+    public static final int INTERACTIVITY_DISALLOWED = 3;
+    public static final int MAX_ACTIONS_REACHED = 4;
```
nesalang: Reaching Max Actions vs Exceeding Max Actions?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/NumberInputRenderer.java
```diff
@@ -45,6 +46,12 @@ public View render(
             HostConfig hostConfig,
             ContainerStyle containerStyle)
     {
+        if (!hostConfig.getSupportsInteractivity())
+        {
+            renderedCard.addWarning(new AdaptiveWarning(AdaptiveWarning.INTERACTIVITY_DISALLOWED, "Input.Number is not allowed"));
```
nesalang: NumberInput for consistency?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
eschavez: yeah. I should change this to long since that is what getMaxActions returns

eschavez: yeah. I should change this to long since that is what getMaxActions returns

---
In reply to: [175174429](https://github.com/Microsoft/AdaptiveCards/pull/1250#discussion_r175174429) [](ancestors = 175174429)

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
eschavez: no. Should be >= .  In fact, i will never be greater than because the previous for loop will exit once i == maxActions 

---
In reply to: [175174739](https://github.com/Microsoft/AdaptiveCards/pull/1250#discussion_r175174739) [](ancestors = 175174739)

eschavez: Talked offline. Need to also make sure size != maxactions. Good catch!

---
In reply to: [175221222](https://github.com/Microsoft/AdaptiveCards/pull/1250#discussion_r175221222) [](ancestors = 175221222,175174739)

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
eschavez: Keeping i as an int but changing maxActions to long.

---
In reply to: [175220676](https://github.com/Microsoft/AdaptiveCards/pull/1250#discussion_r175220676) [](ancestors = 175220676,175174429)

PR #1252, opened 2018-03-13
testing

nesalang:


nesalang:
Verified that sharedModel CI build Unit Test is working.

PR #1251, opened 2018-03-12
Add HasKey check before LookingUp values in ResouceDictionaryMap

almedina-ms:

PR #1262, opened 2018-03-15
fixes image set issues (#1251)

### LINKED ISSUE #1261: image set is scrollable, and some images are partially hidden

nesalang:
fixes #1261


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRImageSetUICollectionView.mm
```diff
@@ -79,6 +86,14 @@ - (CGFloat)collectionView:(UICollectionView *)collectionView
 
 - (CGSize)intrinsicContentSize
 {
-    return [self.collectionViewLayout collectionViewContentSize];
+    return [self collectionViewContentSize];
+}
+
+- (CGSize) collectionViewContentSize
```
bviglietta: >collectionViewContentSize [](start = 11, length = 25)

I don't really understand the context here well enough to review this code. For instance, I can see that if the image set contains two images, we're going to return a CGSize large enough to hold four images, but I don't know whether that's the intended behavior.

nesalang: that's good and valid point. This method doesn't return exact size. It only give UICollectionView to figure out approximate of good size, and UIStackView also resize the container. 

---
In reply to: [175205307](https://github.com/Microsoft/AdaptiveCards/pull/1262#discussion_r175205307) [](ancestors = 175205307)

PR #1269, opened 2018-03-15
Allowing empty items in Column

eschavez:
Also Added check in Android for empty items list


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ContainerRenderer.java
```diff
@@ -76,6 +81,7 @@ else if ((container = Container.dynamic_cast(baseCardElement)) == null)
             containerView.setOnClickListener(new ActionElementRenderer.ButtonOnClickListener(renderedCard, container.GetSelectAction(), cardActionHandler));
         }
 
+        viewGroup.addView(containerView);
```
nesalang: why do you want to add linearlayout if it's empty?

PR #1275, opened 2018-03-19
Move inline ShowCards out of the card padding

### LINKED ISSUE #371: UWP Renderer: Inline ShowCards are within the card margins rather than being edge-to-edge

RebeccaAnne:
Add an additional panel which does not apply the card padding so that show cards are not within the card padding.

Fixes #371


RebeccaAnne:
Updated the test app to expand show cards so that changes in show card behavior (such as this one) are covered by the app.

PR #1276, opened 2018-03-19
Fix annotations for UWP

almedina-ms:
Fixes the annotation issues for UWP


source/shared/cpp/ObjectModel/jsoncpp.cpp
```diff
@@ -263,6 +263,9 @@ static inline void fixNumericLocaleInput(char* begin, char* end) {
 #pragma warning(disable : 4996)
 #endif
 
+#pragma warning(disable : 28726)
+#pragma warning(disable : 28719)
```
RebeccaAnne: Consider adding a comment saying what these warnings are.

Issue #1144, opened 2018-02-20
Android and UWP font family needs to be applied to Card


eschavez:
Host Config specifies FontFamily but android does not respect that option.

matthidinger:
@eschavez any ETA on when this will get checked in?

RebeccaAnne:
UWP doesn't support this either

almedina-ms:
iOS does support it

eschavez:
Fixed.

PR #1310, opened 2018-03-27
Fix ignore of font family in UWP and Android

almedina-ms:
Fixes font family host config parameter from being ignored in UWP and Android renderer. 

https://github.com/Microsoft/AdaptiveCards/issues/1144


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1120,6 +1120,17 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
...
+        HString fontFamilyName;
+        THROW_IF_FAILED(hostConfig->get_FontFamily(fontFamilyName.GetAddressOf()));
+
+        ComPtr<IInspectable> inspectable;
+        ComPtr<IFontFamily> fontFamily;
+        ComPtr<IFontFamilyFactory> fontFamilyFactory;
+        THROW_IF_FAILED(Windows::Foundation::GetActivationFactory(HStringReference(L"Windows.UI.Xaml.Media.FontFamily").Get(), &fontFamilyFactory));
+        THROW_IF_FAILED(fontFamilyFactory->CreateInstanceWithName(fontFamilyName.Get(), nullptr, inspectable.ReleaseAndGetAddressOf(), &fontFamily));
+        THROW_IF_FAILED(xamlTextBlock->put_FontFamily(fontFamily.Get()));
+
```
khouzam: We should have a simple test that specifies the font family.

PR #1312, opened 2018-03-28
Add ActionSet to shared model and UWP

RebeccaAnne:


source/shared/cpp/ObjectModel/Enums.h
```diff
@@ -268,7 +271,8 @@ enum class Spacing {
 
 enum class ActionsOrientation {
     Vertical = 0,
-    Horizontal
+    Horizontal,
+    None,
```
nesalang: feel weird to have None as orientation type

RebeccaAnne: I use this when an orientation hasn't been explicitly set on an ActionSet, so that i know to fall back on whatever's in the host config. There's some precedent for this (image size also has a "none")

---
In reply to: [178390247](https://github.com/Microsoft/AdaptiveCards/pull/1312#discussion_r178390247) [](ancestors = 178390247)

RebeccaAnne: I could change it to Default though, that might seem less weird.

---
In reply to: [178391172](https://github.com/Microsoft/AdaptiveCards/pull/1312#discussion_r178391172) [](ancestors = 178391172,178390247)

nesalang: no, it makes sense to grab one from host config when it's set None.

---
In reply to: [178391349](https://github.com/Microsoft/AdaptiveCards/pull/1312#discussion_r178391349) [](ancestors = 178391349,178391172,178390247)

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -400,7 +390,17 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
         ComPtr<IAdaptiveCardElement> adaptiveCardElement;
         THROW_IF_FAILED(adaptiveImage.As(&adaptiveCardElement));
         ComPtr<IUIElement> backgroundImage;
-        BuildImage(adaptiveCardElement.Get(), renderContext, renderArgs, &backgroundImage);
+
+        ComPtr<IAdaptiveElementRendererRegistration> elementRenderers;
```
nesalang: nit: elementRendererRegistartion maybe? very confusing with elementRenderers vs elementRenderer

RebeccaAnne:
Per discussion with Gilles, putting this into a vnext branch rather than master.

PR #1321, opened 2018-03-30
Update test files related to notification card and different font hos…

RebeccaAnne:
…t config


source/uwp/AdaptiveCardTestApp/Expected/v1.0.sampleDifferentFont.Tests.Action.Title.json
```diff
@@ -0,0 +1 @@
+{"HostConfigHash":"5284388","CardHash":"cb85211","Error":"Parsing card failed"}
```
nesalang: looks like this and two other files below it are missing images

source/uwp/AdaptiveCardTestApp/Expected/v1.0.windows-timeline.v1.0.Scenarios.NotificationCard.json
```diff
@@ -0,0 +1 @@
+{"HostConfigHash":"0164943","CardHash":"8dadcfb","Error":null}
```
nesalang: does this card's image supposed to have buttons like other host config versions do?

source/uwp/AdaptiveCardTestApp/Expected/v1.0.sampleDifferentFont.Tests.Action.Title.json
RebeccaAnne: These cards fail to parse (by design) because they have invalid card json (this one is missing action titles and the next one is missing version). As a result, there's no rendered card. The "Error":"Parsing card failed" in this file will tell the test app that the failure is expected.

---
In reply to: [178395259](https://github.com/Microsoft/AdaptiveCards/pull/1321#discussion_r178395259) [](ancestors = 178395259)

source/uwp/AdaptiveCardTestApp/Expected/v1.0.windows-timeline.v1.0.Scenarios.NotificationCard.json
RebeccaAnne: This host config doesn't allow interactivity, so the buttons don't render.

---
In reply to: [178395476](https://github.com/Microsoft/AdaptiveCards/pull/1321#discussion_r178395476) [](ancestors = 178395476)

PR #1323, opened 2018-03-31
Updating shared model and UWP code to use defined macros for namespace.

eschavez:


source/shared/cpp/ObjectModel/ActionParserRegistration.cpp
```diff
@@ -4,8 +4,7 @@
 #include "ShowCardAction.h"
 #include "SubmitAction.h"
 
-namespace AdaptiveCards
-{
+AdaptiveSharedNamespaceStart
```
RebeccaAnne: Could this be:
namespace AdaptiveSharedNamespaceRef
{

source/shared/cpp/ObjectModel/pch.h
```diff
@@ -1,4 +1,15 @@
 #pragma once
+
+#ifdef ADAPTIVE_CARDS_WINDOWS
+#include "winPch.h"
+#endif
+
+#ifndef AdaptiveSharedNamespaceStart
+#define AdaptiveSharedNamespaceStart namespace AdaptiveCards { 
+#define AdaptiveSharedNamespaceRef AdaptiveCards
```
RebeccaAnne: >AdaptiveSharedNamespaceRef [](start = 8, length = 26)

Maybe just "AdaptiveSharedNamespace"? (no Ref)

source/uwp/AdaptiveCards.sln
```diff
@@ -36,20 +36,20 @@ Global
 		{CF169157-F487-420C-8B9B-CA3ABE2BB209}.Release|x64.Build.0 = Release|x64
 		{CF169157-F487-420C-8B9B-CA3ABE2BB209}.Release|x86.ActiveCfg = Release|Win32
 		{CF169157-F487-420C-8B9B-CA3ABE2BB209}.Release|x86.Build.0 = Release|Win32
-		{B5DECA6E-D9EA-476D-9B7E-34D3B3044CD8}.Debug|ARM.ActiveCfg = Debug|ARM
```
RebeccaAnne: Don't include these changes.

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -1041,17 +1397,28 @@ namespace AdaptiveCards { namespace Rendering { namespace Uwp
 
     [
         version(NTDDI_WIN10_RS1),
-        activatable(NTDDI_WIN10_RS1)
+        activatable(NTDDI_WIN10_RS1),
+#ifdef ADAPTIVE_CARDS_WINDOWS
+        contract(InternalContract, 1),
+        internal,
+        activatable(InternalContract, 1)
+#endif
```
RebeccaAnne: Something to consider: would it be possible/better to have this whole chunk of stuff be a single define? Something like WINDOW_INTERNAL_ACTIVATABLE_CONTRACT that would add just one line here and include all of this so we don't have to repeat it on each runtime class? I realize there might need be a few different ones for activatable vs. not/runtimeclass vs. interface so it might be more trouble than it's worth, but just an idea.

PR #1364, opened 2018-04-09
Fix undesirable separation when hiding first element in Container

dclaux:
Fix lingering visualizer bug introduced by previous change in renderer

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -118,7 +118,12 @@ export abstract class CardElement {
         }
 
         if (this._separatorElement) {
-            this._separatorElement.style.display = this._isVisible ? this._defaultRenderedElementDisplayMode : "none";
+            if (this.parent && this.parent.isFirstElement(this)) {
+                this._separatorElement.style.display = "none";                
+            }
+            else {
+                this._separatorElement.style.display = this._isVisible ? this._defaultRenderedElementDisplayMode : "none";
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -118,7 +118,12 @@ export abstract class CardElement {
         }
 
         if (this._separatorElement) {
-            this._separatorElement.style.display = this._isVisible ? this._defaultRenderedElementDisplayMode : "none";
+            if (this.parent && this.parent.isFirstElement(this)) {
+                this._separatorElement.style.display = "none";                
+            }
+            else {
+                this._separatorElement.style.display = this._isVisible ? this._defaultRenderedElementDisplayMode : "none";
+            }
```
tongyliu: Could simplify into one conditional:

```
let shouldShowSeparator = this._isVisible && !(this.parent && this.parent.isFirstElement(this));
```
(or something similar).

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2879,7 +2885,7 @@ export class Container extends CardElement {
         for (let item of this._items) {
             handleElement(item);
         }
-
+
```
tongyliu: Nit: unnecessary whitespace

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Actually I think that would make it harder to read and understand, which is why I did it in a more explicit way.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: This is followed by "return true". There should always be a blank line after a closing curly bracket unless there is another closing curly bracket on the next line or if there is an else statement.

tongyliu: Oh I was referring to the extra spaces that got added to the blank line (darker green highlighted section), not the line itself.

dclaux: Oh! Sorry :)

PR #1366, opened 2018-04-10
Remove validation of images on actions until it is implemented.

khouzam:
The addition of images on actions has added iconUrl to actions. The dotNet render issues a warning for the unknown element. For the time being disable the no warning on the specific payload until the feature is added.


khouzam:
Simple text fix

PR #1368, opened 2018-04-11
explicit image changes

nesalang:
add parameters for image width and height for explicit initialization
this is draft version to get feedback on the changes.


source/shared/cpp/ObjectModel/Image.cpp
```diff
@@ -145,6 +174,11 @@ std::shared_ptr<BaseCardElement> ImageParser::DeserializeWithoutCheckingType(
     image->SetImageSize(ParseUtil::GetEnumValue<ImageSize>(json, AdaptiveCardSchemaKey::Size, ImageSize::None, ImageSizeFromString));
     image->SetAltText(ParseUtil::GetString(json, AdaptiveCardSchemaKey::AltText));
     image->SetHorizontalAlignment(ParseUtil::GetEnumValue<HorizontalAlignment>(json, AdaptiveCardSchemaKey::HorizontalAlignment, HorizontalAlignment::Left, HorizontalAlignmentFromString));
+    image->SetPixelWidth(ParseUtil::GetFloat(json, AdaptiveCardSchemaKey::PixelWidth, 0.0f, false));
+    image->SetPixelHeight(ParseUtil::GetFloat(json, AdaptiveCardSchemaKey::PixelHeight, 0.0f, false));
+    if(image->GetPixelWidth() || image->GetPixelHeight()) {
+        image->SetImageSize(ImageSize::Explicit);
+    }
```
RebeccaAnne: What's the spec for this? Is the card author supposed to set the image size to explicit if they're using explicit sizes?

nesalang: No, card author is not required to set this value. They can specify either width or height or both, and either case, we explicitly set the wdith and height according to their input values. 

---
In reply to: [181459672](https://github.com/Microsoft/AdaptiveCards/pull/1368#discussion_r181459672) [](ancestors = 181459672)

RebeccaAnne: Is setting this to "Explicit" is supported in the schema? I'm wondering what happens if they set the size to "explicit" and *don't* set width and height? Would it be better if image size doesn't have an explicit option and we just ignore image size if height and width are set?

---
In reply to: [182288887](https://github.com/Microsoft/AdaptiveCards/pull/1368#discussion_r182288887) [](ancestors = 182288887,181459672)

nesalang: that's very good point. It's not in the schema. in such case, we can create a parse warning and set the size to "None". having it makes code cleaner in other platforms. But I do see your concern, and if you feel strong about it, I will remove the enum value. let me know

---
In reply to: [182480765](https://github.com/Microsoft/AdaptiveCards/pull/1368#discussion_r182480765) [](ancestors = 182480765,182288887,181459672)

RebeccaAnne: I do like your idea of adding a parse warning when these aren't consistent. Maybe check with Matt and see what he thinks about adding a parse warning in cases where these related properties are not set consistently (explicit w/h properties and size are both set, for example)

---
In reply to: [182541849](https://github.com/Microsoft/AdaptiveCards/pull/1368#discussion_r182541849) [](ancestors = 182541849,182513104,182480765,182288887,181459672)

PR #1379, opened 2018-04-16
iOS: Do not call render at viewDidLoad

yeohan:
## Defect
We wrongly called the render method in viewDidLoad. This resulted in double rendering of the same adaptive card, and multiple unnecessary calls to viewLoadElements.

This reproduces with renderAsViewController.

## Fix
Do not call render in viewDidLoad.

## Tests
Using a custom app, renderAsViewC renders the adaptive card with no dupulication.
Using the Visualizer app, adaptive cards are rendered as previous.


nesalang:
sorry, didn't realize, you can't merge

PR #1383, opened 2018-04-18
Reduce calls to expensive locale functions

RebeccaAnne:
Profiling showed that the locale functions were taking a significant amount of time during rendering. Updated the date time parser code to not call locale functions until we actually encounter a date. This change results in a ~20-25% reduction in the overall render time.

### DIFF
--- a/source/uwp/Renderer/lib/DateTimeParser.cpp
+++ b/source/uwp/Renderer/lib/DateTimeParser.cpp
@@ -14,7 +14,20 @@ DateTimeParser::DateTimeParser(const std::string& language)
 std::string DateTimeParser::GenerateString(DateTimePreparser text)
 {
     std::wostringstream parsedostr;
-    bool haveSetLanguage = false;
+
+    if (text.HasDateTokens())
+    {
+        std::locale language;
+        try
+        {
+            language = std::locale(m_languageString.c_str());
+        }
+        catch (...)
+        {
+            language = std::locale("");
+        }
+        parsedostr.imbue(language);
+    }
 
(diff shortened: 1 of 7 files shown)


source/uwp/Renderer/lib/DateTimeParser.cpp
```diff
@@ -29,9 +23,24 @@ std::string DateTimeParser::GenerateString(DateTimePreparser text)
...
+        {
+            std::locale language;
+            try
+            {
+                language = std::locale(m_languageString.c_str());
+            }
+            catch (...)
+            {
+                language = std::locale("");
+            }
```
almedina-ms: For every date we have to make the locale call, wouldnt it be better to have a flag in the DateTimePreparser that checks if at least one of the elements is a date that must be localized and so we could move this piece of code outside the for loop so the locale call will not be generated for each date?

RebeccaAnne: The haveSetLanguage boolean prevents us from calling this more than once.

---
In reply to: [182553935](https://github.com/Microsoft/AdaptiveCards/pull/1383#discussion_r182553935) [](ancestors = 182553935)

RebeccaAnne: That said, i do think your suggestion is cleaner, updated to use that approach

---
In reply to: [182561429](https://github.com/Microsoft/AdaptiveCards/pull/1383#discussion_r182561429) [](ancestors = 182561429,182553935)

PR #1388, opened 2018-04-19
Improve FactSet behavior in .NET/UWP/Android/iOS renderers

khouzam:
It looked like we had some inconsistencies in our processing of facts and fact sets in our renderers.

This tries to consolidate the behaviors for wrapping text as well as respect the MaxWidth property from the host config for the Fact Title.

If wrap (default) is set, the fact title and/or value will wrap when getting to the end of the column.
If MaxWidth (default 150) is set on the Title in host config, the title column will not extend beyond that value and wrap if necessary. The value column will take the remaining space in the FactSet.


source/shared/cpp/ObjectModel/HostConfig.h
```diff
@@ -54,7 +54,7 @@ struct TextConfig
     ForegroundColor color = ForegroundColor::Default;
     bool isSubtle = false;
     bool wrap = true;
-    unsigned int maxWidth = 150;
+    unsigned int maxWidth = ((unsigned int)((unsigned int)~0));
```
RebeccaAnne: >((unsigned int)((unsigned int) [](start = 28, length = 30)

What does the double casting do? 

Would std::numeric_limits<unsigned int>::max() work here?

samples/v1.0/Elements/FactSet.json
```diff
@@ -7,8 +7,8 @@
             "type": "FactSet",
             "facts": [
                 {
-                    "title": "Fact 1",
-                    "value": "Value 1"
+                    "title": "This is a long FactSet that should wrap the title",
```
RebeccaAnne: >"title": "This is a long FactSet that should wrap the title", [](start = 20, length = 61)

Should we make this a test file rather than changing our main FactSet example? I believe this is the one that will show up on [http://adaptivecards.io/explorer/FactSet.html](http://adaptivecards.io/explorer/FactSet.html) so we might want to keep it "nice looking"

khouzam: Good point. Moved it to test.

source/shared/cpp/ObjectModel/HostConfig.h
khouzam: Left over from development. Removed the extra cast. Tried std:numeric_limits<unsigned int>::max() but it seems that a function call in a struct definition doesn't work.

source/dotnet/Library/AdaptiveCards/Rendering/TextBlockConfig.cs
```diff
@@ -34,6 +34,12 @@ public class TextBlockConfig
...
         /// </summary>
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public bool Wrap { get; set; }       
+        public bool Wrap { get; set; }
+
+        /// <summary>
+        /// The maximum width for text
+        /// </summary>
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public uint MaxWidth { get; set; }
```
matthidinger: While `uint` is a better type I'm pretty sure we've been using `int` everywhere for this. Up to you

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
```diff
@@ -446,17 +448,22 @@ protected static HtmlTag FactSetRender(AdaptiveFactSet factSet, AdaptiveRenderCo
...
+                    Wrap = context.Config.FactSet.Value.Wrap,
+                    // MaxWidth is not supported on the Value of FactSet. Do not set it.
                 };
                 var uiValue = context.Render(factValue)
                     .AddClass("ac-factvalue");
 
                 // create row in factset 
                 var uiRow = uiFactSet
                     .AddBodyRow();
+                uiRow.Style("height", "1px");
```
matthidinger: What is this for?

source/dotnet/Library/AdaptiveCards/AdaptiveTextBlock.cs
```diff
@@ -108,5 +108,15 @@ public AdaptiveTextBlock(string text)
...
+
+        /// <summary>
+        ///   The maximum width of the textblock
+        /// </summary>
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+#if !NETSTANDARD1_3
+        [XmlAttribute]
+#endif
+        [DefaultValue(0)]
+        public uint MaxWidth { get; set; }
```
matthidinger: Although `uint` is better, we used `int` previously as seen in `MaxLines` - maybe better to stay consistent

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
khouzam: This has the facts be top aligned instead of centered aligned. 

---
In reply to: [185054118](https://github.com/Microsoft/AdaptiveCards/pull/1388#discussion_r185054118) [](ancestors = 185054118)

PR #1399, opened 2018-04-20
 Don't create the xml document if there's no html

RebeccaAnne:
Profiling showed significant time spent in msxml. Updated code to not create the xml document if the text doesn't have markdown in it. Thanks to Joseph for the update to the markdown parser to allow me to detect this. Change results in ~15% improvement in renderer performance.

### DIFF
--- a/source/shared/cpp/ObjectModel/MarkDownParsedResult.cpp
+++ b/source/shared/cpp/ObjectModel/MarkDownParsedResult.cpp
@@ -44,6 +44,7 @@ void MarkDownParsedResult::AppendParseResult(MarkDownParsedResult &x)
     }
     m_codeGenTokens.splice(m_codeGenTokens.end(), x.m_codeGenTokens);
     m_emphasisLookUpTable.splice(m_emphasisLookUpTable.end(), x.m_emphasisLookUpTable);
+    m_isHTMLTagsAdded = m_isHTMLTagsAdded || x.HasHtmlTags();
 }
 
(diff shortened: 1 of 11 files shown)


source/shared/cpp/ObjectModel/MarkDownParsedResult.cpp
```diff
@@ -228,7 +229,7 @@ void MarkDownParsedResult::MatchLeftAndRightEmphasises()
                 }
             }
             // check which one has leftover delims
-            (*currentLeftEmphasis)->GenerateTags(*currentEmphasis);
+            m_isHTMLTagsAdded = m_isHTMLTagsAdded | (*currentLeftEmphasis)->GenerateTags(*currentEmphasis);
```
khouzam: >ed | (*cu [](start = 47, length = 9)

Should be logical or not bitwise

khouzam: This code never calls GenerateTags when m_isHTMLTagsAdded is true. The condition should either be flipped or separated into two statements.

---
In reply to: [183227326](https://github.com/Microsoft/AdaptiveCards/pull/1399#discussion_r183227326) [](ancestors = 183227326)

nesalang: thanks for catching this.

---
In reply to: [183489214](https://github.com/Microsoft/AdaptiveCards/pull/1399#discussion_r183489214) [](ancestors = 183489214,183227326)

PR #1404, opened 2018-04-26
Making the UWP Test app, convert all line endings to LF for consistency

khouzam:
In order to avoid false positive failures when files are checked out with Unix line endings versus Windows line endings, I've make the UWP test app convert all line endings to Unix (easier to replace \r\n with \n than the other way around).

This required all the expected results to be regenerated.


RebeccaAnne:
I ran this locally and I see source changed results for Solitare and AutoImageSet. Do those need to be updated?

khouzam:
Updated, they were already failing sometimes, so yeah they should be updated anyway.

---
In reply to: [385105550](https://github.com/Microsoft/AdaptiveCards/pull/1404#issuecomment-385105550) [](ancestors = 385105550)

PR #1427, opened 2018-05-04
explicit column width for uwp

nesalang:
externally, the property is expressed as width
internally, the property is expressed as pixelWidth


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1491,10 +1491,10 @@ AdaptiveNamespaceStart
         ComPtr<IAdaptiveCardResourceResolvers> resourceResolvers;
         THROW_IF_FAILED(renderContext->get_ResourceResolvers(&resourceResolvers));
 
-        UINT32 explicitWidth = 0, explicitHeight = 0;
-        THROW_IF_FAILED(adaptiveImage->get_Width(&explicitWidth));
+        UINT32 pixelWidth = 0, explicitHeight = 0;
```
RebeccaAnne: >explicitHeight [](start = 31, length = 14)

If we're going to call it pixelWidth, let's call it pixelHeight as well

samples/Tests/Column.Explicit.Size.json
```diff
@@ -0,0 +1,52 @@
...
+	"version": "1.0",
+	"body": [
+		{
+		    "type": "ColumnSet",
+		    "columns": [
+		        {
+		            "width": "50px",
+		            "items": [
+		                {
+		                    "type": "Image",
```
RebeccaAnne: >		                     [](start = 0, length = 22)

Make the spacing consistent. You should use tabs in the json samples.

source/uwp/Renderer/lib/XamlBuilder.cpp
nesalang: thanks for catching this; this shouldn't have changed.

---
In reply to: [186201969](https://github.com/Microsoft/AdaptiveCards/pull/1427#discussion_r186201969) [](ancestors = 186201969)

samples/Tests/Column.Explicit.Size.json
nesalang: o.k.


---
In reply to: [186202407](https://github.com/Microsoft/AdaptiveCards/pull/1427#discussion_r186202407) [](ancestors = 186202407)

PR #1430, opened 2018-05-04
Fix AutoImageSet inconsistencies when loading images asynchronously.

khouzam:
We had logic to make sure that we don't stretch an image when set to autosize.
The logic would check the size of the parent and set the MaxWidth and MaxHeight if the parent was bigger.
When in a ImageSet, the parent would be the whole grid and that logic would not work properly.
This changes the logic to simply force the MaxWidth and MaxHeight of Auto images to their pixel size.

I've also added a new comparison on the TestApp to show the live XAML card.
This is the card before being rendered to a bitmap.
It can then be examined to see the root cause of test failures using the live tree viewer.


source/uwp/AdaptiveCards.sln
```diff
@@ -36,20 +36,20 @@ Global
 		{CF169157-F487-420C-8B9B-CA3ABE2BB209}.Release|x64.Build.0 = Release|x64
 		{CF169157-F487-420C-8B9B-CA3ABE2BB209}.Release|x86.ActiveCfg = Release|Win32
 		{CF169157-F487-420C-8B9B-CA3ABE2BB209}.Release|x86.Build.0 = Release|Win32
-		{B5DECA6E-D9EA-476D-9B7E-34D3B3044CD8}.Debug|ARM.ActiveCfg = Debug|ARM
```
RebeccaAnne: >{B5DECA6E-D9EA-476D-9B7E-34D3B3044CD8}.Debug|ARM.ActiveCfg = Debug|ARM [](start = 2, length = 70)

Remove changes to the solution file

PR #1434, opened 2018-05-04
Add testing for serialization to json and fix bugs

RebeccaAnne:
Added serialization testing to the UWP test app. The test app will now round trip the json through AdatpiveCard.FromJsonString->card.ToJsonString->AdaptiveCard.FromJsonString before rendering to ensure that all json necessary for rendering is serialized correctly.  Additionally, the test app will keep expected results from json serialization and report changes as failures.

As part of this work, I updated the serialization code to be less verbose by not writing empty strings and default values when not needed.

As a result of the testing, I found and fixed a number of bugs. Specifically:
- AdditionalProperties were not serialized on elements or actions
- The separator value for elements was uninitialized, leading to unexpected results when constructing elements using the shared model API (rather than deserializing from json)
- Cards without a select action failed to serialize due to a null pointer
- Card language was not stored at the UWP level (causing it not to be serialized), nor exposed through the API. 
- Pixel sizes for images were not serialized at the shared model layer
- Inputs were using a version of our string conversion which failed on null pointers, causing them not to serialize in many cases. Additionally, the values of choice sets and number inputs were not set back on the shared model by the UWP layer, causing them not to serialize.
- Icon url was not set on the shared model by the uwp layer, causing it not to be serialized.
- The explicit image size code had a bug which caused it not to work if the non-explicit size wasn't also set. Updated the xaml builder code to address.

Finally, when fixing some variable initialization issues, realized that we had several extraneous constructors for our types at the shared model layer that are not being used. Removed them.


tuanh118:
:100:

Issue #945, opened 2017-11-14
Action.Submit "data" property as string


matthidinger:
Re #652 

The docs for Action.Submit today say that `data` can be a string or object. WebChat currently uses a string to trigger an "imBack" to the bot, and they use data as an object to post back more complicated data.

http://adaptivecards.io/explorer/Action.Submit.html

The following currently causes nothing to happen when I press submit. This seems like a recent bug?

I believe the expected response would be:
```json
{
 "data": "hello",
 "firstName": "",
 "lastName": ""
}
```

Payload:
```json
{
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "Input.Text",
            "id":"lastName",
            "placeholder": "What is your last name?"
        }
    ],
    "actions": [
        {
            "type": "Action.Submit",
            "title": "Action.Submit",
            "data": "hello"
        }
    ]
}
```

This payload works:
```json
            "data": {
                "x" : 13
            }      
```

mrbrennonwilliams:
Probably because of the sample, I limited the data property to that of a JSON string. I probably incorrectly assumed that a complex object should be serializable.

dclaux:
The contract _has_ to be that data is an object. Otherwise, what would be the contract with the host as to the object it receives when the action is executed?

In the above example, with data defined as
```
"data": {
    "x" : "hello"
}
```
the host gets the following object back as the action is executed:
```
"data": {
    "x": "hello",
    "lastName": null
}
```
If data were defined like this:
```
"data": "hello"
```
then what would the host receive?
```
"data": {
    "????": "hello"
    "lastName": null
}
```

If we want to support any type for the data property, then the contract with the host would have to be something like the below.

If data is a string (or any value type):
```
"data": {
    "originalData": "hello",
    "lastName": null
}
```

If data is an object:
```
"data": {
    "originalData": {
        "x": "hello"
    },
    "lastName": null
}
```

PR #1447, opened 2018-05-12
supporting strings and objects for SubmitAction data

ignacionr:
Fixes: the C++ model supported SubmitAction data property roundtrip only if a string.
Likely to be related to https://github.com/Microsoft/AdaptiveCards/issues/945 or similar issues. The original code would parse an object but would only set a string.


ignacionr:
Also notice a separate method for getting the property value as a Json::Value was created (so to not break existing code that may depend on the string-only interface).

ignacionr:
Caught a bug with the card empty constructor leaving the style uninitialized (this was causing random exceptions). Removed unnecessary string copying on parameter passing.

Shy to remove some obvious misinterpretations of the 'const' keyword (like passing shared pointers as const copy on the pointer, and passing copies of enums as const) because I miss what the intent might have been (other than just lack of understanding of what const means). They're just bad coding but at least don't hurt performance or semantics.

ignacionr:
Hey guys, this one is getting cold and will soon need manual merge. If you don't like it just let me know, we can discuss it. @RebeccaAnne I am shy to put things on your already big plate, but if you have a second to track this, do comment. Thanks!

RebeccaAnne:
I approved this one back on May 18th, are you waiting on something else from me?

ignacionr:
@RebeccaAnne do you think you can merge and close? I don't have write access...

RebeccaAnne:
Ah, will do, didn't realize you needed us to do that. Thanks for the fixes!

PR #1448, opened 2018-05-13
Cpp improvements

ignacionr:
You are making a beautiful thing happen, truly thankful for that. Please consider this collaboration.

- ~~Bug fix: RemoveParser was backwards (unlikely to be called ever, yet...)~~ (my bad, the code wasn't backwards but the check was legit -whether the parser was for a "known"/standard element).

- Bug fixes: many empty constructors would forget to initialize enum members.

- Semantic fix: it's customary to use toupper (not tolower) when computing case-insensitive things. In English it makes no difference, but in languages like Greek, the lower case symbol will depend of the position of the letter within the word (not true for uppercase).

- Performance improvement: a majority of strings were being passed in by copy or const copy; made to pass-in by const ref (of course not for the parser work).

Additionally, a suggestion on parsing; you may be too young or too far from compiler teams to know Bison or lexx/yacc, but there are better ways to construct these machine states without so much manual work.

### DIFF
--- a/source/shared/cpp/ObjectModel/ActionParserRegistration.cpp
+++ b/source/shared/cpp/ObjectModel/ActionParserRegistration.cpp
@@ -20,7 +20,7 @@ AdaptiveSharedNamespaceStart
         });
     }
 
-    void ActionParserRegistration::AddParser(std::string elementType, std::shared_ptr<ActionElementParser> parser)
+    void ActionParserRegistration::AddParser(std::string const &elementType, std::shared_ptr<ActionElementParser> parser)
     {
         if (m_knownElements.find(elementType) == m_knownElements.end())
         {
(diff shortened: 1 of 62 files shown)


source/shared/cpp/ObjectModel/ActionParserRegistration.cpp
```diff
@@ -32,19 +32,15 @@ AdaptiveSharedNamespaceStart
         }
     }
 
-    void ActionParserRegistration::RemoveParser(std::string elementType)
+    void ActionParserRegistration::RemoveParser(std::string const &elementType)
     {
-        if (m_knownElements.find(elementType) == m_knownElements.end())
```
RebeccaAnne: The intent here is to not allow callers to remove parsers for the known element types. We want to remove the parser only if they type is not in m_knownElements.

ignacionr: Thanks for checking this and for helping me understand. I will revert that part.

ignacionr: Got the original code back on https://github.com/Microsoft/AdaptiveCards/pull/1448/commits/d3017a3c0398ee43c18c772a960202d5738e40b3 ; please let me know if that's alright.

ignacionr:
Bummer I missed that! Will fix as soon as I get to the office. Sorry the time zone diff adds for the unintended suspense :)

ignacionr:
This is done @paulcam206 thanks for your review ; by the way welcome to the team, I understand you're the Taniwha guy, right?

Issue #1449, opened 2018-05-13
[Android][Bug][Teams] Multichoice input does no return the correct value and returns title in the case of style: compact 


rednivrug15:
Multichoice inputwhich have compact style does not return the correct value when the Submit button is pressed. It returns the title instead of the value . 

Sample JSON payload :
`{
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "type": "AdaptiveCard",
  "version": "1.0",
  "speak": "Your  meeting about \"Adaptive Card design session\" is starting at 12:30pmDo you want to snooze  or do you want to send a late notification to the attendees?",
  "body": [
    {
      "type": "TextBlock",
      "text": "Adaptive Card design session",
      "size": "large",
      "weight": "bolder"
    },
    {
      "type": "TextBlock",
      "text": "Conf Room 112/3377 (10)",
      "isSubtle": true
    },
    {
      "type": "TextBlock",
      "text": "12:30 PM - 1:30 PM",
      "isSubtle": true,
      "spacing": "none"
    },
    {
      "type": "TextBlock",
      "text": "Snooze for"
    },
    {
      "type": "Input.ChoiceSet",
      "id": "snooze",
      "style": "compact",
      "value": "5",
      "choices": [
        {
          "title": "5 minutes",
          "value": "5",
          "isSelected": true
        },
        {
          "title": "15 minutes",
          "value": "15"
        },
        {
          "title": "30 minutes",
          "value": "30"
        }
      ]
    }
  ],
  "actions": [
    {
      "type": "Action.Submit",
      "title": "Snooze",
      "data": {
        "x": "snooze"
      }
    },
    {
      "type": "Action.Submit",
      "title": "I'll be late",
      "data": {
        "x": "late"
      }
    }
  ]
}`

Issue #1432, opened 2018-05-04
[Android][HostConfig] Currently the android SDK does not follow the host config property for spaces between buttons . 


rednivrug15:
Android SDK does not obey the host config property for spaces between the buttons .

Issue #1426, opened 2018-05-03
[Android][Bug][Teams] We would want buttons to be horizontally scrollable . Currently the buttons are just placed in the width available . 


rednivrug15:
Since we cannot override the whole actions rendering , i am unable to make the buttons horizontally scrollable . Please suggest an alternative if there is any .

PR #1452, opened 2018-05-13
Fix Button spacing issues and add horizontal scroll view in case actions orientation is horizontal . 

rednivrug15:
1. Fix spacing between two buttons . It should follow the hostConfig property for spacing between two buttons [#1432](https://github.com/Microsoft/AdaptiveCards/issues/1432)

2. Fix the spacing between the buttons layout and the last card element . This should also follow the host config  property [#1432](https://github.com/Microsoft/AdaptiveCards/issues/1432)

3. Add horizontal scroll view to make buttons scrollable  in case they exceed the width available [#1426](https://github.com/Microsoft/AdaptiveCards/issues/1426). 

4. Fix a bug in multiple choice input where title is returned instead of returning the value when Action.Submit button is pressed [#1449](https://github.com/Microsoft/AdaptiveCards/issues/1449).

5. Fix fact set rendering issue where the content is clipped [#1472](https://github.com/Microsoft/AdaptiveCards/issues/1472)

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
@@ -235,7 +235,6 @@ public class AdaptiveCardRenderer
             actionButtonsLayout.setGravity(Gravity.CENTER_HORIZONTAL);
         }
 
-
         if (hostConfig.getActions().getActionsOrientation().swigValue() == ActionsOrientation.Vertical.swigValue())
         {
             actionButtonsLayout.setOrientation(LinearLayout.VERTICAL);
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
@@ -244,8 +244,10 @@ public class AdaptiveCardRenderer
             actionButtonsLayout.setOrientation(LinearLayout.HORIZONTAL);
         }
 
+
         Spacing spacing = hostConfig.getActions().getSpacing();
-        BaseCardElementRenderer.setSpacingAndSeparator(context, viewGroup, spacing, false, hostConfig, true);
+        /* Passing false for seperator since we do not have any configuration for seperator in actionsConfig */
+        BaseCardElementRenderer.setSpacingAndSeparator(context, viewGroup, spacing, false, hostConfig, true /* Horizontal Line */);
 
         if (viewGroup != null)
         {
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
@@ -260,7 +260,7 @@ public class AdaptiveCardRenderer
             ActionElementRenderer.getInstance().render(renderedCard, context, fragmentManager, actionButtonsLayout, actionElement, cardActionHandler, hostConfig);
         }
 
-        if (viewGroup != null)
+        if (viewGroup != null && hostConfig.getActions().getActionsOrientation().swigValue() == ActionsOrientation.Horizontal.swigValue())
         {
             HorizontalScrollView horizontalScrollView = new HorizontalScrollView(context);
             horizontalScrollView.setHorizontalScrollBarEnabled(false);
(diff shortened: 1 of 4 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
```diff
@@ -255,6 +260,15 @@ else if (alignment == ActionAlignment.Center.swigValue())
             ActionElementRenderer.getInstance().render(renderedCard, context, fragmentManager, actionButtonsLayout, actionElement, cardActionHandler, hostConfig);
         }
 
+        if (viewGroup != null && hostConfig.getActions().getActionsOrientation().swigValue() == ActionsOrientation.Horizontal.swigValue())
+        {
```
msmoosa: We are adding actionbuttonslayout above and then removing it from view group. Can you change the code above to selectively add to scroll view if required and avoid the add and removal

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/ComboBoxInputHandler.java
```diff
@@ -24,6 +24,12 @@ public String getInput()
     {
         // no need to validate
         ChoiceSetInput choiceSetInput = (ChoiceSetInput) m_baseInputElement;
-        return (String) getSpinner().getSelectedItem();
+        int index = getSpinner().getSelectedItemPosition();
+        String selectedItem = "";
+        if (index >= 0)
```
msmoosa: nit: check index < getchoices().length just to be super safe?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/ComboBoxInputHandler.java
```diff
@@ -24,6 +24,12 @@ public String getInput()
...
+        {
+            selectedItem = choiceSetInput.GetChoices().get(index).GetValue();
```
msmoosa: Simply return here.. why store?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
```diff
@@ -242,6 +244,9 @@ else if (alignment == ActionAlignment.Center.swigValue())
             actionButtonsLayout.setOrientation(LinearLayout.HORIZONTAL);
         }
 
+        Spacing spacing = hostConfig.getActions().getSpacing();
+        BaseCardElementRenderer.setSpacingAndSeparator(context, viewGroup, spacing, false, hostConfig, true);
```
almedina-ms: Can you add a comment next to the false and true values to know what do they stand for?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/ComboBoxInputHandler.java
almedina-ms: I would prefer having one single return point in this case, after all it's just one extra line of code

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
rednivrug15: Yes. Will add the same comment as has been used  elsewhere .

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
almedina-ms: @rednivrug15  I agree with this comment, can you change it?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/ComboBoxInputHandler.java
rednivrug15: Not needed as the SDK prefers single point of return .

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
rednivrug15: @almedina-ms  Since the horizontal scroll has been implemented I will revert the change for horizontal scroll and keep rest of the bug fixes here . Can you please look into this ?

PR #1467, opened 2018-05-16
W4 warnings for shared cpp

paulcam206:
Bump warning level to `W4` and make warnings errors. Fix issue found by `W4`. Note that we disable `C4702` because of issues in jsoncpp.


paulcam206:
Thanks @RebeccaAnne!

BoomerHudson:
I notice a lot of email from you my main computer has been getting fixed twice this month my second computer been freezing up and if that's not good enough my ui on my phone has been stopping so I not been able to do much I am waiting on my main computer to arrive today and I can finally get back in the swing of things again. I don't know much about what I should be doing to learn what I need to be learning thanks for the help yeal gives

Get Outlook for Android

PR #1469, opened 2018-05-16
Add support for more properties in property sheet

dclaux:
Add way to easily style buttons in CSS over containers with "style": "emphasis"


dclaux:
The last couple of commits add:
- FactSet.facts editing
- ImageSet.images editing
- Action.Http.headers edition
- Input.ChoiceSet.choices editing
- Card parsing and validation warnings

PR #1471, opened 2018-05-17
[Android] Add horizontal scroll bar for horizontal layout actions

almedina-ms:
Fix for this https://github.com/Microsoft/AdaptiveCards/issues/1426

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
@@ -10,6 +10,7 @@ import android.support.v4.app.FragmentManager;
 import android.view.Gravity;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.HorizontalScrollView;
 import android.widget.LinearLayout;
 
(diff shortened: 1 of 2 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
```diff
@@ -244,7 +246,17 @@ else if (alignment == ActionAlignment.Center.swigValue())
 
         if (viewGroup != null)
         {
-            viewGroup.addView(actionButtonsLayout);
+            if(actionButtonsLayoutOrientation == ActionsOrientation.Horizontal.swigValue())
+            {
```
rednivrug15: Also this does not work on my machine . The show card action code : 
` 
            ViewGroup parent = (ViewGroup) viewGroup.getParent().getParent(); 
            ViewGroup hiddenCards = (ViewGroup) parent.getChildAt(1);
            hiddenCards.addView(invisibleCard);
`

Looks for the parents's parent first child ( Expected to be show card layout ) . In this case horizontal scroll view is direct parent whose parent is the card layout ( not the root card layout which is expected ) . Hence any action which is show card will not render here

PR #1445, opened 2018-05-12
share shared model's c++ header

nesalang:
resolves issue #1463

PR #1444, opened 2018-05-11
fix for #1419

nesalang:
fixes issues for #1419 where when UIImage size is zero, app crashes.

Issue #1419, opened 2018-05-02
Crash when image size is zero


an888ha:
App Crashing when scrolling fast in a view which has multiple adaptive cards iOS
App is crashing when we have multiple adaptive cards in a view having sanitized image urls. The reason being image.width = 0.
Multiplier is not finite!  That's illegal.  multiplier:nan
(
	0   CoreFoundation                      0x000000011bf091e6 __exceptionPreprocess + 294
	1   libobjc.A.dylib                     0x000000011b5a6031 objc_exception_throw + 48
	2   CoreFoundation                      0x000000011bf0e472 +[NSException raise:format:arguments:] + 98
	3   Foundation                          0x00000001168ae652 -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 193
	4   Foundation                          0x0000000116832840 -[NSLayoutConstraint _setMultiplier:] + 336
	5   Foundation                          0x0000000116832567 +[NSLayoutConstraint constraintWithItem:attribute:relatedBy:toItem:attribute:multiplier:constant:] + 248
	6   AdaptiveCards                       0x000000010ed16087 __36-[ACRView processImageConcurrently:]_block_invoke_2 + 421
	7   libdispatch.dylib                   0x000000011796c807 _dispatch_call_block_and_release + 12
	8   libdispatch.dylib                   0x000000011796d848 _dispatch_client_callout + 8
	9   libdispatch.dylib                   0x000000011797892b _dispatch_main_queue_callback_4CF + 628
	10  CoreFoundation                      0x000000011becbc99 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
	11  CoreFoundation                      0x000000011be8fea6 __CFRunLoopRun + 2342
	12  CoreFoundation                      0x000000011be8f30b CFRunLoopRunSpecific + 635
	13  GraphicsServices                    0x000000011e68ca73 GSEventRunModal + 62
	14  UIKit                               0x0000000117ad30b7 UIApplicationMain + 159
	15  TeamsAppCore                        0x000000010f75d4aa AXPApplicationMain + 58
	16  TeamSpaceApp                        0x000000010e2c0ec9 main + 393
	17  libdyld.dylib                       0x000000011c402955 start + 1
)

msmoosa:
The reason being image.width = 0.

PR #1479, opened 2018-05-18
Adaptive Cards Patch for 1.0.2 release

khouzam:
This cherry-picks changes taken from our master branch for issues that are blocking customers and should address the following issues:

- Do not call render at viewDidLoad (#1379)
- Updating the swig file to include vector<string> and running swig (#1405)
- Fixing crash where no item in RadioGroup is selected (#1406)
- Improve FactSet behavior in .NET/UWP/Android/iOS renderers (#1388)
- Missing changes from code review feedback for FactSets (#1417)
- Fix for Zero-Width/Height images #1419 (#1444)
- Showcard to call delegate (#1458)
- [Android] Add horizontal scroll bar for horizontal layout actions (#1471)
- Fix Button spacing issues and add horizontal scroll view in case actions orientation is horizontal .  (#1452)
- Fixes adding newline char issue in the text (#1465)
- Share shared model's c++ header (#1445)
- Updated sample app to use didLoadElements delegate (#1473)
- Fix double button rendering (#1475)
- Jwoo/ios text clipping fix (#1477)
- Update Android build version to 1.0.2
- Update ios build version


almedina-ms:
I tested all the fixes for ANDROID I could test here and they worked

PR #1498, opened 2018-05-22
added background color to image shared object

tuanh118:


nesalang:
would be nice to have unit test.:smile:

tuanh118:
@nesalang sorry I wanted to speed up in order for the UWP implementation to be done today. What kind of test do you think would be appropriate? I can't seem to be able to find a good example.

PR #1499, opened 2018-05-22
Introduce `.clang-format` rules to enforce style consistency

paulcam206:
We will eventually be moving this standardization up the tree, but for now we're just enforcing it in `shared/cpp`.

In order for the tools to work correctly, I've reworked our namespacing macro.

There are lots of different ways to utilize this functionality. Visual Studio support is built in to recent versions and available as a plugin in older versions. Lots of other editors have native or pluggable support for this format. Additionally, the `clang-format` console application is available for pretty much all platforms.

### DIFF
--- a/source/shared/cpp/ObjectModel/pch.h
+++ b/source/shared/cpp/ObjectModel/pch.h
@@ -4,10 +4,8 @@
 #include "winPch.h"
 #endif
 
-#ifndef AdaptiveSharedNamespaceStart
-#define AdaptiveSharedNamespaceStart namespace AdaptiveCards {
+#ifndef AdaptiveSharedNamespace
 #define AdaptiveSharedNamespace AdaptiveCards
-#define AdaptiveSharedNamespaceEnd }
 #endif
 
 #include <memory>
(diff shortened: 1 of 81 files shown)


source/shared/cpp/ObjectModel/pch.h
```diff
@@ -4,10 +4,8 @@
 #include "winPch.h"
 #endif
 
-#ifndef AdaptiveSharedNamespaceStart
-#define AdaptiveSharedNamespaceStart namespace AdaptiveCards {
```
RebeccaAnne: The problem this was solving was that the namespace on windows is nested and has more closing braces. We need a fix that allows us to copy this to windows without it breaking, which is why we had the braces defined here.

On windows we have:

#define AdaptiveSharedNamespaceStart namespace Windows { namespace Internal { namespace AdaptiveCards { 
#define AdaptiveSharedNamespace Windows::Internal::AdaptiveCards
#define AdaptiveSharedNamespaceEnd }}}

paulcam206: When this comes down to Windows, we just need to remove `AdaptiveSharedNamespaceStart` and `AdaptiveNamespaceEnd` and then change to build using c++17, which allows nested namespace declarations (e.g. `namespace Foo::Bar::Baz { class Bleh; }`)

khouzam:
BTW I'd like to wait until we are done with our Windows features and the code has been merged upstream before taking this change.

paulcam206:
I'm going to rework this into two checkins for ease of checkin:
one submit for namespace changes, and one submit of .clang-format with formatting applied :)

PR #1500, opened 2018-05-22
Tupha/background color image uwp

tuanh118:
- Added support for image background color in 2 styles
  - Normal style: Created a Border element with background color to wrap around Image
  - Person style: Created a Grid element to contain 2 ellipses, one filled with the SolidColorBrush, the other filled with ImageBrush
- Added a new card payload to with its test results


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1622,7 +1654,27 @@ AdaptiveNamespaceStart
...
+            {
+                // Create a surronding border with solid color background to contain the image
+                ComPtr<IBorder> border = XamlHelpers::CreateXamlClass<IBorder>(HStringReference(RuntimeClass_Windows_UI_Xaml_Controls_Border));
+
+                ABI::Windows::UI::Color color;
+                THROW_IF_FAILED(GetColorFromString(HStringToUTF8(backgroundColor), &color));
+                ComPtr<IBrush> backgroundColorBrush = GetSolidColorBrush(color);
+                THROW_IF_FAILED(border->put_Background(backgroundColorBrush.Get()));
+
+                IUIElement* imageAsUiElement;
```
almedina-ms: ComPtr<IUIElement>

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1622,7 +1654,27 @@ AdaptiveNamespaceStart
...
+                THROW_IF_FAILED(xamlImage.CopyTo(&imageAsUiElement));
+                border->put_Child(imageAsUiElement);
```
almedina-ms: Check the return value

source/uwp/Renderer/lib/AdaptiveImage.cpp
```diff
@@ -55,6 +57,18 @@ AdaptiveNamespaceStart
         return m_url.Set(url);
     }
 
+    _Use_decl_annotations_
+        HRESULT AdaptiveImage::get_BackgroundColor(HSTRING* backgroundColor)
```
almedina-ms: Spacing issue here and in line 66

almedina-ms:
You also need to add the check for .NET not running the tests for your change as the tests are failing

PR #1503, opened 2018-05-23
show only one showcard at a time

nesalang:
corrected to desired behavior:
show only one show card at a time when show card button is touched.

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRContentStackView.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRContentStackView.mm
@@ -7,12 +7,14 @@
 
 #include "ACRContentStackView.h"
 #include "ACOHostConfigPrivate.h"
+#import "ACRShowCardTarget.h"
 
 using namespace AdaptiveCards;
 
 @implementation ACRContentStackView
 {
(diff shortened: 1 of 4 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRContentStackView.mm
```diff
@@ -7,12 +7,14 @@
 
 #include "ACRContentStackView.h"
 #include "ACOHostConfigPrivate.h"
+#import "ACRShowCardTarget.h"
 
 using namespace AdaptiveCards;
 
 @implementation ACRContentStackView
 {
     NSMutableArray* _targets;
```
bviglietta: >NSMutableArray* _targets; [](start = 4, length = 25)

What's this array for? It doesn't look like it's ever used.

nesalang: it's used in line 141. this view will be parent view of buttons that use the target, and the parent view will keep reference for target alive. 

---
In reply to: [191554601](https://github.com/Microsoft/AdaptiveCards/pull/1503#discussion_r191554601) [](ancestors = 191554601)

PR #1509, opened 2018-05-23
UWP: Host Config support for default poster and custom play button 

RebeccaAnne:
Also includes improvements to the default play button.


source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -49,45 +66,154 @@ void GetMediaPosterAsImage(
...
+    // Set the size
+    ComPtr<IFrameworkElement> rectangleAsFrameworkElement;
+    THROW_IF_FAILED(rectangle.As(&rectangleAsFrameworkElement));
+    THROW_IF_FAILED(rectangleAsFrameworkElement->put_Height(c_playIconSize));
+    THROW_IF_FAILED(rectangleAsFrameworkElement->put_Width(c_playIconSize));
+
+    // Round the corners
+    THROW_IF_FAILED(rectangle->put_RadiusX(c_playIconCornerRadius));
+    THROW_IF_FAILED(rectangle->put_RadiusY(c_playIconCornerRadius));
+
```
khouzam: I like the icon, but let's have a validation that this is how we want to ship.

RebeccaAnne: I'll follow up with Matt.

---
In reply to: [192467417](https://github.com/Microsoft/AdaptiveCards/pull/1509#discussion_r192467417) [](ancestors = 192467417)

PR #1510, opened 2018-05-23
corrected "auto" and "stretch" behavior of column width

nesalang:


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRColumnSetRenderer.mm
```diff
@@ -57,15 +57,28 @@ - (UIView* )render:(UIView<ACRIContentHoldingView> *)viewGroup
...
+                                    toItem:stretchView
+                                    attribute:NSLayoutAttributeWidth
+                                    multiplier:1
+                                    constant:0].active = YES;
+            }
+            stretchView = curView;
+        } else {
+            try
+            {
+                relativeColumnWidth = std::stof(column->GetWidth());
```
bviglietta: >GetWidth [](start = 56, length = 8)

What's supposed to happen if width is "auto"? Throw here, then reuse the multiplier of the previous column?

I don't like using exceptions like this, or having the "auto" behavior be so implicit. And I'm not sure the behavior is what you want either: a series of auto columns in a row could have geometrically increasing or decreasing width, as they keep reusing the same multiplier.

nesalang: for auto, no adjustment is needed, it will show its intrinsic content size. 

---
In reply to: [191563142](https://github.com/Microsoft/AdaptiveCards/pull/1510#discussion_r191563142) [](ancestors = 191563142)

Issue #389, opened 2017-06-05
Actions should allow an image to be displayed in addition to text for the title


weretygr:
# Implementation status
- [x] TypeScript (#1071)
- [x] UWP (#1278)
- [x] WPF (#1072)
- [x] iOS (#1279)
- [x] Android (#1280)
- [x] Documentation

# Problem
There's no way to display an image in addition to text on an action. Some hosts, like toast notifications, are built around this experience where developers can provide "image buttons" for their actions.

![image](https://user-images.githubusercontent.com/13246069/30493475-153a4d50-99f9-11e7-83d6-1cd3c476e504.png)

This can NOT be part of the card body. The actions need to be consistently rendered just like actions are today - with the only difference being that an image can be provided.

Card authors should NOT have control over the placement of the image. Otherwise the end user experience would be inconsistent, since some toast notifications would have the image on the left, some on the right, etc.

# Spec

## Schema changes

We add an `iconUrl` property to action...

| Property | Type | Required | Description |
|---|---|---|---|
| **type**| **IAction** | true | |
| **title** | string | true | Label for button or link that represents this action |
| **speak** | [Speak](/Microsoft/AdaptiveCards/blob/master/docs/SpeechAndAdvancedCustomization.md) | false | Specifies what should be spoken for this entire element.  This is simple text or SSML fragment |
| **iconUrl** | string | false | Optional icon to be shown on the action in conjunction with the title |

```json
"actions": [
  {
    "type": "Action.Submit",
    "title": "Snooze",
    "iconUrl": "ms-appx:///Assets/Snooze.png",
    "data": "snooze"
  },
  {
    "type": "Action.Submit",
    "title": "Dismiss",
    "iconUrl": "ms-appx:///Assets/Dismiss.png",
    "data": "dismiss"
  }

]
```

## Host Config

```json
"actions": {
    "iconPlacement": "aboveTitle|leftOfTitle",
    "iconSize": 30
}
```

Defaults: `aboveTitle` and `30`

**Update 8/3**: `iconSize` was changed from a string "30px" to a number [30], to match the unit conventions throughout host config. Also updated spec to clarify defaults and some behavior.

## Renderer Requirements

1. If `iconPlacement` is `aboveTitle` then the `iconSize` should be used as the **height** of the image, while maintaining aspect ratio. Card Authors should use square images for ideal portability between Hosts.
[...]

PR #1511, opened 2018-05-23
[HTML] Add support for icons in action buttons

dclaux:


dclaux:
These latest commits implement https://github.com/Microsoft/AdaptiveCards/issues/389

PR #1518, opened 2018-05-29
Tupha/relative image url uwp renderer android

tuanh118:
- Added support for loading local image (from `/res`)
- Added support for `imageBaseUrl` (both online and local)
- Refactored the updated image loading logic into a new parent class, `GenericImageLoaderAsync`, and renamed other loader classes for clearer intents. Now the image loader classes inheritance looks like this:
```
GenericImageLoaderAsync (common image loading logic)
-- BackgroundImageLoaderAsync (background image loader)
-- InnerImageLoaderAsync (images inside adaptive cards)
  -- ImageRendererImageLoaderAsync (image element loader)
  -- ActionElementRendererIconImageLoaderAsync (icon image inside action element loader)
```


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/GenericImageLoaderAsync.java
```diff
@@ -0,0 +1,145 @@
...
+        catch (Exception excep)
+        {
+            return new HttpRequestResult<>(excep);
+        }
+    }
+
+    // Helper function to load online image URL
+    private HttpRequestResult<Bitmap> loadOnlineImage(String url) throws IOException, URISyntaxException
+    {
+        Bitmap bitmap;
```
almedina-ms: >Bitmap bitmap; [](start = 8, length = 14)

[nit] Move this declaration closer to the first use or do: 

Bitmap bitmap = BitmapFactory.decode ...

Issue #1516, opened 2018-05-25
[HTML] Add ability to style ActionSet via CSS


dptejas:
# Problem
Currently, ActionSet elements can't be custom styled in CSS because no explicit class is applied to them by the HTML renderer.

# Solution
The HTML renderer should add the **ac-actionSet** class to an ActionSet's rendered DOM element.

dptejas:
Implemented the issue in #1521

Issue #1515, opened 2018-05-25
[HTML] Add isPrimary boolean property to the Action class


dptejas:
# Problem
For backwards compatibility with its previous Actionable Message card format, Outlook needs the ability to mark an action as "primary" so that it can get a different visual treatment.

# Solution
In the HTML renderer:
- Introduce an **isPrimary** boolean property on the Action class
- Introduce a way to custom parse Action objects (similarly to the custom parsing of Element objects)

# Design
New isPrimary property on the Action class:
```typescript
export class Action {
    ...
    isPrimary: boolean = false;
    ...
}
```

Custom action parsing events (global and per-card, like other events):
```typescript
export class AdaptiveCard {
    ...
    static onParseAction: (action: Action, json: any) => void = null;
    onParseAction: (action: Action, json: any) => void = null;
    ...
}
```

It is the responsibility of the host to handle the onParseAction event to custom parse the isPrimary action. isPrimary is not parsed out of the box and is not an official Adaptive property.

ignacionr:
Hello @dptejas 
The property you mention does not exist on the current http://adaptivecards.io/schemas/adaptive-card.json ; can you share where you've found it documented?

dclaux:
@ignacionr the property is indeed not documented. It will be introduced in the HTML renderer only, not in the schema (at least not for now.) Hosts who want to support this property will have to do so via extensibility (e.g. by custom parsing it.) THis property will make it possible to give a different visual treatment to the "primary".


I have updated the title and will add a slittle spec.

ignacionr:
This makes sense. Can you guys make sure that when rendered on HTML, pressing “enter” while the focus is on a control will execute this action?
It could be convenient.

dclaux:
Yes, we can do that. I just opened https://github.com/Microsoft/AdaptiveCards/issues/1517

matthidinger:
For reference, we have a few related issues on this topic: #861 and #272

dclaux:
https://github.com/Microsoft/AdaptiveCards/issues/861 is obsolete IMO, I no longer believe the solution I proposed in there is the right one. I think we should close it.

dptejas:
Implemented solution in #1521

PR #1521, opened 2018-05-30
Introduce style class for primary Action and ActionSet

dptejas:
This PR addresses issues #1516 & #1515


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2185,14 +2189,15 @@ export abstract class Action {
 
     id: string;
     title: string;
+    isPrimaryAction: boolean;
```
dclaux: Should be isPrimary, not isPrimaryAction

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2185,14 +2189,15 @@ export abstract class Action {
...
 
     toJSON() {
         let result = {};
 
         Utils.setProperty(result, "type", this.getJsonTypeName());
         Utils.setProperty(result, "id", this.id);
         Utils.setProperty(result, "title", this.title);
-
+        Utils.setProperty(result, "isPrimaryAction", this.isPrimaryAction);
```
dclaux: Please remove. This is not an official Adaptive property so it can't just be emitted in toJSON

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2243,6 +2248,10 @@ export abstract class Action {
     parse(json: any) {
         this.id = json["id"];
         this.title = json["title"];
+        this.isPrimaryAction = false;
```
dclaux: Not necessary

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2243,6 +2248,10 @@ export abstract class Action {
...
+        if (json["isPrimaryAction"] != null) {
```
dclaux: JSON property name should also be isPrimary.
We shouldn't parse isPrimary in the library, as it is not an official Adaptive property. It's ok to have support for it, but it is up to the host to custom parse it.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2780,6 +2789,8 @@ class ActionCollection {
         }
         else {
             var buttonStrip = document.createElement("div");
+            let hostConfig = this._owner.hostConfig;
```
dclaux: Not necessary

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2780,6 +2789,8 @@ class ActionCollection {
...
+            buttonStrip.className = hostConfig.makeCssClassName("ac-actionStrip");
```
dclaux: Should be ac-actionSet

source/nodejs/adaptivecards/src/card-elements.ts
dptejas: Done

source/nodejs/adaptivecards/src/card-elements.ts
dptejas: Changing it

source/nodejs/adaptivecards/src/card-elements.ts
dptejas: This is outdated code. It is not from latest commit

source/nodejs/adaptivecards/src/card-elements.ts
dptejas: This is outdated code. It is not from latest commit

source/nodejs/adaptivecards/src/card-elements.ts
dptejas: This is outdated code. It is not from latest commit

source/nodejs/adaptivecards/src/card-elements.ts
dptejas: This is outdated code. It is not from latest commit

Issue #1522, opened 2018-05-30
[HTML] Add support to modify lineHeight


dptejas:
# **Problem**
Currently, line height of text is blindly calculated by multiplying font size with 1.33. For better look and feel of the card, there should be option for client to specify line Heights for each of fontSizes

# **Solution**
In the HostConfig:

Introduce a section for lineHeights which specifies line Height for each of the fontSize

# **Design**
       Modify HostConfig to accommodate below property

```typescript
            'lineHeights': {
                'small':<line height for small text size in px> ,
                'default':<line height for small text size in px> ,
                'medium': <line height for small text size in px>,
                'large': <line height for small text size in px>,
                'extraLarge': <line height for small text size in px> 
            }
```

     Rendering of TextBlock can be done in following way:

```typescript
            switch (this.size) {
                case Enums.TextSize.Small:
                    lineHeight = this.hostConfig.lineHeights.small;
                    break;
                case Enums.TextSize.Medium:
                    lineHeight = this.hostConfig.lineHeights.medium;
                    break;
                case Enums.TextSize.Large:
                    lineHeight = this.hostConfig.lineHeights.large;
                    break;
                case Enums.TextSize.ExtraLarge:
                    lineHeight = this.hostConfig.lineHeights.extraLarge;
                    break;
                default:
                    lineHeight = this.hostConfig.lineHeights.default;
                    break;
            }
```

dptejas:
Implemented solution in #1524

PR #1524, opened 2018-05-30
Support to modify lineHeights

dptejas:
Addresses issue #1522


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -644,14 +644,14 @@ export abstract class CardElement {
 }
 
 export class TextBlock extends CardElement {
-    private _computedLineHeight: number;
+    private lineHeight: number;
```
dclaux: All private fields should start with _

Why change the name anyway? Can we please keep _computedLineHeight?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -722,12 +722,26 @@ export class TextBlock extends CardElement {
                     break;
             }
 
-            // Looks like 1.33 is the magic number to compute line-height
-            // from font size.
-            this._computedLineHeight = fontSize * 1.33;
```
dclaux: No, we can't remove that. lineHeights in HostConfig isn't a documented feature, things need to continue working exactly as before in the absence of the lineHeights definition:
- lineHeights in HostConfig should be optional
- When specified, lineHeights must be fully defined, e.g. Small, Medium, etc. are all required
- When lineHeights isn't defined in HostConfig, line height should be computed exactly as before to maintain 100% backward compatibility

source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -322,6 +322,15 @@ export class HostConfig {
         large: 21,
         extraLarge: 26
     };
+
+    readonly lineHeights = {
```
dclaux: No, we can't do that. lineHeights needs to be optional to maintain backwards compat. Please add a ILineHeightDefinitions interface, then declare the lineHeights property in HostConfig as follows:

```
lineHeights?: ILineHeightDefinitions,
```

source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -358,6 +367,14 @@ export class HostConfig {
                 extraLarge: obj.fontSizes && obj.fontSizes["extraLarge"] || this.fontSizes.extraLarge
             };
 
+            this.lineHeights = {
```
dclaux: How about adding an `if (obj.lineHeights)` statement around this instead of repeating it 5 times? I know the same pattern is used in other places, but it's never too late to start :)

source/nodejs/adaptivecards/src/card-elements.ts
dptejas: have taken this aspects in latest commit

source/nodejs/adaptivecards/src/host-config.ts
dptejas: done

source/nodejs/adaptivecards/src/host-config.ts
dptejas: done

PR #1525, opened 2018-05-30
[Android] Add vertical stretch property

almedina-ms:
Add the height property that allows items to stretch to fill empty spaces. Also adds new tests for every item that may be part of an adaptive card

Completes this issue: https://github.com/microsoft/adaptivecards/issues/1297


samples/Tests/ColumnsDifferentWidths.json
```diff
@@ -0,0 +1,54 @@
+{
```
RebeccaAnne: Can you re-run the UWP test app on these and checkin the baseline files? Thanks!

Issue #1528, opened 2018-05-31
Hint the browser to suggest emails (input type=email)


ignacionr:
Render HTML <input type="email"> for Input.Text of style "email," so that most browsers will help the user autocomplete.
At the moment, only Safari on the MacBook Pro with Touch Bar will hint email addresses.

PR #1538, opened 2018-05-31
added InputTextStyle enum, connected to email input type

ignacionr:
TS, not my strong suit. Can you guys please check?
This takes care of https://github.com/Microsoft/AdaptiveCards/issues/1528

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -1593,7 +1593,7 @@ export class TextInput extends Input {
         }
         else {
             this._inputElement = document.createElement("input");
-            this._inputElement.type = (this.style == Enums.InputTextStyle.Email) ? "email" : "text";
+            this._inputElement.type = Enums.InputTextStyle[this.style].toLowerCase();
             this._inputElement.className = this.hostConfig.makeCssClassName("ac-input", "ac-textInput");
             this._inputElement.style.width = "100%";
             this._inputElement.tabIndex = 0;
@@ -1643,7 +1643,7 @@ export class TextInput extends Input {
         this.maxLength = json["maxLength"];
         this.isMultiline = json["isMultiline"];
         this.placeholder = json["placeholder"];
-        this.style = json["style"];
+        this.style =  Utils.getEnumValueOrDefault(Enums.InputTextStyle, json["style"], this.style);
     }
 
     get value(): string {
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -1593,7 +1593,7 @@ export class TextInput extends Input {
         }
         else {
             this._inputElement = document.createElement("input");
-            this._inputElement.type = "text";
+            this._inputElement.type = (this.style == Enums.InputTextStyle.Email) ? "email" : "text";
             this._inputElement.className = this.hostConfig.makeCssClassName("ac-input", "ac-textInput");
             this._inputElement.style.width = "100%";
             this._inputElement.tabIndex = 0;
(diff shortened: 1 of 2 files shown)


matthidinger:
Thanks @ignacionr. Seems this was an oversight in the TS renderer, not just email, but the other input types as well. 

This PR should be updated to include `url` and `tel` - which are the [supported Input.Text styles](http://adaptivecards.io/explorer/Input.Text.html)

@dclaux can you confirm this? 

https://www.w3schools.com/tags/tag_input.asp

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -1593,7 +1593,7 @@ export class TextInput extends Input {
         }
         else {
             this._inputElement = document.createElement("input");
-            this._inputElement.type = "text";
+            this._inputElement.type = (this.style == Enums.InputTextStyle.Email) ? "email" : "text";
```
dclaux: Why not just set type to this.style? Without that most valid styles will simply be ignored.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -1620,6 +1620,7 @@ export class TextInput extends Input {
     maxLength: number;
     isMultiline: boolean;
     placeholder: string;
+    style: Enums.InputTextStyle = Enums.InputTextStyle.Text;
```
dclaux: This won't work, Enums.InputTextStyle is a class, and the Text field is a string, can't assign string to class.

InputTextStyle needs to be defined as an enumeration like pretty much everything else in enums.ts

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -1641,6 +1643,7 @@ export class TextInput extends Input {
         this.maxLength = json["maxLength"];
         this.isMultiline = json["isMultiline"];
         this.placeholder = json["placeholder"];
+        this.style = json["style"];
```
dclaux: This should be
```
this.style = Utils.getEnumValueOrDefault(Enums.InputTextStyle, json["style"], this.style);
```

source/nodejs/adaptivecards/src/enums.ts
```diff
@@ -109,6 +109,13 @@ export class ContainerStyle {
     static readonly Emphasis = "emphasis";
 }
 
+export class InputTextStyle {
```
dclaux: Needs to be defined as an enum. ContainerStyle is defined as a class because the style property of Container is a string that can have values that are not defined in ContainerStyle.

dclaux:
Thanks for your help on this. Please review my comments.

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: Will do that; I was being selfish only thinking about the email. I will make sure that parse(json) gets a correct default if needed.

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: Yep. Exactly. Will do.

source/nodejs/adaptivecards/src/enums.ts
ignacionr: Got it. On it.

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: Funny how Github can't track the fact that the class was changed on a different file and that fulfills your request; would you need to remove the feedback @dclaux ?

ignacionr:
Please notice I'm using direct reverse mapping and then toLower() for the point where the element gets created and assigned a type. I believe this to be safe as defined (parse will reset to default if the json value is not predefined so seems safe).

dclaux:
That's fine at least for now given that the enum values map exactly to supported types.

dclaux:
Thanks for your contribution!

ignacionr:
Thanks for reviewing and accepting. As I said on Twitter, I don't know where you find these 28hr days.

PR #1543, opened 2018-05-31
Fix up namespacing and switch to C++17 language spec

paulcam206:
Rather than conflate introducing `.clang-format` with namespacing changes, I'm going to do this as two separate PRs.

In this PR we get rid of the `AdaptiveSharedNamespaceStart` and `AdaptiveSharedNamespaceEnd` macros and move the VS project over to using C++17.


source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModel/AdaptiveCardsSharedModel.vcxproj
```diff
@@ -90,6 +90,7 @@
       <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
       <TreatWarningAsError>true</TreatWarningAsError>
       <DisableSpecificWarnings>4702</DisableSpecificWarnings>
+      <LanguageStandard>stdcpp17</LanguageStandard>
     </ClCompile>
```
khouzam: If this is necessary, shouldn't it also be included in the uwp version of the vcxproj?

paulcam206: It's not strictly required, but since we're going to need to build C++17 on our end, we need to make sure this build remains happy. That being said, we should definitely switch UWP over. I'll push a change shortly.

PR #1546, opened 2018-06-01
Add Events for media play and end

RebeccaAnne:


source/uwp/Renderer/lib/AdaptiveMediaConfig.cpp
```diff
@@ -44,4 +46,18 @@ AdaptiveNamespaceStart
         return m_playButton.Set(playButton);
     }
 
+    _Use_decl_annotations_
```
paulcam206: It's recommended by current standards that we explicitly annotate both declaration and definition (i.e. don't use `_Use_decl_annotations_`).

paulcam206: If you want, you can ignore this and we can clean all uses up at a later time. :)

RebeccaAnne: Logged issue #1553 to track.

---
In reply to: [192874207](https://github.com/Microsoft/AdaptiveCards/pull/1546#discussion_r192874207) [](ancestors = 192874207)

Issue #1550, opened 2018-06-03
Availability of CppCoreChecks should depend on build system, not platform


ignacionr:
E.g. I'm not building on Android, Apple, or Windows, and yet I don't have access to the Visual Studio-specific CppCoreCheck headers.

For now I will build defining ```__APPLE__``` but if you get into a third round of this, @paulcam206 I know you will come up with a proper way to isolate CppCoreChecks. For fellow C++ users?

https://github.com/Microsoft/AdaptiveCards/blob/17eeb8adb25b71482f00ca080b9ca3cc253b26a1/source/shared/cpp/ObjectModel/pch.h#L26

paulcam206:
I think I should be able to add compiler detection to this :)

paulcam206:
See #1552

paulcam206:
Should be fixed with #1552. Please confirm and close :)

ignacionr:
That's some show of love. Thank you all!

PR #1552, opened 2018-06-04
Add MSC directive to prevent cppcorecheck build failures on non-VS builds

paulcam206:
Simple fix for #1550 -- Only include cppcorecheck headers when compiling with MS toolchain.


ignacionr:
Thank you!

PR #1556, opened 2018-06-05
Tupha/support card level select action dotnet

tuanh118:
- Added support for card-level `selectAction` in WPF and HTML
- Fixed `Action.ShowCard` in HTML. Previously, the additional cards were not added to the card
- Cleaned up UI (spacing, hover color) for Button elements created for `selectAction`

### DIFF
--- a/source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionRenderer.cs
+++ b/source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionRenderer.cs
@@ -24,9 +24,7 @@ namespace AdaptiveCards.Rendering.Wpf
         {
             var uiButton = new Button
             {
-                //HorizontalAlignment = HorizontalAlignment.Stretch,
                 Style = context.GetStyle($"Adaptive.{action.Type}"),
-                Padding = new Thickness(6, 4, 6, 4),
                 Content = new TextBlock
                 {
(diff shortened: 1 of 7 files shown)


source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionRenderer.cs
```diff
@@ -24,9 +24,7 @@ public static Button CreateActionButton(AdaptiveAction action, AdaptiveRenderCon
         {
             var uiButton = new Button
             {
-                //HorizontalAlignment = HorizontalAlignment.Stretch,
                 Style = context.GetStyle($"Adaptive.{action.Type}"),
```
khouzam: This has nothing to do with this change. It should go in a separate change.

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveContainerRenderer.cs
```diff
@@ -20,7 +20,6 @@ public static FrameworkElement Render(AdaptiveContainer container, AdaptiveRende
             }
 
             Grid uiOuterContainer = new Grid();
-            uiOuterContainer.Background = context.GetColorBrush(containerStyle.BackgroundColor);
```
khouzam: This should be a separate change

PR #1561, opened 2018-06-05
[.NET] Prevent async exceptions from tearing down the process

khouzam:
This helps catch exceptions in async calls that might tear down the process.

### DIFF
--- a/source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveCardRenderer.cs
+++ b/source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveCardRenderer.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveCardRenderer.cs
```diff
@@ -165,20 +166,29 @@ void Callback(object sender, AdaptiveActionEventArgs args)
...
+                    Resources = Resources,
+                    ElementRenderers = ElementRenderers
+                };
+
+                var stream = context.Render(card).RenderToImage(width);
+                renderCard = new RenderedAdaptiveCardImage(stream, card, context.Warnings);
+            }
+            catch(Exception e)
+            {
+                Debug.WriteLine($"RENDER Failed. {e.Message}");
```
matthidinger: Only thing that seems weird is swallowing the exception here. Should we throw it? Or do we need to use a TaskCompletionSource and .SetException() ?

vincec-msft: Throwing the exception is what causes the unhandled exception to take down the process. We definitely don't want to do that.

PR #1567, opened 2018-06-06
Make `ParseUtil` a namespace

paulcam206:
A class consisting of public static methods, private ctor/dtor, and no state should just be a namespace.


paulcam206:
👍

Issue #1569, opened 2018-06-07
[HTML] Custom Parsing of Action Objects


dptejas:
# Problem

There is no way to custom parse actions similar to how it is possible to custom parse elements.

# Design
Have a global and per-card action parsing eventHandler(like other events)

```typescript
export class AdaptiveCard {
    ...
    static onParseAction: (action: Action, json: any) => void = null;
    onParseAction: (action: Action, json: any) => void = null;
    ...
}
```

ignacionr:
In our -external, 3rd party- work with creating hosts, "Action.Submit" fits most cases.

- The mechanics of submitting are defined by the host.
- There's a "data" property that can hold any configuration relative to how to execute the submittion. [e.g any equivalent to endpoint indication, conversation tracking, cookie equivalents, client-held state, could go here]

Curious to know what case couldn't be accommodated. Can you provide an example of such situation?

dptejas:
In outlook actionable message, developer can mark an action as "primary" so that it can get a different visual treatment. Since "primary" is not a open schema for adaptiveCards , in outlook we need support to parse actions custom similar to custom parse elements. 
Let me know if you need more information on the same

ignacionr:
@dptejas I'm so sorry you guys probably have internal conversations. If you have some time to illustrate how this "primary" property is different to the "isPrimary" property that you already proposed in https://github.com/Microsoft/AdaptiveCards/issues/1515 ?

dclaux:
@ignacionr `Action.Submit`, at least as loosely defined as it is defined right now, actually doesn't suit the needs of some clients. Bots have an implicit underlying communication channel over which all messages between the client and the "server" (the Bot) can be sent and received. Outlook, for one, doesn't have that. Outlook does have its implicit communication channel with Exchange, but Exchange isn't the third party service that needs to be reached when an action is executed. Outlook wanted/needed another type of more explicit action model, and that model the `Action.Http` action type. More info here: https://docs.microsoft.com/en-us/outlook/actionable-messages/adaptive-card#actionhttp

While it is true that technically each host can document that the `data` property of an `Action.Submit` is to be formatted this way or that way, it remains a not-well-defined property bag and Outlook favored a more strongly-typed approach.

All that said, the action model is unrelated to this change which is only about visuals.

ignacionr:
This info you’re sharing is really important. Helps me get an idea of the driving forces. Thanks for taking the time to do so.

[...]

PR #1570, opened 2018-06-07
Support OnParseAction Handler

dptejas:
Addresses issue in #1569


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2321,6 +2321,7 @@ export abstract class Action {
     };
 
     parse(json: any) {
+        raiseParseActionEvent(this, json);
         this.id = json["id"];
```
dclaux: Nit: insert blank line after raiseParseActionEvent.

Issue #1361, opened 2018-04-09
[TS] Media player support


khouzam:

dclaux:
@ignacionr would you want to help (again)? This feature is somewhat expensive because to do it properly the playback controls have to be owned by the Adaptive renderer, the standard HTML5 player controls just won't do because:

1. They can't be styled, and will end up looking bad in many cards
2. They are rendered differently in every browser

You might argue that's ok, but on top of that the standard controls for the <audio> tag don't even resize, making it a requirement for Adaptive to render its own controls.

So far I haven't had time to implement this, so if you want to give it a shot, you are more than welcome.

ignacionr:
@dclaux I had read your previous comment on those required limitations, and understand the reason.
Super happy to help; let's check again by Friday 8th see where I've gotten. Sounds like a plan? (By the way I thought you'd reply after the Monday meeting -didn't mean to have you working on a Sunday.)

dclaux:
@ignacionr sounds good.

ignacionr:
@dclaux I'm going with controls rendered as anchors, _under_ the video for starters (will share a screenshot soon); can you explain what is the intention when there are multiple sources (is it intended as alternatives or as a sequence?)

dclaux:
@ignacionr whatever works, feel free. Let me just give you a dump of what matters most:
- Don't use third party controls (e.g. no dependency on another library)
- Don't have dependencies on external images for buttons
- Don't have a dependency on an external color font
- Make sure that the control bar in its entirety reflows well when space is limited. I don't think we want controls to wrap, for instance

- Make sure the whole control bar can be styled via the Adaptive CSS file


As for how things are supposed to work with sources:

- Go through the list
- Valid sources will have their `mimeType` starting with either **audio/** or **video/**. All other MIME types are ignored
- The first valid MIME type determines the type of media the player will play (audio or video)
- At that point, drop all sources that are not of that type, and feed the list of remaining sources to your `<audio>` or `<video>` HTML element
- If you don't find any valid sources, render the poster instead. Poster should likely be rendered like a stretched image would be. If there is no posted on the Media element, rendere the default poster from HostConfig. If there is no default poster, feel free to come up with a creative fallback strategy :)
- When you do render the poster, overlay a "play" button in its center; when clicked, this button would open the first of the original sources (regardless of what it is) in a separate browser window (or tab)

andrewleader:
Looks like we forgot to close this, closing

PR #1577, opened 2018-06-08
Media now supports parsing; first round HTML

ignacionr:
This is to share the progress (Media is parsed, tests are in place, and a shy initial HTML rendering with anchors with Unicode media control buttons -just text- as described in https://en.wikipedia.org/wiki/Media_controls ); again TS not my best suit but trying to keep consistent with what I see done. For tests I'm practically copying those used by @RebeccaAnne on the WUP side.

As for the whitespace changes, I had to work hard for Visual Studio Code to keep them to the minimum. The ones it kept seem reasonable to me, but let me know I can roll those back.

All feedback is welcome!


ignacionr:
https://github.com/Microsoft/AdaptiveCards/issues/1361

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+
+export class Media extends CardElement {
+    poster: string;
+    altText: string;
+    sources: Array<Source> = [];
+    private _mediaType: string;
+    private _mediaElement: HTMLMediaElement;
+    static readonly acceptedMediaTypes = ["audio", "video"];
+
+    private processSources(): Array<Source> {
```
dclaux: This might be my personal taste, but I find this a bit more complex and hard to read than probably necessary (especially the type filter that modifies itself). How about something like this?

```
let goodSources: Array<MediaSource> = [];

for (let source of this.sources) {
	let mimeComponents = source.mimeType.split('/');

	if (mimeComponents.length == 2) {
		if (!this._mediaType) {
			let index = Media.acceptedMediaTypes.indexof(mimeComponents[0]);

			if (index >= 0) {
				this._mediaType = Media.acceptedMediaTypes[index];
			}
		}

		if (mimeComponent[0] == this._mediaType) {
			goodSourdces.push(source);
		}
	} 
}

return goodSources;
```

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+        let typeFilter = (t): boolean => {
+            if (Media.acceptedMediaTypes.indexOf(t)) {
+                this._mediaType = t;
+                typeFilter = (tt) => tt === this._mediaType;
+                return true;
```
dclaux: I usually put a blank line above a return statement if it is not the only (first) one in the block.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+            }
```
dclaux: Empty line after closed curly bracket.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+            return false;
+        }
+        return this.sources.filter((s): boolean => {
```
dclaux: Empty line after closed curly bracket.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+            if (s.mimeType) {
+                let mimeComponents = s.mimeType.split('/');
+                if (mimeComponents.length == 2) {
```
dclaux: Empty line above if statement if it isn't the first in the block.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+    poster: string;
+    altText: string;
+    sources: Array<Source> = [];
+    private _mediaType: string;
+    private _mediaElement: HTMLMediaElement;
+    static readonly acceptedMediaTypes = ["audio", "video"];
+
+    private processSources(): Array<Source> {
+        let typeFilter = (t): boolean => {
+            if (Media.acceptedMediaTypes.indexOf(t)) {
```
dclaux: Does this actually work? indexOf will return 0 for the first element in the array, right? Won't that eval to false when it should eval to true?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+            return false;
+        }
+        return this.sources.filter((s): boolean => {
+            if (s.mimeType) {
+                let mimeComponents = s.mimeType.split('/');
+                if (mimeComponents.length == 2) {
+                    return typeFilter(mimeComponents[0]);
+                }
+            }
+            return false;
```
dclaux: Empty line after closed curly bracket

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+        });
+    }
+
+    private makeButton(label: string): HTMLAnchorElement {
+        let button = document.createElement("a");
+        button.className = "ac-media-button";
+        button.text = label;
+        return button;
```
dclaux: Empty line above return

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+    }
+
+    protected internalRender(): HTMLElement {
+        var element = document.createElement("div");
+
```
dclaux: Remove empty line

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+        element.className = this.hostConfig.makeCssClassName("ac-media");
+
+        let goodSources = this.processSources();
+
+        if (goodSources.length) {
+
```
dclaux: Remove empty line

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+            this._mediaElement = <HTMLMediaElement>document.createElement(this._mediaType);
+            this._mediaElement.style.width = "100%";
+
+            goodSources.forEach(s => {
+                let src: HTMLSourceElement = document.createElement("source");
+                src.src = s.url;
+                src.type = s.mimeType;
+                this._mediaElement.appendChild(src);
+            });
+            element.appendChild(this._mediaElement);
```
dclaux: I would put an empty line above, but that's just me.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+    pause() {
+        this._mediaElement.pause();
+    }
+
+    toJSON() {
+        let result = super.toJSON();
+
+        Utils.setProperty(result, "poster", this.poster);
+        Utils.setProperty(result, "altText", this.altText);
+        if (this.sources.length) {
```
dclaux: Empty line above if statement

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+        return "Media";
+    }
+
+    renderSpeech(): string {
+        if (this.speak != null)
+            return this.speak + '\n';
+
+        return null;
+    }
+
```
dclaux: Remove empty line

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+        }
+    }
+
+    getJsonTypeName(): string {
+        return "Media";
+    }
+
+    renderSpeech(): string {
+        if (this.speak != null)
+            return this.speak + '\n';
```
dclaux: Please always use open and close curly brackets, even if there is only one statement.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+            var stop_button = this.makeButton("⏹️");
+            stop_button.onclick = e => this.stop();
+            element.appendChild(stop_button);
+        }
+
+        return element;
+    }
+
+    play() {
+        this._mediaElement.play();
```
dclaux: We probably want to hide the play button and show the pause button instead

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+    }
+
+    stop() {
+        this._mediaElement.load();
```
dclaux: I don't think we need "stop", play and pause should be enough.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+        this._mediaElement.pause();
+    }
+
+    pause() {
+        this._mediaElement.pause();
```
dclaux: We probably want to hide the pause button and show the play button instead

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+
+            var play_button = this.makeButton("▶️");
+            play_button.onclick = e => this.play();
+            element.appendChild(play_button);
+
+            var pause_button = this.makeButton("⏸️");
+            pause_button.onclick = e => this.pause();
+            element.appendChild(pause_button);
+
+            var stop_button = this.makeButton("⏹️");
```
dclaux: I don't think we need a stop button. Maybe a "go back to beginning" button instead.

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: This is the kind of feedback I was after at this time, thanks for it!

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
...
+        return this;
+    }
+
+    toJSON() {
+        return { mimeType: this.mimeType, url: this.url };
+    }
+}
+
+export class Media extends CardElement {
+    poster: string;
```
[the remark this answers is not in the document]
ignacionr: This helps a lot, since TS is not my strongest suit. Thanks for sharing.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4351,6 +4351,148 @@ export class ColumnSet extends CardElementContainer {
     }
 }
 
+export class Source {
```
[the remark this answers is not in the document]
ignacionr: Much better, will change.

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: Sure can change it.

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: Not a fan of vertical spaces, but this seems to bother many people (some good friends even mock me for this). I will follow the guidelines you're expressing here.

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: I won't fall for the temptation of saying "I copied from elsewhere on the same file" because it's not a justification.

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: You are right, index 0 is falsey, so the code is wrong.
My idea was to first create some tests for the rules on audio/video and accepted types, then test it.

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: The simpler the better. Agreed!

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Feel free to fix that elsewhere as well ;)

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: This was implemented (lines that follow).

ignacionr:
@dclaux would like you to comment, mainly on the poster topics:
- Since you said "render just like an Image does" I decided to actually create one and use as-is. I think it can be a good idea, since it'd leverage any flexibility we may want to add in the future.
- Expanding on that approach, I also made it so the opening of the first original media on poster-mode, is actually an OpenUrlAction object.

Those are not yet working, but it'd help a lot if you confirm that's an approach you'd ad-imprimatur.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4460,42 +4464,51 @@ export class Source {
...
+		}
+
+		return goodSources;
     }
 
-    private makeButton(label: string): HTMLAnchorElement {
+    private static makeButton(label: string, buttonHint: string, handler: ()=>void): HTMLAnchorElement {
         let button = document.createElement("a");
-        button.className = "ac-media-button";
+        button.classList.add("ac-media-button", buttonHint);
```
dclaux: You will need to build the class name by calling HostConfig.makeCssClassName, which means you will need a reference to the button owner's HostConfig. This is necessary so that multiple cards rendered on the same screen can be styled differently using separate stylesheets.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4460,42 +4464,51 @@ export class Source {
...
+					goodSources.push(source);
+				}
+			} 
+		}
+
+		return goodSources;
     }
 
-    private makeButton(label: string): HTMLAnchorElement {
+    private static makeButton(label: string, buttonHint: string, handler: ()=>void): HTMLAnchorElement {
```
dclaux: Nit: "()=>void" should be "() => void"

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4460,42 +4464,51 @@ export class Source {
...
         let button = document.createElement("a");
-        button.className = "ac-media-button";
+        button.classList.add("ac-media-button", buttonHint);
+		button.title = buttonHint;
+		button.onclick = e => { handler(); e.preventDefault(); return false; };
```
dclaux: A couple things:

- Can you spread the statements over three lines below. In general that's better because it makes it possible to set a breakpoint on these statements.
- Event handlers are called onXxx in general in this codebase. Yours here would be onClick
- It would basically never happen because of the way you use makeButton, but ideall you would still check for null before invoking the handler.


Also, ActionButton do not call preventDefault in their handler, and that's a problem. Would you take this opportunity and fix that with this feature? Totally ok if you'd rather not.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4460,42 +4464,51 @@ export class Source {
...
         button.text = label;
+		
         return button;
     }
 
     protected internalRender(): HTMLElement {
-        var element = document.createElement("div");
+        let element = <HTMLElement>document.createElement("div");
```
dclaux: Why cast? It's no necessary.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4513,34 +4526,57 @@ export class Media extends CardElement {
...
-            var stop_button = this.makeButton("⏹️");
-            stop_button.onclick = e => this.stop();
-            element.appendChild(stop_button);
-        }
+			
+			let controlBar = <HTMLDivElement>document.createElement("div");
+			controlBar.className = this.hostConfig.makeCssClassName("ac-controlbar");
+			controlBar.style.whiteSpace = "nowrap";
+
+            this._playButton = Media.makeButton("▶️", "play", ()=>this.play());
```
dclaux: There seems to be a one character misalignment in the next few lines. VS Code has had that bug of introducing these for a while now.

dclaux: Let's start "play", "pause" and "reset" with capital letters given they will be displayed.

dclaux: "()=>this.play()" should be "() => this.play()"

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4513,34 +4526,57 @@ export class Media extends CardElement {
...
+
+            this._resetButton = Media.makeButton("⏮️", "reset", ()=>this.reset());
+            controlBar.appendChild(this._resetButton);
+			
+			element.appendChild(controlBar);
+        }
+		else {
+			let img = new Image();
+			img.url = this.poster;
+			if (this.sources.length > 0) {
```
dclaux: Empty line before if statement

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4513,34 +4526,57 @@ export class Media extends CardElement {
...
+				let openAction = new OpenUrlAction();
+				openAction.url = this.sources[0].url;
+				img.selectAction = openAction;
+			}
+			let renderedImage = img.render();
```
dclaux: Empty line after closing }

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: It only becomes unnecessary if I declare the type of the variable explicitly; if I don't, then the type of ```element``` becomes HTMLDivElement and further down for the poster representation I'm invoking render (which returns an HTMLElement); since I prefer to have one rather than two return statements, I need to make the type of 'element' be the base class.

But I think I may remove the concern if I do ```let element: HTMLElement = document.createElement("div")``` instead.

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: - Whitespaces, yes. Thanks for understanding. I will see to this.
- Will do the Play, Pause, Reset for title and lowercase for the CSS classes.
- Yes, thanks for noticing.

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: (Y)

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: I need to get these vertical whitespaces right. Doesn't come naturally, sorry.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: I don't think you need to do anything. HTMLDivElement extends HTMLElement, so it *is* an HTMLElement and can be handled naturally as such. In other word you can have a method that is supposed to return an HTMLElement return HTMLDivElement.

ignacionr: You got it exactly right, only backwards. If I leave it as 
```let element = document.createElement("div");```
(element remains declared as HTMLDivElement) and further down
```element = something-that-is-HTMLElement-but-possibly-not-HTMLDivElement;```
that other one breaks.

I put it there because tsc was complaining, not because I felt like writing more code today 😃

source/nodejs/adaptivecards/src/card-elements.ts
ignacionr: by the way, how do you feel about having CSS classes like that? I normally like very much the fact that it blends well to selecting and changing stuff on any level of the stylesheet. On the other hand it can look too verbose. Would love your opinion.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Sorry, I don't get it. There is nothing in your code that requires the cast. It would work just the same without the cast. What is the "complaint" VS Code is giving you?

ignacionr: You're completely right. I had the previous round in the back of my head. Sorry for that. Owe you a beer next time I go to Redmond.

ignacionr:
Sure will prepare a couple of outputs for reference.

ignacionr:
Will need to redo this -problems merging. Will reopen in two days. Thanks and sorry for the wait.

PR #1579, opened 2018-06-08
Enable CPPCORECHECK_DECLARATION_WARNINGS and fix errors

paulcam206:
There are effectively two class of error that we have in our codebase that was caught by `CPPCORECHECK_DECLARATION_WARNINGS`:

* [I.22: Avoid complex initialization of global objects](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#i22-avoid-complex-initialization-of-global-objects)
This check is a good one -- it ensures that global variable initialization is only done with `constexpr`s - meaning that there's no concern about the order in which global variables init. Unfortunately, the checker has a bug where it sees function-static variables as global, so basically all of `Enums.cpp` shows up as hits. I reviewed static variable use in this file and then disabled the check in that file's scope.
* [ES.84: Don't (try to) declare a local variable with no name](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es84-dont-try-to-declare-a-local-variable-with-no-name)
Basically, some functions return an object and some operations create temporary objects. These objects may have complex ctors or dtors which run in a way and at a time that's not obvious to the reader. There are 3 classes of error found in our codebase that needed to be addressed:
   - `std::unordered_set::insert()` has a number of overloads. When tracking known properties, we were adding each element individually, which is calling an overload that returns a `std::pair<>`. Since we don't actually care about anything in the `pair`, we can just call the overload that takes an initializer list, which has a `void` return type. And since it's a list we're passing in, we might as well just make one call to `insert()` with all of the properties.
   - When using an iterator in a `for` loop, it's best practice to increment using the prefix operator rather than the postfix, since postfix [causes a temporary object to be constructed](https://en.cppreference.com/w/cpp/language/operator_incdec) (see `Notes` section).
   - There are a few places where we were calling either `std::transform()` or an `insert()` function where it didn't make sense to change to a different form. For these, I assigned the return to a new local variable, which squelches the warning and also alerts the reader to the fact that there's an object being created (even if it's not going to be used).

### DIFF
--- a/source/shared/cpp/ObjectModel/BaseActionElement.cpp
+++ b/source/shared/cpp/ObjectModel/BaseActionElement.cpp
@@ -93,10 +93,10 @@ void BaseActionElement::SetAdditionalProperties(Json::Value const &value)
 
 void BaseActionElement::PopulateKnownPropertiesSet()
 {
-    m_knownProperties.insert(AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Type));
-    m_knownProperties.insert(AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Title));
-    m_knownProperties.insert(AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Id));
-    m_knownProperties.insert(AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::IconUrl));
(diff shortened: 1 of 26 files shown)


paulcam206:
Oops -- still updating description :)

PR #1581, opened 2018-06-08
added ability to scroll when there are more buttons than viewing area

nesalang:
enabled the containing view to be scrollable when there are more buttons than viewable area.


Assaya:
https://www.evernote.com/shard/s486/sh/d1733d06-f560-43bb-936f-5ac7d539afa7/346e50efc0900bdb20cea136ffa68e1b


angela@thehelm

[cid:1b22a6f5-a007-4bc8-b52c-2b645e9af8aa]

PR #1603, opened 2018-06-13
[iOS] Add missing test for Vertical Stretch

almedina-ms:
Adds a missing file that was included in the solution but wasn't added to the commit. This missing file may make the build process to fail

### DIFF
--- a/samples/Tests/InputsFormWithHeightStretch.json
+++ b/samples/Tests/InputsFormWithHeightStretch.json
@@ -1,0 +1,86 @@
+{
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"type": "AdaptiveCard",
+	"version": "1.0",
+	"body": [
+		{
+			"type": "ColumnSet",
+			"columns": [
+				{
+					"type": "Column",
+					"width": 2,
+					"items": [
+						{
+							"type": "TextBlock",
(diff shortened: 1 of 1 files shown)


nesalang:
adding this won't affect UWP's test app outputs, right?

samples/Tests/InputsFormWithHeightStretch.json
```diff
@@ -0,0 +1,86 @@
...
+				},
+				{
+					"type": "Column",
+					"width": 1,
+					"items": [
+						{
+							"type": "Image",
+							"url": "https://upload.wikimedia.org/wikipedia/commons/b/b2/Diver_Silhouette%2C_Great_Barrier_Reef.jpg",
+							"size": "auto",
+                            "height": "stretch"
```
nesalang: tabs?

paulcam206:
Is this what's causing the build break on #1611?

nesalang:
i think so

---
In reply to: [397431440](https://github.com/Microsoft/AdaptiveCards/pull/1603#issuecomment-397431440) [](ancestors = 397431440)

almedina-ms:
@paulcam206 #1611 is unrelated, the issue was that the PR that I passed yesterday included a file in the solution that I forgot to add to the commit, so now if you try to build the iOS solution then you get the file missing error

PR #1612, opened 2018-06-14
[UWP][Android][iOS] Add aesthetic fix Icons in Actions

almedina-ms:
Add the fix in UWP, android and iOS for the case where the icon placement is above the title in actions but not all actions define an icon so the behavior turns into changing the placement to the left of the title for the buttons that do define an icon


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
```diff
@@ -253,10 +254,27 @@ else if (alignment == ActionAlignment.Center.swigValue())
 
         int i = 0;
         long maxActions = hostConfig.getActions().getMaxActions();
-        for (; i < size && i < maxActions; i++)
+
+        boolean allActionsHaveIcons = true;
+        for(; i < size && i < maxActions; ++i)
         {
             BaseActionElement actionElement = baseActionElementList.get(i);
+            if(actionElement.GetIconUrl().isEmpty()){
```
khouzam: >{ [](start = 52, length = 1)

Nit: bracket

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
```diff
@@ -253,10 +254,27 @@ else if (alignment == ActionAlignment.Center.swigValue())
...
+                allActionsHaveIcons = false;
```
khouzam: break. You don't need to keep going

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.mm
```diff
@@ -258,6 +258,27 @@ - (void)addTasksToConcurrentQueue:(std::vector<std::shared_ptr<BaseCardElement>>
...
+        if( action->GetIconUrl().empty() ){
+            allActionsHaveIcons = NO;
+        }
+    }
+    
+    for(auto &action : actions)
+    {
+        std::string iconUrl = action->GetIconUrl();
+        if(!iconUrl.empty())
+        {
```
khouzam: Nit: format

PR #1613, opened 2018-06-14
Adds element tree view

dclaux:
Allows selection of elements using a tree view. Useful to select containers for instance.


dclaux:
image

PR #1628, opened 2018-06-20
fixes extra spacing issue in UICollectionView

nesalang:
Fixes extra spacing issue in UICollectionView by more accurately calculating its intrinsic content size.

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRImageSetUICollectionView.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRImageSetUICollectionView.mm
@@ -97,6 +97,10 @@ using namespace AdaptiveCards;
     size_t cellCounts = _imgSet->GetImages().size();
     int dimension = (ceil(sqrt(cellCounts)));
     CGSize imageSize = ((UICollectionViewFlowLayout *)self.collectionViewLayout).itemSize;
-    return CGSizeMake(dimension * imageSize.width, dimension * imageSize.height);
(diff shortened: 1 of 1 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRImageSetUICollectionView.mm
```diff
@@ -95,8 +95,25 @@ - (CGSize)intrinsicContentSize
...
+
+    // sanity check
+    if(!imageSize.width || !self.frame.size.width || !cellCounts){
+        return CGSizeMake(0, 0);
+    }
+    int frameWidth = self.frame.size.width;
+    int imageWidthWithSpacing = imageSize.width + spacing;
+
+    // if there is spacing to the right edge, it's o.k.
+    int numbersOfItemsInRow = frameWidth / imageWidthWithSpacing;
```
bviglietta: >int numbersOfItemsInRow = frameWidth / imageWidthWithSpacing; [](start = 4, length = 61)

I don't think this math is quite right, because of the truncation from float to int. Imagine you have a frame width of 100, image width of 10, and spacing of 0.5. You should be able to fit nine images into the frame, but this math will say you can fit 10.

nesalang: thanks, that's great find. really appreciate it.

---
In reply to: [197204662](https://github.com/Microsoft/AdaptiveCards/pull/1628#discussion_r197204662) [](ancestors = 197204662)

PR #1631, opened 2018-06-21
changed toggle view to UISwitch

nesalang:
toggle view is updated with UISwitch


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRToggleInputDataSource.mm
```diff
@@ -7,35 +7,41 @@
...
+    NSString *_title;
 }
 
 - (instancetype)initWithInputToggle:(std::shared_ptr<ToggleInput> const&)toggleInput
       WithHostConfig:(std::shared_ptr<HostConfig> const&)hostConfig
 {
     self = [super init];
+    _title = [NSString stringWithCString:toggleInput->GetTitle().c_str()
+                                encoding:NSUTF8StringEncoding];
+    _toggleSwitch = [[UISwitch alloc] init];
```
bviglietta: These assignments should go inside the `if (self)` block. If self is nil (unlikely though that may be) we don't want to segfault.

Issue #1633, opened 2018-06-21
[TS] Input.Text - Two Parsing Roundtrip Issues


ignacionr:
Found two issues with round trips.

- ```style``` becomes a number on toJSON(). This can be truly bad and it's my fault. Sorry.
- ```horizontalAlignment``` is being forced when undefined.

<img width="541" alt="screen shot 2018-06-21 at 2 14 00 pm" src="https://user-images.githubusercontent.com/1210985/41703461-92984f78-755d-11e8-9337-732ce760ae17.png">

Let me take care of both.

ignacionr:
@dclaux if you can spare a few secs for this one; I don't think there's anybody else with some minutes to review TypeScript? For reasons too long to explain, this is blocking me for finishing the media rendering.

PR #1636, opened 2018-06-21
[TS] Fixed Input.Text style and horizontalAlignment roundtrip; added test

ignacionr:
As per https://github.com/Microsoft/AdaptiveCards/issues/1633 there were a couple of issues related to parse/toJSON roundtrips.

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -347,9 +347,11 @@ export abstract class CardElement {
 
         Utils.setProperty(result, "type", this.getJsonTypeName());
         Utils.setProperty(result, "id", this.id);
+
         if (this.horizontalAlignment !== null) {
             Utils.setEnumProperty(Enums.HorizontalAlignment, result, "horizontalAlignment", this.horizontalAlignment);
         }
+
         Utils.setEnumProperty(Enums.Spacing, result, "spacing", this.spacing, Enums.Spacing.Default);
         Utils.setProperty(result, "separator", this.separator, false);
         Utils.setProperty(result, "height", this.height, "auto");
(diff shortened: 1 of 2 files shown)


ignacionr:
@dclaux if you can spare a few secs for this one; I don't think there's anybody else with some minutes to review TypeScript? For reasons too long to explain, this is blocking me for finishing the media rendering. Merci d'avance!

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -347,7 +347,9 @@ export abstract class CardElement {
 
         Utils.setProperty(result, "type", this.getJsonTypeName());
         Utils.setProperty(result, "id", this.id);
-        Utils.setEnumProperty(Enums.HorizontalAlignment, result, "horizontalAlignment", this.horizontalAlignment);
+        if (this.horizontalAlignment !== null) {
```
dclaux: Blank line before if

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -347,7 +347,9 @@ export abstract class CardElement {
...
+            Utils.setEnumProperty(Enums.HorizontalAlignment, result, "horizontalAlignment", this.horizontalAlignment);
+        }
```
dclaux: Blank line after closing }

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: I think this should be `if (this.horizontalAlignment)`

ignacionr: I thought that too. But "Left" is equivalent to 0, which is falsey.

ignacionr: Want me to shift the values and make "Left" be 1? I normally prefer to have an "Undefined"/"Invalid"/"Uninitialized" zero-equivalent on enums.

dclaux: Indeed

Issue #1645, opened 2018-06-24
Inline emotes/ gifs in adaptive card


Tyagit3:
Is there a way to use a gif/emotes in an adaptive card?

Kartik1607:
Adaptive cards support gif. Add the url of gif inside the image element something like 
```JSNO
{
    "type":"Image",
    "url": "https://gifurl/"
}
```

ignacionr:
I believe @Tyagit3 means to mix 🗣 emoji inside 🙌 text. I can confirm that at least Unicode emoji works perfectly well if properly encoded into the JSON. See the image.

<img width="432" alt="screen shot 2018-06-25 at 9 15 55 am" src="https://user-images.githubusercontent.com/1210985/41827236-907b8f0e-7858-11e8-8c3f-79c9f4377198.png">

ignacionr:
@Tyagit3 FYI I created a test that renders a card with a text block that includes Emoji. The test checks that the length of the text rendered is the same as the original string (making sure it didn't chop the chars or misinterpreted the sequence).

You can check it out at https://github.com/Microsoft/AdaptiveCards/pull/1647/commits/6f5404489efa6d2e355eb4a5b1d40c1340711d89 ; if you feel this fulfills your doubt, go ahead and close this issue.

Thank you!

Tyagit3:
@ignacionr This is a simple emoji. I was asking about something similar but to show animated emojis/ animojis.

Kartik1607:
> Or you want to replace text in grammatically significant points of the phrasing, with animations
Can't wait for it.

@Tyagit3 The closest you can get is to use markdown and add gif's in place of emojis. You would have to take care of image sizing though.

andrewleader:
Tracking this feature request for inline animated emojis/GIFs here: #2525

Closing this question thread. Markdown with GIFs is a great workaround for now Kartik (but definitely has image sizing issues as you point out), thanks for that though!

PR #1647, opened 2018-06-25
[TS] added a test to ensure Unicode/Emoji support

ignacionr:
Just a test (passes currently) to ensure that we transfer text as-is to HTML, including Unicode.
Addressing the question posed at https://github.com/Microsoft/AdaptiveCards/issues/1645


ignacionr:
@paulcam206 last effort could you merge? (As an external collab I don't have write privileges, and that's quite alright only I have to bother you.)

paulcam206:
No problem :)

Issue #1539, opened 2018-05-31
[Android]Constructors should be protected instead of private to allow extending of renderers instead of overwriting everything.


rednivrug15:
All the constructor in the renderers currently have private constructors. Making it protected would help us to extend those renderers instead of having to override everything

matthidinger:
Thanks for reporting this. I believe we were tracking this change previously and it may already be in master. 

@khouzam @tuanh118 could you confirm, and whether or not this will be in the next maven push?

PR #1649, opened 2018-06-25
[Android] Change private for protected in all Renderers constructors

almedina-ms:
This PR fixes this: https://github.com/Microsoft/AdaptiveCards/issues/1539


almedina-ms:
@rednivrug15

PR #1653, opened 2018-06-27
Revert "[Android] Refocus when a show card changes visibility"

almedina-ms:
Reverts Microsoft/AdaptiveCards#1651


almedina-ms:
@rednivrug15

Issue #1635, opened 2018-06-21
[Android] ElementVisibilityChanged event


rednivrug15:
Android task tracking overall scenario #1638

Dependencies

- Microsoft Teams

==Original Text==

The show card button when clicked opens a new inline card . The user should be directly scrolled to see the card instead of requiring manual scrolling.

khouzam:
This has been merged.

Issue #1381, opened 2018-04-18
[Android][Bug][Teams] Clicking on Showcard card button takes back the scroll to edit text if the cursor was still in the edit text


rednivrug15:
On any card with show card action  and Input.Text , once the show card button is clicked , the scroll takes you back to the cursor position.

khouzam:
This has been merged.

PR #1654, opened 2018-06-27
[Android] Refocus when a show card changes visibility

almedina-ms:
This PR fixes this: #1635

And maybe this: #1381


almedina-ms:
@rednivrug15

PR #1674, opened 2018-07-02
BitCode Added in Project configuration

an888ha:
While Submitting the Teams App to app store we were getting the following error 
**ERROR ITMS-90668: "Invalid Bundle Executable. The executable file 'TeamSpaceApp.app/Frameworks/AdaptiveCards.framework/AdaptiveCards' contains incomplete bitcode. To compile binaries with complete bitcode, open Xcode and choose Archive in the Product menu."**
Enable BitCode is set to yes in Adaptive Cards , but it is putting marker bit code not the actual one. 
To fix the issue we added one more property in User Defined Setting BITCODE_GENERATION_MODE = bitcode;


nesalang:
Thanks!

PR #1678, opened 2018-07-02
Add a GitHub issue template

anbare:
These are simply GitHub changes. Adds a default template that appears when users create a new issue, so that we can help gather more actionable info from them.

Also moved CONTRIBUTING.md to the .GitHub folder to reduce clutter (GitHub supports it being in that folder).


ignacionr:
While you guys are looking at that, there's a section at the CONTRIBUTING.md that suggests creating issues with the "Schema" label. It's not possible to apply labels if you don't have write permissions so maybe correct that wording? That is, since the audience is greater than the local team.

PR #1681, opened 2018-07-03
[CPP] Fix/cpp small fixes

ignacionr:
Only one little bit was breaking, the others are just removing warnings for unused and reorder.

@paulcam206 and @RebeccaAnne keep the awesome work, you have an avid reader here.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,6 @@
 {
-    "editor.insertSpaces": false,
-    "editor.formatOnSave": true, // only if you want auto fomattting on saving the file
+	"editor.insertSpaces": false,
+	"editor.formatOnSave": false, // only if you want auto fomattting on saving the file
 	"editor.detectIndentation": false,
-	
+	"C_Cpp.default.cppStandard": "c++17"
 }
--- a/source/shared/cpp/ObjectModel/DateTimePreparsedToken.cpp
+++ b/source/shared/cpp/ObjectModel/DateTimePreparsedToken.cpp
@@ -3,13 +3,13 @@
 
 using namespace AdaptiveSharedNamespace;
 
-DateTimePreparsedToken::DateTimePreparsedToken() : m_text(""), m_format(DateTimePreparsedTokenFormat::RegularString),
-    m_date{}
+DateTimePreparsedToken::DateTimePreparsedToken() : m_text(""), m_date{},
+    m_format(DateTimePreparsedTokenFormat::RegularString)
 {
 }
 
 DateTimePreparsedToken::DateTimePreparsedToken(std::string const &text, DateTimePreparsedTokenFormat format) :
-    m_text(text), m_format(format), m_date{}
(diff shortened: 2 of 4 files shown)


source/shared/cpp/ObjectModel/ShowCardAction.cpp
```diff
@@ -51,7 +51,7 @@ std::shared_ptr<BaseActionElement> ShowCardActionParser::Deserialize(
     auto parseResult = AdaptiveCard::Deserialize(json.get(propertyName, Json::Value()), std::numeric_limits<double>::max(), elementParserRegistration, actionParserRegistration);
 
     auto showCardWarnings = parseResult->GetWarnings();
-    auto warningsEnd = warnings.insert(warnings.end(), showCardWarnings.begin(), showCardWarnings.end());
+    warnings.insert(warnings.end(), showCardWarnings.begin(), showCardWarnings.end());
```
paulcam206: Same comment as the ParseUtil.cpp change :)

source/shared/cpp/ObjectModel/ParseUtil.cpp
```diff
@@ -306,7 +306,7 @@ std::string ParseUtil::ToLowercase(std::string const &value)
 {
     std::string new_value;
     new_value.resize(value.size());
-    auto new_end = std::transform(value.begin(), value.end(), new_value.begin(), [](char c) { return std::tolower(c, std::locale()); });
```
paulcam206: I actually added this one on purpose. This was part of PR #1579, and was introduced due to core guideline [ES.84](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es84-dont-try-to-declare-a-local-variable-with-no-name) - _"Don't (try to) declare a local variable with no name"_.

Per original PR:
> Basically, some functions return an object and some operations create temporary objects. These objects may have complex ctors or dtors which run in a way and at a time that's not obvious to the reader. 

> There are a few places where we were calling either `std::transform()` or an `insert()` function where it didn't make sense to change to a different form. For these, I assigned the return to a new local variable, which squelches the warning and also alerts the reader to the fact that there's an object being created (even if it's not going to be used).

ignacionr: I see, and understand. Still curious how do you feel about it at a personal level? (As a programmer I’ve used unnamed objects and don’t agree that adding names makes anything other than clutter up.)
Again, agree to the value of following Core so will change back.

paulcam206: I'm a little on the fence about it. In the specific instances you changed back, I agree that it introduces a bit of clutter. However, I think it's worth it to force future devs to recognize that there's a complex object being created. It also makes scoping clear (objects that construct and destruct in place are weird) :)

ignacionr:
Simply reverted that commit. Please recheck @paulcam206 and thanks so much for the guidance.

PR #1687, opened 2018-07-05
Jwoo/variablenumoflines choiceset

nesalang:
Made ChoiceSet titles to be displayed in multiple lines instead of adjusting font size to fit.
Disabled loading nib for date and time input view.
Removed logic for using custom image renderer for loading background image. 
Fixed the issues of too little or much gaps by refining the logic for intrinsic content size method.

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRChoiceSetViewDataSource.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRChoiceSetViewDataSource.mm
@@ -10,6 +10,9 @@
 
 using namespace AdaptiveCards;
 
+const CGFloat padding = 16.0f;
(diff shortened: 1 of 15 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRChoiceSetViewDataSource.mm
```diff
@@ -17,6 +17,8 @@ @implementation ACRChoiceSetViewDataSource
     NSIndexPath *_lastSelectedIndexPath;
     NSMutableSet *_defaultValuesSet;
     NSArray *_defaultValuesArray;
+    CGFloat _padding;
+    CGFloat _accessoryViewWidth;
```
bviglietta: These should just be constants, instead of fields on every instance of the class.

Same for the analogous fields on other classes.

PR #1696, opened 2018-07-05
.NET Accessibility updates

matthidinger:


grahamehorner:
@matthidinger just wondering if hardcoded strings should be avoided and these should come via a localised resource file.

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
```diff
@@ -762,7 +768,8 @@ protected static HtmlTag DateInputRender(AdaptiveDateInput input, AdaptiveRender
                 .Attr("type", "date")
                 .AddClass("ac-input")
                 .AddClass("ac-dateInput")
-                .Style("width", "100%");
+                .Style("width", "100%")
+                .Attr("aria-name", input.Placeholder ?? "Select date in mm/dd/yyyy format");
```
grahamehorner: would be nice if these default strings came via a localised resource file.

Issue #1652, opened 2018-06-27
[TS] Consider splitting card-elements.ts into per-class files


ignacionr:
It would take some coordination. I know it's painful. Also it's likely in your head you have some broader plan (e.g. to separate the parsing from the rendering, as happens with the C++ code base).

But let's rather do this before reaching 10k lines of code?

dclaux:
No plan to separate the parsing from the rendering, that is by design. Splitting into per class files has its challenges, with CardElement referencing Action and Action referencing CardElement, but yes, I think eventually we'll want to do that.

ignacionr:
Right. The parent could become an interface instead of a class. Something that provides the services in CardElement we're using from Action.
If this change will happen, wouldn't you agree we rather do it sooner than later?

ignacionr:
This simple change:

<img width="367" alt="screen shot 2018-06-28 at 11 44 22 am" src="https://user-images.githubusercontent.com/1210985/42013697-b06c4c32-7ac8-11e8-9ede-de2883a07e5b.png">

plus changing the declaration of setParent() and parent() only on that class. This works AOK (tsc produces no issues, the derived setters still work, and tests pass).

dclaux:
Sorry for the delay, I had missed your new comments on this. Yes, if we're going to do it, better sooner than later.

The goal for this work is that things work exactly as before after the changes though. E.g. no breaking change for developers who are using the renderer in their own app. Likely this is possible by introducing IAction as you suggest, but it might be more complicated.

I don't know when I'll have time for this (and honestly, to me it is not that important. I find that often people don't really know why they even want classes split into separate files other than "because it's better".) Is this something you would want to take a look at in addition to the Media element?

ignacionr:
Thanks for your comments. I agree in general to what you say, as I value more “fluency” than file structure, and also couldn’t care less for an assumed value in just conventions.

On the other hand, while working on the Media element, it did happen that merging changes confused git causing extra work (a long file worked on different parts has greater chances of merge conflicts).

Yes, I would be happy to give this a try. Because it requires some coordination, is the code normally frozen over the weekend? Can we issue a request to the team to push this Friday, so they won’t find themselves losing stuff? Also the commit would need to be reviewed and merged with priority so to avoid missing changes in between.

Again, I’m game.

PR #1710, opened 2018-07-13
Refactor/card elements

ignacionr:
Hey @dclaux please take a look at this. Took more than I anticipated but just because this thing of having to work for a living, you know. Still I'm pretty happy and hope you like it.

Considering:

- https://github.com/Microsoft/AdaptiveCards/issues/1652
- this refactoring is on a moving target (by definition)
- folks may have their changes done on the big file already, before rebasing

So I'm offering this refactor assistant, that you can run ``` node refactor.js ``` and will create 42 new files, and patch the global include (adaptivecards.ts) to export from each.

The Good:
- After running the refactoring, you can ``` jest ``` and all tests pass. This is important because you could say "usage style that we want to support needs to be covered by tests || die()".
- Say you have some changes on the big file, but the project has already moved on to per-class; you can still run this refactor on your code base and it will redistribute code on the same fashion, so there will be effectively no changes on what you haven't changed.
The Bad:
- It produces a long list of unused imports. Removing them manually would lose the idempotence (is that a word?) and remove the second benefit listed.
The Ugly:
- All that's not a class ends on a new 'utility.ts' file. Some "export" clauses look bit weirdly stated.

Eager to read your thoughts on this.
Sincerely, Ignacio.//

### DIFF
--- a/source/nodejs/adaptivecards/package.json
+++ b/source/nodejs/adaptivecards/package.json
@@ -44,6 +44,7 @@
     "dts-generator": "^2.1.0",
     "jest": "^21.2.1",
     "json-loader": "0.5.4",
+    "prepend-file": "^1.3.1",
     "regenerator-runtime": "^0.11.0",
     "ts-jest": "^21.2.3",
     "tslint": "^4.4.2",
--- a/source/nodejs/adaptivecards/src/refactor.js
+++ b/source/nodejs/adaptivecards/src/refactor.js
@@ -2,6 +2,7 @@
 exports.__esModule = true;
 var fs = require("fs");
 var ts = require("typescript");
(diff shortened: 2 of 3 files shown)


dclaux:
@ignacionr very cool! Let me look into it, will get back to you.

ignacionr:
Definitely! This is due to the moving target nature (pretty sure once I merge down from master I'll see the new thing in source). Will get it done.

ignacionr:
Soon as I corrected that, and the full cycle started to work, I finally entered the zone you had warned about. Now getting to a bigger refactoring (manual) adding interfaces; it's looking good to me. Looking forward to your review (it'll be ready Monday before morning -the beauty of time zones).

PR #1713, opened 2018-07-13
Media fixes

RebeccaAnne:
- Fixed a bug which lead to missing posters when interactivity was off
- Made sure the entire touch target was made collapsed rather than just the poster
- Added a height to the media element in the audio case to ensure it was displayed when there is no poster


source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -316,12 +317,13 @@ HRESULT HandleMediaClick(
     HSTRING mimeType,
     IAdaptiveMediaEventInvoker* mediaInvoker)
 {
+    ComPtr<IMediaElement> localMediaElement{ mediaElement };
+
     // When the user clicks: hide the poster, show the media element, open and play the media
```
khouzam: Did this need to move? The access to the variable remains in the same scope as before.

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -370,8 +372,23 @@ HRESULT HandleMediaClick(
...
+            ComPtr<IImageSource> posterSource;
+            RETURN_IF_FAILED(localMediaElement->get_PosterSource(&posterSource));
+
+            if (audioOnly && posterSource == nullptr)
+            {
+                // If this is audio only and there's no poster, set the height so that the 
+                // controls are visible.
+                ComPtr<IFrameworkElement> mediaAsUiElement;
+                RETURN_IF_FAILED(localMediaElement.As(&mediaAsUiElement));
+                RETURN_IF_FAILED(mediaAsUiElement->put_Height(c_audioHeight));
```
khouzam: Don't we already have a mediaAsUiElement variable above that we could use?

source/uwp/Renderer/lib/MediaHelpers.cpp
RebeccaAnne: Moved it back.

---
In reply to: [202800202](https://github.com/Microsoft/AdaptiveCards/pull/1713#discussion_r202800202) [](ancestors = 202800202)

source/uwp/Renderer/lib/MediaHelpers.cpp
RebeccaAnne: This variable is misnamed, it's actually a FrameworkElement. Updated to the correct name.

---
In reply to: [202800736](https://github.com/Microsoft/AdaptiveCards/pull/1713#discussion_r202800736) [](ancestors = 202800736)

PR #1717, opened 2018-07-16
allows action set view to be accessed

nesalang:
Added ActionsSet renderer
Adjusted inline showcard padding
Fixed UIScrollView interfering with ACRTextView's scrolling


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCardPrivate.h
```diff
@@ -12,5 +12,6 @@ using namespace AdaptiveCards;
 @interface ACOAdaptiveCard()
 
 - (std::shared_ptr<AdaptiveCard> const &)card;
+- (void)card:(std::shared_ptr<AdaptiveCard> const &)card;
```
bviglietta: >card: [](start = 8, length = 5)

Setters are almost universally named like `setCard:` in Objective-C.

But it also seems like it would be simpler to make this a property:
`@property(nonatomic) std::shared_ptr<AdaptiveCard> card;`

PR #1683, opened 2018-07-03
Adding support for height="stretch" to WPF renderer (resolves issue #1665)

Identifier:

PR #1734, opened 2018-07-21
 Aligning rows within ColumnSets by default, unless element height explicitly = "auto"

Identifier:
This is an optional addition on top of #1683 that also aligns elements inside ColumnSets by default, unless users explicitly set the height property to "auto", which seems like a much better default behavior.


matthidinger:
Hey @Identifier, please accept my apologies that this has existed over a year :)

Is this still something you want us to look at it? I see a // TODO in the code near the `Height` property, so any additional information would be helpful

Identifier:
Hi @matthidinger, I believe this was already integrated at some point, so I'll just close this now.

PR #1745, opened 2018-07-31
Tupha/api enumerate all images dotnet

tuanh118:
Created and exposed a new function from `AdaptiveCard` object named `GetResourceInformation` that returns a list of information for all Images and potentially Medias in the card


source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
```diff
@@ -108,5 +108,116 @@ public void TestParsingInvalidCardMissingType()
...
+                    typeof(AdaptiveImage),
+                    null
+                ),
+                new RemoteResourceInformation(
+                    "http://adaptivecards.io/content/cats/3.png",
+                    typeof(AdaptiveImage),
+                    null
+                ),
+                new RemoteResourceInformation(
+                    "http://adaptivecards.io/content/cats/2.png",
```
nesalang: what is expected behavior when we have multiple of same resources? although we are not getting the resources, it seems wasteful to have redundant resources listed.

source/dotnet/Library/AdaptiveCards/AdaptiveCard.cs
```diff
@@ -218,5 +218,95 @@ public bool ShouldSerializeJsonSchema()
...
+                ));
+            }
+
+            // Get all images in body
+            foreach (AdaptiveElement bodyElement in card.Body)
+            {
+                resourceInformationList.AddRange(GetResourceInformationInElement(bodyElement));
+            }
+
+            // Get all images in ShowCard actions
```
nesalang: actions have icon urls; i checked shared model; its actions elements doesn't do anything; let me check with Becky

tuanh118: Right thanks for pointing out

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
tuanh118: Yeah I think returning everything in order (current solution) is the safer way, since the caller can process the array after getting the result in any way they want. Do you know what other renderers are handling it?

source/dotnet/Library/AdaptiveCards/AdaptiveCard.cs
```diff
@@ -218,5 +218,96 @@ public bool ShouldSerializeJsonSchema()
...
+
+            // Base case
+            if (element is AdaptiveImage imageElement && !String.IsNullOrEmpty(imageElement.UrlString))
+            {
+                resourceInformationList.Add(new RemoteResourceInformation(
+                    imageElement.UrlString,
+                    typeof(AdaptiveImage),
+                    null
+                ));
+            }
```
nesalang: nit: extra space at LN279

tuanh118: hm I can't see it

tuanh118:
It's probably better to wait for `iconUrl` implementation for .NET is done (Gilles and I are working on it). Should I just keep this PR open or close it and create a new one when everything's good to go?

source/dotnet/Library/AdaptiveCards/AdaptiveCard.cs
nesalang: sorry i meant new line between if and else if stmt.

---
In reply to: [207416337](https://github.com/Microsoft/AdaptiveCards/pull/1745#discussion_r207416337) [](ancestors = 207416337)

Issue #1026, opened 2017-12-04
Support explicit column width in pixels


dclaux:
# Implementation status
- [x] JavaScript 
- [x] UWP (#1392)
- [x] .NET (#1662)
- [x] iOS (#1664)
- [x] Android (#1663)
- [x] Documentation

# Problem
Achieving table-like layouts with Adaptive is hard because one has to use several ColumnSets on top of each other and there are only two ways to ensure that all columns in all those set have the same widths:
- Use weights
- Use auto and make sure the content always has the same width. That is easy enough with images, but very hard (near impossible) with text

# Solution
Allow columns to have an explicit width, expressed in pixels. Essentially, make it possible in an easy and straightforward way to do what is already possible with a column containing an explicitly sized invisible image, which card authors _will_ do to achieve the results they want.

|Property|Type|Description|
|--|--|--|
|Column.**width**|string|Defines the width of the column in pixels. Width can be either a number or a string. To specify a width in pixels, set the unit px. eg: "width": "50px".|

Example use:
```
{
	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
	"type": "AdaptiveCard",
	"version": "1.0",
	"body": [
		{
		    "type": "ColumnSet",
		    "columns": [
		        {
		            "width": "50px",
		            "items": [
		                {
		                    "type": "Image",
		                    "url": "http://3.bp.blogspot.com/-Xo0EuTNYNQg/UEI1zqGDUTI/AAAAAAAAAYE/PLYx5H4J4-k/s1600/smiley+face+super+happy.jpg",
		                    "size": "stretch"
		                }
		            ]
		        },
		        {
		            "width": "stretch",
		            "verticalContentAlignment": "center",
		            "items": [
		                {
		                    "type": "TextBlock",
		                    "text": "This card has two ColumnSets on top of each other. In each, the left column is explicitly sized to be 50 pixels wide.",
		                    "wrap": true
		                }
		            ]
		        }
		    ]
		},
		{
		    "type": "ColumnSet",
		    "columns": [
		        {
		            "width": "50px"
		        },
		        {
		            "width": "stretch",
		            "verticalContentAlignment": "center",
		            "items": [
		                {
		                    "type": "TextBlock",
		                    "text": "In this second ColumnSet, columns align perfectly even though there is nothing in the left column.",
		                    "wrap": true
		                }
		            ]
		        }
		    ]
		}
	]
}
```

As rendered:
[...]

PR #1746, opened 2018-07-31
added handler for column width in pixels

tuanh118:
Not sure if double unit makes sense for pixel (e.g., `50.3px`) and also for weighted proportion (already in use)

For testing, you can use the payload below. I noticed the payload provided in the [epic](https://github.com/Microsoft/AdaptiveCards/issues/1026) does not have `"type": "Column"` anywhere and the .NET renderer throws an error for it. Is that something missing from the .NET renderer? 
```
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "ColumnSet",
      "columns": [
        {
          "type": "Column",
          "width": "50px",
          "items": [
            {
              "type": "Image",
              "url": "http://3.bp.blogspot.com/-Xo0EuTNYNQg/UEI1zqGDUTI/AAAAAAAAAYE/PLYx5H4J4-k/s1600/smiley+face+super+happy.jpg",
              "size": "stretch"
            }
          ]
        },
        {
          "type": "Column",
          "width": "stretch",
          "items": [
            {
              "type": "TextBlock",
              "text": "This card has two ColumnSets on top of each other. In each, the left column is explicitly sized to be 50 pixels wide.",
              "wrap": true
            }
          ]
        }
      ]
    },
    {
      "type": "ColumnSet",
      "columns": [
        {
          "type": "Column",
          "width": "50px"
        },
        {
          "type": "Column",
          "width": "stretch",
          "items": [
            {
              "type": "TextBlock",
              "text": "In this second ColumnSet, columns align perfectly even though there is nothing in the left column.",
              "wrap": true
            }
          ]
        }
      ]
    }
  ]
}
```

### DIFF
--- a/source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
+++ b/source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
@@ -385,11 +385,14 @@ namespace AdaptiveCards.Rendering.Html
                 }
                 else
                 {
-                    double val;
-                    if (double.TryParse(width, out val))
+                    if (double.TryParse(width, out double val) && val >= 0)
                     {
                         var percent = Convert.ToInt32(100 * (val / max));
                         uiColumn = uiColumn.Style("flex", $"1 1 {percent}%");
+                    }
+                    else if (width.EndsWith("px") && double.TryParse(width.Substring(0, width.Length-2), out double pxVal) && pxVal >= 0)
+                    {
+                        uiColumn = uiColumn.Style("flex", $"0 0 {(int)pxVal}px");
                     }
                     else
                     {
--- a/source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveColumnSetRenderer.cs
+++ b/source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveColumnSetRenderer.cs
@@ -53,10 +53,10 @@ namespace AdaptiveCards.Rendering.Wpf
                     uiColumnSet.ColumnDefinitions.Add(new ColumnDefinition() { Width = GridLength.Auto });
                 else
                 {
-                    if (double.TryParse(width, out double val))
+                    if (double.TryParse(width, out double val) && val >= 0)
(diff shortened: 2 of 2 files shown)


matthidinger:
At some point we stopped requiring a type on Column, the .NET Renderer is probably out of date there. I think int makes sense for pixel unit, not sure where fraction of a pixel applies?

khouzam:
We need to make sure that we're handling parsing of the units in the same way on all renderers. @paulcam206 noticed that also and is looking at fixing it.

Fractional pixels can make sense since we're using logical pixels, but the width is exposed as an int in the shared model. In terms of functionality, fractional pixels don't really make a that much of a difference (until we add transforms and scales then could really affect the widths). My main concern here is consistency in how we parse the value having 23.5px be valid in one renderer but not in others will make for a bad sharing experience.

paulcam206:
I have a change in #1750 that deals with fractional pixels. FWIW, the visualizer supports (or at least allows) fractional pixels.

tuanh118:
So the consensus is we allow "23.5px" and also render exactly 23.5px right?

paulcam206:
Well, as @khouzam noted, the shared model exposes the width as an `unsigned int`, so we can't currently have non-integral sizes on the rendering side. It's something we might want to pursue in the future, but can't today without a breaking change to the object model.

paulcam206:
the way I've addressed it in #1750 is that the parser will accept fractional values, but still only exposes integral values through the object model. let's have a chat later about what to do.

one thing I *don't* know is whether the typescript visualizer is actually using fractional units, or just snapping to the nearest integer.

tuanh118:
https://github.com/Microsoft/AdaptiveCards/blob/master/source/nodejs/adaptivecards/src/card-elements.ts#L101
It's using the int part of the number. For example, 11.8 -> 11, -11.8 -> -11

paulcam206:
I think this lines up with what `std::stoi` does, so both would be in agreement.

tuanh118:
I just added final modifications so that it aligns with the TypeScript renderer: allow fractional values for pixel and truncates it to use only its integral part.
Also added a check to handle negative width

PR #1748, opened 2018-08-01
implemented API for enumerating images in a card (1094)

nesalang:
- implemented API for enumerating images in a card (1094)
- added ios UnitTest for the feature


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.h
```diff
@@ -15,5 +16,6 @@
 - (NSArray *)getInputs;
 - (void)setInputs:(NSArray *)inputs;
 - (void)appendInputs:(NSArray *)inputs;
+- (ACORemoteResourceInformation *)remoteResourceInformation;
```
bviglietta: Why is this on the public API? It doesn't look like callers can do anything with it. Is there a plan to add something into the ACORemoteResourceInformation interface later on?

nesalang: so for developers, first option will be including c++ compatible header which is ACORemoteResourceInformationPrivate.h, and use its c++ intefaces.  second option of having object-c wrapper for c++ interfaces will come later when there are enough demands. 

---
In reply to: [208670082](https://github.com/Microsoft/AdaptiveCards/pull/1748#discussion_r208670082) [](ancestors = 208670082)

bviglietta: Is the C++ object model really open to developers? It doesn't appear to be documented, and it's in headers named "Private". If you start advertising it, that implies you'll support it and make efforts to maintain backward compatibility.

In the other platform SDKs for AdaptiveCards, it looks like RemoteResourceInformation is fully projected into the target environment.

---
In reply to: [208711745](https://github.com/Microsoft/AdaptiveCards/pull/1748#discussion_r208711745) [](ancestors = 208711745,208670082)

nesalang: yes, C++ object model is open to developers; and I need to update the spec. I will change the header name. and also it probably is wise to fully project RemoteResourceInformation to target environment. 

---
In reply to: [208736676](https://github.com/Microsoft/AdaptiveCards/pull/1748#discussion_r208736676) [](ancestors = 208736676,208711745,208670082)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACORemoteResourceInformation.h
```diff
@@ -0,0 +1,17 @@
...
+//
+//  Copyright © 2018 Microsoft. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "ACOBaseCardElement.h"
+
+@interface ACORemoteResourceInformation:NSObject
+
+@property (readonly) NSString* url;
```
bviglietta: >NSString* url [](start = 21, length = 13)

I would recommend exposing this as an `NSURL*`.

PR #1749, opened 2018-08-01
[Android] Add support for background image color

almedina-ms:
Add support for background image color in Android

Fixes this: https://github.com/Microsoft/AdaptiveCards/issues/1413

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
@@ -141,14 +141,7 @@ public class ImageRenderer extends BaseCardElementRenderer
         int backgroundColor = -1;
         if(!imageBackgroundColor.isEmpty())
         {
-            try
-            {
-                // The valid color formats are "#RRGGBB" and "#AARRGGBB", the string doesn't comply with the format, then keep the -1
-                backgroundColor = Color.parseColor(imageBackgroundColor);
-            }
-            catch (IllegalArgumentException e)
-            {
-            }
+            backgroundColor = Color.parseColor(imageBackgroundColor);
         }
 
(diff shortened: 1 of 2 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
```diff
@@ -126,7 +136,20 @@ else if ((image = Image.dynamic_cast(baseCardElement)) == null)
 
         ImageView imageView = new ImageView(context);
         imageView.setTag(image);
-        ImageRendererImageLoaderAsync imageLoaderAsync = new ImageRendererImageLoaderAsync(renderedCard, imageView, hostConfig.getImageBaseUrl(), image.GetImageStyle());
+
+        String imageBackgroundColor = image.GetBackgroundColor();
+        int backgroundColor = -1;
+        if(!imageBackgroundColor.isEmpty())
+        {
+            backgroundColor = Color.parseColor(imageBackgroundColor);
```
nesalang: try & catch blocks?

almedina-ms: currently the UWP renderer drops the card if the color is not valid, will check with Gilles the fix for it

---
In reply to: [207017510](https://github.com/Microsoft/AdaptiveCards/pull/1749#discussion_r207017510) [](ancestors = 207017510)

nesalang: If i were you, if android renderer has infrastructures for warnings, i would use that; otherwise, i would just log it and put it on to-do list. 

---
In reply to: [207045249](https://github.com/Microsoft/AdaptiveCards/pull/1749#discussion_r207045249) [](ancestors = 207045249,207017510)

PR #1750, opened 2018-08-01
More cppcore changes

paulcam206:
Enable the following checks:
* [CPPCORECHECK_BOUNDS_WARNINGS](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#probounds-bounds-safety-profile)
   - The check that we had the most errors in deals with unchecked bounds access via `operator[]`. When indexing into a container type via `[]`, no bounds checking occurs, making it easy to walk off the end of, say, a `std::vector<>`. Access via `at()` **is** bounds checked and will throw if an attempt is made to retrieve an element at an invalid index. The fix is to switch to using `at()` when possible.
* [CPPCORECHECK_OWNER_POINTER_WARNINGS, CPPCORECHECK_RAW_POINTER_WARNINGS, and CPPCORECHECK_SHARED_POINTER_WARNINGS](http://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r-resource-management)
  - There were actually few issues caught by these checks. They were all to do with a ban on the explicit use of `operator new` -- prefer using `std::make_unique<>` and `std::make_shared<>`.
* [CPPCORECHECK_TYPE_WARNINGS](http://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#prosafety-type-safety-profile)
   - This profile was responsible for most changes in this PR. Most of the issues caught were due to either the use of c-style casts (rather than `static_cast<>`) or the use of casts (of any kind) when doing arithmetic operations. The trickiest of these issues was in `ValidateUserInputForDimensionWithUnit()`, where we were doing some calisthenics around the conversion of a dimensional unit in a string (e.g. `17px`) to an `int` (`17`). The existing code allowed for the use of fractional values, but only if they converted directly to an `int`. For example, `20.0px` and `20.0000000000000000001px` are accepted by the model, but `20.1px` is not. However, the [visualizer](http://adaptivecards.io/visualizer/) allows fractional pixels, so the fix is to just switch to `std::stoi()` for our conversion.

### DIFF
--- a/source/shared/cpp/ObjectModel/MarkDownBlockParser.cpp
+++ b/source/shared/cpp/ObjectModel/MarkDownBlockParser.cpp
@@ -24,13 +24,17 @@ void MarkDownBlockParser::ParseBlock(std::stringstream &stream)
     case ']': case ')':
     {
         // add these char as token to code gen list
-        m_parsedResult.AddNewTokenToParsedResult((char)stream.get());
+		char streamChar;
+		stream.get(streamChar);
+        m_parsedResult.AddNewTokenToParsedResult(streamChar);
         break;
     }
     case '\n': case '\r':
     {
         // add new line char as token to code gen list
-        m_parsedResult.AddNewLineTokenToParsedResult((char)stream.get());
+		char streamChar;
+		stream.get(streamChar);
+        m_parsedResult.AddNewLineTokenToParsedResult(streamChar);
         break;
     }
     // handles list block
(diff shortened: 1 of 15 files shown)


source/shared/cpp/ObjectModel/MarkDownBlockParser.cpp
```diff
@@ -24,13 +24,17 @@ void MarkDownBlockParser::ParseBlock(std::stringstream &stream)
     case ']': case ')':
     {
         // add these char as token to code gen list
-        m_parsedResult.AddNewTokenToParsedResult((char)stream.get());
+		char streamChar;
```
nesalang: is it to remove static cast?

paulcam206: there are two forms of `get` function relevant here. the first takes no arguments and returns an `unsigned char`. the second can take a `char&` as an out parameter. rather than hard casting from a signed value to an unsigned one, using the latter form puts the responsibility on stl to Do The Right Thing :)

nesalang: cool; thanks

---
In reply to: [207014840](https://github.com/Microsoft/AdaptiveCards/pull/1750#discussion_r207014840) [](ancestors = 207014840)

nesalang:
thanks for your great work.

PR #1752, opened 2018-08-01
[Android] Add support for explicit image size

almedina-ms:
Add support for explicit image size for android


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
```diff
@@ -184,11 +184,11 @@ else if (horizontalAlignment == HorizontalAlignment.Center)
...
 
-        if(hasExplicitSize)
+        if (hasExplicitSize)
         {
             int widthInPixels = Util.dpToPixels(context, pixelWidth);
             int heightInPixels = Util.dpToPixels(context, pixelHeight);
-            if(isAspectRatioNeeded)
+            if (isAspectRatioNeeded)
             {
                 if (pixelWidth != 0) {
```
paulcam206: there's still a mix of `\n{` and `{\n` :)

paulcam206:
:shipit:

PR #1758, opened 2018-08-02
stopped requiring type property for Columns

tuanh118:
- **Comment:** Not totally happy with the modifications, but it's the best solution I have to make it work. Definitely open for a better solution.
- **Modifications:**
  - Removed `[JsonRequired]` from `AdaptiveTypedElement.cs` since it already has `Required = Required.Always` in the line above and `[JsonRequired]` seems to block children to overriding it
  - Overriding `Required` in `AdaptiveColumn`
  - Added a special case in `AdaptiveTypedElementConverter.cs`. If we want to do something similar for other elements in the future, we'll also need to add corresponding special cases.


tuanh118:
@khouzam Not very elegant but that is the only way (supported in .NET 1.3) I can find for
- Determining if `Type` is required
- Get the correct `TypeName` to use when `Type` is not required

Issue #1658, opened 2018-06-28
[Android] API to enumerate all images contained in a card


andrewleader:
See parent #1094 for details

PR #1759, opened 2018-08-02
[Android] GetResourceUris API

almedina-ms:
Add the GetResources API to android

Solves this: https://github.com/Microsoft/AdaptiveCards/issues/1658

### DIFF
--- a/source/android/adaptivecards/src/AdaptiveCardObjectModel.i
+++ b/source/android/adaptivecards/src/AdaptiveCardObjectModel.i
@@ -124,6 +124,7 @@ struct tm {
 %shared_ptr(AdaptiveCards::SubmitAction)
 %shared_ptr(AdaptiveCards::AdaptiveCardParseWarning)
 %shared_ptr(AdaptiveCards::ParseResult)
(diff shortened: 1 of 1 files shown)


source/android/adaptivecards/src/AdaptiveCardObjectModel.i
```diff
@@ -139,7 +141,6 @@ struct tm {
 %shared_ptr(AdaptiveCards::SubmitActionParser)
 %shared_ptr(AdaptiveCards::ImageSetParser)
 %shared_ptr(AdaptiveCards::DateInputParser)
-%shared_ptr(AdaptiveCards::DateTimePreparsedToken)
```
RebeccaAnne: >) [](start = 49, length = 1)

Why are we removing this?

almedina-ms: I dont even know how did this happen, reverting this deletion

---
In reply to: [207609313](https://github.com/Microsoft/AdaptiveCards/pull/1759#discussion_r207609313) [](ancestors = 207609313)

PR #1764, opened 2018-08-03
added fallback for invalid color

tuanh118:
**NOTE**: This change affects all color hex values, including ones in host config and image's background color in payload

Added a check to use color only when its format is `#AARRGGBB`. Otherwise, simply ignore it

### DIFF
--- a/source/uwp/Renderer/lib/Util.cpp
+++ b/source/uwp/Renderer/lib/Util.cpp
@@ -547,6 +547,15 @@ HRESULT GenerateSharedSeparator(
 
 HRESULT GetColorFromString(std::string colorString, ABI::Windows::UI::Color *color) noexcept try
 {
+    // Expected format for color string: "#AARRGGBB"
+    // Other format is ignored (set alpha to 0)
+    if (colorString.length() != 9)
+    {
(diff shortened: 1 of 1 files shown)


source/uwp/Renderer/lib/Util.cpp
```diff
@@ -547,6 +547,15 @@ HRESULT GenerateSharedSeparator(
...
 HRESULT GetColorFromString(std::string colorString, ABI::Windows::UI::Color *color) noexcept try
 {
+    // Expected format for color string: "#AARRGGBB"
+    // Other format is ignored (set alpha to 0)
+    if (colorString.length() != 9)
+    {
+        color->A = static_cast<BYTE>(0);
+
+        return S_OK;
+    }
```
almedina-ms: this do validate that the color complies with the format #AARRGGBB, but how are we validating that the values for ARGB are valid hex characters??

tuanh118: you're right! Fixed in 0b983df

tuanh118:
per our email discussion, I also added support for 6-digit hex values, which is quite a bit of change. What do you guys think?

PR #1768, opened 2018-08-03
Added support for icon in actions

tuanh118:
- **WPF:** when icon is left of title, its height is adjusted to match title's height
- **HTML:** icon height is always set to `iconSize` since it's difficult to detect and match title's height

Added a few parsing tests


khouzam:
>         }

Need to restore the oldIconPlacement value to not have changed the hostconfig for other cards.

---
Refers to: source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs:331 in b32b64a. [](commit_id = b32b64abe8c4d6954870e244e908638175529c93, deletion_comment = False)

PR #1770, opened 2018-08-07
Removed the ListBox in AdaptiveChoiceSetRenderer since it interferes …

Identifier:
…with the ability to click and use the keyboard on the radio buttons and checkboxes inside.  Also the style matches the WebChat rendering now as well.


khouzam:
Thanks @Identifier 

I've merged this change in. It's the same behavior that we had in our UWP renderer.

PR #1771, opened 2018-08-07
Enabling TextBlock wrapping when a card has no constrained size.

Identifier:
Currently, when an adaptive card does not have a constrained size (i.e. is allowed to grow to fit its contents), TextBlocks do not wrap even if "wrap" is set to true.  Now, when a TextBlock has "wrap": true it will give way so that the container will only grow to fit its other children and the TextBlock will wrap to fit accordingly.


khouzam:
Thanks @Identifier,

It looks like this breaks cards such as [this one](https://github.com/Microsoft/AdaptiveCards/files/2354807/adaptivecard.json.txt) from issue #1861 which seems very much the same problem that you're trying to address. We are looking to have the same behavior in all renderers to help with consistency. We'll be looking at the problem for our 1.2 release.

shalinijoshi19:
@RebeccaAnne , is this PR still going to be needed after the 1.2 work we are doing around rich text?

PR #1774, opened 2018-08-07
Disallow Strings to Ints

tuanh118:
The `.NET` parser has the same bug as reported in the original issue. This is the default behavior of NewtonSoft. So we just need a custom Integer converter with stricter rule that disallows String to be accepted for all Int attributes


source/dotnet/Library/AdaptiveCards/AdaptiveTextBlock.cs
```diff
@@ -103,6 +103,7 @@ public AdaptiveTextBlock(string text)
         ///     When Wrap is true, you can specify the maximum number of lines to allow the textBlock to use.
         /// </summary>
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+        //[JsonConverter(typeof(StrictIntConverter))]
```
matthidinger: Looks good to me. If it's all working I think we can remove this commented line

PR #1775, opened 2018-08-08
Tupha/support icon size uwp

tuanh118:
Most of the changes in `XamlBuilder.cpp` is rearranging code blocks in order to correctly apply styling. I've added more comments to each part so it might be clearer to review the new file rather than the diff


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -829,9 +823,43 @@ AdaptiveNamespaceStart
...
+            }
+
+            ComPtr<IPanel> buttonContentsPanel;
+            THROW_IF_FAILED(buttonContentsStackPanel.As(&buttonContentsPanel));
+
+            // Add image to stack panel
+            XamlHelpers::AppendXamlElementToPanel(buttonIcon.Get(), buttonContentsPanel.Get());
 
             // Just add spacing when the icon must be located at the left of the title
             if (iconPlacement == ABI::AdaptiveNamespace::IconPlacement::LeftOfTitle)
```
khouzam: >iconPlacement == ABI::AdaptiveNamespace::IconPlacement::LeftOfTitle [](start = 16, length = 67)

Shouldn't the spacing be added when not al actions have icons also? Since that will result in LeftOfTitle behavior also? Shouldn't this block be in the if clause above?

tuanh118: you're right. I didn't notice it since it was there before. Fixed in d8f2554

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -829,38 +823,62 @@ AdaptiveNamespaceStart
...
             THROW_IF_FAILED(buttonIcon.As(&buttonIconAsFrameworkElement));
+            ComPtr<IUIElement> separator;
+            if (iconPlacement == ABI::AdaptiveNamespace::IconPlacement::AboveTitle && allActionsHaveIcons)
+            {
+                THROW_IF_FAILED(buttonContentsStackPanel->put_Orientation(Orientation::Orientation_Vertical));
 
-            // Just add spacing when the icon must be located at the left of the title
-            if (iconPlacement == ABI::AdaptiveNamespace::IconPlacement::LeftOfTitle)
+                // Set height to iconSize (aspect ratio is automatically maintained)
+                THROW_IF_FAILED(buttonIconAsFrameworkElement->put_Height(iconSize));
```
nesalang: having just one dimension for rectangle seems indicating square; maintaining the aspect ratio is the right approach? shouldn't we explicitly set the height and width to iconsize?

RebeccaAnne:
Could you add a test file to demonstrate icon size and add it to the test app?

samples/Tests/Action.IconUrl.json
```diff
@@ -0,0 +1,20 @@
...
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"type": "AdaptiveCard",
+	"version": "1.0",
+	"body": [
+		{
+			"type": "TextBlock",
+			"text": "This card has an action with an icon. Icon placement and size depend on host.",
+			"wrap": true
+		}
+	],
```
khouzam: Need to run the test app and add all the new expected files.

tuanh118: Yeah I've been resolving conflicts in the Test app and working on the new tests

PR #1776, opened 2018-08-08
Introduce parser registration, parser warning, and parser error tests

paulcam206:
Bolster code coverage for `ActionParserRegistration`, `ElementParserRegistration`, `AdaptiveCardParseWarning`, and `AdaptiveCardParseException`. Fix logic flaw in `ActionParserRegistration::RemoveParser()` where a caller can remove known parsers, but can't remove custom parsers. Also throw when attempting to remove a known action parser, to be consistent with `ElementParserRegistration`.

### DIFF
--- a/source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ParserRegistrationTest.cpp
+++ b/source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ParserRegistrationTest.cpp
@@ -28,15 +28,7 @@ namespace AdaptiveCardsSharedModelUnitTest
 					BaseCardElement(AdaptiveCards::CardElementType::Custom),
 					BaseActionElement(AdaptiveCards::ActionType::Custom)
 				{
-					m_customImage = value.get("customImageProperty", Json::Value()).asString();
-				}
-
-				virtual void GetResourceInformation(std::vector<RemoteResourceInformation>& resourceUris) override
-				{
-					RemoteResourceInformation resourceInfo;
-					resourceInfo.url = m_customImage;
-					resourceInfo.resourceType = CardElementType::Image;
-					resourceUris.push_back(resourceInfo);
+					m_customImage = value.get("customProperty", Json::Value()).asString();
 				}
 
 			private:
--- a/source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ParserRegistrationTest.cpp
+++ b/source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ParserRegistrationTest.cpp
@@ -44,7 +44,7 @@ namespace AdaptiveCardsSharedModelUnitTest
 			};
 
 			// Define custom element parser
-			class TestCustomParser : public BaseCardElementParser
+			class TestCustomElementParser : public BaseCardElementParser
 			{
 			public:
 				virtual std::shared_ptr<BaseCardElement> Deserialize(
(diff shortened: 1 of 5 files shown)


source/shared/cpp/ObjectModel/ActionParserRegistration.cpp
```diff
@@ -20,21 +20,25 @@ namespace AdaptiveSharedNamespace {
...
+            throw AdaptiveCardParseException(ErrorStatusCode::UnsupportedParserOverride,
+                "Overriding known action parsers is unsupported");
         }
     }
 
     void ActionParserRegistration::RemoveParser(std::string const &elementType)
     {
-        if (m_knownElements.find(elementType) != m_knownElements.end())
+        // make sure caller isn't attempting to remove a known element's parser
+        if (m_knownElements.find(elementType) == m_knownElements.end())
```
paulcam206: I'm assuming that this was the intended behavior?

RebeccaAnne: Looks right. In ElementParserRegistration we throw in an else (as we do in AddParser). Maybe add that here as well?

---
In reply to: [208670589](https://github.com/Microsoft/AdaptiveCards/pull/1776#discussion_r208670589) [](ancestors = 208670589)

RebeccaAnne: Or alternatively don't throw in the element parser registration.

---
In reply to: [208674103](https://github.com/Microsoft/AdaptiveCards/pull/1776#discussion_r208674103) [](ancestors = 208674103,208670589)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ParserRegistrationTest.cpp
```diff
@@ -0,0 +1,105 @@
...
+					m_customImage = value.get("customImageProperty", Json::Value()).asString();
+				}
+
+				virtual void GetResourceInformation(std::vector<RemoteResourceInformation>& resourceUris) override
+				{
+					RemoteResourceInformation resourceInfo;
+					resourceInfo.url = m_customImage;
+					resourceInfo.resourceType = CardElementType::Image;
+					resourceUris.push_back(resourceInfo);
+				}
```
RebeccaAnne: Is this needed?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ParserRegistrationTest.cpp
```diff
@@ -0,0 +1,105 @@
...
+					std::shared_ptr<AdaptiveCards::ElementParserRegistration> elementParserRegistration,
+					std::shared_ptr<AdaptiveCards::ActionParserRegistration> actionParserRegistration,
+					std::vector<std::shared_ptr<AdaptiveCards::AdaptiveCardParseWarning>>& warnings,
+					const Json::Value& value) override
+				{
+					return std::make_shared<TestCustomElement>(value);
+				}
+			};
+
+			ActionParserRegistration parser;
```
RebeccaAnne: >ActionParserRegistration [](start = 3, length = 24)

Shall we test element parser registration as well?

source/shared/cpp/ObjectModel/ActionParserRegistration.cpp
paulcam206: I think throwing makes sense. I'll add it in :)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ParserRegistrationTest.cpp
paulcam206: I was going to get to it in a future checkin, but I suppose I can just add it in now 👍

source/shared/cpp/ObjectModel/ActionParserRegistration.cpp
paulcam206: No only does it make sense, but it is consistent with how ElementParserRegistration works :)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ParserRegistrationTest.cpp
paulcam206: nope!

PR #1779, opened 2018-08-08
Tupha/disallow strings for ints nodejs

tuanh118:


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -990,8 +990,10 @@ export class TextBlock extends CardElement {
         this.color = Utils.getEnumValueOrDefault(Enums.TextColor, json["color"], this.color);
         this.isSubtle = json["isSubtle"];
         this.wrap = json["wrap"] === undefined ? false : json["wrap"];
+        if (typeof (json["maxLines"]) === "number") {
```
dclaux: You seem to have an indentation problem, the body of the if block isn't indented properly.

tuanh118: yeah my commit editor was acting weird. I also fixed to use the current `typeof` practice

dclaux:
Hopefully that won't break any existing scenario...

tuanh118:
@dclaux I checked all the examples we have in https://adaptivecards.io. All of them use integer `maxLines`. We can look into building a test suite to compare final html using Jest if we don't have anything built in already

PR #1785, opened 2018-08-09
require title for all actions

tuanh118:
This includes normal actions and selectActions

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -23,6 +23,18 @@ function isActionAllowed(action: Action, forbiddenActionTypes: Array<string>): b
 }
 
 export function createActionInstance(json: any, errors: Array<IValidationError>): Action {
+    if (!json["title"] && json["title"] !== "") {
+        raiseParseError(
+            {
+                error: Enums.ValidationError.ActionTitleCantBeNull,
+                message: "Action title is null"
+            },
+            errors
+        );
+
+        return null;
+    }
+
     var actionType = json["type"];
 
     var result = AdaptiveCard.actionTypeRegistry.createInstance(actionType);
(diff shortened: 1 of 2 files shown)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -23,6 +23,18 @@ function isActionAllowed(action: Action, forbiddenActionTypes: Array<string>): b
 }
 
 export function createActionInstance(json: any, errors: Array<IValidationError>): Action {
+    if (!json["title"] && json["title"] !== "") {
+        raiseParseError(
+            {
+                error: Enums.ValidationError.ActionTitleCantBeNull,
```
dclaux: I think we can reuse PropertyCantBeNull here instead of defining a new enum?

source/nodejs/adaptivecards/src/enums.ts
```diff
@@ -119,6 +119,7 @@ export class ContainerStyle {
 export enum ValidationError {
     Hint,
     ActionTypeNotAllowed,
+    ActionTitleCantBeNull,
```
dclaux: I think we might not need to add this.

source/nodejs/adaptivecards/src/card-elements.ts
tuanh118: Thanks for pointing that out. I wasn't sure what exact error to raise

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -26,7 +26,7 @@ export function createActionInstance(json: any, errors: Array<IValidationError>)
     if (!json["title"] && json["title"] !== "") {
         raiseParseError(
             {
-                error: Enums.ValidationError.ActionTitleCantBeNull,
+                error: Enums.ValidationError.PropertyCantBeNull,
                 message: "Action title is null"
```
dclaux: Actually, one more change if you dont mind. I think the message should probably more like "Actions should always have a title."

QingpingMeng:
Any reason why title is required for all actions here? This is a breaking change as some selectAction do not need to have title and doesn't functions at all.

PR #1801, opened 2018-08-15
Fix checkboxes and radio buttons labels not being truncated

dclaux:
Also make it possible to click label to select checkboxes and radio buttons
Also fix radio button and checkbox visual alignment in Chrome and Opera


matthidinger:
Dang, I was going to try and review this for you but it's got a massive changeset for some reason :/

dclaux:
Must be because of the merge... It's an old branch I hadn't used in a while. All the changes are in card-elements.ts you can ignore everything else.

Get Outlook for Android<https://aka.ms/ghei36>

dclaux:
Actually you just need to look at the last commit

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -975,6 +980,34 @@ export class TextBlock extends CardElement {
...
+            renderedElement.htmlFor = this.forElementId;
+        }
+
+        return renderedElement;
+    }
+
+    forElementId: string;
+
+    static generateUniqueInputId(): string {
+        let uniqueId = "__ac-input-id" + Label.uniqueIdCounter;
```
matthidinger: I'm not sure how valid this concern is but since IDs need to be unique throughout the entire DOM I wonder if we run the risk of conflicts for any cards that may be delivered via server-side HTML and then appended using the JS lib? in the .NET HTML renderer I did a Guid.NewGuid substring 8 to try reduce the chance of conflict 
https://github.com/Microsoft/AdaptiveCards/blob/5577cd2802fb0fa21f2319c6cb6229bed55a0a0d/source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs#L23

Seems JS doesn't have such a facility but something like this function may work well enough:
https://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript

dclaux:
Changed to use a UUID generator. Question is, is it OK to have a copy of that code [written by a third party](https://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript/21963136#21963136)? The code in question is under MIT license so I think it's OK, but I'm not sure.

khouzam:
@dclaux, it looks like this merged way more than the intended changes. I'm going to revert this change.

PR #1812, opened 2018-08-17
gathers images in actions

nesalang:
gathers images in actions

### DIFF
--- a/source/shared/cpp/ObjectModel/BaseActionElement.cpp
+++ b/source/shared/cpp/ObjectModel/BaseActionElement.cpp
@@ -95,7 +95,12 @@ void BaseActionElement::PopulateKnownPropertiesSet()
          AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::IconUrl)});
 }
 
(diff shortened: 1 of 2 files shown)


source/shared/cpp/ObjectModel/BaseActionElement.cpp
```diff
@@ -95,7 +95,12 @@ void BaseActionElement::PopulateKnownPropertiesSet()
          AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::IconUrl)});
 }
 
-void BaseActionElement::GetResourceInformation(std::vector<RemoteResourceInformation>&)
+void BaseActionElement::GetResourceInformation(std::vector<RemoteResourceInformation>& resourceInfo)
 {
-    return;
+    if (!m_iconUrl.empty()) {
```
paulcam206: the `{` should be on the next line :)

nesalang: you're right!

thanks

---
In reply to: [211026141](https://github.com/Microsoft/AdaptiveCards/pull/1812#discussion_r211026141) [](ancestors = 211026141)

PR #1819, opened 2018-08-22
[Android] Android test app with RecyclerViews

almedina-ms:
I made this app for testing the bugs that may find when rendering adaptive cards in recyclerviews.


source/android/mobilechatapp/src/main/assets/Action.OpenUrl.json
```diff
@@ -0,0 +1,18 @@
...
+		}
+	],
+	"actions": [
+		{
+			"type": "Action.OpenUrl",
+			"title": "Action.OpenUrl",
+			"url": "http://adaptivecards.io"
+		}
+	]
+}
```
khouzam: Can we split these into folders (Scenarios, Tests, Elements) and have a step that copies them from the \Samples location. Having multiple copies is just going to lead to discrepancies in the the tests between this renderer and the other renderers.

PR #1823, opened 2018-08-23
Add a method to get all actions for a card

rcasto:


matthidinger:
Code looks good, @dclaux could you please double check this when you're back

dclaux:
Sorry I can't do a proper review, but we can't expose the items property of Action Collection as is done in this pr. That property should not be exposed directly because it makes it possible to mess with the array and break the renderer logic (actions are parented and setting up the parent is done in addAction) There should already be getActionAt and getActionCount so not sure why this change is needed in the first place.

Get Outlook for Android<https://aka.ms/ghei36>

rcasto:
Yeah you're totally right, you can utilize getActionCount and getActionAt to do this functionality. Was mainly adding as an easy way to get actions for a card to then disable them. Will go ahead and close this pull request.

PR #1826, opened 2018-08-24
[Android] Issues in saving and restoring input form due to recycling of views

rednivrug15:
Inputs are lost due to recycling of views . The change here is that the client can implement an interface which will invoked with any change in the inputs by the library . Also exposing setters so that the client can call these to set the desired values


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/CheckBoxSetInputHandler.java
```diff
@@ -7,41 +7,61 @@
...
         // no need to validate
         ChoiceSetInput choiceSetInput = (ChoiceSetInput) m_baseInputElement;
 
         Vector<String> resultList = new Vector<String>();
         ChoiceInputVector choiceInputVector = choiceSetInput.GetChoices();
-        for (int index = 0;index < m_checkBoxList.size(); index++)
-        {
-            if (m_checkBoxList.get(index).isChecked())
-            {
+        for (int index = 0; index < m_checkBoxList.size(); index++) {
```
almedina-ms: [nit] modify from 
for(){
}
to 
for()
{
}

almedina-ms: also do for if and functions

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/registration/CardRendererRegistration.java
```diff
@@ -234,7 +251,7 @@ else if(tag instanceof AdaptiveCard)
     }
 
     private static CardRendererRegistration s_instance = null;
-
+    private static IInputWatcher m_InputWatcher = null;
```
almedina-ms: static variables are prefixed with s_

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/RenderedAdaptiveCard.java
```diff
@@ -53,6 +53,22 @@ public void registerInputHandler(IInputHandler handler)
         return input;
     }
 
+    public void setInputs(Map <String, String> inputs)
```
almedina-ms: [nit] extra space between "Map" and "<key, val>"

rednivrug15: Resolved

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/registration/CardRendererRegistration.java
rednivrug15: Removed static keyword

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/CheckBoxSetInputHandler.java
rednivrug15: Done

PR #1835, opened 2018-08-28
Add XSD for XML intellisense authoring of adaptive cards using XML/Razor

tomlm:
XML is super useful serialization format as it allows XML based template engines such as Razor to be used to do data binding to cards.

This XSD gives you intellisense while authoring XML cards


matthidinger:
Thanks Tom! I don't suppose there's any reliable way to generate this from the OM or otherwise as we make schema changes along the way?

Issue #1829, opened 2018-08-25
Renderers don't accept 1.1 cards


khouzam:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* .NET HTML
* .NET WPF
* Android
* iOS

# Sample

```json
{
	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
	"type": "AdaptiveCard",
	"version": "1.1",
	"body": [
		{
			"type": "TextBlock",
			"text": "color: default",
			"color": "default"
		}
	]
}
```

# Version of SDK
While the renderers are 1.1, most don't haven't been updated to accept 1.1 cards.

# Issue

Setting the card version to 1.1 fails on current renderers.

PR #1838, opened 2018-08-29
Update renderer versions

carlos-zamora:
Fix for bug #1829 

Updated following renderers to 1.1:
* .NET (WPF & HTML)
* iOS
* Android


khouzam:
We should add a test card that has a version of 1.1 that should be parsed properly and perhaps one that has 1.2 that should have a warning.

carlos-zamora:
-Added json test for versions 1.1, 1.2, and 1.3 
-Renamed AdaptiveCard.Version.json ("Contains" call in AllPayloadTests.cs detected new json tests)
-Removed "EnsureCanRender" and all references
-Replaced "EnsureCanRender" functionality with parse-time enforcement (used to be render-time)
-Updated "KnownSchemaVersion"
-Updated "TestVersionParsing"
-Updated "AllPayloardTests" to catch null cards
-Added failing json tests to "AllPayloadTests"

source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs
```diff
@@ -21,21 +21,25 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
+                // The depth checks that cards within a Action.ShowCard don't require the version
+                if (jObject.Value<string>("version") == "")
+                    throw new AdaptiveSerializationException("Property is required but was found empty: version");
 
-            // If this is the root AdaptiveCard and missing a version we fail parsing. 
-            // The depth checks that cards within a Action.ShowCard don't require the version
-            if (reader.Depth == 0 && card.Version == null)
-            {
-                // TODO: HACK for BF needing to deserialize legacy payloads that did not have a version
-                card.Version = "0.5";
```
paulcam206: Do we no longer need this fallback?

source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs
```diff
@@ -21,21 +21,25 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
-                //throw new AdaptiveSerializationException("Required property 'version' not found on AdaptiveCard");
+                if (new AdaptiveSchemaVersion(jObject.Value<string>("version")) > AdaptiveCard.KnownSchemaVersion)
```
paulcam206: always enclose blocks of code in `{` `}` -- even single-line blocks

paulcam206: (same below)

source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs
```diff
@@ -44,5 +48,30 @@ public override bool CanConvert(Type objectType)
...
             return typeof(AdaptiveCard).GetTypeInfo().IsAssignableFrom(objectType.GetTypeInfo());
         }
+
+        private AdaptiveCard MakeFallbackTextCard(JObject jObject)
+        {
+            // Retrieve fallbackText and language defined by parsed json
+            string language = jObject.Value<string>("language");
+            string fallbackText = jObject.Value<string>("fallbackText");
+
+            if (fallbackText == null || fallbackText == "")
```
paulcam206: Should be using `String.IsNullOrEmpty()` here. ([docs](https://docs.microsoft.com/en-us/dotnet/api/system.string.isnullorempty?view=netframework-4.7.2#System_String_IsNullOrEmpty_System_String_))

source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs
```diff
@@ -44,5 +48,30 @@ public override bool CanConvert(Type objectType)
...
+                fallbackText = "We're sorry, this card couldn't be displayed";
+
+            // Define AdaptiveCard to return
+            AdaptiveCard fallbackTextCard = new AdaptiveCard("1.0");
+            fallbackTextCard.BackgroundImage = null;
+            fallbackTextCard.Lang = language ?? "en";
+            fallbackTextCard.Speak = "";
```
paulcam206: I'm not sure we need to explicitly set this here

source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs
```diff
@@ -44,5 +48,30 @@ public override bool CanConvert(Type objectType)
...
+            fallbackTextCard.Body.Add(new AdaptiveTextBlock
+            {
+                Text = fallbackText
+            });
+
+            // Add relevant warning
+            Warnings.Add(new AdaptiveWarning(7, "Schema version is not supported"));
```
paulcam206: why `7`?

khouzam: We should have a common set of error codes that are shared amongst the platform. But that's beyond this scope. I would define the error codes in an enum and use those values so at least it's explicit and consolidated.

source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs
khouzam: Yes, we need to be consistent across the the different platforms.

samples/Tests/AdaptiveCard.Version1.1.json
```diff
@@ -0,0 +1,13 @@
+{
+    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+    "type": "AdaptiveCard",
+    "version": "1.1",
```
khouzam: Because your changing some test cards and adding new ones, you need to run the UWP TestApp to generate new payload to compare against.

source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs
```diff
@@ -10,6 +10,9 @@ public class AdaptiveCardConverter : JsonConverter, ILogWarnings
     {
         public List<AdaptiveWarning> Warnings { get; set; } = new List<AdaptiveWarning>();
 
+        // TODO: temporary warning code for fallback card. Remove when common set of error codes created and integrated.
+        private enum WarningStatusCode { UnsupportedSchemaVersion = 7 };
```
paulcam206: I'm fine with this for this commit so we don't muddy up the rest of this PR. However, I think as your next change we should change `AdaptiveWarning` to take a `WarningStatusCode` and fix it up. :)

Issue #1828, opened 2018-08-24
Version enforcement in shared model is treating version as double instead of semantic


khouzam:
# Platform

What platform is your issue or question related to?

* UWP
* iOS
* Android

# Issue

Putting a version of 1.10 in a card is accepted by the shared model which should only accept 1.1 and lower cards. A version of 1.2 fails as expected.

PR #1841, opened 2018-08-30
Move to semantic versioning (no more double)

paulcam206:
Fix for #1828, where `1.1` and `1.10` were considered equivalent due to the version number being treated as a `double`. This change mostly impacts renderer versioning -- the version number was being passed by renderers as a parameter to the `AdaptiveCards::Deserialize*()` family of methods. If a caller passed in `std::numeric_limits<double>::max()` as the version, version checking was disabled (only known user is `ShowCardAction`). Since all of the callers hardcode their version information, I've decided that `""` (that is, the empty string) will stand in as the "disable version checking" sentinel value. 

In this change:
* `CMakeLists.txt`
  * Add `SemanticVersion.cpp` to the list
  * Sort list
* Re-SWIG Android files
* `AdaptiveCardRenderer.java`
  * Update version to `1.1`
  * Change version type to `String`
* Add `SemanticVersion.[cpp|h]` to:
  * xcode project
  * AdaptiveCardsSharedModel project
* `ACOAdaptiveCard.mm`
  * Update version to `1.1`
  * Pass version as a string
* Add tests for new `SemanticVersion` class (100% code coverage)
* Update existing tests to pass versions as a string
* Move `<regex>` include to `pch.h`
* Found a few spots where `\t`s were in source files rather than spaces, so fixed those


source/shared/cpp/ObjectModel/SemanticVersion.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+    bool versionValid = false;
+#pragma warning(push)
+#pragma warning(disable:26426)
+    // disable spurious warning (Global initializer calls a non-constexpr function.) -- this isn't global
+    static const std::regex versionMatch(R"regex(^([\d]+)(?:\.([\d]+))?(?:\.([\d]+))?(?:\.([\d]+))?$)regex");
+#pragma warning(pop)
+    std::smatch subMatches;
+    if (std::regex_match(version, subMatches, versionMatch) && subMatches.size() >= 5)
+    {
```
khouzam: Is this necessary?

PR #1844, opened 2018-09-02
Add CSS class-names to FactSets and Fact card-elements

shmuelgutman:
Like in other card-elements, add CSS class names to FactSets and Facts


dclaux:
Thank you!

PR #1847, opened 2018-09-04
Revert "Fix checkboxes and radio buttons labels not being truncated"

khouzam:
Reverts Microsoft/AdaptiveCards#1801

Somehow this fix touched way more stuff than it should have.


khouzam:
Because the change was so big, we removed the bad commit to avoid confusion.

PR #1859, opened 2018-09-05
Media support for .NET WPF

tuanh118:
This change is big so I'm creating a separate PR just for WPF.

Created a media player to support Media element. The small change in `AdaptiveCards.Sample.WPFVisualizer` project is to illustrate how a host can handle raised media element when inline playback is not allowed.

**NOTE**:
- `https` is not supported by WPF, so if a Media element contains all `https` sources, a warning is logged, and clicking on thumbnail would raise an event for host to handle the media


khouzam:
> 				}

Need to run the UWP test app to update the validation assets since you've changed a card.

---
Refers to: samples/Tests/Media.json:18 in 1a5e3cb. [](commit_id = 1a5e3cbdae5d68a9acf929c2e4a60e38bc53c9a0, deletion_comment = False)

source/dotnet/Test/AdaptiveCards.Test/SerializationTests.cs
```diff
@@ -415,40 +415,102 @@ public void ContainerStyle()
...
   ""type"": ""AdaptiveCard"",
   ""version"": ""1.0"",
   ""body"": [
+	{
+	    ""type"": ""Media"",
+	    ""sources"": [
+		    {
+			    ""mimeType"": ""video/mp4"",
+			    ""url"": ""http://adaptivecardsblob.blob.core.windows.net/assets/AdaptiveCardsOverviewVideo.mp4""
+            }
```
khouzam: Please remove tabs

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveCardRenderer.cs
```diff
@@ -165,12 +166,17 @@ public RenderedAdaptiveCard RenderCard(AdaptiveCard card)
...
-            void Callback(object sender, AdaptiveActionEventArgs args)
+            void ActionCallback(object sender, AdaptiveActionEventArgs args)
             {
                 renderCard?.InvokeOnAction(args);
             }
 
-            var context = new AdaptiveRenderContext(Callback, null)
+            void MediaClickCallback(object sender, AdaptiveMediaEventArgs args)
+            {
+                renderCard.InvokeOnMediaClick(args);
```
khouzam: >. [](start = 26, length = 1)

Do you need the null check ? (rendercard?.InvokeOnMediaClick)

Issue #1832, opened 2018-08-25
[.NET] Fails to render card if text style is not understood


khouzam:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* .NET HTML
* .NET WPF

# Example

```json
{
	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
	"type": "AdaptiveCard",
	"version": "1.0",
	"body": [
		{
			"type": "TextBlock",
			"text": "color: default",
			"color": "default",
			"size": "smaller"
		}
	]
}
```
# Expectations

On other platforms invalid values are ignored and the default is used. This should be consistent.
It prevents newer values from being used and sent to older renderers.

PR #1863, opened 2018-09-06
Handle invalid values

carlos-zamora:
Fix for bug #1832 

Catch invalid values with converter and (1) create warning + (2) set to default value.
Test creates card with invalid values on elements that have default value defined. Card then serialized and checked to see if properties were set to default.


source/dotnet/Library/AdaptiveCards/IgnoreDefaultStringEnumConverter.cs
```diff
@@ -18,6 +21,21 @@ public IgnoreDefaultStringEnumConverter(bool camelCaseText) : base(camelCaseText
...
+            try
+            {
+                // Try to read regularly
+                return base.ReadJson(reader, objectType, existingValue, serializer);
+            }
+            catch
+            {
+                // Catch invalid values and replace them with default value
+                // Add warning stating behavior
+                Warnings.Add(new AdaptiveWarning(0, $"Value \"{reader.Value.ToString()}\" could not be converted to \"{typeof(TEnum).ToString()}\", using the default value of \"{defaultValue}\" instead."));
```
khouzam: This is not the right value. It should be defined in the enum.

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
```diff
@@ -225,5 +225,57 @@ public void TestResourceInformation()
...
+            var json = @"{
+	            ""$schema"": ""http://adaptivecards.io/schemas/adaptive-card.json"",
+	            ""type"": ""AdaptiveCard"",
+	            ""version"": ""1.0"",
+	            ""body"": [
+		            {
+			            ""type"": ""Container"",
+			            ""style"": ""asdf"",
+			            ""spacing"": ""asdf"",
+			            ""items"": [
```
khouzam: Please fix tabs

Issue #196, opened 2017-04-04
Media player support


matthidinger:
# Implementation status
- [x] .NET (#1356)
- [x] Android (#1353)
- [x] iOS (#1352)
- [x] JavaScript (#1361) (in PR: #1864)
- [x] UWP (#1354)
- [x] Documentation

## Problem

It is not possible to embed media elements in an Adaptive Card today. Media would initially support **audio** and **video**.

## Scenario Requirements

* Authors need to be able to provide both video and audio. Some bots only send audio snippets.
* Media does NOT auto play.
* **NEW**: Media is NOT downloaded until user clicks play
  * That means in HTML, set `preload="none"`
* Hosts can choose to allow inline playback (or disable interactivity completely as usual).
* By default, without any host changes...
  * Inline playback is enabled
  * Videos/audio are fully functional, hosts don't need to do any work!
* For video...
  * Authors can provide a thumbnail (poster) to be shown
  * If author didn't provide a poster, host's default poster is used (or blank/grey/black if host didn't provide a default poster)
  * If host disabled interactivity, stop here
  * Otherwise, a default play button should be overlayed and centered horizontally/vertically (hosts can override this play button)
  * Clicking the entire thumbnail+play button should...
    * If inline playback allowed...
      * Overlay play button disappears
      * Media controls appear (native OS media controls)
      * Video starts playing
    * Else if inline not allowed...
      * Simply raises an event to the host, host is then responsible for playing the video

* For audio...
  * Authors can provide a thumbnail (poster) to be shown
  * If author didn't provide a poster, host's default poster is used (or blank/grey/black if host didn't provide a default poster)
  * If host disabled interactivity, stop here
  * Otherwise, a default play button should be overlayed and centered horizontally/vertically (hosts can override this play button)
  * Clicking the entire thumbnail+play button should...
    * If inline playback allowed...
      * Overlay play button disappears
      * Media controls appear (native OS media controls)
      * Displayed poster should remain displayed
      * Audio starts playing
    * Else if inline not allowed...
      * Simply raises an event to the host, host is then responsible for playing the audio

* If hosts disabled inline playback, hosts need to know when the media element was tapped, so that they can handle playing the media themselves
* Hosts need to be able to limit the max download size of the http media, so that a 1 GB file doesn't get downloaded, for example.
* Authors can provide AltText for accessibility purposes
* Authors can provide closed captions file

## Asks
* Cortana
* Objects
* Skype (in header, not body)

[...]

PR #1864, opened 2018-09-06
[JS] Media element support

dclaux:
This PR implements https://github.com/Microsoft/AdaptiveCards/issues/196 and https://github.com/Microsoft/AdaptiveCards/issues/1361

Notes:
- **HostConfig.playButton isn't supported**: a better experience is achieved with native styling via CSS
- **HostConfig.maxMediaSize isn't supported**: as far as I can tell, there is no way to retrieve the size of the actual stream for several reasons:
  - I can't know which stream from the sources will be picked by the Media control (since it depends on the codecs installed on the machine)
  - Determining the size of the stream would require making a HEAD request on the stream's URL which would generate a cross-domain error anyway

Here is what things look like in Chrome (note that all browsers have their own UI for controls).
image

Notes:
- The default Play button is somewhat customizable in CSS and is designed to mimic Chrome's media player button, which provides for a smooth transition between the poster and the actual player.
- When the is no poster, the background is also customizable in CSS.

Once video is playing (controls disappear once mouse is idle or moves out of the player's box):
image

Audio player works just the same. Here is what it looks like while playing in Chrome:
image

When inline playback isn't allowed, the host application can handle the static **Media.onPlay** event.


andrewleader:
Great screenshots, thanks David!

@matthidinger I kinda agree that it's okay the HostConfig playButton isn't supported... if hosts are customizing, they probably would want to use CSS rather than an image which doesn't scale well.

@RebeccaAnne on UWP were you able to implement the HostConfig maxMediaSize? I'm curious if anyone was able to implement that. If not, it should just be removed.

matthidinger:
Looks good, thanks David!

tuanh118:
I think for an audio with poster, it should display the static poster image while the audio is playing.

dclaux:
If that's what we want then I'd say that playback controls should always appear below the poster for Audio, with no default play button overlay. The `<audio>` tag can be made to only load metadata or no data at all until the play button is pressed.

Issue #1845, opened 2018-09-03
JSON content not properly rendered on iOS devices.


shriyamgupta:
For the following card content the JSON strings are rendered in HTML encoding rather than JSON strings on iOS devices for MS Teams App

Card Content file is attached below.
[CardContent.txt](https://github.com/Microsoft/AdaptiveCards/files/2344351/CardContent.txt)

<img width="497" alt="screen shot 2018-09-03 at 11 10 59 am" src="https://user-images.githubusercontent.com/42666174/44969502-1f856f00-af6b-11e8-8268-c1a986670cfa.png">
.

<img width="505" alt="screen shot 2018-09-03 at 11 11 18 am" src="https://user-images.githubusercontent.com/42666174/44969485-0ed4f900-af6b-11e8-910d-e3e8cb19f7bd.png">

nesalang:
@shriyamgupta, is the screenshot at the bottom what the renderer should render?
Sample app's textBlock renderer will ignore all html tags as shown in the first screenshot.
Team's app use custom textBlock renderer that renders html tags.

rednivrug15:
@nesalang the below text block just contains JSON 
`				"text": "{\"task\":{\"type\":\"continue\",\"value\":{\"title\":\"Custom Form\",\"height\":430,\"width\":510,\"fallbackUrl\":\"https:\/\/taskmoduletest.azurewebsites.net\/customform\",\"url\":\"https:\/\/taskmoduletest.azurewebsites.net\/customform\"}}}",
`

It should render as plain text .It renders correctly on Android and Desktop .

shriyamgupta:
> @shriyamgupta, is the screenshot at the bottom what the renderer should render?
> Sample app's textBlock renderer will ignore all html tags as shown in the first screenshot.
> Team's app use custom textBlock renderer that renders html tags.
Sorry for the late reply but we were not using the custom label renderers in Teams iOS app for adaptive cards.

PR #1865, opened 2018-09-06
allows escaped chars to be renderered as html text

nesalang:
fixes issue #1845.

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.mm
@@ -308,7 +308,7 @@ typedef UIImage* (^ImageLoadBlock)(NSURL *url);
             NSDictionary *data = nil;
 
             // use Apple's html rendering only if the string has markdowns
-            if(markDownParser->HasHtmlTags()) {
+            if(markDownParser->HasHtmlTags() || markDownParser->IsEscaped()) {
                 NSString *fontFamilyName = nil;
                 if(!self->_hostConfig.fontFamilyNames){
                     fontFamilyName = @"'-apple-system',  'HelveticaNeue'";
--- a/source/shared/cpp/ObjectModel/MarkDownParser.cpp
+++ b/source/shared/cpp/ObjectModel/MarkDownParser.cpp
@@ -36,7 +36,7 @@ bool MarkDownParser::HasHtmlTags()
     return m_hasHTMLTag;
 }
 
-bool MarkDownParser::IsEscaped()
+bool MarkDownParser::IsEscaped() const
 {
     return m_isEscaped;
 }
@@ -56,7 +56,7 @@ void MarkDownParser::ParseBlock()
 std::string MarkDownParser::EscapeText()
 {
     std::string escaped;
-    unsigned int unescapedCounts = 0;
+    unsigned int nonEscapedCounts = 0;
     
     for (std::string::size_type i = 0; i < m_text.length(); i++)
     {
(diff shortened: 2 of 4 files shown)


source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -69,9 +76,15 @@ std::string MarkDownParser::EscapeText()
             break;
         default:
             escaped += m_text.at(i);
+            unescapedCounts++;
             break;
         }
     }
+    
+    if(!m_isEscaped && unescapedCounts != m_text.length())
```
paulcam206: should be a space between if and the opening `(`

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -35,6 +35,12 @@ bool MarkDownParser::HasHtmlTags()
 {
     return m_hasHTMLTag;
 }
+
+bool MarkDownParser::IsEscaped()
```
paulcam206: this method should be `const`

source/shared/cpp/ObjectModel/MarkDownParser.cpp
```diff
@@ -50,7 +56,8 @@ void MarkDownParser::ParseBlock()
 std::string MarkDownParser::EscapeText()
 {
     std::string escaped;
-
+    unsigned int unescapedCounts = 0;
```
paulcam206: I think something like `nonEscapedCounts` would be a better variable name -- "unescaped" sounds like an action has occurred. :)

source/shared/cpp/ObjectModel/MarkDownParser.h
```diff
@@ -18,12 +18,15 @@ class MarkDownParser
     std::string TransformToHtml();
 
     bool HasHtmlTags();
+
```
paulcam206: don't need this extra newline

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.mm
```diff
@@ -308,7 +308,7 @@ - (void)processTextConcurrently:(std::shared_ptr<BaseCardElement> const &)textEl
             NSDictionary *data = nil;
 
             // use Apple's html rendering only if the string has markdowns
-            if(markDownParser->HasHtmlTags()) {
+            if(markDownParser->HasHtmlTags() || markDownParser->IsEscaped()) {
```
paulcam206: do we know if this behavior shows in other platforms?

source/shared/cpp/ObjectModel/MarkDownParser.cpp
nesalang: thanks :-)

---
In reply to: [215793995](https://github.com/Microsoft/AdaptiveCards/pull/1865#discussion_r215793995) [](ancestors = 215793995)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.mm
nesalang: the issuer mentioned that he or she tested it also on Desktop and Android, and they don't reproduce, but i'm not certain which version of Destop sample app they are referring to.

---
In reply to: [215794246](https://github.com/Microsoft/AdaptiveCards/pull/1865#discussion_r215794246) [](ancestors = 215794246)

source/shared/cpp/ObjectModel/MarkDownParser.h
nesalang: thanks; it's Xcode doing :-)

---
In reply to: [215794155](https://github.com/Microsoft/AdaptiveCards/pull/1865#discussion_r215794155) [](ancestors = 215794155)

PR #1869, opened 2018-09-07
[JS] Fix button titles not being truncated properly anymore

dclaux:
This regression was introduced when support for icons was added. With this fix, buttons are now properly truncated again:
image

This fix also adds a new parameter in HostConfig to allow button titles to wrap even when they have no icons or when the icon is displayed on the left of the title. This allows for better scaling in narrower cards:
image


matthidinger:
That seems like a useful host config option for sure. Have we considered perhaps maybe it adjustable with something like `titleMaxLines` or similar? So to achieve the result in the screenshot would be `titleMaxLines: 2`

dclaux:
@matthidinger my opinion on titleMaxLines:
- The reason I allowed for the title to not wrap is so that it wouldn't be truncated. Adding titleMaxLines would apply truncation. So it feels a little like it would defeat the purpose
- It's hard to implement in HTML. I had already made the same remarks when we did maxLines for TextBlock. It requires two passes because you have to dynamically retrieved the computed line height after the card is fully laid out, then calculate the total height the title should have based on that. For textBlock the saving grace is we control the line height in code.


@andrewleader had suggested at some point that we add a HostConfig setting to control button text size; if we did that, titleMaxLines would be easy to do.

matthidinger:
Gotcha, ok something we can revisit later

PR #1872, opened 2018-09-08
- enforces stricter parsing 

nesalang:
- enforces stricter parsing 
- generates warnings instead of errors


source/shared/cpp/ObjectModel/Util.cpp
```diff
@@ -75,41 +75,127 @@ std::string ValidateColor(const std::string& backgroundColor,
...
-        parsedDimension = 0;
+        unsigned int cnts = 1;
+        for (cnts; cnts <= unit.length(); cnts++)
+        { 
+            unsigned int offset = cnts;
+            if (*(requestedDimension.end() - offset) != *(unit.end() - offset))
+            {
+                warnings.emplace_back(std::make_shared<AdaptiveCardParseWarning>(
+                        AdaptiveSharedNamespace::WarningStatusCode::InvalidDimensionSpecified,
+                        "expected input arugment to be specified as \\d+(.\\d+)?px with no spaces, but received " + requestedDimension));
```
RebeccaAnne: >"expected input arugment to be specified as \\d+(.\\d+)?px with no spaces, but received " [](start = 24, length = 89)

Consider defining this string as a constant

source/shared/cpp/ObjectModel/Util.cpp
```diff
@@ -75,41 +75,127 @@ std::string ValidateColor(const std::string& backgroundColor,
...
+    }
+
+    if (requestedDimension.length() >= unit.length()) 
     {
-        parsedDimension = 0;
+        unsigned int cnts = 1;
+        for (cnts; cnts <= unit.length(); cnts++)
+        { 
+            unsigned int offset = cnts;
+            if (*(requestedDimension.end() - offset) != *(unit.end() - offset))
```
RebeccaAnne: Please add a comment here to document what this is checking.

source/shared/cpp/ObjectModel/Util.cpp
```diff
@@ -75,41 +75,127 @@ std::string ValidateColor(const std::string& backgroundColor,
 }
 
 void ValidateUserInputForDimensionWithUnit(const std::string &unit, const std::string &requestedDimension,
-    int &parsedDimension)
+    int &parsedDimension, std::vector<std::shared_ptr<AdaptiveCardParseWarning>>& warnings)
 {
-    if (requestedDimension.empty())
+    int parsedVal = 0;
+    parsedDimension = 0;
```
RebeccaAnne: Do the renderers ignore the pixel width if it's 0?

nesalang: yes, for column and image renderer will ignore 0 val when both width & height are set to zeroes

---
In reply to: [216388662](https://github.com/Microsoft/AdaptiveCards/pull/1872#discussion_r216388662) [](ancestors = 216388662)

source/shared/cpp/ObjectModel/Util.cpp
```diff
@@ -76,43 +76,116 @@ std::string ValidateColor(const std::string& backgroundColor,
...
+                    AdaptiveSharedNamespace::WarningStatusCode::InvalidDimensionSpecified,
+                    warningMessage + requestedDimension)); 
+            parsedVal = 0;
         }
-        catch (const std::out_of_range &e)
+        parsedDimension = parsedVal;
+    }
+    catch (const std::invalid_argument &e)
+    {
+        (void)e;
```
paulcam206: if you don't need `e`, it should probably be left out of the `catch` conditional (i.e. `catch (const std::invalid_argument &)`)

(same below)

source/shared/cpp/ObjectModel/Util.cpp
```diff
@@ -76,43 +76,116 @@ std::string ValidateColor(const std::string& backgroundColor,
 }
 
 void ValidateUserInputForDimensionWithUnit(const std::string &unit, const std::string &requestedDimension,
```
paulcam206: any reason we wouldn't want to use a regular expression here? seems like it would be simpler...

source/shared/cpp/ObjectModel/Util.cpp
nesalang: thanks!

---
In reply to: [217200327](https://github.com/Microsoft/AdaptiveCards/pull/1872#discussion_r217200327) [](ancestors = 217200327)

source/shared/cpp/ObjectModel/Util.cpp
nesalang: i had the same thought after I finished it :-)

---
In reply to: [217201334](https://github.com/Microsoft/AdaptiveCards/pull/1872#discussion_r217201334) [](ancestors = 217201334)

nesalang: changed it to use regex

---
In reply to: [217207111](https://github.com/Microsoft/AdaptiveCards/pull/1872#discussion_r217207111) [](ancestors = 217207111,217201334)

source/shared/cpp/ObjectModel/Util.cpp
```diff
@@ -79,83 +79,38 @@ void ValidateUserInputForDimensionWithUnit(const std::string &unit, const std::s
...
-                warningMessage + requestedDimension)); 
-        return;
-    }
+    std::string stringPattern = "^([1-9]+\\d*)(\\.\\d+)?";
+    stringPattern += ("(" + unit + ")$");
+    std::regex pattern(stringPattern);
+    std::smatch matches;
 
-    try
+    if(std::regex_search(requestedDimension, matches, pattern))
```
paulcam206: nitpick - need a space after the `if` :)

source/shared/cpp/ObjectModel/Util.cpp
```diff
@@ -76,43 +76,69 @@ std::string ValidateColor(const std::string& backgroundColor,
...
+    }
+    else 
+    {
+        warnings.emplace_back(std::make_shared<AdaptiveCardParseWarning>(
+                AdaptiveSharedNamespace::WarningStatusCode::InvalidDimensionSpecified,
+                warningMessage + requestedDimension)); 
+    }
+}
+
+bool ShouldParseForExplicitDimension(const std::string &input)
```
paulcam206: can't the logic in this function be replaced with a regex as well?

nesalang: it would have been great fit if we were looking for specific pattern, but we are checking if the string contains specific char combinations. we are doing this, so we can produce relavant warning messages. if user misspelled "auto", we don't want to say "it doesn't contain px". 

---
In reply to: [217520663](https://github.com/Microsoft/AdaptiveCards/pull/1872#discussion_r217520663) [](ancestors = 217520663)

Issue #1848, opened 2018-09-04
.NET Date formatting not enforcing rules properly


matthidinger:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* .NET HTML
* .NET WPF

Using payload `DateTimeTestTranslation.json` there are a handful of **valid** date formats, and **invalid** date formats. 1 of the valid formats is failing, and 3 of the invalid formats are not failing when they are expected to.

To repro:

```shell
PS C:\Dev\Projects\AdaptiveCards\source\dotnet\Samples\AdaptiveCards.Sample.Html> dotnet run ../../../../samples -r -i -o ./samples.html
PS C:\Dev\Projects\AdaptiveCards\source\dotnet\Samples\AdaptiveCards.Sample.Html> start .\samples.html
```

![image](https://user-images.githubusercontent.com/1432195/45058589-f8539e00-b04d-11e8-9787-e0f850759bc0.png)

PR #1874, opened 2018-09-10
Update Date Formatting 

carlos-zamora:
Fix for bug #1848 
* DateTime is now case-sensitive
* Added COMPACT implementation
* Fixed "Missing seconds" portion of test


samples/Tests/DateTimeTestTranslation.json
```diff
@@ -84,7 +84,7 @@
                 },                                
                 {
                     "type": "TextBlock",
-                    "text": "Missing seconds {{DATE(2017-10-27T22:27Z)}} {{TIME(2017-10-27T22:27:00Z)}}",
+                    "text": "Missing seconds {{DATE(2017-10-27T22:27Z)}} {{TIME(2017-10-27T22:27Z)}}",
```
matthidinger: This is actually supposed to be missing the seconds. That payload has 2 groups of formats, the first group (VALID) should all work as intended. The second group (INVALID) are supposed to fail parsing and should return the literal string without any date-formatting logic

carlos-zamora: Yeah. This is in the INVALID group. `{{DATE(…)}}` is missing the seconds but `{{TIME(…)}}` isn't. So, the expected behavior is `output: Missing seconds {{DATE(…)}} 10:27 PM` when it should be `output: Missing seconds {{DATE(…)}} {{TIME(…)}}`.

Removing the seconds from the second part helps test how the `TIME` function handles the same scenario.

khouzam: You'll need to update the UWP TestApp expected results since this will change the rendering of the card as well as the serialization.

---
In reply to: [216468846](https://github.com/Microsoft/AdaptiveCards/pull/1874#discussion_r216468846) [](ancestors = 216468846)

source/dotnet/Library/AdaptiveCards/Rendering/RendererUtilities.cs
```diff
@@ -13,8 +13,8 @@ namespace AdaptiveCards.Rendering
     public static class RendererUtilities
     {
         private static readonly Regex TextFunctionRegex =
-            new Regex(@"\{\{(?<func>DATE|TIME){1}\((?<date>.+?){1}(?:,\s*(?<hint>Short|Long){1}\s*)??\)\}\}",
-                RegexOptions.ExplicitCapture | RegexOptions.IgnoreCase);
+            new Regex(@"\{\{(?<func>DATE|TIME){1}\((?<date>\d\d\d\d-\d\d-\d\dT\d\d:\d\d:\d\d[Z+-].*?){1}(?:,\s*(?<hint>SHORT|LONG|COMPACT){1}\s*)??\)\}\}",
```
paulcam206: This regex doesn't look quite right to me -- for the date, shouldn't we have something like this?
```regex
(?<date>[\d]{4}-[\d]{2}-[\d]{2}T[\d]{2}:[\d]{2}:[\d]{2}[Z+-].*?){1}
```

carlos-zamora: Both are the same exact regex. Not sure which is more clear or preferred though.

paulcam206: I think the counted character class version is a bit more intuitive to a reader -- you don't have to count the digits :)

PR #1877, opened 2018-09-10
Update nodejs's notion of max version to 1.1

paulcam206:


dclaux:
I was actually going to work on this. There is more to do; if the renderer receives a v1.0 card, it needs to not parse properties/elements/actions introduced in subsequent versions. Similarly, toJSON needs to only serialize the properties/elements/actions that comprise the target payload version.

So essentially, **parse** and **toJSON** need a version parameter, and element/action registrations need to be associated with a particular version.

One of the reasons these changes are required is that the JS renderer is used for our tools (visualizer and designer.)

paulcam206:
@dclaux - We're actually **not** supposed to disable parsing of new elements in older card versions. @matthidinger or @khouzam should be able to provide more context.

dclaux:
We don't have a choice because the JS renderer is used in our tools, and the tools need to be able to handle different Adaptive versions.

paulcam206:
Current PoR for other platforms like UWP is that a 1.0 card with a media element will happy render and allow media playback.

dclaux:
The other renderers are not used by our tools. The goal for the designer is to have a version selector so users can create v1.0, v1.1 etc cards easily. IMO there are two main solutions to achieve that:
- Either any app (including our tools) is supposed to dynamically load the appropriate version of the renderer corresponding to the version of the card it is handling
- Or the renderer is able to properly handle all versions

I believe the second option is best. As for the POR, I was not consulted as far as I can remember, so consider this my input and me prompting to have the conversation about the design.

paulcam206:
Alright -- I'll leave it to @matthidinger and @khouzam to chime in. I'm not familiar with the designer's specific requirements.

khouzam:
Hi @dclaux,

I understand that the designer might need be enforce some slightly more stricter rules when someone tries to produce a card, but for the renderers that simply display cards, which most of the clients are, this is the best current fallback story that we have.

If I'm a card author, we've introduced a new property in 1.1 that is verticalContentAlignment. This enables me to make my card look better, so I want to take advantage of it. The problem that I'm facing now, is that some of the endpoints that might consume my card are not on 1.1 yet, but still on 1.0. I can make the conscious decision to use the property and have 1.0 renderers will ignore it, but newer renderers will use it and the content will look better. If 1.1 renderers ignore the new property if the card version is set to 1.0, then I'm afraid that most cards will not take advantage of any of the new features unless they are being shared in a closed environment where they know that all the endpoints have a specific version.

There will be a lot of cards that can use new features as enhancements but that still convey the proper author intent when some properties or elements are ignored by older renderers. If indeed the card relies on 1.1 features heavily and ignoring the new elements doesn't give a proper card, the card author would set the version of the card to 1.1 and rely on the fallback text for older renderers.

Until we design a per-element fallback that we can use to have some sore of conditional behavior for new features, I think that this is the best alternative to enable card authors to use the newer features without having a binary decision of feature vs breadth of endpoints.

PR #1889, opened 2018-09-11
Speak and Language for Fallback Card

carlos-zamora:
# Speak for Fallback Card
Added speak to fallback card in .NET and Shared Model. Logic shown in table below:

Text pre-defined? | Speak pre-defined? | Result
------------------- | --------------------- | ------
N | N | Text = Speak = default
N | Y | Text = default, Speak = **SPEAK**'s defined val
Y | N | Text = Speak = **TEXT**'s defined val
Y | Y | Each to their own pre-defined val

# Lang for Fallback Card
Also made fallback card have language.
* **SharedModel**: if lang not defined, pass on as empty/not-defined and have renderer deal with it
* **.NET**: at parse time, if lang not defined, set to `CultureInfo.CurrentCulture.TwoLetterISOLanguageName`

This one seems a little odd, so let me know what your thoughts are.

### DIFF
--- a/source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
+++ b/source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
@@ -232,7 +232,8 @@ Json::Value AdaptiveCard::SerializeToJsonValue() const
 #ifdef __ANDROID__
 std::shared_ptr<AdaptiveCard> AdaptiveCard::MakeFallbackTextCard(
     const std::string& fallbackText,
-    const std::string& language) throw(AdaptiveSharedNamespace::AdaptiveCardParseException)
+    const std::string& language,
+    const std::string& speak) throw(AdaptiveSharedNamespace::AdaptiveCardParseException)
 #else
 std::shared_ptr<AdaptiveCard> AdaptiveCard::MakeFallbackTextCard(
     const std::string& fallbackText,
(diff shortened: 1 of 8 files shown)


source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
```diff
@@ -231,10 +236,11 @@ std::shared_ptr<AdaptiveCard> AdaptiveCard::MakeFallbackTextCard(
 #else
```
khouzam: Won't this break Android? Doesn't it also need the speak parameter?

khouzam:
>     }

Do we want to validate that speak is also properly set?

---
Refers to: source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/EverythingBagel.cpp:435 in 8c3bfee. [](commit_id = 8c3bfee202088c38be1a6e217bc301371e5dffc8, deletion_comment = False)

carlos-zamora:
In EverythingBagel, speak is now checked to see if it's properly set. There is one minor difference between the SharedModel and .NET `MakeFallbackCard` though. In the SharedModel, it only handles the assignment. So, when you call the function, whatever gets passed is explicitly set. The control flow for speak is handled before it's called.

In .NET, the function does that extra stuff whenever it's called (i.e.: speak control flow, lang setting).

So, when EverythingBagel uses `MakeFallbackCard`, the speak control flow never occurs. On the other hand, `MakeFallbackCard` is only called when the version is invalid so, in practice (at least currently), I think there shouldn't be a difference.

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
khouzam: I think you still need to run swig for this change.

---
In reply to: [216856776](https://github.com/Microsoft/AdaptiveCards/pull/1889#discussion_r216856776) [](ancestors = 216856776)

PR #1895, opened 2018-09-12
Add support for Media element in designer

dclaux:
image


andrewleader:
From the screenshot, looks awesome to me! I'm not able to review the code, so if you need a code review approval, Matt or Tuan Anh would have to do that.

Issue #1899, opened 2018-09-12
.NET DateTime not translated to AdaptiveCard language


carlos-zamora:
# Platform

* .NET HTML
* .NET WPF

# Issue

`DateTimeTranslationTest.json` won't translate the date to the language of the AdaptiveCard. Instead, it uses the language of the system running it (`CultureInfo.CurrentCultureSetting()` is the default interpretation when `ToString()`-ing a DateTime). 

The error occurs when DATE is in `SHORT` or `LONG` format.

# Proposed Fix

The `AdaptiveTextBlock` should have a `language` property that is the same as that of the `AdaptiveCard` it is in. If the language is not defined, then use the `CultureInfo.CurrentCultureSetting()` as is done now. 

# Current Output
![image](https://user-images.githubusercontent.com/11050425/45457315-3cc0e880-b6a3-11e8-9cd4-f017e1f19076.png)

# Expected Output
![image](https://user-images.githubusercontent.com/11050425/45457317-42b6c980-b6a3-11e8-8355-6b78fcf7ff1d.png)

PR #1920, opened 2018-09-18
.NET Language passed through context + SharedModel/.NET Invalid Language Warning

carlos-zamora:
Bugfix for #1899.

* **.NET**: Added PropagateLang to pass `lang` throughout AdaptiveCard contents
* **.NET**: `lang` is checked if valid at parsetime. If invalid, warning is thrown (same warning in Javascript)
* **SharedModel**: `language` is checked if valid at parsetime. If invalid, warning is thrown (same warning in Javascript).

**NOTE**: *.NET* validates `lang` and propagates the corrected `lang` at __parsetime__. *SharedModel* only checks if `language` is valid and attaches warning if applicable. The "correct" language is used/changed at __rendertime__.


khouzam:
Please update master and merge it into your branch to fix the build break.

carlos-zamora:
@khouzam Oops. A few mistakes on my end. I'll let you know when it's actually ready. Sorry!

source/dotnet/Library/AdaptiveCards/AdaptiveColumnSet.cs
```diff
@@ -38,5 +38,13 @@ public class AdaptiveColumnSet : AdaptiveElement
 #endif
         [DefaultValue(null)]
         public AdaptiveAction SelectAction { get; set; }
+
+        public void SetLang(string value)
+        {
+            foreach(AdaptiveColumn column in Columns)
```
paulcam206: need a space here

carlos-zamora: So like this?:
```
public void SetLang (string value) 
{ 
    foreach (AdaptiveColumn column in Columns)
```

source/dotnet/Library/AdaptiveCards/Util.cs
```diff
@@ -0,0 +1,45 @@
...
+    static class Util
+    {
+        // iterates through each element of the body and sets the language on non-null ColumnSets, Containers, and TextBlocks
+        internal static void PropagateLang(string lang, List<AdaptiveElement> body)
+        {
+            foreach (AdaptiveElement element in body)
+            {
+                string type = element.Type;
+
+                if (type.Equals(AdaptiveColumnSet.TypeName))
```
paulcam206: >if (type.Equals(AdaptiveColumnSet.TypeName)) [](start = 16, length = 44)

this feels like a `switch` statement

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
```diff
@@ -105,6 +105,16 @@ std::shared_ptr<ParseResult> AdaptiveCard::Deserialize(
     std::string language = ParseUtil::GetString(json, AdaptiveCardSchemaKey::Language);
     std::string speak = ParseUtil::GetString(json, AdaptiveCardSchemaKey::Speak);
 
+    // check if language is valid
+    try 
+    {
+        std::locale(language.c_str());
+    }
+    catch (...)
```
paulcam206: >(...) [](start = 10, length = 5)

can we catch a more specific exception here?

carlos-zamora:
# Valid Languages
Valid languages include the following:
* `null`
* empty string (`""`)
* 2 letters
* 3 letters

Any strings with invalid characters result in warning. Any strings composed of letters only but that are invalid do not throw warning. This matches Javascript behavior.

# .NET: Language ~~Propagation~~ Handling
Language propagation got removed from the PR. Now the language is just passed into the context and used by the textblock... waaay easier.

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
carlos-zamora: replaced by `std::runtime_error`. Still super not specific but that's the best I could find.

source/dotnet/Library/AdaptiveCards/Util.cs
carlos-zamora: Removed from PR.

PR #1923, opened 2018-09-18
Designer update

dclaux:
Adds support for `backgroundImage` on **AdaptiveCard**, **Container** and **Column**
Adds support for `style` on **Container** and **Column**
Adds support for `lang`, `fallbackText` and `speak` on **AdaptiveCard**
Adds missing support for weighted column widths
Remove irrelevant properties from property sheet for some element types


dclaux:
@andrewleader I have just published the changes to http://acdesignerstaging.azurewebsites.net/

dclaux:
Good catch - fixed, staging site updated.

PR #1944, opened 2018-09-22
Jwoo/ios markdown hyperlink

nesalang:
- made markdown hyperlink clickable.
- fixed background padding
- added performance tests


source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizerTests/ADCIOSVisualizerTests.mm
```diff
@@ -108,10 +121,231 @@ - (void)testChoiceSetInputCanGatherDefaultValues {
     }
 }
 
-- (void)testPerformanceExample {
+- (void)testPerformanceOnComplicatedCards {
```
bviglietta: >testPerformanceOnComplicatedCards [](start = 8, length = 33)

Since all of these performance test methods are identical except for the list of payload names, it may be worthwhile to factor the common code into its own method that takes the payload list as an argument.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTextBlockRenderer.mm
```diff
@@ -78,25 +84,49 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
...
+    } else if (txtBlck->GetHorizontalAlignment() == HorizontalAlignment::Center) {
+        horizontalAlignment = NSLayoutAttributeCenterX;
+    }
+
+    [NSLayoutConstraint constraintWithItem:lab attribute:horizontalAlignment relatedBy:NSLayoutRelationEqual toItem:wrappingview attribute:horizontalAlignment multiplier:1.0 constant:0].active = YES;
+    [NSLayoutConstraint constraintWithItem:lab attribute:NSLayoutAttributeBottom relatedBy:NSLayoutRelationEqual toItem:wrappingview attribute:NSLayoutAttributeBottom multiplier:1.0 constant:0].active = YES;
+    [NSLayoutConstraint constraintWithItem:lab attribute:NSLayoutAttributeTop relatedBy:NSLayoutRelationEqual toItem:wrappingview attribute:NSLayoutAttributeTop multiplier:1.0 constant:0].active = YES;
+
+    lab.textContainer.maximumNumberOfLines = int(txtBlck->GetMaxLines());
+    if(!lab.textContainer.maximumNumberOfLines and !txtBlck->GetWrap()){
```
bviglietta: >and [](start = 47, length = 3)

Nit: really prefer to use `&&` operator. Spelling it `and` is legal but so uncommon.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRUIImageView.mm
```diff
@@ -6,10 +6,28 @@
...
+    }
+    return self;
+}
+
+- (instancetype)initWithFrame:(CGRect)frame {
+    self = [super initWithFrame:frame];
+    if(self) {
+        self.tag = eACRUIImageTag;
+    }
+    self.desiredSize = frame.size;
```
bviglietta: I don't see why one of these statements is inside the `if` block and the other isn't.

source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizerTests/ADCIOSVisualizerTests.mm
nesalang: refactoring the measureBlock has one con. Xcode overlays test results and stats on the mesaureBlock. If mesaureBlock is refactored, then only the last run test stats is shown. I will investigate how to fix that, and update the test in the future.

---
In reply to: [219909885](https://github.com/Microsoft/AdaptiveCards/pull/1944#discussion_r219909885) [](ancestors = 219909885)

Issue #1851, opened 2018-09-05
Input.ChoiceSet sends back ; separated values in Mobile renderers ( iOS and Android )


rednivrug15:
# Platform

* Android
* iOS

# Host

Teams

# Version of SDK

1.0.0

# Issue

When an Action.Submit button is pressed Mobile SDK's ( Android and iOS ) return back ; separated values while the HTML renderer return back , separated values . This happens when multiple choice selection is allowed in Input.ChoiceSet . There should be a common experience for all the platforms .

matthidinger:
@khouzam looks like we missed this for the 1.0 release. Per #652 they are indeed supposed to be comma separated. This fix to align all the renderers will be a breaking change so let's make sure we include that in the release notes for 1.1.

PR #1948, opened 2018-09-24
Fix choiceset separators on android and ios

paulcam206:
Fix for #1851.


paulcam206:
Note that this still needs to be tested on iOS.

paulcam206:
@carlos-zamora verified iOS works correctly 🤘

PR #1953, opened 2018-09-25
Allow for empty body in iOS and Android

khouzam:
The android and iOS renderers had a check that the body of the card was not empty.

This was inconsistent with the other renderers. Removing the check.
Adding a test.


samples/Tests/Action.NoBody.json
```diff
@@ -0,0 +1,40 @@
+{
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
```
paulcam206: >	 [](start = 0, length = 1)

do we care about `\t` in json files?

khouzam: Half of them are \t the other half are spaces. So not currently.

Issue #1837, opened 2018-08-29
Ordered list not supported on android and are converted into unordered list


rednivrug15:
# Platform

What platform is your issue or question related to? (Delete other platforms).
* Android

# Author or host

Host ( Teams )

# Version of SDK

1.0

# Issue

Ordered list not supported on android and are converted into unordered list

Sample json 
```json
{
 	"type": "AdaptiveCard",
 	"body": [{
 		"text": "adaptiveCardMarkdown",
 		"type": "TextBlock"
 	}, {
 		"text": "**bold**",
 		"type": "TextBlock"
 	}, {
 		"text": "_Italic_",
 		"type": "TextBlock"
 	}, {
 		"text": "[hyperlink](https://microsoft.com)",
 		"type": "TextBlock"
 	}, {
 		"text": "[unsafe link](javascript:alert(''))",
 		"type": "TextBlock"
 	}, {
 		"text": "![image](https://cdn2.iconfinder.com/data/icons/social-icons-33/128/Trello-128.png)",
 		"type": "TextBlock"
 	}, {
 		"text": "`code`",
 		"type": "TextBlock"
 	}, {
 		"text": "```code block```",
 		"type": "TextBlock"
 	}, {
 		"text": "- Item 1\r- Item 2\r- Item 3",
 		"type": "TextBlock"
 	}, {
 		"text": "1. Green\r2. Orange\r3. Blue",
 		"type": "TextBlock"
 	}, {
 		"text": "<a href='https://abc.com'>html should not allow</a>",
 		"type": "TextBlock"
 	}, {
 		"facts": [{
 			"title": "**bold**",
 			"value": "_Italic_"
 		}],
 		"type": "FactSet"
 	}, {
 		"choices": [{
 			"title": "**bold**  _Italic_",
 			"value": "1"
 		}],
 		"style": "expanded",
 		"value": "1",
 		"id": "myColor2",
 		"type": "Input.ChoiceSet"
 	}, {
 		"title": "**bold**  _Italic_",
 		"valueOff": "false",
 		"valueOn": "true",
 		"value": "true",
 		"id": "acceptTerms",
 		"type": "Input.Toggle"
 	}],
 	"version": "1.0",
 	"fallbackText": "{\"cardId\":\"ec3f0ff6-0452-4206-ae08-2163682f2def\"}"
 }
```

paulcam206:
should be fixed 👍

PR #1956, opened 2018-09-25
Add markdown ordered list support to android

paulcam206:
Fix for #1837.

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
@@ -134,11 +134,37 @@ public class TextBlockRenderer extends BaseCardElementRenderer
 
     // Class to replace ul and li tags
     public class UlTagHandler implements Html.TagHandler{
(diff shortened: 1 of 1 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
```diff
@@ -271,6 +297,10 @@ else if ((textBlock = TextBlock.dynamic_cast(baseCardElement)) == null)
 
         MarkDownParser markDownParser = new MarkDownParser(textWithFormattedDates);
         String textString = markDownParser.TransformToHtml();
+
+        // preprocess string to change <li> to <foo> so we get a chance to handle them
```
khouzam: ><foo> [](start = 47, length = 5)

listItem?

paulcam206: better than `someRandomThing`...

paulcam206: should be sorted 🤘

PR #1958, opened 2018-09-26
choicesetresouceresolvers

an888ha:


an888ha:
1.Choice Set renderer will ask acrChoiceSetResourceDelegate for choice set images in different states.
2. Tint Color for toggle exposed so that anyone can set tint color , if he/she over-rides the toggle input renderer.

source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ADCChoiceSetResourceResolver.m
```diff
@@ -0,0 +1,30 @@
...
+//  Created by Ankit Maurya on 9/21/18.
+//  Copyright © 2018 Microsoft. All rights reserved.
+//
+
+#import "ADCChoiceSetResourceResolver.h"
+
+@implementation ADCChoiceSetResourceResolver
+
+- (UIImage*) getRadioButtonForSelectedState{
+    return [UIImage imageNamed:@"checked.png" inBundle:[NSBundle bundleWithIdentifier:@"MSFT.AdaptiveCards"] compatibleWithTraitCollection:nil];
```
kratijain-microsoft: is it possible to extract these into instance variables like radioButtonCheckedImage, radioButtonUncheckedImage, etc and initialize only once and return everytime?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRChoiceSetViewDataSource.mm
```diff
@@ -100,24 +87,56 @@ - (NSInteger)numberOfSectionsInTableView:(UITableView *)tableView
...
 
-- (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath
-{
-    UITableViewCell *cell = nil;
-
+- (void)setImageForCell:(UITableViewCell *)cell choiceSetResourceResolver:(NSObject<ACOIResourceResolver> *)choiceSetResourceResolver indexPath:(NSIndexPath * _Nonnull)indexPath {
     if(_userSelections[[NSNumber numberWithInteger:indexPath.row]] == [NSNumber numberWithBool:YES]){
         if(_isMultiChoicesAllowed) {
-            cell = [tableView dequeueReusableCellWithIdentifier:checkedCheckboxReuseID];
+            if(choiceSetResourceResolver && [choiceSetResourceResolver respondsToSelector:@selector(getCheckBoxButtonSelectedState)])
```
kratijain-microsoft: Can we simplify the code and nested if-else here? May be return early, may be initialize default images only once and return ?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRChoiceSetViewDataSource.mm
```diff
@@ -100,24 +87,56 @@ - (NSInteger)numberOfSectionsInTableView:(UITableView *)tableView
...
+            {
+                cell.imageView.image = [UIImage imageNamed:@"unchecked.png" inBundle:[NSBundle bundleWithIdentifier:@"MSFT.AdaptiveCards"] compatibleWithTraitCollection:nil];
+            }
         }
     }
+}
 
+- (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:choiceSetCellReuseID];;
```
kratijain-microsoft: nit : double ;

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRChoiceSetViewDataSource.h
```diff
@@ -10,20 +10,21 @@
...
 @end
 
 @interface ACRChoiceSetViewDataSource:NSObject<UITableViewDataSource, UITableViewDelegate, ACRIBaseInputHandler>
 @property NSString *id;
+@property ACOHostConfig *hostConfig;
 @property BOOL isMultiChoicesAllowed;
 
-- (instancetype)initWithInputChoiceSet:(std::shared_ptr<AdaptiveCards::ChoiceSetInput> const&)choiceSet;
+- (instancetype)initWithInputChoiceSet:(std::shared_ptr<AdaptiveCards::ChoiceSetInput> const&)choiceSet
+                         andHostConfig:(ACOHostConfig *)hostConfig;
```
nesalang: please remove "and", you can leave it as HostConfig

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRChoiceSetViewDataSource.mm
```diff
@@ -24,17 +21,6 @@ - (instancetype)initWithStyle:(UITableViewCellStyle)style reuseIdentifier:(nulla
 {
     self = [super initWithStyle:style reuseIdentifier:reuseIdentifier];
     if(self) {
-        UIImage *iconImage = nil;
```
nesalang: this initWithStyle method gets called when dequeueReusableCellWithIdentifier is called and it can't get reusable cell.  This ensures that the cell with the correct image will get reused, and the code changes only the label. With your change, we are loading and setting the image each time cell is asked.

Since initWithStyle is get called by dequeReusableCellWithIdentifier, and thus we can't pass our resource resolver, you will have to add a method to ACRChoiceSetCell that takes resource resolver as argument. and you call this method once the deque call return on the returned cell. 
inside the method, you can simply check if icon image's.imge is set, if not, we load it there.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOIResourceResolver.h
```diff
@@ -9,6 +9,11 @@
 
 @protocol ACOIResourceResolver
 
+@optional
 - (UIImage *)resolveImageResource:(NSURL *)url;
+- (UIImage*) getRadioButtonForSelectedState;
```
nesalang: could we overload resolveImageResource:(NSURL *)url to take NSString and remove new interfaces. It seems too speicific. specifics of resource types being resolved is determined by scheme.

an888ha: I think this is more clean. As in later approach I will have to create 4 schema for 4 states.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRChoiceSetViewDataSource.mm
an888ha: What about storing the images as properties and use them ?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRChoiceSetViewDataSource.h
an888ha: ok

khouzam:
This seems to be very convoluted way of solving this issue. The resource resolver object is there to help with images that the cards need to consume, not for overriding the rendering of resources that the renderer uses. I think it would be easier to have a specific way of directly setting the images that are needed for choice sets and not have to query them for every rendering operation that needs them. This is not something that I would expect to change during the lifetime of the application and if it does, then they could be overridden at runtime using the same API that was initially used to set them.

PR #1976, opened 2018-09-28
Add performance test app to UWP

RebeccaAnne:
This change brings in a UWP performance app I wrote a few months ago while doing the xaml basic investigations.


source/uwp/PerfApp/App.xaml.cpp
```diff
@@ -0,0 +1,116 @@
...
+/// Invoked when application execution is being suspended.  Application state is saved
+/// without knowing whether the application will be terminated or resumed with the contents
+/// of memory still intact.
+/// </summary>
+/// <param name="sender">The source of the suspend request.</param>
+/// <param name="e">Details about the suspend request.</param>
+void App::OnSuspending(Object^ sender, SuspendingEventArgs^ e)
+{
+    (void) sender;  // Unused parameter
+    (void) e;   // Unused parameter
```
paulcam206: it's preferred to comment out the variable names in the signature

RebeccaAnne: Fixed, although this was code generated by visual studio :)

---
In reply to: [221771986](https://github.com/Microsoft/AdaptiveCards/pull/1976#discussion_r221771986) [](ancestors = 221771986)

PR #1994, opened 2018-10-04
Introduce simple editorconfig

paulcam206:
Add a simple `.editorconfig` to set indenting and charset standards across the codebase. Also, apply rules to existing files.

There will be a followup checkin later to introduce a `.clang-format` file for more in-depth style enforcement.


khouzam:
> 		}

This still has tabs

---
Refers to: source/dotnet/Samples/XamarinFormsBotClient/AdaptiveCards.Sample.BotClient.Droid/Resources/Resource.Designer.cs:30 in 0a79ab2. [](commit_id = 0a79ab2e53c172597c3f06c3428287163e9e132b, deletion_comment = False)

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/NumberInputRenderer.java
```diff
@@ -77,6 +77,6 @@ else if ((numberInput = NumberInput.dynamic_cast(baseCardElement)) == null)
 
         return editText;
     }
-
```
RebeccaAnne: Any idea what's causing this extra spacing to be added?

paulcam206: It's changing from

```
<spaces><linefeed>
```

to 

```
<linefeed>
```

samples/Tests/InvalidMediaMix.json
```diff
@@ -1 +1 @@
...
					"url": "https://adaptivecardsblob.blob.core.windows.net/assets/AdaptiveCardsOverviewVideo.mp4"
				},
				{
					"mimeType": "audio/mpeg",
					"url": "https://adaptivecardsblob.blob.core.windows.net/assets/AdaptiveCardsOverviewVideo.mp4"
				}
			]
		}
	]
}
```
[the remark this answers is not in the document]
paulcam206: This file had Mac-style linefeeds which apparently didn't agree with the tool I was using.

paulcam206:
I blame the node.js tool that I was using (`eclint`). `.pfx` files should definitely not be affected by the `.editorconfig` since there are only explicit directives for common source filetypes.

PR #1996, opened 2018-10-05
Bring shared model under clang-format control

paulcam206:
Introduce new `.clang-format` file (for shared model only) and apply changes.

Please review for issues where the new formatting is incorrect or worse than it used to be.


source/shared/cpp/ObjectModel/ActionParserRegistration.h
```diff
@@ -16,19 +17,17 @@ namespace AdaptiveSharedNamespace {
         virtual std::shared_ptr<BaseActionElement> Deserialize(
             std::shared_ptr<AdaptiveSharedNamespace::ElementParserRegistration> elementParserRegistration,
             std::shared_ptr<AdaptiveSharedNamespace::ActionParserRegistration> actionParserRegistration,
-            std::vector<std::shared_ptr<AdaptiveCardParseWarning>>& warnings,
-            const Json::Value& value) = 0;
+            std::vector<std::shared_ptr<AdaptiveCardParseWarning>>& warnings, const Json::Value& value) = 0;
```
RebeccaAnne: >, const Json::Value& value [](start = 76, length = 26)

We talked offline, but for posterity: I don't like that this has some of the parameters on their own lines and some that share a line. My preference would be that if they all fit on one line the all go on one line, but if not that each gets it's own line. Glancing at this i would likely miss seeing this value param and read it as having 3 parameters.

source/shared/cpp/ObjectModel/BaseActionElement.cpp
```diff
@@ -86,17 +85,16 @@ Json::Value BaseActionElement::GetAdditionalProperties() const
...
 }
 
 void BaseActionElement::PopulateKnownPropertiesSet()
 {
-    m_knownProperties.insert({ AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Type),
-         AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Title),
-         AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Id),
-         AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::IconUrl)});
+    m_knownProperties.insert({AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Type),
+        AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Title), AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Id),
```
RebeccaAnne: Here too, at the caller. This seems less readable/obvious with the long arguments sharing a line.

source/shared/cpp/ObjectModel/Enums.cpp
```diff
@@ -8,1137 +8,1023 @@
...
-    static std::unordered_map<std::string, AdaptiveCardSchemaKey, CaseInsensitiveHash, CaseInsensitiveEqualTo> adaptiveCardSchemaKeyNameToEnum = GenerateStringToEnumMap<AdaptiveCardSchemaKey>(adaptiveCardSchemaKeyEnumToName);
-
-    if (adaptiveCardSchemaKeyEnumToNameOut != nullptr)
-    {
-        *adaptiveCardSchemaKeyEnumToNameOut = adaptiveCardSchemaKeyEnumToName;
-    }
+    void GetAdaptiveCardSchemaKeyEnumMappings(std::unordered_map<AdaptiveCardSchemaKey, std::string, EnumHash>* adaptiveCardSchemaKeyEnumToNameOut,
+        std::unordered_map<std::string, AdaptiveCardSchemaKey, CaseInsensitiveHash, CaseInsensitiveEqualTo>* adaptiveCardSchemaKeyNameToEnumOut)
+    {
+        static std::unordered_map<AdaptiveCardSchemaKey, std::string, EnumHash> adaptiveCardSchemaKeyEnumToName = {
```
RebeccaAnne: > static std::unordered_map<AdaptiveCardSchemaKey, std::string, EnumHash> adaptiveCardSchemaKeyEnumToName [](start = 7, length = 104)

This mapping definitely works better as one per line.

source/shared/cpp/ObjectModel/ActionParserRegistration.h
RebeccaAnne: I like this better :)

---
In reply to: [223460135](https://github.com/Microsoft/AdaptiveCards/pull/1996#discussion_r223460135) [](ancestors = 223460135)

Issue #1282, opened 2018-03-20
[Shared] Parser support for Quick Actions


khouzam:

RebeccaAnne:
#147

PR #1997, opened 2018-10-05
Add InlineAction on Input.Text to shared model and UWP (QuickReply)

RebeccaAnne:
#1282 and #1283


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -2654,10 +2655,202 @@ AdaptiveNamespaceStart
...
+
+        ABI::AdaptiveNamespace::ActionType actionType;
+        THROW_IF_FAILED(localInlineAction->get_ActionType(&actionType));
+
+        // Show cards are not supported for inline actions
+        if (actionType == ActionType_ShowCard)
+        {
+            THROW_IF_FAILED(renderContext->AddWarning(
+                ABI::AdaptiveNamespace::WarningStatusCode::UnsupportedValue,
+                HStringReference(L"Show card not supported for InlineAction").Get()));
```
paulcam206: >Show card [](start = 35, length = 9)

shouldn't this be `ShowCard`?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -2654,10 +2655,202 @@ AdaptiveNamespaceStart
...
+        ComPtr<IGridStatics> gridStatics;
+        THROW_IF_FAILED(GetActivationFactory(HStringReference(RuntimeClass_Windows_UI_Xaml_Controls_Grid).Get(), &gridStatics));
+
+        ComPtr<IGrid> xamlGrid = XamlHelpers::CreateXamlClass<IGrid>(HStringReference(RuntimeClass_Windows_UI_Xaml_Controls_Grid));
+        ComPtr<IVector<ColumnDefinition*>> columnDefinitions;
+        THROW_IF_FAILED(xamlGrid->get_ColumnDefinitions(&columnDefinitions));
+        ComPtr<IPanel> gridAsPanel;
+        THROW_IF_FAILED(xamlGrid.As(&gridAsPanel));
+
+        // Craete the first column and add the text box to it
```
paulcam206: >Craete [](start = 11, length = 6)

super nit: typo :)

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -2654,10 +2655,202 @@ AdaptiveNamespaceStart
...
+        ComPtr<IColumnDefinition> textBoxColumnDefinition = XamlHelpers::CreateXamlClass<IColumnDefinition>(HStringReference(RuntimeClass_Windows_UI_Xaml_Controls_ColumnDefinition));
+        THROW_IF_FAILED(textBoxColumnDefinition->put_Width({ 1, GridUnitType::GridUnitType_Star }));
+        THROW_IF_FAILED(columnDefinitions->Append(textBoxColumnDefinition.Get()));
+
+        ComPtr<IUIElement> textBoxAsUIElement;
+        THROW_IF_FAILED(localTextBox.As(&textBoxAsUIElement));
```
paulcam206: we don't use this until quite a bit further down... shouldn't we be declaring closer to use?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -2654,10 +2655,202 @@ AdaptiveNamespaceStart
...
+        if (!isMultiLine)
+        {
+            EventRegistrationToken keyDownEventToken;
+            THROW_IF_FAILED(textBoxAsUIElement->add_KeyDown(Callback<IKeyEventHandler>(
+                [touchTargetUIElement, actionInvoker, localInlineAction](IInspectable* /*sender*/, IKeyRoutedEventArgs* args) -> HRESULT
+            {
+                ABI::Windows::System::VirtualKey key;
+                RETURN_IF_FAILED(args->get_Key(&key));
+
+                if (key == ABI::Windows::System::VirtualKey::VirtualKey_Enter)
```
paulcam206: >if (key == ABI::Windows::System::VirtualKey::VirtualKey_Enter) [](start = 16, length = 62)

more of a spec question -- do we care about shift-enter?

paulcam206:
Looks good to me overall!

source/uwp/Renderer/lib/XamlBuilder.cpp
RebeccaAnne: So, in fact, the spec calls for multiline text boxes to submit on enter and add a newline on shift-enter, but only if the user is using a physical keyboard. I ran into a few issues with that, and I've got a mail out to Andrew to clarify, so when he returns there may be more changes here in another PR. The behavior implemented in this PR is that in multiline text boxes enter is a newline and does not trigger the action (which is the spec'd behavior for touch keyboards).

That said, even in this case, which is a non-multiline text box, we shouldn't submit on shift-enter. I'll update this to not invoke the action if shift is pressed.

---
In reply to: [223483878](https://github.com/Microsoft/AdaptiveCards/pull/1997#discussion_r223483878) [](ancestors = 223483878)

PR #2014, opened 2018-10-10
updated gradle task

nesalang:
added a new task to create pom

### DIFF
--- a/source/android/adaptivecards/build.gradle
+++ b/source/android/adaptivecards/build.gradle
@@ -42,6 +42,40 @@ task sourceJar(type: Jar) {
     classifier "sources"
 }
 
+task createPom {
+    doLast {
+        pom {
+            project {
+                groupId 'io.adaptivecards'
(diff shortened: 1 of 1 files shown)


source/android/adaptivecards/build.gradle
```diff
@@ -42,6 +42,40 @@ task sourceJar(type: Jar) {
     classifier "sources"
 }
 
+task createPom {
+    doLast {
+        pom {
+            project {
+                groupId 'io.adaptivecards'
+                artifactId 'adaptivecards-android'
+                version System.getenv('ADAPTIVECARDSVERSION') ?: '1.1.0'
```
khouzam: I'd rather if we have an invalid version such as 0.0.0 if the setup is wrong then to potentially create a package that could be interpreted as valid. We now have 2 places that need to be updated when we bump up the version.

khouzam:
This also needs to go to the master branch. The process should be master then cherry-pick to release.

PR #2039, opened 2018-10-13
Jwoo/ios internal release

nesalang:
adds python script that updates podspec per internal ios release


UpdatePodSpec.py
```diff
@@ -0,0 +1,58 @@
...
+
+# get AdaptiveCard Version from custom.prop
+with open('custom.props', 'r') as f:
+    lines = f.readlines();
+    for line in lines:
+        if 'NuGetPackVersion' in line:
+            p = re.compile(r'\d+\.\d+\.\d+')
+            m = p.search(line)
+            if m != None:
+                acversion = m.group()
```
paulcam206: should we break once we find the line containing the version?

nesalang: you're absolutely correct, thanks! 

---
In reply to: [225654731](https://github.com/Microsoft/AdaptiveCards/pull/2039#discussion_r225654731) [](ancestors = 225654731)

Issue #2040, opened 2018-10-15
Schema 1.1 SelectAction


Benjathing:
# Platform

* .NET HTML
* JavaScript

# Author or host

Author, sending cards to WebChat emulator and DirectLine

# Version of SDK

NuGet 1.0.3

# Issue

As of this morning, the card JSON I had written in my database is no longer passing through JSON Schema validation.
The cards sent are still version 1.0; they're not using any of the new 1.1 features; however, the schema information I was using was pulling directly from the url https://adaptivecards.io/schemas/adaptive-card.json. I noticed that it has changed since last Friday with the new 1.1 features.

Playing around with the JSON and the schema in https://www.jsonschemavalidator.net/, I noticed that my cards have images with 'selectAction' properties. The old and new schemas are differing in these values.

The old schema just mentioned that Action.ShowCard is not supported in the description, but set the Action as any from Action:

    "selectAction": {
      "description": "An Action that will be invoked when the Image is tapped or selected. Action.ShowCard not supported",
      "$ref": "#/definitions/Action"
    }

The new schema explicitly specifies that this isn't supported:

    "selectAction": {
      "description": "An Action that will be invoked when the `Image` is tapped or selected. `Action.ShowCard` is not supported.",
      "type": "object",
      "enum": [
        "Action.Submit",
        "Action.OpenUrl"
      ]
    }

I've only just started playing with json schemas, so I'm not sure, but the enum there doesn't actually specify what the selectAction can be. Isn't that expecting a string with any of those values?

Indeed, I got the following issue from the validator:
![image](https://user-images.githubusercontent.com/10608747/46933176-a6675480-d09e-11e8-83df-aacc9d2b4473.png)

I changed the relevant section to:

    "selectAction": {
      "description": "An Action that will be invoked when the `Image` is tapped or selected. `Action.ShowCard` is not supported.",
      "type": "object",
      "item": {
        "anyOf":[
            { "$ref": "#/definitions/Action.Submit" },
            { "$ref": "#/definitions/Action.OpenUrl" }
        ]
      }
    }

and the problem went away.

Is this correct? If not, this block is in the schema 4 times, Image, ColumnSet, Column. and Container.

Benjathing:
Ironically, the new designer allows "ShowCard" actions on select:
![image](https://user-images.githubusercontent.com/10608747/46934695-6efba680-d0a4-11e8-9de6-587babd258db.png)

andrewleader:
[...]

PR #2097, opened 2018-10-16
Fix actions definitions and subcard versioning

paulcam206:
Fix for #2040


paulcam206:
yeah, it probably needs to be picked to a couple of places. I'll take care of it 👍

PR #2099, opened 2018-10-16
Cherry pick schema fix (#2040)

paulcam206:


paulcam206:
Note that change looks large because of spaces->tabs conversion that occurred in master

Issue #1911, opened 2018-09-14
AdaptiveCards - selectAction property didn't work on iOS Teams app


levotragiang:
# Platform
* iOS
* Node.js
* MS Teams

Hi guys,

I'm developing a chat bot, using MS Teams as a channel on both desktop app and mobile app, and Adaptive Cards of course.
Everything just worked smoothly until yesterday, i did updated the latest version of MS Teams on iOS device. Then, **selectAction** property of my Adaptive Cards didn't work anymore :'(

I think maybe MS got this bug while releasing this version...

Could anyone help me, or to raise this issue on MS teams at least?!

Thanks in advance!

vishmonne:
Hi Team,

How was the issue fixed ?

Even i am facing the same problem. My adaptive card submit action button is not working in Ms TEAM IOS app

shalinijoshi19:
@nesalang can you take a look at this?

nesalang:
> Hi Team,
> How was the issue fixed ?
> Even i am facing the same problem. My adaptive card submit action button is not working in Ms TEAM IOS app

@vishmonne 
Can you share your json payload?

nesalang:
@shalinijoshi19  Since the issue is raised in the context of Teams, I will check first if this is reproducible on our SDK, if not reproducible, I will hand it over to Teams.

Jcragons:
Same here on ios teams 1.0.75 (french locale if it matters) here the json response send back to ms bot platform (we use rest calls), this card is not showing, just blank :

```
{
   "type":"message",
   "from":{
      "id":"xxxxx",
      "name":"Beesy-Local"
   },
   "attachments":[
      {
         "contentType":"application/vnd.microsoft.card.adaptive",
         "content":{
            "type":"AdaptiveCard",
            "body":[
               {
                  "type":"TextBlock",
                  "text":"Et voici la liste des actions pour vous  :",
                  "wrap":true
               },
               {
                  "type":"ColumnSet",
                  "columns":[
                     {
                        "type":"Column",
                        "items":[
                           {
                              "type":"TextBlock",
                              "text":"**1** - test",
                              "wrap":true
                           }
                        ],
                        "width":"stretch"
                     },
                     {
                        "type":"Column",
                        "selectAction":{
                           "type":"Action.Submit",
                           "data":{
                              "action":"showaction",
                              "actionData":{
                                 "id":"2254803"
                              }
                           }
                        },
                        "items":[
                           {
[...]

Issue #2105, opened 2018-10-18
Adaptive cards v1.1 - iconUrl in action.openUrl do not work for "selectAction" on a column


sravanthialuri:
Excited about the new Adaptive cards v1.1. Thanks to the team for the great useful features. I'm trying to use one of the most useful one for our project the "iconUrl". 

The "iconUrl" works great when included in the "Actions" "action.openUrl" but I'm unable to see the icon image when used on a Column. Below is the sample code.

```
{
	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
	"type": "AdaptiveCard",
	"version": "1.1",
	"body": [
		{
			"type": "ColumnSet",
			"columns": [
				{
					"type": "Column",
					"items": [
						{
							"type": "TextBlock",
							"text": "Column 1"
						}						
					],
					"selectAction": {
						"type": "Action.OpenUrl",
                        "iconUrl": "http://adaptivecards.io/content/cats/1.png",
						"title": "cool link",
						"url": "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
					}
				}
		    ]
		}
	],
    "actions":[
        {
            "type": "Action.OpenUrl",
            "iconUrl": "http://adaptivecards.io/content/cats/1.png",
            "title": "cool link",
            "url": "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
        }
    ]
}
```

nesalang:
Can you specify which platform(s) you are seeing this issue?

andrewleader:
Hey @sravanthialuri 

selectActions do NOT show any visuals. They are purely touch targets. The `title` property can be used for accessibility purposes (screen readers, tooltips, etc), but the `iconUrl` is not used by design. It is only seen in buttons within the `actions` array (just as `title` is really only seen in buttons).

If you need an image, you can add an `Image` element.

If that doesn't work for your scenario, let us know why! Closing this for now. Thanks!

PR #2106, opened 2018-10-18
Jwoo/ios selection action fix

nesalang:
fixes issue #1911 and possibly #2105 if the issue occurs on iOS.

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRUILabel.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRUILabel.mm
@@ -62,24 +62,12 @@
     CGPoint location = point;
     location.x -= self.textContainerInset.left;
     location.y -= self.textContainerInset.top;
-    CGFloat fraction = 20.0f;
-    // find character under touch
-    NSUInteger characterIndex = [self.layoutManager characterIndexForPoint:location inTextContainer:self.textContainer fractionOfDistanceBetweenInsertionPoints:&fraction];
-    CGPoint endPoint =[self.layoutManager locationForGlyphAtIndex:[self.layoutManager glyphIndexForCharacterAtIndex:self.textStorage.length - 1]];
-    CGFloat characterWidth = 2 * endPoint.x - [self.layoutManager locationForGlyphAtIndex:[self.layoutManager glyphIndexForCharacterAtIndex:self.textStorage.length - 2]].x;
-
-    NSLog(@"character index x= %tu", characterIndex	);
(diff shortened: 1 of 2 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRUILabel.mm
```diff
@@ -57,4 +57,21 @@ - (void)layoutSubviews
...
+    location.y -= self.textContainerInset.top;
+    CGFloat fraction = 0.0f;
+    NSUInteger characterIndex = [self.layoutManager characterIndexForPoint:location
+                                                           inTextContainer:self.textContainer
+                                  fractionOfDistanceBetweenInsertionPoints:&fraction];
+    if (!(fraction == 0.0 || fraction == 1.0) && characterIndex < self.textStorage.length) {
+        if ([self.textStorage attribute:NSLinkAttributeName atIndex:characterIndex effectiveRange:NULL]) {
+            return self;
+        }
+    }
```
bviglietta: I'm sorry, I don't understand the intention of this change. We're ignoring UI interactions unless they land on a character that's part of a hyperlink?

But only if they land inside the character, and not on the left or right edge of it?

nesalang: that's correct. this method is part of UIView API, and it's used for returning the farthest UIView that should receive the given UIEvent in UIView tree.
This UIView is subclass of UITextView, and only UIEvent it should handle is touch event on hyperlink. This change is needed because our container classes should receive AdaptiveCards events, and the textview should handle touch event  on hyperlink. 

---
In reply to: [226504988](https://github.com/Microsoft/AdaptiveCards/pull/2106#discussion_r226504988) [](ancestors = 226504988)

nesalang: i'm not exactly sure what you meant by "dropping inputs that happen to land exactly between two characters". 

characterIndexForPoint method returns nearest character index in the text storage, so the returned index will be always valid character index.

the fraction tell you how far the touch is from the insertion point. if it's zero or 1, it's outside of characters in line fragment. linefragment is filled with characters, but linefragment will occupy one line of textview, and its horizontal position is determined by LayoutManager. The fraction will have value between 0 and 1.0 when touch lands in the frame of the line fragement. 

---
In reply to: [226796362](https://github.com/Microsoft/AdaptiveCards/pull/2106#discussion_r226796362) [](ancestors = 226796362,226741749,226504988)

PR #2133, opened 2018-10-29
[Android] Add data URI support

almedina-ms:
Solves this issue: https://github.com/Microsoft/AdaptiveCards/issues/2129

I also took the liberty to fix a couple issues that I hadnt seen/thought about before, mainly about moving the default online loader method back to the GenericImageLoader and small extra stuff

This is now based on the UWP pull request, so that one must be merged first


samples/v1.2/Elements/Image.DataUri.json
```diff
@@ -0,0 +1,11 @@
+{
+    "type": "AdaptiveCard",
+    "version": "1.0",
+    "body": [
+        {
+            "type": "Image",
```
paulcam206: >             [](start = 0, length = 12)

We're supposed to be using tabs here. You should probably take a moment to enable [EditorConfig support for Android Studio](https://stackoverflow.com/questions/44694132/editorconfig-in-android-studio). :)

PR #2134, opened 2018-10-29
Upgrade to new gradle version to fix NDK break

paulcam206:
Updates to the NDK cause build breaks in older versions of gradle. The only way to get us building is for users to downgrade the NDK version, which isn't optimal. This change upgrades us to a newer version of gradle which doesn't have problems with the newer NDK version.


paulcam206:
oops -- linting issue that needs to be fixed

PR #2141, opened 2018-10-30
Added the new adaptive cards sample for hotel booking

arafattehsin:
I tried to find out something similar to hotels (not restaurant) booking but could not find one so I created myself.

Please have a look at it and let me know in case of any issue.


arafattehsin:
Closing it for another PR.

Issue #628, opened 2017-09-15
Image data URI support


matthidinger:
<table>
<tr>
<th>Release</th>
<th>Renderer status</th>
<th>Tasks status</th>
</tr>
<tr>
<td>1.2
<td>✔️ .NET (#2131)<br/>✔️ Android (#2129)<br/>✔️ iOS (#2130)<br/>✔️ TS (#2442)<br/>✔️ UWP (#2128)</td>
<td>✔️ Shared (#2127)</td>
</tr>
</table>

# Solves requests

* Embedding image data locally within card (#1991)

# Solution
The [data URI scheme](https://en.wikipedia.org/wiki/Data_URI_scheme
) is a uniform resource identifier (URI) scheme that provides a way to include data in-line in web pages as if they were external resources.

## Schema
**Updated** support on all image url properties to support data urls

**AdaptiveCard**

Property|Type|Required|Description
---|---|---|---
backgroundImage|string|false|An image to use as the background of the card. Supports data URI.

**Image**

Property|Type|Required|Description
---|---|---|---
url|string|false|The URL to the image. Supports data URI.

**Action**

Property|Type|Required|Description
---|---|---|---
iconUrl|string|false|The URL to the image. Supports data URI.

## Example
Note: this already works in the HTML renderer since browsers support this. We would need to add support in the other renderers.

```json
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "Image",
[...]

PR #2150, opened 2018-11-05
[UWP] Add data uri for images support

almedina-ms:
Solves this issue: https://github.com/Microsoft/AdaptiveCards/issues/2128


source/shared/cpp/ObjectModel/AdaptiveBase64Util.cpp
```diff
@@ -0,0 +1,180 @@
...
+    size_t enc_len = 0;
+    unsigned char a3[3];
+    unsigned char a4[4];
+
+    out->resize(EncodedLength(in));
+
+    size_t input_len = in.size();
+    std::string::const_iterator input = in.begin();
+
+    while (input_len--) {
```
khouzam: Needs to follow the clang-format

source/shared/cpp/ObjectModel/AdaptiveBase64Util.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+std::string AdaptiveBase64Util::Encode(const std::string& decodedBase64)
+{
+    std::string encodedString;
+    Encode(decodedBase64, &encodedString);
+    return encodedString;
+}
+
+// Format for DataURI is data:[<MediaType>][;base64],data with MediaType and base64 being optional and data is composed of [A-Z a-z 0-9 + /] characters
+std::string AdaptiveBase64Util::ExtractDataFromUri(const std::string& dataUri)
```
khouzam: >std::string [](start = 0, length = 11)

This should return a vector<char> or similar

source/shared/cpp/ObjectModel/AdaptiveBase64Util.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+    if (base64Alphabet.find(c) == base64Alphabet.end())
+    {
+        return 255;
+    }
+    return base64Alphabet[c];
+}
+
+bool AdaptiveBase64Util::Encode(const std::vector<char>& in, std::string* out)
+{
```
khouzam: Can we have a some tests in the shared model to cover this code?

khouzam:
One more minor thing. Can you move the files to v1.1. We're going to backport this because Timeline needs it in Windows.

bandmask:
@almedina-ms @khouzam 
Currently having problem with the functionality this is intended to solve.
We cannot render data uris in format of `data:image/png:base64 [base64EncodedString]` without implementing our own image renderer (converting the AdaptiveImage to and UWP Image with source as BitmapImage).

I would like to request some enlighting pointers for how to do this without using a custom renderer.

We have tried images both as a direct child to the adaptiveCard body, and as a child element in a nested hierarchy, without any success.

Since the #2128 along with #628 both are closed, one would think this should work out of the box?

Best regards

PR #2151, opened 2018-11-06
Fix our enums

paulcam206:
Here are changes that are not really part of the enum change, or are just follow-on changes:
* `source\android\...\objectmodel\*`: SWIG'd changes
* `source\android\...\renderer\registration\CardRendererRegistration.java`: Slight change to how object model is projected
* `source\ios\*`: changes for build
* `source\shared\cpp\.clang-format`: auto-format escaped newlines
* `ColumnSet.h`: remove unused `std::unordered_map<>` member
* `FactSet.cpp`: use `const`
* `SharedAdaptiveCard.cpp`: create local var to make it clear what's happening (cppcoreguidelines rule)
* `Util.h`: remove `using` statement that would leak our namespace usage, fix namespacing
* `Util.cpp`: add missing `using` statement, fix up namespacing, add `const`
* `uwp\renderer\lib\util.h`: fix namespacing (previously leaked here by shared model's `Util.h`)
* `pch.h`: add `Enums.h` to save on build perf
* `EverythingBagel.json`: mix a couple of cases for values to make sure enum mapping is case insensitive

## Why?
`Enums.[cpp|h]` is a bit tricky from an authoring perspective. It's easy to copy/paste the wrong thing and end up with incorrect behavior. It's also difficult to make changes across the entirety of enums usage without lots of editor heavy lifting.

I recently found a minor mistake that caused our enums to be copy constructed every time a lookup is performed. After initially fixing the problem by robofixing all of our implementations, it became clear that this was an unsustainable model. To that end, I've redone our enums with an eye towards maintainability.

## The Problem Space
The problem that we need to solve breaks down like this: C++ doesn't have native support for mapping `enum values <-> string representations of enum values`. This is a problem for us because we need to be able to perform this mapping in both directions to support serialization and deserialization. Unfortunately, the STL doesn't provide a data structure that can provide this functionality. So instead, we have to maintain two `std::unordered_map<>`s -- one mapping from the enum type to a string, and the other providing mapping in the other direction. We _could_ just use one map, but then we would have O(1) lookup in one direction, and O(N) lookup in the other, which isn't ideal.

It's worth noting that there are a couple of cases where the mapping isn't 1:1 -- there are a couple of enums where we map additional string(s) to an enum value for backwards compatibility reasons (see e.g. `ImageSize`).

## The Old Way
The way that this previously worked was that we had a series of `Get*EnumMappings()` functions that had static instances of maps. The mapping from enum type to string was handled via a hand-rolled initializer list. Mapping in the other direction was either generated via the `GenerateStringToEnumMap<>()` function (1:1 mapping case) or was likewise hand-rolled (back-compat case). `Get*EnumMappings()` would hand out copies of our `unordered_map`s to callers. Those callers? Hand-rolled `*FromString()` and `*ToString()` functions. The `*FromString()` functions were responsible for handling what to do in the case of a string that wasn't in the map. This handling falls into three classes:
* Throw an exception
* Return `*::Default`
* Return some other enum value as the default

To be crystal clear: If one needs to add a new enum in this system, one must:
1. Create a new `enum class` in `Enums.h`
1. Add forward declarations for `*ToString()` and `*FromString()` functions in `Enums.h`
1. Create `Get*EnumMappings()` function in `Enums.cpp`
1. Create `*ToString()` function in `Enums.cpp`
1. Create `*FromString()` function in `Enums.cpp`

### Old Example
`Enums.h`:
```c++
enum class ImageStyle
{
    Default = 0,
    Person
};
const std::string ImageStyleToString(ImageStyle style);
ImageStyle ImageStyleFromString(const std::string& style);
```

`Enums.cpp`:
```c++
void GetImageStyleEnumMappings(std::unordered_map<ImageStyle, std::string, EnumHash>* imageStyleEnumToNameOut,
                               std::unordered_map<std::string, ImageStyle, CaseInsensitiveHash, CaseInsensitiveEqualTo>* imageStyleNameToEnumOut)
{
    static std::unordered_map<ImageStyle, std::string, EnumHash> imageStyleEnumToName = {{ImageStyle::Default, "default"},
                                                                                         {ImageStyle::Person, "person"}};

    static std::unordered_map<std::string, ImageStyle, CaseInsensitiveHash, CaseInsensitiveEqualTo> imageStyleNameToEnum = {
        {"default", ImageStyle::Default},
        {"person", ImageStyle::Person},
        {"normal", ImageStyle::Default} // Back compat to support "Normal" for "Default" for pre V1.0 payloads
    };

    if (imageStyleEnumToNameOut != nullptr)
    {
        *imageStyleEnumToNameOut = imageStyleEnumToName;
    }

    if (imageStyleNameToEnumOut != nullptr)
    {
        *imageStyleNameToEnumOut = imageStyleNameToEnum;
    }
}

const std::string ImageStyleToString(ImageStyle style)
{
    std::unordered_map<ImageStyle, std::string, EnumHash> imageStyleEnumToName;
    GetImageStyleEnumMappings(&imageStyleEnumToName, nullptr);

    if (imageStyleEnumToName.find(style) == imageStyleEnumToName.end())
    {
        throw std::out_of_range("Invalid ImageStyle style");
    }
    return imageStyleEnumToName[style];
}

ImageStyle ImageStyleFromString(const std::string& style)
{
    std::unordered_map<std::string, ImageStyle, CaseInsensitiveHash, CaseInsensitiveEqualTo> imageStyleNameToEnum;
    GetImageStyleEnumMappings(nullptr, &imageStyleNameToEnum);

    if (imageStyleNameToEnum.find(style) == imageStyleNameToEnum.end())
    {
        return ImageStyle::Default;
    }

    return imageStyleNameToEnum[style];
}
```

## The Magic
The core set of changes to get us out of dealing with enums comes from the contents of `EnumMagic.h`. Its contents can be broken down into these parts:
* A templated class, `EnumMapping<>`, responsible for constructing and providing access to our two maps.
  * Can be constructed either with a single initializer list, which generates both maps (1:1 case), or with two initializer lists, which will create the string to enum map as the union of the reverse mapping and the supplied list.
* A macro, `DECLARE_ADAPTIVECARD_ENUM`, which provides forward declarations of our mapping functions
* A set of macros for defining our mapping functions
  * `_DEFINE_ADAPTIVECARD_ENUM_INVARIANT`: Helper macro used by other `DEFINE_*` macros. Supplies implementations that don't vary by enum style.
    * Generated function to provide access to the appropriate `static const EnumMapping<>` instance, named `get*Enum()`
    * Generated `*ToString()` function
  * `DEFINE_ADAPTIVECARD_ENUM_THROW`: Defines `*FromString()` function that throws an exception if supplied string doesn't map
  * `DEFINE_ADAPTIVECARD_ENUM_DEFAULT`: Defines `*FromString()` function that returns a specified default value if supplied string doesn't map
  * `DEFINE_ADAPTIVECARD_ENUM`: Defines `*FromString()` function that returns `ENUM::Default` if supplied string doesn't map

## The New Way
Adding an enum in the new system looks like this:

### Enums.h
```c++
enum class NewThingy
{
    thing1,
    thing2,
    Default
}
DECLARE_ADAPTIVECARD_ENUM(NewThingy); // provides all necessary fwd decls
```

### Enums.cpp

In the case where there exists `NewThingy::Default`:
```c++
DEFINE_ADAPTIVECARD_ENUM(NewThingy, {
    { NewThingy::thing1, "thing1" },
    { NewThingy::thing2, "thing2" },
    { NewThingy::Default, "Default" }});
```

That's it!

If there's no `::Default`, it might look something like this (here we're saying that `NewThingy::thing1` should be used as the default):
```c++
DEFINE_ADAPTIVECARD_ENUM_DEFAULT(NewThingy, NewThingy::thing1, {
    { NewThingy::thing1, "thing1" },
    { NewThingy::thing2, "thing2" }});
```

If you need to supply a list of additional mappings for back-compat:
```c++
DEFINE_ADAPTIVECARD_ENUM(ImageStyle, {
    {
        { NewThingy::thing1, "thing1" },
        { NewThingy::thing2, "thing2" }
    },
    {
        {"normal", NewThingy::thing1} // NB: other reverse mappings computed automatically
    }});
```

Or if you need to throw on a failed mapping, it would look like this:
```c++
DEFINE_ADAPTIVECARD_ENUM_THROW(NewThingy, {
    { NewThingy::thing1, "thing1" },
    { NewThingy::thing2, "thing2" }});
```

### Our `ImageStyle` example from above in new form
`Enums.h`:
```c++
enum class ImageStyle
{
    Default = 0,
    Person
};
DECLARE_ADAPTIVECARD_ENUM(ImageStyle);

```

`Enums.cpp`:
```c++
    DEFINE_ADAPTIVECARD_ENUM(ImageStyle, {
            {
                {ImageStyle::Default, "default"},
                {ImageStyle::Person, "person"}
            },
            {
                {"normal", ImageStyle::Default} // Back compat to support "Normal" for "Default" for pre V1.0 payloads
            }});
```

### Benefits
* In developing this change, I took some (admittedly unscientific) perf stats. Overall runtime for our unit tests (run in serial, N=12) came down about 10% with this change.
* Future improvements can be made across all of our enums without having to edit every single function
* No chance of a whoopsie where a new enum is added that doesn't have the same behavior as all of the others

### Downsides
As with anything, there _are_ tradeoffs:
* Somewhat slower build time (due to heavy use of macros and templates)
* Less debuggable than the hand-rolled versions, at least in Visual Studio
* If a new class of enum or new scenario arises, we may have to write some new macros, and they can be a bit of a pain


source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/EverythingBagel.json
```diff
@@ -29,7 +29,7 @@
 			"isSubtle": false,
 			"maxLines": 1,
 			"size": "default",
-			"weight": "default",
+			"weight": "normAl",
```
khouzam: >normAl", [](start = 14, length = 8)

is this intended?

paulcam206: yep -- just making sure we have coverage for our back-compat reverse mapping :)

---
In reply to: [231253792](https://github.com/Microsoft/AdaptiveCards/pull/2151#discussion_r231253792) [](ancestors = 231253792)

source/shared/cpp/ObjectModel/EnumMagic.h
```diff
@@ -0,0 +1,128 @@
...
+    _DEFINE_ADAPTIVECARD_ENUM_INVARIANT(ENUMTYPE, __VA_ARGS__);      \
+    ENUMTYPE ENUMTYPE##FromString(const std::string& t)              \
+    {                                                                \
+        try                                                          \
+        {                                                            \
+            return EnumHelpers::get##ENUMTYPE##Enum().fromString(t); \
+        }                                                            \
+        catch (const std::out_of_range&)                             \
+        {                                                            \
+            return DEFAULT;                                          \
```
nesalang: is DEFAULT needed to be verified that it is correct type?

paulcam206: The build will fail if it's invalid

---
In reply to: [231267703](https://github.com/Microsoft/AdaptiveCards/pull/2151#discussion_r231267703) [](ancestors = 231267703)

paulcam206:
Anyone else care to look?

PR #2153, opened 2018-11-07
[.Net][HTML + WPF] Add support for image dataUri

almedina-ms:
Solves this issue: https://github.com/Microsoft/AdaptiveCards/issues/2131

The html renderer didnt need any work as it's already supported (https://github.com/Microsoft/AdaptiveCards/issues/628)


source/dotnet/Library/AdaptiveCards.Rendering.Wpf/Helpers/ImageExtensions.cs
```diff
@@ -125,7 +138,16 @@ public static async void SetBackgroundSource(this Grid grid, Uri uri, AdaptiveRe
...
+            BitmapImage bi = null;
+            if (finalUri.Scheme == "data")
+            {
+                bi = GetBitmapFromBase64(finalUri);
+            }
+            else
+            {
+                bi = await context.ResolveImageSource(finalUri);
+            }
+
```
khouzam: You probably want to invoke the resource resolver in the case of DataURIs in case the host has registered their own.

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveImageRenderer.cs
```diff
@@ -22,6 +22,7 @@ public static FrameworkElement Render(AdaptiveImage image, AdaptiveRenderContext
             }
 
             uiImage.SetSource(image, finalUri, context);
+
```
paulcam206: don't think we need this change

PR #2154, opened 2018-11-07
Jwoo/ios quick action

nesalang:
added quick action


nesalang:
added quick action to text input
1. button is added next to text filed in case of single line, to text view in case of multiline
2. action was modified, so the code can be reused
3. text view as refactored to include default settings
4. xibs are added for style customization.

PR #2156, opened 2018-11-07
Fix title in action for sentiment test 

almedina-ms:
This wont matter because I'm just testing stuff


almedina-ms:
Don't review this, I'm just playing around with github

PR #2160, opened 2018-11-08
Updating designer link in readme

andrewleader:


paulcam206:
Looks like this is no longer needed.

Issue #2019, opened 2018-10-10
Fallback and requires properties on elements


andrewleader:
<table>
<tr>
<th>Release</th>
<th>Renderer status</th>
<th>Tasks status</th>
</tr>
<tr>
<td>1.2
<td>✔️ .NET (#2418)<br/>✔️ Android (#2419)<br/>✔️ iOS (#2420)<br/>✔️ TS (#2421)<br/>✔️ UWP (#2125)</td>
<td>✔️ Shared (#2624)<br/>🔄 Designer (#2502)</td>
</tr>
</table>

# Solves requests

* Ability to provide fallback elements when elements aren't supported #2018
* Ability to declare required host features for content within card #2028

# Summary
Add a `fallback` and `requires` property to all elements

# Schema
**New** properties on all elements

Property | Type | Required | Description
---|---|---|---
fallback | `"drop"` or AdaptiveElement | false | Fallback content or action if the element cannot be rendered. If not provided and an ancestor element has a fallback, the ancestor's fallback will be triggered. Otherwise, the element itself will be dropped.
requires | Dictionary<string, string> | false | Requirements for the element to be rendered. In the format of `<feature, version>`, where the version number uses [NuGet package manager version syntax](https://docs.microsoft.com/en-us/nuget/reference/package-versioning#version-ranges-and-wildcards) to allow specifying min/max/etc

**New** properties on all actions (only difference is the allowed fallback object type of Action rather than Element)

Property | Type | Required | Description
---|---|---|---
fallback | `"drop"` or AdaptiveAction | false | Fallback content or action if the element cannot be rendered. If not provided and an ancestor element has a fallback, the ancestor's fallback will be triggered. Otherwise, the element itself will be dropped.
requires | Dictionary<string, string> | false | Requirements for the element to be rendered. In the format of `<feature, version>`, where the version number uses [NuGet package manager version syntax](https://docs.microsoft.com/en-us/nuget/reference/package-versioning#version-ranges-and-wildcards) to allow specifying min/max/etc

# Example

## Alternative media content

```json
{
  "type": "Media",
  "poster": "https://...",
  "sources": [ ],
  "fallback": {
    "type": "Image",
    "url": "https://movies.com/thumbnail"
  }
}
```

## Drop a group of dependent content

The text doesn't make sense without the graph, so we want all of this to be dropped together.

On the entire container, the author specifies `"fallback": "drop"`.

[...]

PR #2161, opened 2018-11-09
[HTML] Fallback and Requires support

dclaux:
Implements https://github.com/Microsoft/AdaptiveCards/issues/2019


dclaux:
Sample payload to test with:

```
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "fallbackText": "This is the fallback text",
    "fallback": {
        "type": "TextBlock",
        "text": "This is the fallback element"
    },
    "body": [
        {
            "type": "Container",
            "fallback": {
                "type": "TextBlock",
                "text": "The container is falling back."
            },
            "selectAction": {
                "type": "Action.OpenUrl",
                "title": "Hello world",
                "url": "http://www.microsoft.com"
            },
            "items": [
                {
                    "type": "TextBlock",
                    "text": "Publish Adaptive Card schema",
                    "weight": "bolder",
                    "size": "medium"
                },
                {
                    "type": "ActionSet",
                    "actions": [
                        {
                            "type": "Action.OpenfUrl",
                            "title": "Hello world",
                            "url": "http://www.microsoft.com"
                        }
                    ]
                },
                {
                    "type": "Image",
                    "size": "large",
                    "url": "http://cdn.nirmaltv.com/images/Powerpoint-2013.png",
                    "fallback": "drop",
                    "selectAction": {
                        "type": "Action.OpenUrl",
                        "title": "Hello world",
                        "url": "http://www.microsoft.com"
                    }
                }
            ]
        },
        {
            "type": "ColumnSet",
            "selectAction": {
                "type": "Action.OpenUrl",
                "title": "Hello world",
                "url": "http://www.microsoft.com"
            },
            "columns": [
                {
                    "width": 1,
                    "style": "emphasis",
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "Column 1",
                            "horizontalAlignment": "center"
                        }
                    ]
                },
                {
                    "width": 1,
                    "style": "emphasis",
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "Column 2",
                            "horizontalAlignment": "center"
                        }
                    ]
                },
                {
                    "width": 1,
                    "style": "emphasis",
                    "fallback": "drop",
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "Column 3",
                            "horizontalAlignment": "center"
                        }
                    ]
                }
            ]
        }
    ],
    "actions": [
        {
            "type": "Action.OpefnUrl",
            "title": "Hello world",
            "url": "http://www.microsoft.com",
            "fallback": {
                "type": "Action.OpenUrl",
                "title": "Fallback action",
                "url": "http://www.microsoft.com"
            }
        }
    ]
}
```

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -26,19 +26,101 @@ function generateUniqueId(): string {
...
+            if (typeof fallback === "string" && fallback.toLowerCase() === "drop") {
+                result = null;
+            }
+            else if (typeof fallback === "object") {
+                result = createElementInstance(
+                    parent,
+                    fallback,
+                    errors);
+            }
+        }
```
paulcam206: a lot of this logic is duplicated from above. helper function?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3325,6 +3412,23 @@ class ActionCollection {
         this._owner = owner;
     }
 
+    parse(json: any, errors?: Array<IValidationError>) {
```
paulcam206: >         [](start = 56, length = 8)

looks like you haven't configured your editor to consume `.editorconfig` yet? instructions [here](https://editorconfig.org/#download).

paulcam206:
>             throw new Error("The action already belongs to another element.")

or is null? also, `;`

---
Refers to: source/nodejs/adaptivecards/src/card-elements.ts:3666 in 077c274. [](commit_id = 077c274d9d8be96db7f6948b3221a21e0d1d1815, deletion_comment = False)

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Done.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Done

dclaux:
Now supports "requires" clause. Can be tested via the Outlook host in the visualizer, which defines two "capabilities":
- helloWorld, any version
- whatNot, version 2.12

To test, select the Outloook Actionable Messages host container, and add a "requires" property to any element, action or the card itself. For instance:
```
"requires": {
  "whatNow": "3.0"
}
```

The above will cause the element/action/card to fall back.

source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -330,7 +330,166 @@ export class ContainerStyleSet {
     }
 }
 
+export class Version {
+    private _versionString: string;
+    private _major: number;
+    private _minor: number;
```
paulcam206: shouldn't we be supporting semantic versioning? there should be a `build` and `revision` property

source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -330,7 +330,166 @@ export class ContainerStyleSet {
...
+                break;
+            }
+        }
+
+        minorVersionString = minorVersionString.substr(0, minorVersionString.length - trailingZeros);
+
+        return !this._isValid ? this._versionString : this._major + "." + minorVersionString;
+    }
+
+    compareTo(otherVersion: Version): number {
```
paulcam206: > compareTo(otherVersion: Version): [](start = 3, length = 34)

looks like this won't work for wildcards or version ranges (per spec)

source/nodejs/adaptivecards/src/host-config.ts
dclaux: There is an active discussion on the side about this. I think we should keep it simple, but I will update the code according to the conclusion of the conversation.

source/nodejs/adaptivecards/src/host-config.ts
dclaux: My implementation doesn't support ranges at the moment, and uses the explicit string "any" for wildcard. There is an active discussion on the side about versioning. I think we should keep things simple, but I will update the code according to the conclusion of the conversation.

dclaux: Per discussion, we are going with a simple model for 1.2.

source/nodejs/adaptivecards/src/host-config.ts
dclaux: Per discussion we are going with a simple model for 1.2.

Issue #2104, opened 2018-10-18
Bug: .NET fails to parse card with selectAction of type Action.ToggleVisibility


andrewleader:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* .NET

# Author or host

Host

# Version of SDK

NuGet 1.1.0

# Issue

When trying to parse a card that has an `Image.selectAction` set to `Action.ToggleVisibility` (a custom Outlook action), the card fails to parse and throws the following exception...

```
AdaptiveCards.AdaptiveSerializationException: Error setting value in extension data for type 'AdaptiveCards.AdaptiveImage'. Path 'selectAction', line 1, position 4297. ---> Newtonsoft.Json.JsonSerializationException: Error setting value in extension data for type 'AdaptiveCards.AdaptiveImage'. Path 'selectAction', line 1, position 4297. ---> Newtonsoft.Json.JsonSerializationException: Unexpected token while deserializing object: EndObject. Path 'selectAction', line 1, position 4297.

   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)

   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.ReadExtensionDataValue(JsonObjectContract contract, JsonProperty member, JsonReader reader)

   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.SetExtensionData(JsonObjectContract contract, JsonProperty member, JsonReader reader, String memberName, Object o)
```

The card JSON to repro this issue is as follows

```json
{
	"type": "AdaptiveCard",
	"$schema": "https://adaptivecards.io/schemas/adaptive-card.json",
	"version": "1.0",
	"body": [
		{
			"type": "ColumnSet",
			"columns": [
				{
					"type": "Column",
					"items": [
						{
							"type": "TextBlock",
							"horizontalAlignment": "Center",
							"text": "20/8/18",
							"wrap": true
						}
					],
					"width": "auto"
				},
				{
					"type": "Column",
					"horizontalAlignment": "Left",
					"spacing": "medium",
					"items": [
						{
							"type": "TextBlock",
							"text": "Air Travel Expense",
							"wrap": true
						}
					],
					"width": "stretch"
				},
				{
					"type": "Column",
					"items": [
						{
							"type": "TextBlock",
[...]

PR #2163, opened 2018-11-09
Handle invalid Image.SelectAction

carlos-zamora:
Bugfix for #2104.

## Explanation of Issue
For whatever reason, Json.net mishandles passing `null` up to an `object` (Image) expecting an `object` (SelectAction). This is because the `JsonReader` expects more tokens, but is handed an unhandled `EndObject` token and just throws an error. This all happens in `JsonSerializerInternalReader` when trying to set the properties of the `Image`.

## Fix
Since we can't control anything other than the `contract` and the `converter`, it looks like the best we can do is catch the exception thrown when deserializing the Image. Then just pass up what we were able to deserialize of the Image so far.

## Side Effect
Unfortunately, anything written after the faulty `SelectAction` is lost. This means that the following two JSON would return different AdaptiveCards. This is still different behavior from the SharedModel and JavaScript because those platforms would output Card 1's result, regardless of which JSON is passed.

---Card 1---
image
```
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "Image",
      "url": "http://adaptivecards.io/content/cats/1.png",
      "selectAction": {
        "type": "INVALID ACTION",
        "title": "cool link",
        "url": "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
      },
    }
  ]
}
```

---Card 2---
image
```
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "Image",
      "selectAction": {
        "type": "INVALID ACTION",
        "title": "cool link",
        "url": "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
      },
    }
  ]
}
```

Both cards throw the proper warning:
image

### DIFF
--- a/source/dotnet/Test/AdaptiveCards.Test/AdaptiveActionTests.cs
+++ b/source/dotnet/Test/AdaptiveCards.Test/AdaptiveActionTests.cs
@@ -30,6 +30,39 @@ namespace AdaptiveCards.Test
         }
 
         [TestMethod]
+        public void ParseInvalidAction()
+        {
+            string url = "http://adaptivecards.io/content/cats/1.png";
+            var json = @"{
+	""$schema"": ""http://adaptivecards.io/schemas/adaptive-card.json"",
+	""type"": ""AdaptiveCard"",
+	""version"": ""1.0"",
+	""body"": [
+		{
+			""type"": ""Image"",
+			""url"": """+ url + @""",
+			""selectAction"": {
+				""type"": ""Invalid Action"",
+				""url"": ""https://www.youtube.com/watch?v=dQw4w9WgXcQ""
+			}
+		}
+	]
+}";
+
+            var result = AdaptiveCard.FromJson(json);
+
+            var card = result.Card;
+            Assert.IsNotNull(card);
+
+            Assert.IsTrue(result.Warnings.Count == 1);
+
+            AdaptiveImage image = (AdaptiveImage) card.Body[0];
+            Assert.AreEqual(image.Url, url);
+            Assert.IsNull(image.SelectAction);
+
+        }
+
+        [TestMethod]
         public void ParseSentiment()
         {
             var json = @"{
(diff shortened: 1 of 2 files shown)


source/dotnet/Test/AdaptiveCards.Test/AdaptiveActionTests.cs
```diff
@@ -29,6 +29,39 @@ public void ParseIconUrl()
...
+            string url = "http://adaptivecards.io/content/cats/1.png";
+            var json = @"{
+	""$schema"": ""http://adaptivecards.io/schemas/adaptive-card.json"",
+	""type"": ""AdaptiveCard"",
+	""version"": ""1.0"",
+	""body"": [
+		{
+			""type"": ""Image"",
+			""url"": """+ url + @""",
+			""selectAction"": {
```
andrewleader: What happens if there's a property placed AFTER selectAction? Please try the following

```json
{
  "type": "Image",
  "selectAction": {
    "type": "InvalidAction"
  },
  "url": "https://msn.com"
}
```

And validate that the Url property gets populated correctly.

carlos-zamora: Yeah. Mentioned in original comment. It drops the url. The problem is that the bug is in JSON.NET.

andrewleader: Ah I see, so it's intended to only be a partial fix. Makes sense.

khouzam: Does changing the version of JSON.NET address the issue? That might be another way to handle this.

---
In reply to: [232431897](https://github.com/Microsoft/AdaptiveCards/pull/2163#discussion_r232431897) [](ancestors = 232431897)

shalinijoshi19:
Possibly a Json.NET update can fix this. Will look into this

PR #2166, opened 2018-11-12
[iOS] Add data Uri support

almedina-ms:
Solves this issue: https://github.com/Microsoft/AdaptiveCards/issues/2130


nesalang:
it looks like this PR picked up some previous commits.

almedina-ms:
Rebuilt the android version and it worked

image

PR #2177, opened 2018-11-16
Update markdown processing logic to prevent script injection

dclaux:
# Breaking change
With this change, TextBlock.text is treated as plain text if there is no markdown processor. Sanitation is deferred to the markdown processing logic (Markdown-it does it for instance.) Hosts that handle markdown processing will have to be updated.

# Updating host code
Previously, hosts would have to write this to handle markdown processing:
```
AdaptiveCard.processMarkdown = (text: string) => {
    return <processed text property>;
}
```

With this change, hosts now have to update to the following code:
```
AdaptiveCard.onProcessMarkdown = (text: string, result: IMarkdownProcessingResult) => {
    result.outputHtml = <processed text property>;
    result.didProcess = true;
}
```


matthidinger:
What will happen when people update the package and this break occurs? Does `processMarkdown` still exist? Will it silently fail and just assign a new property onto AdaptiveCard? Can you also update the nodejs\adaptivecards\README.md to include a "**Breaking Change Notice**? That README is what is displayed on NPM

dclaux:
@matthidinger processMarkdown still exists under the covers, it's now a method and not an assignable event handler. But JS is so permissive, it can still be replaced, which is what would happen. The end result would be broken card rendering. Maybe we want something a little more radical, e.g. when processMarkdown is set by the developer we throw an exception with an error that tells them to move to onProcessMarkdown. Would you prefer that? I think I would.

andrewleader:
That sounds like a good idea to me!

PR #2178, opened 2018-11-17
html quick actions 

nesalang:


source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
```diff
@@ -111,6 +111,17 @@ protected static HtmlTag AddActionAttributes(AdaptiveAction action, HtmlTag tag,
             return tag;
         }
 
+        protected static HtmlTag AddInlineActionAttributes(AdaptiveAction action, HtmlTag tag, AdaptiveRenderContext context)
+        {
+            tag.AddClass("ac-action-inlineAction")
```
paulcam206: >AddClass("ac-action-inlineAction") [](start = 16, length = 34)

can there be more than one `inlineAction`? if so, we might need a `GetInlineActionCssClass` method.

regardless, we should probably pick a better classname here. classnames on tags are case sensitive, but CSS selectors are not, so it would be difficult to style a non-inline action named "inlineaction" differently than an actual inline action.

nesalang: thanks

---
In reply to: [236835643](https://github.com/Microsoft/AdaptiveCards/pull/2178#discussion_r236835643) [](ancestors = 236835643)

Issue #2176, opened 2018-11-16
[adaptivecards.io/samples] Typo in Food Order Json


englishextra:
I believe that JSON sample here https://adaptivecards.io/samples/FoodOrder.html

misses a
```json
 "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
```
Thanks

paulcam206:
This has been addressed in master

PR #2180, opened 2018-11-19
Add schema ref to foodorder.json

paulcam206:
Fix for #2176


paulcam206:
Updated expected card sources :)

PR #2191, opened 2018-11-27
[UWP] Fix problem with async tasks not finishing before Auto size calculations are made

almedina-ms:
Fixes the crash where some parts of the auto image size calculations are made before the image has been loaded. I changed the auto sizing to be done after the image is loaded so we're sure the image source has been set and the properties are actually accesible.

Also, a resource resolver test was added to the test app so the resource resolver code is also checked when the test app is ran.


source/uwp/AdaptiveCardTestApp/ViewModels/RunningTestsViewModel.cs
```diff
@@ -151,7 +153,11 @@ public async Task AddCardToTimeline(FileViewModel card)
 
         private async Task<TestResultViewModel> TestCard(FileViewModel cardFile, FileViewModel hostConfigFile)
         {
-            var renderResult = await RenderCard(cardFile, hostConfigFile);
+            Dictionary<string, IAdaptiveCardResourceResolver> resourceResolvers = new Dictionary<string, IAdaptiveCardResourceResolver>();
+            resourceResolvers.Add("symbol", new SampleResourceResolver());
+
+
```
paulcam206: one `\n` too many :)

source/uwp/Renderer/lib/XamlBuilder.h
```diff
@@ -144,8 +151,13 @@ namespace AdaptiveNamespace
         void SetImageOnUIElement(_In_ ABI::Windows::Foundation::IUriRuntimeClass* imageUrl,
                                  T* uiElement,
                                  ABI::AdaptiveNamespace::IAdaptiveCardResourceResolvers* resolvers,
+                                 bool isSizeAuto,
+                                 IInspectable* parentElement,
+                                 IInspectable* imageContainer,
                                  _Out_ bool* mustHideElement,
-                                 ABI::Windows::UI::Xaml::Media::Stretch stretch = Stretch_UniformToFill);
+                                 ABI::Windows::UI::Xaml::Media::Stretch stretch = Stretch_UniformToFill
+            );
```
paulcam206: is this what format document (Ctrl-K Ctrl-D) gave you?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -570,9 +573,9 @@ namespace AdaptiveNamespace
 
                 ComPtr<T> strongImageControl(uiElement);
                 ComPtr<XamlBuilder> strongThis(this);
-                THROW_IF_FAILED(getResourceStreamOperation->put_Completed(
+                HRESULT hr = (getResourceStreamOperation->put_Completed(
```
paulcam206: >HRESULT hr = [](start = 16, length = 12)

why aren't we throwing here on failure any more?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -653,7 +663,7 @@ namespace AdaptiveNamespace
             ComPtr<XamlBuilder> strongThis(this);
             THROW_IF_FAILED(bufferWriteOperation->put_Completed(
                 Callback<Implements<RuntimeClassFlags<WinRtClassicComMix>, IAsyncOperationWithProgressCompletedHandler<UINT32, UINT32>>>(
-                    [strongThis, this, bitmapSource, randomAccessStream, strongImageControl](
+                    [strongThis, this, bitmapSource, randomAccessStream, strongImageControl, isAutoSize, parentElement, imageContainer /*ellipseAsShape, frameworkElement*/](
```
paulcam206: >/*ellipseAsShape, frameworkElement*/ [](start = 135, length = 36)

remove

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1710,6 +1731,94 @@ namespace AdaptiveNamespace
...
+                    &eventToken));
+            }
+        }
+    }
+
+    _Use_decl_annotations_ template<typename T>
+    void XamlBuilder::SetAutoSize(T* destination, IInspectable* parentElement, IInspectable* imageContainer, bool mustHideElement)
+    {
+        ComPtr<IInspectable> parentElement2(parentElement);
+        //if container is ellipse
```
paulcam206: >//if container is ellipse [](start = 8, length = 25)

is this accurate?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1710,6 +1731,94 @@ namespace AdaptiveNamespace
...
+                    // what it would be otherwise, just set the visibility when we get the image
+                    return ellipseAsUIElement->put_Visibility(Visibility::Visibility_Visible);
+                })
+                    .Get(),
+                    &eventToken));
+            }
+        }
+    }
+
+    _Use_decl_annotations_ template<typename T>
```
paulcam206: >_Use_decl_annotations_ [](start = 4, length = 22)

there aren't any declared annotations

paulcam206: (also above)

---
In reply to: [236882414](https://github.com/Microsoft/AdaptiveCards/pull/2191#discussion_r236882414) [](ancestors = 236882414)

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1710,6 +1731,94 @@ namespace AdaptiveNamespace
...
+    void XamlBuilder::SetAutoSize(T* destination, IInspectable* parentElement, IInspectable* imageContainer, bool mustHideElement)
+    {
+        ComPtr<IInspectable> parentElement2(parentElement);
```
paulcam206: >ComPtr<IInspectable> parentElement2(parentElement); [](start = 8, length = 51)

this should be scoped inside the conditional block below. also, consider using a better name

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1710,6 +1731,94 @@ namespace AdaptiveNamespace
...
+        //if container is ellipse
+        if (parentElement != nullptr && m_enableXamlImageHandling)
+        {
+            ComPtr<IImage> xamlImage(destination);
```
paulcam206: >ComPtr<IImage> xamlImage(destination); [](start = 12, length = 38)

declare this right before it's used

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1710,6 +1731,94 @@ namespace AdaptiveNamespace
...
+
+            // If the image hasn't loaded yet
+            if (mustHideElement)
+            {
+                // Collapse the Image control while the image loads, so that resizing is not noticeable
+                THROW_IF_FAILED(imageAsUIElement->put_Visibility(Visibility::Visibility_Collapsed));
+
+                // Handle ImageOpened event so we can check the imageSource's size to determine if it fits in its parent
+                EventRegistrationToken eventToken;
+                HRESULT hr = (xamlImage->add_ImageOpened(
```
paulcam206: >HRESULT hr [](start = 16, length = 10)

are we doing anything with this? should we throw?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1710,6 +1731,92 @@ namespace AdaptiveNamespace
         return S_OK;
     }
 
+    template<>
+    void XamlBuilder::SetAutoSize<IEllipse>(IEllipse* destination, IInspectable* parentElement, IInspectable* imageContainer, bool mustHideElement)
```
RebeccaAnne: >mustHideElement [](start = 131, length = 15)

Can we name this variable something more descriptive? It took me a bit to figure out what this represents. Maybe reverse it and call it IsImageLoaded.

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1710,6 +1731,92 @@ namespace AdaptiveNamespace
...
+            THROW_IF_FAILED(imageSource.As(&imageSourceAsBitmap));
+
+            // If the image hasn't loaded yet
+            if (mustHideElement)
+            {
+                // Collapse the Ellipse while the image loads, so that resizing is not noticeable
+                THROW_IF_FAILED(ellipseAsUIElement->put_Visibility(Visibility::Visibility_Collapsed));
+                // Handle ImageOpened event so we can check the imageSource's size to determine if it fits in its parent
+                EventRegistrationToken eventToken;
+                THROW_IF_FAILED(brushAsImageBrush->add_ImageOpened(
```
RebeccaAnne: >add_ImageOpened [](start = 51, length = 15)

Is it possible for the image to open before you set this handler?

almedina-ms: The problem we found was that the dataUri images and the images retrieved by resource resolvers wouldn't hit this event

---
In reply to: [239865550](https://github.com/Microsoft/AdaptiveCards/pull/2191#discussion_r239865550) [](ancestors = 239865550)

PR #2199, opened 2018-11-30
Introduce version range checking

paulcam206:
Part of work needed for #2019 in shared model -- introduce version range/wildcard support a la [NuGet](https://docs.microsoft.com/en-us/nuget/reference/package-versioning#version-ranges-and-wildcards).


source/shared/cpp/ObjectModel/SemanticVersion.h
```diff
@@ -4,55 +4,105 @@
...
+                return true;
+            }
+            else if (lhs.get(currSegment) != rhs.get(currSegment))
+            {
+                // not equal, no need to examine further segments.
+                return false;
+            }
+            else
+            {
+                // current segment matches. continue to next.
```
carlos-zamora: Add `continue;`?

source/shared/cpp/ObjectModel/SemanticVersion.h
```diff
@@ -4,55 +4,105 @@
...
+            const SemanticVersionSegment currSegment = static_cast<SemanticVersionSegment>(i);
+            const auto lhsVersion = lhs.get(currSegment);
+            const auto rhsVersion = rhs.get(currSegment);
+            if (lhs.hasWildcard(currSegment) || rhs.hasWildcard(currSegment))
+            {
+                return false;
+            }
+            else if (lhsVersion == rhsVersion)
+            {
+                // current segment matches. continue to next.
```
carlos-zamora: Add `continue;`?

source/shared/cpp/ObjectModel/SemanticVersion.h
paulcam206: Sadly, the examples in cppcoreguidelines don't seem to be available. However, according to [ES.77](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-continue), you're supposed to avoid `break;` and `continue;` where possible.

source/shared/cpp/ObjectModel/SemanticVersion.h
paulcam206: see other comment

source/shared/cpp/ObjectModel/SemanticVersion.h
```diff
@@ -4,55 +4,105 @@
...
+    enum class SemanticVersionSegment : unsigned int
+    {
+        Major = 0,
+        Minor,
+        Build,
+        Revision,
+        Max // must always be last
+    };
+
+    // Indicates if a SemanticVersion instance supports wildcards. We could use a bool here, but it complicates the
```
RebeccaAnne: >Indicates if a SemanticVersion instance supports wildcards [](start = 7, length = 58)

In what scenarios do we support wildcards vs not supporting them?

RebeccaAnne: >complicates [](start = 100, length = 11)

For my own information, what's the complication?

source/shared/cpp/ObjectModel/SemanticVersion.h
RebeccaAnne: Maybe remove the else block then? It looks like it's just asking for someone to add a continue to it :)


---
In reply to: [240842144](https://github.com/Microsoft/AdaptiveCards/pull/2199#discussion_r240842144) [](ancestors = 240842144)

source/shared/cpp/ObjectModel/SemanticVersion.h
```diff
@@ -4,55 +4,105 @@
...
-        {
-            return false;
+            const SemanticVersionSegment currSegment = static_cast<SemanticVersionSegment>(i);
+            const auto lhsVersion = lhs.get(currSegment);
+            const auto rhsVersion = rhs.get(currSegment);
+            if (lhs.hasWildcard(currSegment) || rhs.hasWildcard(currSegment))
+            {
+                return false;
+            }
+            else if (lhsVersion == rhsVersion)
```
RebeccaAnne: >lhsVersion == rhsVersion [](start = 21, length = 24)

Make this a != and remove the empty clause?

paulcam206:
This work, sadly, is now moot after a spec change.

Issue #2037, opened 2018-10-12
Bug: When using .NET SDK and setting Spacing=Large, it also sets deprecated Separation property


andrewleader:
# Platform

* .NET

# Author or host

Author creating cards with .NET SDK

# Version of SDK

AdaptiveCards 1.1.0

# Issue

When adding a text block with Spacing=Large and then serializing to JSON, it also adds the deprecated Separation property, generating a warning in the Visualizer (and incorrectly adding a separator when I didn't want one).

![image](https://user-images.githubusercontent.com/13246069/46897863-1a54f300-ce3a-11e8-9104-1c90085742c1.png)

![image](https://user-images.githubusercontent.com/13246069/46897856-12954e80-ce3a-11e8-867d-bf9aaf090628.png)

khouzam:
This has been merged through PR #2275

PR #2200, opened 2018-11-30
[.NET] Fix separation strong issue

almedina-ms:
Changes the requisite for writing the separation as Strong by requiring spacing to be large and separator to be true, this doesnt modify the general behaviour for spacing so it isnt a breaking change

Fixes this: https://github.com/Microsoft/AdaptiveCards/issues/2037

### DIFF
--- a/source/dotnet/Library/AdaptiveCards/AdaptiveElement.cs
+++ b/source/dotnet/Library/AdaptiveCards/AdaptiveElement.cs
@@ -54,7 +54,7 @@ namespace AdaptiveCards
                         return AdaptiveSeparationStyle.None;
 
                     case AdaptiveSpacing.Large:
-                        return AdaptiveSeparationStyle.Strong;
+                        return Separator ? AdaptiveSeparationStyle.Strong : AdaptiveSeparationStyle.Default;
 
                     default:
                         return AdaptiveSeparationStyle.Default;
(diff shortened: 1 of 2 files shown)


andrewleader:
While this fixes the visual rendering issue, this will still mean people will get warnings about using the separation property even though they never assigned it (they only assigned spacing=none, for example). So I personally disagree with this fix.

I still don't understand the scenario as to why we're back-porting this property. There's some webchat **hosts** that are still using version 0.5? And the concern is that a card author uses Adaptive 1.1 and therefore doesn't provide the deprecated `separation` property and therefore it doesn't render correctly?

Isn't that true with any new attributes/elements we add? If a card author uses Adaptive 1.1 media element, that isn't going to render on webchat hosts running 0.5 either.

matthidinger:
This fix actually does address the serialization issue as intended. For 0.5 back compat in 1.0 the intention was that if a developer set Spacing to Large AND Separator to true then it would also set the legacy Separation property to Strong. The bug is that it wasn't checking the Separator as well.

The reason was this was so a .NET backend could use the 1.0 package while still targeting 0.5 clients. 

That said, the properties have been Obsolete a while now and our plan is to remove them in 1.2.

khouzam:
Since our plan is to remove them from 1.2, why not do it right now since this is the 1.2 codebase?

PR #2202, opened 2018-12-03
[.NET] Remove spacing property for v1.2

almedina-ms:
Removes the spacing property from version 1.2 as it's obsolete 

Not made in the same PR (https://github.com/Microsoft/AdaptiveCards/pull/2200 ) as fixing the separation bug as this may allow to cherrypick the fix easier


andrewleader:
@matthidinger I thought we're not doing breaking changes till 2.0?

If we're removing obsolete properties in 1.2, why don't we remove the obsolete CardElement.Speak property too, right above it? We might as well get rid of all of the obsolete properties at once (there might be others), so that we only burden developers one time, rather than multiple times.

```
[Obsolete("CardElement.Speak has been deprecated.  Use AdaptiveCard.Speak", false)]
public string Speak { get; set; }
```

PR #2203, opened 2018-12-03
[UWP] Fix negative column ratio

almedina-ms:
Looks like we were actually using the value 0 for stretch as it may mean that something as Auto or Stretch or any string that is not a number may have been written 

Earlier PR: https://github.com/Microsoft/AdaptiveCards/pull/2194


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -2294,13 +2294,13 @@ namespace AdaptiveNamespace
...
                 }
-                else if ((isAutoResult == 0) || (widthAsDouble <= 0))
+                else if ((isAutoResult == 0) || (widthAsDouble < 0))
                 {
                     // If auto specified, use auto width
                     columnWidth.GridUnitType = GridUnitType::GridUnitType_Auto;
                     columnWidth.Value = 0;
                 }
-                else if (isStretchResult == 0 || !adaptiveColumnWidth.IsValid())
+                else if (isStretchResult == 0 || !adaptiveColumnWidth.IsValid() || (widthAsDouble < 0))
```
khouzam: >< [](start = 98, length = 1)

Should this be ==

RebeccaAnne: What do we want the behavior to be if someone puts "width" : 0 in their card? or "width" : "foobar"? Seems like those should have the same behavior as negative width, basically that all invalid results should be the same.

I made a fix locally before i saw this where i had the following logic:

                else if (isStretchResult == 0 || !adaptiveColumnWidth.IsValid())
                {
                    // If stretch specified, or column width invalid (unset), use stretch with default of 1
                    columnWidth.GridUnitType = GridUnitType::GridUnitType_Star;
                    columnWidth.Value = 1;
                }
                else if ((isAutoResult == 0) || (widthAsDouble <= 0))
                {
                    // If auto specified, or set to non-positive, use auto width
                    columnWidth.GridUnitType = GridUnitType::GridUnitType_Auto;
                    columnWidth.Value = 0;
                }

---
In reply to: [238446521](https://github.com/Microsoft/AdaptiveCards/pull/2203#discussion_r238446521) [](ancestors = 238446521)

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -2294,13 +2294,13 @@ namespace AdaptiveNamespace
...
                 {
                     // If stretch specified, or column width invalid or set to non-positive, use stretch with default of 1
```
RebeccaAnne: >or set to non-positive [](start = 69, length = 22)

Update comment

source/uwp/Renderer/lib/XamlBuilder.cpp
almedina-ms: The problem is that wtof sets the value 0 to widthAsDouble if the string couldn't be parsed to a number (https://docs.microsoft.com/en-us/previous-versions/hc25t012(v=vs.140) )  so validating with <= 0 was what actually created this issue

Issue #2206, opened 2018-12-05
[UWP] Action.ToggleVisibility


khouzam:

RebeccaAnne:
Implemented by #2221

Issue #2205, opened 2018-12-05
[Shared] Action.ToggleVisibility


khouzam:

RebeccaAnne:
Implemented by #2221

PR #2221, opened 2018-12-07
Add Action.ToggleVisibility to the shared model and UWP

RebeccaAnne:
Implements #2205 and #2206


source/uwp/Renderer/lib/AdaptiveSubmitAction.h
```diff
@@ -40,8 +40,14 @@ namespace AdaptiveNamespace
...
-        IFACEMETHODIMP get_Sentiment(_Out_ ABI::AdaptiveNamespace::Sentiment *sentiment) { return AdaptiveActionElementBase::get_Sentiment(sentiment); }
-        IFACEMETHODIMP put_Sentiment(_In_ ABI::AdaptiveNamespace::Sentiment sentiment) { return AdaptiveActionElementBase::put_Sentiment(sentiment); }
+        IFACEMETHODIMP get_Sentiment(_Out_ ABI::AdaptiveNamespace::Sentiment* sentiment)
+        {
+            return AdaptiveActionElementBase::get_Sentiment(sentiment);
+        }
+        IFACEMETHODIMP put_Sentiment(_In_ ABI::AdaptiveNamespace::Sentiment sentiment)
+        {
+            return AdaptiveActionElementBase::put_Sentiment(sentiment);
+        }
```
almedina-ms: Should we worry about the consistency of 

get_property(prop){ return base::get_property(prop) }
vs
get_property(prop)
{ 
    return base::get_property(prop) 
}

as the style differs based on the length of the property name?

RebeccaAnne: This is a result of using automatic formatting. I agree that it's not great that these are not consistent (it's formatting based on a max line length we've specified) but i think the consistency benefit of having automatic formatting in general outweighs the downside of cases like these where it's not consistent. @paulcam206 in case he has something to add as he's focused on this more than i have.

---
In reply to: [242272890](https://github.com/Microsoft/AdaptiveCards/pull/2221#discussion_r242272890) [](ancestors = 242272890)

PR #2233, opened 2018-12-14
[UWP] Add the parsers for known elements to the UWP parser registrations

### LINKED ISSUE #2196: Bug: [UWP] AdaptiveActionParserRegistration doesn't have default parsers for Action.Submit / etc
# Platform

* UWP

# Author or host

Host trying to parse a card in which I've created my own ActionSet element

# Version of SDK

1.1 from NuGet

# Issue

I've added a custom element parser registration for my own ActionSet element... that all works correctly.

However, when I then try to use the default actions parser registration, it fails to get any parser for `Action.Submit`, even though it definitely should have a default parser for that. The element parser registration has default parsers for everything.

image

image

I tried getting the parser for `Action.OpenUrl` too, but it also returns null.

RebeccaAnne:
Fixes #2196

### DIFF
--- a/source/uwp/UWPUnitTests/ParserRegistration.cs
+++ b/source/uwp/UWPUnitTests/ParserRegistration.cs
@@ -227,7 +227,7 @@ namespace UWPUnitTests
                 "           \"type\":\"TestCustomAction\"," +
                 "           \"internalSubmitAction\":" +
                 "           {" +
-                "			    \"type\": \"Action.Submit\"" +
+                "               \"type\": \"Action.Submit\"" +
                 "           }" +
                 "       }" +
                 "   ]" +


source/uwp/UWPUnitTests/ParserRegistration.cs
```diff
@@ -0,0 +1,253 @@
...
+                "   [" +
+                "   ]," +
+                "   \"actions\":" +
+                "   [" +
+                "       {" +
+                "           \"type\":\"TestCustomAction\"," +
+                "           \"internalSubmitAction\":" +
+                "           {" +
+                "			    \"type\": \"Action.Submit\"" +
+                "           }" +
```
khouzam: Couple of tabs

PR #2250, opened 2018-12-19
Change RenderResults to be WeakReferences

khouzam:
Because the RenderedAdaptiveCard holds on to the root of the tree.
The ClickHandler for the action button hangs on to the ActionInvoker.
This creates a cycle causing the whole AdaptiveCard and XAML tree to leak.
By switching to a WeakRef on the ActionInvoker, this will allow the RenderedAdaptiveCard to be released and the whole tree to go away.
Also setting the RenderContext to use a weakRef on the RenderedAdaptiveCard to prevent another leak.

With the change #2249, we are now leak free in the AdaptiveCardTestApp.

### DIFF
--- a/source/uwp/Renderer/lib/AdaptiveActionInvoker.cpp
+++ b/source/uwp/Renderer/lib/AdaptiveActionInvoker.cpp
@@ -15,14 +15,18 @@ namespace AdaptiveNamespace
 
     HRESULT AdaptiveActionInvoker::RuntimeClassInitialize(_In_ RenderedAdaptiveCard* renderResult) noexcept try
     {
-        m_renderResult = renderResult;
-        return S_OK;
-    }
-    CATCH_RETURN;
+        ComPtr<IRenderedAdaptiveCard> strongRenderResult = renderResult;
+        return strongRenderResult.AsWeak(&m_weakRenderResult);
+    } CATCH_RETURN;
 
     HRESULT AdaptiveActionInvoker::SendActionEvent(_In_ IAdaptiveActionElement* actionElement)
     {
-        return m_renderResult->SendActionEvent(actionElement);
+        ComPtr<IRenderedAdaptiveCard> strongRenderResult;
+        RETURN_IF_FAILED(m_weakRenderResult.As(&strongRenderResult));
+        RETURN_IF_FAILED(strongRenderResult == nullptr ? E_FAIL : S_OK);
+        ComPtr<RenderedAdaptiveCard> renderResult = PeekInnards<RenderedAdaptiveCard>(strongRenderResult);
+        RETURN_IF_FAILED(renderResult == nullptr ? E_FAIL : S_OK);
+        RETURN_IF_FAILED(renderResult->SendActionEvent(actionElement));
+        return S_OK;
     }
-
 }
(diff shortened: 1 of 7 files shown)


matthidinger:
Nice work! Any idea if we have a similar concern in the .NET renderers?

khouzam:
I've checked .NET has automatic garbage collection and we don't leak cards. Even if the same cycle was done in .NET, once the tree got released, everything else would also be released. With native code, since there's a reference cycle, there isn't a way for that to be garbage collected.

source/uwp/Renderer/lib/AdaptiveActionInvoker.cpp
```diff
@@ -15,14 +15,18 @@ namespace AdaptiveNamespace
...
+        ComPtr<IRenderedAdaptiveCard> strongRenderResult = renderResult;
+        return strongRenderResult.AsWeak(&m_weakRenderResult);
+    } CATCH_RETURN;
 
     _Use_decl_annotations_ HRESULT AdaptiveActionInvoker::SendActionEvent(IAdaptiveActionElement* actionElement)
     {
-        return m_renderResult->SendActionEvent(actionElement);
+        ComPtr<IRenderedAdaptiveCard> strongRenderResult;
+        RETURN_IF_FAILED(m_weakRenderResult.As(&strongRenderResult));
+        RETURN_IF_FAILED(strongRenderResult == nullptr ? E_FAIL : S_OK);
```
MikeHillberg: >RETURN_IF_FAILED(strongRenderResult == nullptr ? E_FAIL : S_OK); [](start = 8, length = 64)

Is this really an error? It's an expected case.

What CX and C# in their event source implementations is special-case when the delegate returns RPC_E_DISCONNECTED (I think that's the right error code).  In that case, the event source doesn't blow up, but does remove the listener from its list.

khouzam:
After thinking about it some more, This can break existing scenarios. If the client doesn't hold on to the RenderedCard then the submit actions will loose their reference and won't be able to be invoked. I need to think about this some more.

paulcam206:
Looks good to me!

source/uwp/Renderer/lib/AdaptiveRenderContext.cpp
```diff
@@ -99,11 +105,24 @@ namespace AdaptiveNamespace
...
     Microsoft::WRL::ComPtr<ABI::Windows::UI::Xaml::IResourceDictionary> AdaptiveRenderContext::GetDefaultActionSentimentDictionary()
     {
         return m_actionSentimentDefaultDictionary;
     }
+
+    HRESULT AdaptiveRenderContext::GetRenderResult(AdaptiveCards::Rendering::Uwp::RenderedAdaptiveCard** renderResult)
+    {
+        ComPtr<IRenderedAdaptiveCard> strongRenderResult;
+        RETURN_IF_FAILED(m_weakRenderResult.As(&strongRenderResult));
+        RETURN_IF_FAILED(strongRenderResult == nullptr ? E_FAIL : S_OK);
```
RebeccaAnne: >RETURN_IF_FAILED(strongRenderResult == nullptr ? E_FAIL : S_OK); [](start = 8, length = 64)

Should we add a RETURN_IF_NULL macro?

RebeccaAnne:
Friendly reminder to log an issue to follow up on this with a "better" fix in a future release.

---
In reply to: [449497451](https://github.com/Microsoft/AdaptiveCards/pull/2250#issuecomment-449497451) [](ancestors = 449497451)

Issue #2208, opened 2018-12-05
[Android] Action.ToggleVisibility


khouzam:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

Android implementation of #1729

PR #2252, opened 2018-12-19
[Android] Add ToggleVisibility action support

almedina-ms:
Completes this work item: https://github.com/Microsoft/AdaptiveCards/issues/2208


samples/Tests/ToogleVisibility.AllElements.json
```diff
@@ -0,0 +1,276 @@
...
+    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+    "type": "AdaptiveCard",
+    "version": "1.0",
+    "body": [
+		{
+			"type": "Image",
+			"url": "http://adaptivecards.io/content/cats/1.png",
+			"id": "id1",
+			"selectAction": {
+                "type": "Action.ToggleVisibility",
```
RebeccaAnne: >                 [](start = 0, length = 16)

use tabs in json files

source/uwp/Renderer/lib/Util.h
```diff
@@ -149,6 +149,10 @@ HRESULT GenerateToggleTargetProjection(
     const std::vector<std::shared_ptr<AdaptiveSharedNamespace::ToggleVisibilityTarget>>& containedElements,
     _In_ ABI::Windows::Foundation::Collections::IVector<ABI::AdaptiveNamespace::AdaptiveToggleVisibilityTarget*>* projectedParentContainer) noexcept;
 
+HRESULT GenerateToggleTargetProjection(
+    const std::vector<std::shared_ptr<AdaptiveSharedNamespace::ToggleVisibilityTarget>>& containedElements,
+    ABI::Windows::Foundation::Collections::IVector<ABI::AdaptiveNamespace::AdaptiveToggleVisibilityTarget*>* projectedParentContainer) noexcept;
+
```
RebeccaAnne: Merge issue? Remove duplicate declaration.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/action/ActionElementRenderer.java
```diff
@@ -68,7 +74,54 @@ public ButtonOnClickListener(RenderedAdaptiveCard renderedCard, BaseActionElemen
...
         {
-            m_cardActionHandler.onAction(m_action, m_renderedAdaptiveCard);
+            if(m_action.GetElementType() == ActionType.ToggleVisibility)
+            {
+                ToggleVisibilityAction toggleVisibilityAction = null;
+                if (m_action instanceof ToggleVisibilityAction)
+                {
+                    toggleVisibilityAction = (ToggleVisibilityAction) m_action;
+                }
+                else if ((toggleVisibilityAction = ToggleVisibilityAction.dynamic_cast(m_action)) == null)
```
RebeccaAnne: Can you tell me why we need both of these checks? What is the case where both the if and the else if are false?

almedina-ms: As for the existence of the two checks, that's done for all the elements and actions when they are rendered, as far as I understand the first check is the Java casting version and the second check is the swig casting method; if both of those checks were to fail that means the object is not of the desired type and the action type was modified by something else but the action is not really a ToggleVisibility Action

---
In reply to: [251115166](https://github.com/Microsoft/AdaptiveCards/pull/2252#discussion_r251115166) [](ancestors = 251115166)

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/TagContent.java
```diff
@@ -0,0 +1,86 @@
...
+    {
+        m_id = id;
+    }
+
+    public TagContent(BaseCardElement baseCardElement)
+    {
+        m_id = baseCardElement.GetId();
+        m_baseElement = baseCardElement;
+    }
+
```
khouzam: Should m_id just read the ID from BaseCardElement to avoid discrepancies between the ID in TagContent and the one in the BaseCardElement. If someone calls SetId and changes the Id, should that be reflected in the basecardelement.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/TagContent.java
```diff
@@ -0,0 +1,86 @@
...
+import io.adaptivecards.renderer.inputhandler.IInputHandler;
+
+public class TagContent
+{
+
+    public TagContent(String id)
+    {
+        m_id = id;
+    }
+
```
khouzam: This shouldn't be necessary

source/uwp/AdaptiveCardTestApp/Expected/SourceCards/Tests.ToogleVisibility.AllElements.beea57a.json
```diff
@@ -0,0 +1,276 @@
...
+		{
+			"type": "Image",
+			"url": "http://adaptivecards.io/content/cats/1.png",
+			"id": "id1",
+			"selectAction": {
+                "type": "Action.ToggleVisibility",
+                "targetElements": [
+                    "id2"
+                ]
+            }
```
RebeccaAnne: >             [](start = 0, length = 12)

This result file still has spaces in it. You should re-run the test app as i think that will count as "source changed" for these results.

samples/Tests/ToogleVisibility.AllElements.json
```diff
@@ -0,0 +1,274 @@
+{
```
paulcam206: nitpick: filename has a typo :)

source/uwp/AdaptiveCardTestApp/Expected/microsoft-teams.Tests.ToogleVisibility.AllElements.json
```diff
@@ -0,0 +1 @@
+{"HostConfigHash":"5da68b8","CardHash":"beea57a","Error":null}
```
RebeccaAnne: We should probably delete the results files associated with the typo name.

source/uwp/AdaptiveCardTestApp/Expected/outlook-desktop.v1.2.Elements.Action.ToggleInput.json
```diff
@@ -0,0 +1 @@
+{"HostConfigHash":"ee139d0","CardHash":"feae65c","Error":null}
```
RebeccaAnne: Where are these ToggleInput files coming from?

PR #2260, opened 2018-12-26
Fix Container layout issue on Android.

xinguohenan:
On Android, LinearLayout's orientation is set as horizontal by default. However, for Container layout, we should set orientation as vertical. So in this PR, I try to set containerView's orientation to vertical explicitly.


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ContainerRenderer.java
```diff
@@ -65,9 +65,14 @@ else if ((container = Container.dynamic_cast(baseCardElement)) == null)
         ContainerStyle styleForThis = container.GetStyle().swigValue() == ContainerStyle.None.swigValue() ? containerStyle : container.GetStyle();
         LinearLayout containerView = new LinearLayout(context);
 
+        containerView.setOrientation(LinearLayout.VERTICAL);
```
xinguohenan: test

PR #2264, opened 2018-12-29
Jwoo/uwp wrap

nesalang:


samples/Tests/SearchResults.json
```diff
@@ -6,18 +6,18 @@
...
+			"horizontalAlignment": "center"
         },
 		{
 			"type": "Input.ChoiceSet",
 			"id": "filament",
 			"style": "expanded",
 			"isMultiSelect": false,
 			"value": "1",
-            "wrap": true,
+			"wrap": true,
```
RebeccaAnne: >"wrap": true, [](start = 3, length = 13)

I'm looking at the test app results, and it doesn't look like these choices are wrapping...

RebeccaAnne:
>     {

Set wrap back on the shared model in this function.

---
Refers to: source/uwp/Renderer/lib/AdaptiveChoiceSetInput.cpp:109 in 52c6148. [](commit_id = 52c6148a00c43fc167aa49a56ae005cdf61cb8e7, deletion_comment = False)

RebeccaAnne:
>     {

Set wrap here too

---
Refers to: source/uwp/Renderer/lib/AdaptiveToggleInput.cpp:77 in 52c6148. [](commit_id = 52c6148a00c43fc167aa49a56ae005cdf61cb8e7, deletion_comment = False)

source/uwp/Renderer/lib/AdaptiveToggleInput.h
```diff
@@ -25,6 +25,9 @@ namespace AdaptiveNamespace
         IFACEMETHODIMP get_Title(_Outptr_ HSTRING* title);
         IFACEMETHODIMP put_Title(_In_ HSTRING title);
 
+        IFACEMETHODIMP get_Wrap(_Outptr_ boolean* wrap);
```
RebeccaAnne: >_Outptr_ [](start = 32, length = 8)

Should be _Out_

source/uwp/Renderer/lib/AdaptiveToggleInput.h
```diff
@@ -25,6 +25,9 @@ namespace AdaptiveNamespace
...
+        IFACEMETHODIMP put_Wrap(_In_ boolean wrap);
```
RebeccaAnne: >_In_ [](start = 32, length = 4)

Remove unneeded annotation on data type

source/uwp/Renderer/lib/AdaptiveChoiceSetInput.h
```diff
@@ -32,6 +32,9 @@ namespace AdaptiveNamespace
         IFACEMETHODIMP get_IsMultiSelect(_Out_ boolean* isMultiSelect);
         IFACEMETHODIMP put_IsMultiSelect(boolean isMultiSelect);
 
+        IFACEMETHODIMP get_Wrap(_Outptr_ boolean* wrap);
+        IFACEMETHODIMP put_Wrap(_In_ boolean wrap);
```
RebeccaAnne: Annotations here too

source/uwp/Renderer/lib/AdaptiveChoiceSetInput.cpp
```diff
@@ -51,6 +52,18 @@ namespace AdaptiveNamespace
         return S_OK;
     }
 
+    HRESULT AdaptiveChoiceSetInput::put_Wrap(boolean wrap)
+    {
+        m_wrap = wrap;
+        return S_OK;
+    }
+
+    HRESULT AdaptiveChoiceSetInput::get_Wrap(boolean* wrap)
```
RebeccaAnne: >boolean* wrap [](start = 45, length = 13)

add _ Out _ annotation

source/uwp/Renderer/lib/AdaptiveToggleInput.cpp
```diff
@@ -54,6 +55,18 @@ namespace AdaptiveNamespace
 
     HRESULT AdaptiveToggleInput::put_ValueOn(_In_ HSTRING valueOn) { return m_valueOn.Set(valueOn); }
 
+    HRESULT AdaptiveToggleInput::get_Wrap(boolean* wrap)
```
RebeccaAnne: >boolean* wrap [](start = 42, length = 13)

Add _ Out _ annotation

source/uwp/Renderer/lib/AdaptiveToggleInput.h
nesalang: what's reason for using Out insteand of Outptr?

---
In reply to: [245450087](https://github.com/Microsoft/AdaptiveCards/pull/2264#discussion_r245450087) [](ancestors = 245450087)

source/uwp/Renderer/lib/AdaptiveToggleInput.h
nesalang: when annotation is necessary?

---
In reply to: [245450181](https://github.com/Microsoft/AdaptiveCards/pull/2264#discussion_r245450181) [](ancestors = 245450181)

source/uwp/Renderer/lib/AdaptiveToggleInput.h
RebeccaAnne: Basically outptr is for when the out parameter is a pointer. From the sal header:

"_Out_  : output parameter, written to by called function, pointed-to location not expected to be initialized prior to call
_Outptr_ : like _Out_ when returned variable is a pointer type(so param is pointer-to-pointer type). Called function provides/allocated space."

---
In reply to: [245450521](https://github.com/Microsoft/AdaptiveCards/pull/2264#discussion_r245450521) [](ancestors = 245450521,245450087)

source/uwp/Renderer/lib/AdaptiveToggleInput.h
RebeccaAnne: The short answer is when it provides information that we can't otherwise encode in the language, such as whether it's expected that the caller will be passing in a value for a pointer (_In_) or if it's expected to be set as an out parameter (_Out_). In the case of a plain data type, the annotation doesn't tell us anything (a plain data type is already definitionally "In"). 

---
In reply to: [245450686](https://github.com/Microsoft/AdaptiveCards/pull/2264#discussion_r245450686) [](ancestors = 245450686,245450181)

nesalang:
>     {

thanks, it fixed round trip failure

---
In reply to: [451599753](https://github.com/Microsoft/AdaptiveCards/pull/2264#issuecomment-451599753) [](ancestors = 451599753,451598451)

---
Refers to: source/uwp/Renderer/lib/AdaptiveToggleInput.cpp:77 in 52c6148. [](commit_id = 52c6148a00c43fc167aa49a56ae005cdf61cb8e7, deletion_comment = False)

PR #2265, opened 2018-12-31
Fix Input.ChoiceSet description

v-kydela:
I noticed a typo in adaptive-card.json. The description of the `type` property of `Input.ChoiceSet` was ``"Must be `\"Input.ChoiceInput\"`."`` instead of ``"Must be `\"Input.ChoiceSet\"`."``


paulcam206:
Thanks!

PR #2266, opened 2019-01-01
Jwoo/wpf wrap

nesalang:


source/shared/cpp/ObjectModel/ChoiceSetInput.cpp
```diff
@@ -6,7 +6,8 @@
 using namespace AdaptiveSharedNamespace;
 
 ChoiceSetInput::ChoiceSetInput() :
-    BaseInputElement(CardElementType::ChoiceSetInput), m_isMultiSelect(false), m_choiceSetStyle(ChoiceSetStyle::Compact)
+    BaseInputElement(CardElementType::ChoiceSetInput), m_isMultiSelect(false),
```
paulcam206: It's best practice to order member initializers by order of declaration.

nesalang: thanks

---
In reply to: [245150591](https://github.com/Microsoft/AdaptiveCards/pull/2266#discussion_r245150591) [](ancestors = 245150591)

PR #2270, opened 2019-01-03
Misc. function signature fixes for UWP

paulcam206:


paulcam206:
Rebased on to @RebeccaAnne 's SAL changes

source/uwp/Renderer/lib/AdaptiveCard.cpp
```diff
@@ -40,7 +40,7 @@ namespace AdaptiveNamespace
         std::string adaptiveJsonString;
         RETURN_IF_FAILED(HStringToUTF8(adaptiveJson, adaptiveJsonString));
 
-        return FromJsonString(adaptiveJsonString, elementParserRegistration, actionParserRegistration, parseResult);
+        return _FromJsonString(adaptiveJsonString, elementParserRegistration, actionParserRegistration, parseResult);
```
RebeccaAnne: >_ [](start = 15, length = 1)

This is not a convention we've been using in general in our code, and it looks like our guidelines are against it ("Do not prefix methods with a scoping prefix.")

Issue #2280, opened 2019-01-08
ClassCastException io.adaptivecards.objectmodel.DateInput cannot be cast to io.adaptivecards.objectmodel.TextInput


xinguohenan:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* Android

# Author or host

# Version of SDK

Latest develop code.

# Issue
When using the following json to render, 
```
{
	"type": "AdaptiveCard",
	"version": "1.0",
	"body": [{
		"type": "Input.Text",
		"id": "title",
		"placeholder": "Subject",
		"value": ""
	},
	{
		"type": "Input.PeoplePicker",
		"id": "contact_name",
		"placeholder": "To",
		"value": "",
		"callback": {
			"type": "Action.Callback",
			"url": "cortana/autosuggest/people/adaptive",
			"targetElementId": null,
			"parameters": {
				"provider": "destination_calendar",
				"term": "contact_name"
			}
		}
	},
	{
		"type": "Input.Date",
		"id": "start_date",
		"placeholder": "Date",
		"value": "2019-01-08"
	},
	{
		"type": "ColumnSet",
		"columns": [{
			"type": "Column",
			"width": "stretch",
			"items": [{
				"type": "Input.Time",
				"id": "start_time",
				"placeholder": "Start Time",
				"value": "15:30"
			}]
		},
		{
			"type": "Column",
			"width": "auto",
			"items": [{
				"type": "TextBlock",
				"text": "To"
			}]
		},
		{
			"type": "Column",
			"width": "stretch",
			"items": [{
				"type": "Input.Time",
				"id": "end_time",
				"placeholder": "End Time",
				"value": "16:30"
			}]
		}]
	},
	{
		"type": "Input.ChoiceSet",
		"id": "destination_calendar",
		"value": "Office365",
		"style": "compact",
		"isMultiSelect": false,
		"choices": [{
			"title": "Calendar-Office365",
			"value": "Office365"
		},
		{
			"title": "Calendar-Outlook",
			"value": "Outlook"
		}]
	},
	{
		"type": "Input.Text",
		"id": "implicit_location",
		"placeholder": "Add a location"
	}],
	"actions": [{
		"type": "Action.Submit",
		"data": {
			"order_ref": "first"
		},
		"title": "Add"
	},
	{
		"type": "Action.Submit",
		"data": {
			"intent:cancel": "true"
		},
		"title": "Cancel"
	}],
	"-ms-autofocus": "start_date"
}
[...]

PR #2281, opened 2019-01-08
fixes issue 2280

nesalang:
fixes issue #2280

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/TextInputRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/TextInputRenderer.java
@@ -187,76 +187,68 @@ public class TextInputRenderer extends BaseCardElementRenderer
             }
         });
 
-        TextInput textInput = (TextInput)baseInputElement;
-        BaseActionElement action = textInput.GetInlineAction();
-
         LinearLayout textInputViewGroup = null;
-        if(action != null)
+
+        if(baseInputElement instanceof TextInput)
         {
-            if (hostConfig.GetActions().getShowCard().getActionMode() == ActionMode.Inline &&
-                (action.GetElementType() == ActionType.ShowCard))
-            {
-                renderedCard.addWarning(new AdaptiveWarning(AdaptiveWarning.INTERACTIVITY_DISALLOWED, "Inline ShowCard not supported for InlineAction"));
-            }
-            else
-            {
-                textInputViewGroup = new LinearLayout(context);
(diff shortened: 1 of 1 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/TextInputRenderer.java
```diff
@@ -187,76 +187,68 @@ public void afterTextChanged(Editable s)
...
-                        inlineButton = new Button(themedContext, null, 0);
-                    }
-                    else
-                    {
-                        inlineButton = new Button(context);
-                        inlineButton.setBackgroundColor(Color.TRANSPARENT);
-                        inlineButton.setTextColor(Color.BLACK);
-                        inlineButton.setPadding(16, 0, 0, 8);
+        if(baseInputElement instanceof TextInput)
+        {
```
paulcam206: should `{` go on this line, or at the end of the line above? inconsistent

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/TextInputRenderer.java
```diff
@@ -187,76 +187,68 @@ public void afterTextChanged(Editable s)
...
-                        Context themedContext = new ContextThemeWrapper(context, R.style.adaptiveInlineAction);
-                        inlineButton = new Button(themedContext, null, 0);
-                    }
-                    else
-                    {
-                        inlineButton = new Button(context);
-                        inlineButton.setBackgroundColor(Color.TRANSPARENT);
-                        inlineButton.setTextColor(Color.BLACK);
-                        inlineButton.setPadding(16, 0, 0, 8);
+        if(baseInputElement instanceof TextInput)
```
paulcam206: what should we do if `baseInputElement` is *not* an instance of `TextInput`? is that an expected case?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/TextInputRenderer.java
nesalang: thanks

nesalang: thanks

---
In reply to: [246190275](https://github.com/Microsoft/AdaptiveCards/pull/2281#discussion_r246190275) [](ancestors = 246190275)

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/TextInputRenderer.java
nesalang: yes, this expected case. NumberInput instance can call this method.


---
In reply to: [246190835](https://github.com/Microsoft/AdaptiveCards/pull/2281#discussion_r246190835) [](ancestors = 246190835)

PR #2283, opened 2019-01-09
fixed compiler warning

nesalang:
minor change


paulcam206:
which compiler is throwing the warning? I actually added this local variable because the cppcoreguidelines checker notes that without the local, it's a violation of [ES.84](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es84-dont-try-to-declare-a-local-variable-with-no-name)

nesalang:
it's xcode, apple clang.

---
In reply to: [452909709](https://github.com/Microsoft/AdaptiveCards/pull/2283#issuecomment-452909709) [](ancestors = 452909709)

Issue #2285, opened 2019-01-10
[.NET] Bug: TextBlock text won't use appropriate ContainerStyle foreground color


carlos-zamora:
# Platform
* .NET HTML
* .NET WPF

# Issue
Text in a `TextBlock` is supposed to inherit `ForegroundColor` from the `ContainerStyle` it is in. Instead, it is getting the one from `ContainerStyle.Default`.

Consider this card that should look different between the two presented HostConfigs:
## Card
```
{
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "type": "AdaptiveCard",
  "version": "1.1",
  "body": [
    {
      "type": "Container",
      "style": "default",
      "items": [
        {
          "type": "TextBlock",
          "text": "This container has the default style"
        }
      ]
    },
    {
      "type": "Container",
      "style": "emphasis",
      "items": [
        {
          "type": "TextBlock",
          "text": "This container has the default style"
        }
      ]
    }
  ]
}
```

## Host Config 1
This one has the `Default.ForegroundColors.Default.Default` == `Emphasis.ForegroundColors.Default.Default`
```
{
	"containerStyles": {
		"default": {
			"backgroundColor": "#00000000",
			"foregroundColors": {
				"default": {
					"default": "#FF333333"
				}
      }
		},
		"emphasis": {
			"backgroundColor": "#08000000",
			"foregroundColors": {
				"default": {
					"default": "#FF333333"
				}
			}
		}
  }
}
```

## Host Config 2
This one has the `Default.ForegroundColors.Default.Default` != `Emphasis.ForegroundColors.Default.Default`
```
{
	"containerStyles": {
		"default": {
			"backgroundColor": "#00000000",
			"foregroundColors": {
				"default": {
					"default": "#FF333333"
				}
      }
		},
		"emphasis": {
			"backgroundColor": "#08000000",
			"foregroundColors": {
				"default": {
					"default": "#FFcc3300"
				}
			}
		}
  }
}
```

Issue #1886, opened 2018-09-11
Additional built-in container styles


dclaux:
<table>
<tr>
<th>Release</th>
<th>Renderer status</th>
<th>Tasks status</th>
</tr>
<tr>
<td>1.2
<td>✔️ .NET (#2214)<br/>✔️ Android (#2213)<br/>✔️ iOS (#2212)<br/>🔄 TS (#2424)<br/>✔️ UWP (#2211)</td>
<td>✔️ Shared (#2210)</td>
</tr>
</table>

# Solves requests

* More container background colors (#2030)

# Proposed solution
Provide additional built-in container styles. I propose to name them the same as our foreground colors, namely **good**, **attention**, **warning** and **accent**.

## Host Config
Additional properties for all these new container styles

## Down-level impact
**Low**. Falls back to the default background color, when author potentially would have preferred emphasis rather than default.

## Host burden
**Medium**. Must specify host config values for all of these (otherwise the default colors used won't match host's color scheme).

## Renderer Requirements

Needed

# Auto-generated task status

- [x] Shared
- [x] .NET
- [x] Android
- [x] iOS
- [ ] TS
- [x] UWP

PR #2289, opened 2019-01-11
Additional built-in container styles

carlos-zamora:
Added `good`, `attention`, `warning`, and `accent` to container styles.
**Spec**: #1886 
**Components**:
* .NET (#2214 )
* Android (#2213)
* iOS (#2212)
* UWP (#2211)
* Shared (#2210)

# Testing
* Updated HostConfigs to include new Container Style Definitions
* Added AdaptiveCards of name `Container.Style.<stylename>.json`

# .NET
* `GetContainerStyleConfig()` used to help get desired style definition
Note: Bug #2285 is still present and makes ForegroundColors not work

# Shared Model
* Added to Enums and HostConfig

# UWP
* UWP Test App results
* Main changes made to `AdaptiveContainerStylesDefinition.h/cpp` and `Util.cpp`
* `GetContainerStyleDefinition()` used to help get desired style definition

# Android
* Expanded control flow to switch-case statement in ContainerRenderer and TextBlockRenderer
Note: FactSet just redirects to pertinent portion of TextBlockRenderer

# iOS
* Added `getColorPalette()` to get  desired style definition easier
* Fixed `sample.json`
* Fixed `getTextBlockColor()`


source/shared/cpp/ObjectModel/HostConfig.h
```diff
@@ -185,6 +185,10 @@ namespace AdaptiveSharedNamespace
                                                         {"#FFFFD700", "#B2FFD700"}, // warning
                                                         {"#FF8B0000", "#B28B0000"}  // attention
                                                     }};
+        ContainerStyleDefinition goodPalette;
+        ContainerStyleDefinition attentionPalette;
+        ContainerStyleDefinition warningPalette;
+        ContainerStyleDefinition accentPalette;
```
RebeccaAnne: Do we want default values for these?

source/uwp/AdaptiveCardTestApp/Expected/microsoft-teams.Tests.Action.DuplicateIds.json
```diff
@@ -1 +1 @@
-{"HostConfigHash":"5da68b8","CardHash":"e423716","Error":"Parsing card failed"}
\ No newline at end of file
+{"HostConfigHash":"1eae351","CardHash":"e423716","Error":"Parsing card failed"}
```
RebeccaAnne: Do you know why all these files changed?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOHostConfig.mm
```diff
@@ -365,28 +365,54 @@ + (UIColor *)convertHexColorCodeToUIColor:(const std::string&)hexColorCode
     return color;
 }
 
-- (UIColor *)getBackgroundColorForContainerStyle:(ACRContainerStyle)style
+- (ContainerStyleDefinition) getColorPalette:(ACRContainerStyle)style
```
nesalang: you are returning a struct by value. i suggest adding another param to diff the request either to foreground or background and return only those string instead of whole structure.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOHostConfig.mm
```diff
@@ -365,28 +365,54 @@ + (UIColor *)convertHexColorCodeToUIColor:(const std::string&)hexColorCode
...
+- (UIColor *)getBackgroundColorForContainerStyle:(ACRContainerStyle)style
+{
+    std::string hexColorCode = [self getColorPalette:style].backgroundColor;
     return [ACOHostConfig convertHexColorCodeToUIColor:hexColorCode];
 }
 
 + (ACRContainerStyle)getPlatformContainerStyle:(ContainerStyle)style
 {
-    ACRContainerStyle containerStyle = ACRDefault;
+    ACRContainerStyle containerStyle;
```
nesalang: why did you remove the initialization with the default value?

carlos-zamora: The default value is assigned in the `default` case for the switch-case. Felt it would be redundant to initialize it, but should I change it?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOHostConfig.mm
carlos-zamora: The problem is that the background is a `string` whereas the foreground is a `ColorsConfig` struct. Having this helper method lets me get the background through `getBackgroundColorForContainerStyle()` (in an appropriately converted format) and the foreground colors directly.

source/shared/cpp/ObjectModel/HostConfig.h
carlos-zamora: As of now, the default value is the same as the predefined `defaultPalette` (NOT the one defined by the user in the imported HostConfig). @matthidinger @andrewleader , thoughts on proper behavior or default values for this?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOHostConfig.mm
nesalang: yes, assignment takes only 2-3 instructions.

---
In reply to: [249849267](https://github.com/Microsoft/AdaptiveCards/pull/2289#discussion_r249849267) [](ancestors = 249849267)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTextBlockRenderer.mm
```diff
@@ -81,11 +81,8 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
 
         // Obtain text color to apply to the attributed string
         ACRContainerStyle style = lab.style;
-        ColorsConfig colorConfig;
-        if (style == ACREmphasis)
-            colorConfig = config->GetContainerStyles().emphasisPalette.foregroundColors;
```
nesalang: need to change to have reference here. I didn't realize that you've changed this in earlier commit.

nesalang:
>         ContainerStylesDefinition GetContainerStyles() const;

every usages of this method, user later choose containerstyledefintion based on style. it would be better add style param instead of returning whole container styles definition.

---
Refers to: source/shared/cpp/ObjectModel/HostConfig.h:282 in d8cbc59. [](commit_id = d8cbc59f300392900cb2d6b49ef5de398b40de27, deletion_comment = False)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOHostConfig.mm
nesalang: can you make it return reference?

---
In reply to: [249862786](https://github.com/Microsoft/AdaptiveCards/pull/2289#discussion_r249862786) [](ancestors = 249862786)

source/shared/cpp/ObjectModel/HostConfig.h
andrewleader: We have a default value for emphasisPalette, correct? I think we should similarly specify default values for goodPalette/etc

carlos-zamora: Sure. What do you want them to be? In the tests, I have them set to the subtle color for each ContainerStyle text.

carlos-zamora:
Added `GetContainerStyle(style)`, `GetBackgroundColor(style)`, and `GetForegroundColor(style)` to move some of the repeated logic back to the SharedModel. Should look a lot cleaner now. Can't make use of this in UWP though because UWP has it's own weird ObjectModel on top of it.

source/shared/cpp/ObjectModel/HostConfig.cpp
```diff
@@ -565,6 +577,36 @@ unsigned int HostConfig::GetFontWeight(FontStyle style, TextWeight weight) const
     return result;
 }
 
+ContainerStyleDefinition HostConfig::GetContainerStyle(ContainerStyle style) const
```
nesalang: Could you please make changes so that it wouldn't return a big struct?

source/shared/cpp/ObjectModel/HostConfig.cpp
```diff
@@ -565,6 +577,36 @@ unsigned int HostConfig::GetFontWeight(FontStyle style, TextWeight weight) const
...
+            return _containerStyles.defaultPalette;
+    }
+}
+
+std::string HostConfig::GetBackgroundColor(ContainerStyle style) const
+{
+    return GetContainerStyle(style).backgroundColor;
+}
+
+ColorsConfig HostConfig::GetForegroundColors(ContainerStyle style) const
```
nesalang: Could you look into returnning ColorsConfig also?

source/shared/cpp/ObjectModel/HostConfig.cpp
carlos-zamora: So, the problem with returning a pointer in the SharedModel is that all of the other similar functions in HostConfig (i.e.: `GetImageSizes()`, `GetImageBaseUrl()`, etc.) return the struct too. @paulcam206 was saying that passing by reference starts to cause issues with the lifetime of the pointer. I'm hoping `GetBackgroundColor(style)` and `GetForegroundColors(style)` help with that.  If not, suggestions with how to tackle this?

nesalang: one suggestion is leaving this function as it is, keep the copy of the returned value with ACOHostConfig, and move GetBackGroundColor and GetForegroundColor to platform codes. 
in case of iOS, GetBackgroundColor and GetForegroundColor could directly access the copied value. in that way we keep the copying as one time cost. 
And we can also remove ColorsConfig param from getTextBlockColor after making it an instance function.

---
In reply to: [250766675](https://github.com/Microsoft/AdaptiveCards/pull/2289#discussion_r250766675) [](ancestors = 250766675)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOHostConfig.mm
nesalang: can get rid of this fuction by making getTextBlock color an instance method and access color config through its host config and add optional style param.

---
In reply to: [249968877](https://github.com/Microsoft/AdaptiveCards/pull/2289#discussion_r249968877) [](ancestors = 249968877,249862786)

nesalang:
iOS looks good :)

source/uwp/AdaptiveCardTestApp/Expected/microsoft-teams.Tests.Action.DuplicateIds.json
carlos-zamora: HostConfigs got updated to include more ContainerStyleDefinitions. 

---
In reply to: [247212915](https://github.com/Microsoft/AdaptiveCards/pull/2289#discussion_r247212915) [](ancestors = 247212915)

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -717,24 +717,47 @@ HRESULT GetColorFromString(const std::string& colorString, _Out_ ABI::Windows::U
 }
 CATCH_RETURN;
 
-HRESULT GetColorFromAdaptiveColor(_In_ ABI::AdaptiveNamespace::IAdaptiveHostConfig* hostConfig,
+HRESULT GetContainerStyleDefinition(ABI::AdaptiveNamespace::ContainerStyle style,
+                                    _In_ ABI::AdaptiveNamespace::IAdaptiveHostConfig* hostConfig,
+                                    _Out_ ABI::AdaptiveNamespace::IAdaptiveContainerStyleDefinition** styleDefinition) noexcept try
```
RebeccaAnne: _Outptr_

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -717,24 +717,47 @@ HRESULT GetColorFromString(const std::string& colorString, _Out_ ABI::Windows::U
...
+    case ABI::AdaptiveNamespace::ContainerStyle::Default:
+    default:
+        RETURN_IF_FAILED(containerStyles->get_Default(styleDefinition));
+        break;
+    }
+    return S_OK;
+}
+CATCH_RETURN;
+
+HRESULT GetColorFromAdaptiveColor(ABI::AdaptiveNamespace::IAdaptiveHostConfig* hostConfig,
```
RebeccaAnne: _In_

source/dotnet/Library/AdaptiveCards/AdaptiveContainerStyle.cs
```diff
@@ -20,6 +20,14 @@ public enum AdaptiveContainerStyle
...
-        Emphasis = 1
+        Emphasis = 1,
+
+        Good = 2,
+
+        Attention = 3,
+
+        Warning = 4,
+
+        Accent = 5
```
paulcam206: nitpick -- I don't think we need the extra newlines here

source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
```diff
@@ -55,57 +55,160 @@
...
+					"default": "#2E89FC",
+					"subtle": "#882E89FC"
+				},
+				"good": {
+					"default": "#54a254",
+					"subtle": "#DD54a254"
+				},
+				"warning": {
+					"default": "#c3ab23",
+					"subtle": "#DDc3ab23"
```
paulcam206: out of curiosity, why are we mixing case here?

carlos-zamora: Didn't realize it's not case sensitive. Oops!

---
In reply to: [257394613](https://github.com/Microsoft/AdaptiveCards/pull/2289#discussion_r257394613) [](ancestors = 257394613)

Issue #2220, opened 2018-12-06
Unable to parse Custom ActionType


tnev:
# Platform

* Any

# Author or host

host

# Version of SDK

1.1

# Issue

The library currently has a **Custom** ActionType, but the ActionParserRegistration doesn't support parsing a Custom action. If the ActionParserRegistration were to add a custom action parser to its cardElementParsers map, such as:

```swift
m_cardElementParsers.insert({{ActionTypeToString(ActionType::OpenUrl), std::make_shared<OpenUrlActionParser>()},
                                     {ActionTypeToString(ActionType::ShowCard), std::make_shared<ShowCardActionParser>()},
                                     {ActionTypeToString(ActionType::Submit), std::make_shared<SubmitActionParser>()},
                                     {ActionTypeToString(ActionType::Custom), std::make_shared<CustomActionParser>()}});
```

then it seems that it would be able to support custom actions. Is there a plan for supporting Custom actions in this way, or is there another suggested way of accomplishing this?

RebeccaAnne:
You can add a custom action parser using the ActionParserRegistration::AddParser method, passing the name of your custom action and something that implements ActionElementParser. This will add your custom parser to that array.

tnev:
Hi @RebeccaAnne , thanks for your response. This is actually something I was trying to do before, but for some reason it doesn't seem like ActionParserRegistration is publicly exposed through the iOS framework (sorry, I'm realizing now I should have specified iOS as the platform above!).

I thought it was possibly an issue with the way my project imported the framework, but I wasn't able to access it in the iOS example project you provide either. When I tried to import "AdaptiveCards/ActionParserRegistration.h" I received compiler errors about some of the frameworks included in the pch.h file under the AdaptiveSharedNamespace definition.

Its seems like the Android framework exposes it so I'm not sure if this is an issue with how I'm trying to access it or if there are limitations to accessing the C++ files on iOS. I appreciate your help!

nesalang:
Hi @tnev, iOS doesn't have custom parser supports for actions. 
we plan to add support for custom action parsing.
It will be supported in the similar way as in base cards elements. 
When unknown type is encountered during parsing, adaptivecards create an Unknown element that stores the raw json enclosed in the custom type.
ACRCustomRenderer is a proxy that get called when the unknown type  element is encountered during rendering. From here, you can parse the raw json and render UI elements. To support more than one type, you can branch out to different types from here. 

[...]

PR #2299, opened 2019-01-12
Jwoo/ios custom element update

nesalang:
- Resolves the issue #2220 
- it will break shared model unit test (I need to update the test).


samples/Tests/Action.CustomParsing.json
```diff
@@ -0,0 +1,22 @@
+{
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"type": "AdaptiveCard",
+	"version": "1.0",
+	"body": [
+		{
+			"type": "TextBlock",
+			"text": "This card has a custom action. If renderer can redner the custom type, it will render a button.",
```
khouzam: >redner [](start = 59, length = 6)

render

source/dotnet/Test/AdaptiveCards.Test/AllPayloadTests.cs
```diff
@@ -126,6 +126,7 @@ public void TestAllTestCards()
                     "FlightItinerary_es_fail",
                     "Action.DuplicateIds",
                     "Action.NestedDuplicateIds",
+                    "Action.CustomParsing",
```
khouzam: I don't think that this test should fail parsing.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRRegistration.mm
```diff
@@ -118,16 +126,21 @@ - (void)setActionSetRenderer:(id<ACRIBaseActionSetRenderer>)actionsetRenderer
     _actionSetRenderer = actionsetRenderer;
 }
 
-- (void) setActionRenderer:(ACRBaseActionElementRenderer *)renderer cardElementType:(NSNumber *)cardElementType
+- (void)setActionRenderer:(ACRBaseActionElementRenderer *)renderer cardElementType:(NSNumber *)cardElementType
 {
+    // custom action must be registered through set custom action renderer method
+    if(cardElementType.longValue > ACRUnknownAction) {
```
khouzam: >cardElementType.longValue > ACRUnknownAction [](start = 7, length = 44)

If this is a requirement of the enum, then at least put a comment in the enum that the order is important and that the ACRUnknownAction needs to always be last.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRCustomRenderer.mm
```diff
@@ -52,6 +52,7 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
             [viewGroup addArrangedSubview:wrappingview];
         }
     }
+ */
```
khouzam: Needs to be removed

khouzam:
>     TEST_CLASS(CustomParsingForIOSTest)

Doesn't need to be IOS specific since this is the shared model

---
Refers to: source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/CustomParsingForIOSTest.cpp:18 in 19d5c92. [](commit_id = 19d5c9267b4ca3eead5718ae93c06c444b271a0b, deletion_comment = False)

nesalang:
replaced by #2325

PR #2315, opened 2019-01-18
Support load local content image

wecheng:
1. Define a new function to load local Content images.
2. Add a special handle on the url for appsdk on mobile side to load local image. Parameter url should be a valid url, so we have to add "package:" to pass the valid URI checking.


wecheng:
> application
@khouzam Here list the steps about how to test this.

1. Try to add read contacts permissions in the AndroidManifest file for the test app, like add line <uses-permission android:name="android.permission.READ_CONTACTS"/>.
2. Try to create a test json file from this link: https://adaptivecards.io/samples/, update that image url to "content://com.android.contacts/contacts/***". Here you should pass a real image url for your test phone.
3. Set edited json file as input to see the rendered view.

wecheng:
> khouzam

Sorry, for the first, like add line: <uses-permission android:name="android.permission.READ_CONTACTS"/>

khouzam:
Hi @wecheng 

Please add this functionality to the test application so that it can be consistently validated.

Feature implementations need to have test coverage.

almedina-ms:
>                     if (m_imageBaseUrl == null || m_imageBaseUrl.isEmpty())

this is the conditional to change

---
Refers to: source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/GenericImageLoaderAsync.java:76 in 128f645. [](commit_id = 128f6458ba5d735c3c25b487dcdaa0a4abd1b7b6, deletion_comment = False)

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/GenericImageLoaderAsync.java
```diff
@@ -126,6 +131,14 @@
 
         // Get image identifier
         Resources resources = context.getResources();
+
+        // For host app only provides image file name as url, host app can pass "package:[IMAGE NAME]" as replacement for
+        // meeting the valid URL format checking. Here we will remove this prefix to get file name.
+        if (url.startsWith("package:"))
+        {
+            url = url.replace("package:", "");
+        }
```
almedina-ms: As far as I know, the assets included with the project are always located in the drawable folder, so we can avoid the use of the baseImageUrl property from the hostConfig json

I think it's better to change this line (line 76)
if (m_imageBaseUrl == null || m_imageBaseUrl.isEmpty())
adding a check for the path not starting with package
if ((m_imageBaseUrl == null || m_imageBaseUrl.isEmpty()) && !path.startsWith("package:"))

And making a change here to assign imageBaseUrl with value "drawable/"

almedina-ms: Since the time you made this PR we figured out that we were doing something wrong about how the resource resolvers worked on android, we changed that to mimic the behavior of resource resolvers to the one in UWP and iOS. I say we should keep the "content:" option as is and just remove the "package:" option, as you can see in the mobile test app, that change is already there as a resource resolver

PR #2317, opened 2019-01-18
Add missing exclusivetos to UWP API

RebeccaAnne:
When the UWP code was originally written, we were using a vector implementation that couldn't handle vectors of RuntimeClasses. Because you can't declare a vector of interfaces that are exclusive to a runtimeclass (only vectors of that runtime class) some of our interfaces were not marked as exclusiveto when they should have been. This leads to some awkward API usage in the projected languages. We've long since fixed the vector issue, so updated our API to correctly mark interfaces as exclusiveto.

### DIFF
--- a/source/uwp/Renderer/lib/AdaptiveCard.cpp
+++ b/source/uwp/Renderer/lib/AdaptiveCard.cpp
@@ -104,7 +104,7 @@ namespace AdaptiveNamespace
             RETURN_IF_FAILED(MakeAndInitialize<AdaptiveCard>(&adaptiveCard, sharedParseResult->GetAdaptiveCard()));
             RETURN_IF_FAILED(adaptiveParseResult->put_AdaptiveCard(adaptiveCard.Get()));
 
-            ComPtr<IVector<IAdaptiveWarning*>> warnings;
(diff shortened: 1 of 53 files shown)


source/uwp/Renderer/lib/AdaptiveCard.cpp
```diff
@@ -113,7 +113,7 @@ namespace AdaptiveNamespace
         }
         catch (const AdaptiveCardParseException& e)
         {
-            ComPtr<IVector<IAdaptiveError*>> errors;
+            ComPtr<IVector<ABI::AdaptiveNamespace::AdaptiveError*>> errors;
```
khouzam: >ABI::AdaptiveNamespace:: [](start = 27, length = 24)

Is there a particular reason we are explicit about ABI::AdpativeNamespace when we have the using directive for the namespace at the top of the file?

RebeccaAnne: It's to distinguish the runtimeclass ABI:AdaptiveNamespace::Foo from our implementation class AdaptiveNamespace::Foo (No ABI::).

---
In reply to: [249209912](https://github.com/Microsoft/AdaptiveCards/pull/2317#discussion_r249209912) [](ancestors = 249209912)

PR #2324, opened 2019-01-23
Fix alwaysBleedSeparators half broken after recent changes.

dclaux:
Fix a regression introduced by full bleeding support. Works better than before though thanks to the full bleeding implementation: now separators will bleed into the first parent that has padding instead of only into their direct parent. This produces more predictable results.

Try this payload:
```
{
	"type": "AdaptiveCard",
	"version": "1.0",
	"body": [
		{
			"type": "TextBlock",
            "text": "First element of the card"
        },
		{
			"type": "TextBlock",
            "separator": true,
            "text": "Second element of the card"
        },
        {
            "type": "Container",
            "separator": true,
			"items": [
				{
					"type": "TextBlock",
					"text": "First element in the Container with default style"
				},
				{
					"type": "TextBlock",
                    "separator": true,
					"text": "Second element in the Container with default style"
				}
			]
		},
        {
            "type": "Container",
            "style": "emphasis",
            "separator": true,
			"items": [
				{
					"type": "TextBlock",
					"text": "First element in the Container with emphasis style"
				},
				{
					"type": "TextBlock",
                    "separator": true,
					"text": "Second element in the Container with emphasis style"
				}
			]
		},
        {
            "type": "ColumnSet",
            "separator": true,
            "style": "emphasis",
            "columns": [
                {
                    "width": 1,
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "First"
                        },
                        {
                            "type": "TextBlock",
                            "separator": true,
                            "text": "Second"
                        }
                    ]
                },
                {
                    "width": 1,
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "First"
                        },
                        {
                            "type": "TextBlock",
                            "text": "Second"
                        }
                    ]
                }
            ]
        }
	]
}
```

It will render as follows:
image

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -135,17 +135,6 @@ export abstract class CardElement implements ICardObject {
 			},
 			this.separatorOrientation);
 
-		if (AdaptiveCard.alwaysBleedSeparators && renderedSeparator && this.separatorOrientation == Enums.Orientation.Horizontal) {
-			let parentContainer = this.getParentContainer();
-
-			if (parentContainer && parentContainer.getEffectivePadding()) {
-				let parentPhysicalPadding = this.hostConfig.paddingDefinitionToSpacingDefinition(parentContainer.getEffectivePadding());
-
-				renderedSeparator.style.marginLeft = "-" + parentPhysicalPadding.left + "px";
-				renderedSeparator.style.marginRight = "-" + parentPhysicalPadding.right + "px";
-			}
-		}
-
 		return renderedSeparator;
 	}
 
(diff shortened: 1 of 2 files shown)


matthidinger:
@dclaux do you think you could save that payload into `~/samples/Tests/BackgroundBleed.json` or similar? It sounds like it will be useful for validating the other renderers as they implement this as well.

dclaux:
@matthidinger this is a custom capability of the JS renderer that goes beyond the spec. It is necessary for Actionable Message support on Outlook mobile.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -241,6 +230,25 @@ export abstract class CardElement implements ICardObject {
...
 	protected abstract internalRender(): HTMLElement;
 
+	protected applyPadding() {
+		if (this.separatorElement) {
+			if (AdaptiveCard.alwaysBleedSeparators && this.separatorOrientation == Enums.Orientation.Horizontal && !this.isBleeding()) {
+				let padding = new Shared.PaddingDefinition();
+
+				this.getImmediateSurroundingPadding(padding);
+		
+				let physiaclPadding = this.hostConfig.paddingDefinitionToSpacingDefinition(padding);
```
matthidinger: typo: physiaclPadding

dclaux: Fixed

PR #2332, opened 2019-01-24
Fix unknown element serialization in shared model and UWP

paulcam206:
Found work while doing fallback -- the shared model doesn't properly roundtrip unknown elements. Specifically, elements are emitted with `"type": "Unknown"` rather than the original `type` encountered. This change fixes said problem. 

Additionally, UWP doesn't even persist unknown (or `Unsupported`) elements in its collections, so it won't roundtrip even if the shared model does the right thing. The fix is to introduce `AdaptiveUnsupportedElement` to represent these unknown elements.


source/uwp/Renderer/lib/AdaptiveUnsupportedElement.h
```diff
@@ -0,0 +1,88 @@
+#pragma once
+
+#include "AdaptiveCards.Rendering.Uwp.h"
+#include "Enums.h"
+#include "UnknownElement.h"
+#include "AdaptiveCardElement.h"
+
+namespace AdaptiveNamespace
+{
+    class DECLSPEC_UUID("e1ab7208-17bc-4fb6-b3c0-c1886987b4cf") AdaptiveUnsupportedElement
```
nesalang: Do we need this class at the platform level? 
They would create their custom class if they need custom parsing and custom object. 

If adaptive card parser is given a json with unsupported type, should we return json with unsupported element in it? I think the answer is YES. O.K it makes sense now. thanks!

If that's the case, shouldn't you add a projection of it in GenerateContainedElementsProjection?

source/shared/cpp/ObjectModel/UnknownElement.cpp
```diff
@@ -13,9 +13,29 @@ UnknownElement::UnknownElement() : BaseCardElement(CardElementType::Unknown)
...
 
+void UnknownElement::SetActualType(const std::string& actualType)
+{
+    m_actualType = actualType;
+}
+
+std::string UnknownElement::GetActualType() const
+{
+    return m_actualType;
+}
```
RebeccaAnne: What if BaseCardElements GetElementTypeString returned the actual type? I think that would be consistent with the way custom elements work.

paulcam206:
Sure thing :)

---
In reply to: [457422464](https://github.com/Microsoft/AdaptiveCards/pull/2332#issuecomment-457422464) [](ancestors = 457422464)

source/shared/cpp/ObjectModel/UnknownElement.cpp
paulcam206: I was on the fence about this one. It felt a little weird to me to have it implicit. If you think it's weird to have it explicit, I can change it :)

---
In reply to: [251111068](https://github.com/Microsoft/AdaptiveCards/pull/2332#discussion_r251111068) [](ancestors = 251111068)

source/uwp/Renderer/lib/AdaptiveUnsupportedElement.h
paulcam206: see `Util.cpp` :)

---
In reply to: [251094931](https://github.com/Microsoft/AdaptiveCards/pull/2332#discussion_r251094931) [](ancestors = 251094931)

source/uwp/AdaptiveCardTestApp/Expected/Tests.Action.TitleToJson.json
```diff
@@ -1 +1 @@
-{"actions":[{"data":"null","id":"","title":"Good","type":"Action.Submit"},{"data":"null","id":"","type":"Action.Submit"},{"data":"null","id":"","type":"Action.Submit"},{"data":"null","id":"","title":" ","type":"Action.Submit"}],"body":[{"text":"Action.Title is highly recommended, but for the time being not enforced","type":"TextBlock","wrap":true}],"type":"AdaptiveCard","version":"1.0"}
```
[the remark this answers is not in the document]
paulcam206: no worries :)

---
In reply to: [251108490](https://github.com/Microsoft/AdaptiveCards/pull/2332#discussion_r251108490) [](ancestors = 251108490)

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -1502,6 +1504,37 @@ AdaptiveNamespaceStart
...
+        contract(InternalContract, 1),
+        internal
+#else
+        uuid(BAD9E18F-ACA2-46A7-A16E-31D0B8866ABA)
+#endif
+    ]
+    interface IAdaptiveUnsupportedElement : IInspectable
+    {
+        //[propget] HRESULT Actions([out, retval] Windows.Foundation.Collections.IVector<IAdaptiveActionElement*>** value);
+    };
```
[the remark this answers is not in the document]
paulcam206: woops!

---
In reply to: [251107992](https://github.com/Microsoft/AdaptiveCards/pull/2332#discussion_r251107992) [](ancestors = 251107992)

source/uwp/Renderer/lib/AdaptiveUnsupportedElement.h
paulcam206: Not sure Why my codeflow comment didn't get published... what I meant to say: see changes in `Util.cpp` 👍

source/shared/cpp/ObjectModel/UnknownElement.cpp
```diff
@@ -13,9 +8,19 @@ UnknownElement::UnknownElement() : BaseCardElement(CardElementType::Unknown)
 {
 }
 
+Json::Value UnknownElement::SerializeToJsonValue() const
+{
+    // Important -- we need override the type value that the base element uses for us
```
RebeccaAnne: >// Important -- we need override the type value that the base element uses for us [](start = 4, length = 81)

Is this still needed? Won't the base element now do the right thing here?

source/shared/cpp/ObjectModel/UnknownElement.cpp
```diff
@@ -13,9 +8,19 @@ UnknownElement::UnknownElement() : BaseCardElement(CardElementType::Unknown)
...
+    Json::Value root = BaseCardElement::SerializeToJsonValue();
+    root[AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Type)] = GetElementTypeString();
+    return root;
+}
+
 std::shared_ptr<BaseCardElement> UnknownElementParser::Deserialize(ParseContext&, const Json::Value& json)
 {
+    std::string actualType = ParseUtil::GetTypeAsString(json);
```
RebeccaAnne: Maybe add a comment here explaining why we need this.

paulcam206: Good idea :)

---
In reply to: [251154183](https://github.com/Microsoft/AdaptiveCards/pull/2332#discussion_r251154183) [](ancestors = 251154183)

source/shared/cpp/ObjectModel/UnknownElement.cpp
paulcam206: It does now -- yes :)

---
In reply to: [251153491](https://github.com/Microsoft/AdaptiveCards/pull/2332#discussion_r251153491) [](ancestors = 251153491)

nesalang:
UnknownElement was created to allow iOS platform to handle custom class and its parsing.
We are now using the element to store the json blob that can't be handled by the registered parsers, so we can serialize without losing data.
Should we make similar changes to other platforms such as dotnet?

paulcam206:
> Should we make similar changes to other platforms such as dotnet to which i can add?

Yes, we need to make similar changes to dotnet... 🔧

nesalang:
if you want, i can make that changes. I've already create PR for android.

---
In reply to: [457770472](https://github.com/Microsoft/AdaptiveCards/pull/2332#issuecomment-457770472) [](ancestors = 457770472)

paulcam206:
sounds good :)

---
In reply to: [457772068](https://github.com/Microsoft/AdaptiveCards/pull/2332#issuecomment-457772068) [](ancestors = 457772068,457770472)

PR #2336, opened 2019-01-25
Fix fact value always "Value" when not set + small FactSet alignment

dclaux:


funkeyfreak:
To give some context, this was causing null fields to be rendered, resulting in cards with a null rendered field.

PR #2346, opened 2019-01-30
Fix for uwp column object model to properly handle pixel width

RebeccaAnne:


source/uwp/UWPUnitTests/ObjectModelTests.cs
```diff
@@ -262,7 +262,7 @@ public void ColumnSet()
                 Height = HeightType.Stretch,
                 Id = "ColumnId",
                 IsVisible = false,
-                PixelWidth = 50,
+                Width = "50px",
```
nesalang: thanks for fixing this

source/uwp/Renderer/lib/AdaptiveColumn.cpp
```diff
@@ -41,7 +41,14 @@ namespace AdaptiveNamespace
 
     HRESULT AdaptiveColumn::get_Width(_Outptr_ HSTRING* width) { return m_width.CopyTo(width); }
 
-    HRESULT AdaptiveColumn::put_Width(_In_ HSTRING width) { return m_width.Set(width); }
+    HRESULT AdaptiveColumn::put_Width(_In_ HSTRING width)
+    {
+        RETURN_IF_FAILED(m_width.Set(width));
+
+        std::vector<std::shared_ptr<AdaptiveCardParseWarning>> warningvector;
```
nesalang: seems this is not going to be used at all, will it be possible not to allocate it at all?

RebeccaAnne: I can change ParseSizeForPixelSize to take a pointer instead of a reference and pass null.

---
In reply to: [252409077](https://github.com/Microsoft/AdaptiveCards/pull/2346#discussion_r252409077) [](ancestors = 252409077)

RebeccaAnne: Actually, how strongly do we feel about these warnings? It occurs to me i should make this change in the shared model as well, but if i move the logic there from Deserialize() to SetWidth() we won't have the warnings there either. Certainly i could code around that, but i want to make sure that we actually need those warnings before i make the code more complicated than i need to. We're not super consistent with warnings already, as it looks like we don't return a warning if they set the width to, for example, "50" (no "px").

---
In reply to: [252427457](https://github.com/Microsoft/AdaptiveCards/pull/2346#discussion_r252427457) [](ancestors = 252427457,252409077)

RebeccaAnne: Alright, reorganized it leaving the warnings in place using a pointer rather than a reference.

---
In reply to: [252430493](https://github.com/Microsoft/AdaptiveCards/pull/2346#discussion_r252430493) [](ancestors = 252430493,252427457,252409077)

RebeccaAnne: Indeed

source/shared/cpp/ObjectModel/Util.cpp
```diff
@@ -105,18 +105,18 @@ void ValidateUserInputForDimensionWithUnit(const std::string& unit,
         }
         catch (const std::invalid_argument&)
         {
-            warnings.emplace_back(std::make_shared<AdaptiveCardParseWarning>(WarningStatusCode::InvalidDimensionSpecified,
+            warnings->emplace_back(std::make_shared<AdaptiveCardParseWarning>(WarningStatusCode::InvalidDimensionSpecified,
```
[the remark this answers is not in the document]
RebeccaAnne: Indeed

---
In reply to: [252449267](https://github.com/Microsoft/AdaptiveCards/pull/2346#discussion_r252449267) [](ancestors = 252449267)

Issue #2245, opened 2018-12-17
Color of input text is hardcoded and cannot be changed by the client


rednivrug15:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* Android

# Author or host

Host
Microsoft Teams

If you're an author, who are you sending cards to?

# Version of SDK

Latest SDK 

# Issue

The color of input text color is hardcoded as black and cannot be changed by the client . Do not hardcode the color instead let the client specify the color by passing it through the android context

PR #2347, opened 2019-01-30
Cherry-pick Remove hardcoded color in Android Input (#2245)

khouzam:
Remove the hard-coded black color from Android enabling dark themes for Android.


khouzam:
UWP break should not be worriesome as the build structure would need to be updated but we aren't currently planning on releasing an update to the UWP renderer,

PR #2351, opened 2019-01-31
Add fallback parsing with collision detection as well as requires par…

paulcam206:
## Shared Model Changes
`BaseElement.[h|cpp]`: Implementation of new base class underlying both `BaseCardElement` and `BaseActionElement` that combines the previously duplicated functionality of each. This was needed to ensure that we only needed a single implementation of `fallback` and `requires` parsing/API access.

`ParseContext.[h|cpp]`: Add parse-time stack to track elements as they're being parsed. This is needed to detect `id` collisions across disparate portions of a card. More details on this implementation will be made available before checkin.

Everything else: Either push/pop calls when parsing child elements (e.g. `Container`) or cleanup/follow-on changes to support fallback. See comment in `ParseContext.cpp` for algorithm details.

## Test Changes
`samples/Tests/FallbackParsing.json`: Test card containing element fallback and requires directives

`ObjectModelTest.cpp`: Contains new tests for `id` collision detection and [de]serialization of new `requires` and `fallback` properties

Various UWP test app JSON file changes to account for the fact that `id` is no longer emitted if it isn't set.


source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ObjectModelTest.cpp
```diff
@@ -450,5 +458,596 @@ namespace AdaptiveCardsSharedModelUnitTest
...
+                "version" : "1.2",
+                "body" : [
+                    {
+                        "type": "TextBlock",
+                        "text" : "test text"
+                    }
+                ],
+               	"actions": [
+               		{
+               			"type": "Action.Submit",
```
RebeccaAnne: >			 [](start = 15, length = 3)

Tabs

source/shared/cpp/ObjectModel/BaseElement.h
```diff
@@ -0,0 +1,128 @@
...
+        static void ParseJsonObject(AdaptiveSharedNamespace::ParseContext& context,
+                                    const Json::Value& json,
+                                    std::shared_ptr<BaseElement>& baseElement)
+        {
+            T::ParseJsonObject(context, json, baseElement);
+        }
+
+        template<typename T> void DeserializeBase(ParseContext& context, const Json::Value& json);
+
+        template<typename T> static std::shared_ptr<T> Deserialize(ParseContext& context, const Json::Value& json);
```
RebeccaAnne: >template<typename T> static std::shared_ptr<T> Deserialize(ParseContext& context, const Json::Value& json); [](start = 8, length = 107)

Looks like this declaration is not in use.

source/shared/cpp/ObjectModel/Enums.cpp
```diff
@@ -128,6 +130,7 @@ namespace AdaptiveSharedNamespace
             {AdaptiveCardSchemaKey::TextWeight, "weight"},
             {AdaptiveCardSchemaKey::Thick, "thick"},
             {AdaptiveCardSchemaKey::Thickness, "thickness"},
+            {AdaptiveCardSchemaKey::Thickness, "thickness"},
```
RebeccaAnne: Duplicate

source/shared/cpp/ObjectModel/Enums.cpp
```diff
@@ -11,6 +11,7 @@ namespace AdaptiveSharedNamespace
             {AdaptiveCardSchemaKey::Actions, "actions"},
             {AdaptiveCardSchemaKey::ActionSet, "ActionSet"},
             {AdaptiveCardSchemaKey::ActionSetConfig, "actionSetConfig"},
+            {AdaptiveCardSchemaKey::Actions, "actions"},
```
RebeccaAnne: "Actions" is now in here twice

source/shared/cpp/ObjectModel/Enums.cpp
RebeccaAnne: Also, while you're in the neighborhood, looks like AdaptiveCardSchemaKey::Thick is not in use anywhere.

---
In reply to: [253137890](https://github.com/Microsoft/AdaptiveCards/pull/2351#discussion_r253137890) [](ancestors = 253137890)

RebeccaAnne:
>         Actions,

If you move Actions in the .cpp we should probably move it here too (nothing depends on these matching up, right?)

---
Refers to: source/shared/cpp/ObjectModel/Enums.h:13 in 2a849bc. [](commit_id = 2a849bc5bf520fc17acc57d50f2b65651013ff29, deletion_comment = False)

source/shared/cpp/ObjectModel/ParseContext.cpp
```diff
@@ -1,20 +1,109 @@
...
+        warnings{}, m_idStack{}, m_elementIds{}
     {
         elementParserRegistration = (elementRegistration) ? elementRegistration : std::make_shared<ElementParserRegistration>();
         actionParserRegistration = (actionRegistration) ? actionRegistration : std::make_shared<ActionParserRegistration>();
-        elementIds = (ids) ? ids : std::make_shared<std::unordered_set<std::string>>();
+    }
+
+    unsigned int ParseContext::GetNearestFallbackId(const unsigned int skipId) const
+    {
+        // Walk stack looking for first element to be marked fallback, then return its internal ID. If none, return
```
RebeccaAnne: >If none, return [](start = 100, length = 15)

... ParseContext::InvalidFallbackId

source/shared/cpp/ObjectModel/ParseContext.h
```diff
@@ -12,12 +12,31 @@ namespace AdaptiveSharedNamespace
...
+        // note: not enum class because we don't want typed values
+        enum TupleIndex : unsigned int
+        {
+            Id = 0U,
+            InternalId,
+            IsFallback
+        };
+
+        // ID -> (internal ID, fallback ID)
+        std::unordered_multimap<std::string, std::pair<unsigned int, unsigned int>> m_elementIds;
```
RebeccaAnne: Can you put m_idStack and m_elementIds next to each other in this file with a comment explaining how they are each used?

source/shared/cpp/ObjectModel/ToggleInput.cpp
```diff
@@ -88,11 +88,11 @@ void ToggleInput::SetWrap(bool value)
...
 
-std::shared_ptr<BaseCardElement> ToggleInputParser::Deserialize(ParseContext&, const Json::Value& json)
+std::shared_ptr<BaseCardElement> ToggleInputParser::Deserialize(ParseContext& context, const Json::Value& json)
 {
     ParseUtil::ExpectTypeString(json, CardElementType::ToggleInput);
 
-    std::shared_ptr<ToggleInput> toggleInput = BaseInputElement::Deserialize<ToggleInput>(json);
+    std::shared_ptr<ToggleInput> toggleInput = BaseInputElement::Deserialize<ToggleInput>(context, json);
 
     toggleInput->SetTitle(ParseUtil::GetString(json, AdaptiveCardSchemaKey::Title, true));
```
RebeccaAnne: Does this function need to push/pop?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/EverythingBagel.h
```diff
@@ -1,3 +1,3 @@
...
...alue\":\"13:00\"},{\"id\":\"Input.Toggle_id\",\"isRequired\":false,\"title\":\"Input.Toggle_title\",\"type\":\"Input.Toggle\",\"value\":\"Input.Toggle_on\",\"valueOff\":\"Input.Toggle_off\",\"valueOn\":\"Input.Toggle_on\"},{\"size\":\"Large\",\"text\":\"Everybody's got choices\",\"type\":\"TextBlock\",\"weight\":\"Bolder\"},{\"choices\":[{\"title\":\"Input.Choice1_title\",\"value\":\"Input.Choice1\"},{\"title\":\"Input.Choice2_title\",\"value\":\"Input.Choice2\"},{\"title\":\"Input.Choice3_title\",\"value\":\"Input.Choice3\"},{\"title\":\"Input.Choice4_title\",\"value\":\"Input.Choice4\"}],\"id\":\"Input.ChoiceSet_id\",\"isMultiSelect\":true,\"isRequired\":false,\"style\":\"Compact\",\"type\":\"Input.ChoiceSet\",\"value\":\"Input.Choice2,Input.Choice4\"}],\"type\":\"Container\"},{\"actions\":[{\"id\":\"ActionSet.Action.Submit_id\",\"title\":\"ActionSet.Action.Submit\",\"type\":\"Action.Submit\"},{\"id\":\"ActionSet.Action.OpenUrl_id\",\"title\":\"ActionSet.Action.OpenUrl\",\"type\":\"Action.OpenUrl\",\"url\":\"https://adaptivecards.io/\"}],\"type\":\"ActionSet\"}],\"fallbackText\":\"fallbackText\",\"lang\":\"en\",\"speak\":\"speak\",\"type\":\"AdaptiveCard\",\"version\":\"1.0\"}\n"
```
RebeccaAnne: >\"id\":\"\" [](start = 939, length = 11)

The UWP object model tests also need updating.

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ObjectModelTest.cpp
paulcam206: ugh

---
In reply to: [253127516](https://github.com/Microsoft/AdaptiveCards/pull/2351#discussion_r253127516) [](ancestors = 253127516)

source/shared/cpp/ObjectModel/BaseCardElement.cpp
```diff
@@ -95,15 +78,9 @@ const CardElementType BaseCardElement::GetElementType() const
...
-    Json::FastWriter writer;
-    return writer.write(SerializeToJsonValue());
-}
-
 Json::Value BaseCardElement::SerializeToJsonValue() const
 {
-    Json::Value root = GetAdditionalProperties();
+    Json::Value root = BaseElement::SerializeToJsonValue();
 
     root[AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Type)] = GetElementTypeString();
```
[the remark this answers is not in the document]
paulcam206: Element type string would work. I originally skipped it because element type is... typed. :)

---
In reply to: [253133611](https://github.com/Microsoft/AdaptiveCards/pull/2351#discussion_r253133611) [](ancestors = 253133611)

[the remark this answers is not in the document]
paulcam206: yes :)

---
In reply to: [253133611](https://github.com/Microsoft/AdaptiveCards/pull/2351#discussion_r253133611) [](ancestors = 253133611)

source/shared/cpp/ObjectModel/ParseContext.cpp
```diff
@@ -1,20 +1,109 @@
...
+        }
+        return ParseContext::InvalidFallbackId;
+    }
+
+    void ParseContext::PushElement(const std::tuple<std::string, unsigned int, bool>& ids)
+    {
+        m_idStack.push_back(ids);
+    }
+
+    void ParseContext::PopElement()
```
RebeccaAnne: >void ParseContext::PopElement() [](start = 4, length = 31)

Could you add a header comment to this function giving an overview of the algorithm?

paulcam206:
>         Actions,

order doesn't matter. I was going for a sorted list to make it easier to spot mistakes. :D

---
In reply to: [459806981](https://github.com/Microsoft/AdaptiveCards/pull/2351#issuecomment-459806981) [](ancestors = 459806981)

---
Refers to: source/shared/cpp/ObjectModel/Enums.h:13 in 2a849bc. [](commit_id = 2a849bc5bf520fc17acc57d50f2b65651013ff29, deletion_comment = False)

source/shared/cpp/ObjectModel/BaseElement.h
```diff
@@ -0,0 +1,127 @@
...
+        void SetTypeString(const std::string& type) { m_typeString = type; }
+        std::string m_typeString;
+        std::unordered_set<std::string> m_knownProperties;
+        Json::Value m_additionalProperties;
+
+    private:
+        static unsigned int GetNextInternalId()
+        {
+            static unsigned int nextId = 1;
+            return nextId++;
```
nesalang: shouldn't we think of overflow strategy since renderers on mobile app usually sleeps and rarely quit?

paulcam206: I'm not too concerned about the overflow collision scenario -- you'd have to have more than 4,294,967,295 elements in a single card to run into a collision. It seems exceedingly unlikely that we'd manage to construct that many elements in a single session. However, there *is* one small bug here that I need to fix -- if we overflow back to 0, we'll hand out our `InvalidId` sentinel value.

source/shared/cpp/ObjectModel/ActionParserRegistration.h
```diff
@@ -14,14 +14,32 @@ namespace AdaptiveSharedNamespace
         virtual std::shared_ptr<BaseActionElement> Deserialize(ParseContext& context, const Json::Value& value) = 0;
     };
 
+    class ActionElementParserWrapper : public ActionElementParser
```
RebeccaAnne: Lets add a comment in here explaining why we need the wrapper

paulcam206: Done!

source/shared/cpp/ObjectModel/ToggleInput.cpp
paulcam206: Not any more! 👍

source/shared/cpp/ObjectModel/ParseContext.cpp
paulcam206: I've added a giant comment to `ParseContext.cpp`

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/EverythingBagel.h
paulcam206: Done :)

paulcam206:
I've theoretically addressed all feedback. :)

source/shared/cpp/ObjectModel/ParseContext.cpp
RebeccaAnne: This is great, thanks!

---
In reply to: [255670264](https://github.com/Microsoft/AdaptiveCards/pull/2351#discussion_r255670264) [](ancestors = 255670264)

source/shared/cpp/ObjectModel/ActionParserRegistration.h
RebeccaAnne: I see that the mega comment covers this, but maybe reference it here ("See ParseContext.cpp for details")

---
In reply to: [255195775](https://github.com/Microsoft/AdaptiveCards/pull/2351#discussion_r255195775) [](ancestors = 255195775)

source/shared/cpp/ObjectModel/BaseElement.cpp
```diff
@@ -0,0 +1,193 @@
...
+namespace AdaptiveSharedNamespace
+{
+    unsigned int InternalId::s_currentInternalId = 1;
+
+    InternalId InternalId::Current()
+    {
+        InternalId internalId{s_currentInternalId};
+        return internalId;
+    }
+
```
RebeccaAnne: I think we talked about this once, but can you remind me of the resolution? Could we increment s_currentInternalId here? Do we need to get it more than once from here?

source/shared/cpp/ObjectModel/ActionParserRegistration.h
RebeccaAnne: Ah, nevermind, i see the comment where it's used which is probably just as good.

---
In reply to: [257416041](https://github.com/Microsoft/AdaptiveCards/pull/2351#discussion_r257416041) [](ancestors = 257416041,255195775)

source/shared/cpp/ObjectModel/BaseElement.cpp
paulcam206: We need a way to get the current internal ID without incrementing because we need to be able to push an internal ID on the stack before `BaseElement` is constructed. `Current()` is used in `BaseElement::BaseElement()`, and `Next()` is used in both element and action parser wrappers.

---
In reply to: [257416836](https://github.com/Microsoft/AdaptiveCards/pull/2351#discussion_r257416836) [](ancestors = 257416836)

RebeccaAnne: Fair enough.

---
In reply to: [257418724](https://github.com/Microsoft/AdaptiveCards/pull/2351#discussion_r257418724) [](ancestors = 257418724,257416836)

PR #2362, opened 2019-02-07
Do not mix ComPtr and std::shared_ptr on the same object

khouzam:
We were tracking the references to the `XamlBuilder` through `std::shared_ptr` and `ComPtr`
This would cause the object to be deleted when one of these would go down to 0 references even if the other count wasn't done yet.

We can only have 1 set of reference count, so consolidating to ComPtr.


paulcam206:
In order to make sure we don't make this mistake again, we should restrict how you can construct a `XamlBuilder`. I think we should mark its ctor private and add `MakeAndInitialize` as a friend:

```c++
friend HRESULT Microsoft::WRL::Details::MakeAndInitialize<AdaptiveNamespace::XamlBuilder, AdaptiveNamespace::XamlBuilder>(AdaptiveNamespace::XamlBuilder**);
```

paulcam206:
We might also want to consider a future change to apply this to other classes implementing `RuntimeClass`...

PR #2382, opened 2019-02-11
[Shared][UWP]Add columnset style support 

almedina-ms:
Completes this work item: https://github.com/Microsoft/AdaptiveCards/issues/2372
And this one too: https://github.com/Microsoft/AdaptiveCards/issues/2371


source/uwp/Renderer/lib/AdaptiveColumnSet.h
```diff
@@ -28,6 +28,9 @@ namespace AdaptiveNamespace
         IFACEMETHODIMP get_SelectAction(_COM_Outptr_ ABI::AdaptiveNamespace::IAdaptiveActionElement** action);
         IFACEMETHODIMP put_SelectAction(_In_ ABI::AdaptiveNamespace::IAdaptiveActionElement* action);
 
+        IFACEMETHODIMP get_Style(_Out_ ABI::AdaptiveNamespace::ContainerStyle* style);
+        IFACEMETHODIMP put_Style(ABI::AdaptiveNamespace::ContainerStyle style);
```
paulcam206: We should get some new tests added to UWP's ObjectModelTest :)

almedina-ms: How can I add unit tests? I tried running the UWPUnitTests but I only see a white screen

---
In reply to: [255733667](https://github.com/Microsoft/AdaptiveCards/pull/2382#discussion_r255733667) [](ancestors = 255733667)

paulcam206: Ah -- if you just run the app that's what you'll get. Try running the tests by first showing the Test Explorer window (`Test | Windows | Test Explorer`), then building, then picking the tests to run in the Test Explorer window. Note that `UnitTest` currently runs all the testapp stuff, so you may want to skip running that if you've already run the testapp tests.

---
In reply to: [256072539](https://github.com/Microsoft/AdaptiveCards/pull/2382#discussion_r256072539) [](ancestors = 256072539,255733667)

PR #2391, opened 2019-02-13
Misc. test app improvements

paulcam206:
During my snow day I came across some issues in the UWP test app that blocked me from getting stuff done efficiently:

* My laptop runs in high DPI and I didn't want to switch to regular DPI just to run JSON serialization tests, so I added a `Run Anyway` button to the DPI warning
* It's really painful to manually inspect JSON diffs where everything is on one line. I introduced a formatting routine for the purposes of diffing to make it easier to see the differences.

### DIFF
--- a/source/uwp/AdaptiveCardTestApp/Views/DiffView.xaml.cs
+++ b/source/uwp/AdaptiveCardTestApp/Views/DiffView.xaml.cs
@@ -1,8 +1,8 @@
 using DiffPlex;
 using DiffPlex.DiffBuilder;
 using DiffPlex.DiffBuilder.Model;
+using Newtonsoft.Json;
 using System.Linq;
-using System.Text;
 using Windows.UI;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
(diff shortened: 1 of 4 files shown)


paulcam206:
I should note that I did this yesterday without any internet access. There's probably a better way to deal with the `UICommand` used in the `MessageDialog`.

Also, @RebeccaAnne or @khouzam, any ideas on making the content dialog hosting the diffview wider? I've tried a bunch of things, but figure I'm missing something simple. :)

paulcam206:
In case it's not clear what we get from this change:

image

source/uwp/AdaptiveCardTestApp/Views/DiffView.xaml.cs
```diff
@@ -18,8 +19,107 @@ public DiffView()
             this.InitializeComponent();
         }
 
+        // a simplistic JSON formatter for use in displaying for diffview
+        private string FormatJson(string json)
```
RebeccaAnne: >FormatJson [](start = 23, length = 10)

I used JsonConvert in my visualizer updates for my templating prototype to achieve something similar without having to have hand rolled code for it. Not sure if this is the best approach but it got the job done:

dynamic parsedJson = JsonConvert.DeserializeObject(templateResult.ToString());
string TemplatedJson = JsonConvert.SerializeObject(parsedJson, Formatting.Indented);

paulcam206: I'll give that a try. I was without internet access when I wrote this :)

---
In reply to: [257412073](https://github.com/Microsoft/AdaptiveCards/pull/2391#discussion_r257412073) [](ancestors = 257412073)

paulcam206: works like a champ!

---
In reply to: [257415073](https://github.com/Microsoft/AdaptiveCards/pull/2391#discussion_r257415073) [](ancestors = 257415073,257412073)

PR #2395, opened 2019-02-13
[.NET] Apply ContainerStyle Foreground Colors + default Foreground Colors

carlos-zamora:
Bugfix for #2285 .

`AdaptiveRenderContext` was modified to keep track of the container style. When TextBlock text is being rendered, it now reads the value in `AdaptiveRenderContext` and bases it off that.

`ContainerStylesConfig` now has appropriate default values that match up with SharedModel.

Test for HTML was created. There is no way to make a test for WPF.

### DIFF
--- a/source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
+++ b/source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
@@ -596,12 +596,14 @@ namespace AdaptiveCards.Rendering.Html
                 .Style("flex", "1 1 100%");
             }
 
+            // Keep track of ContainerStyle.ForegroundColors before Container is rendered
             var outerStyle = context.ForegroundColors;
             if (container.Style != null)
             {
                 // Apply background color
                 ContainerStyleConfig containerStyle = context.Config.ContainerStyles.GetContainerStyleConfig(container.Style);
                 uiContainer.Style("background-color", context.GetRGBColor(containerStyle.BackgroundColor));
+
                 context.ForegroundColors = containerStyle.ForegroundColors;
             }
 
@@ -623,6 +625,7 @@ namespace AdaptiveCards.Rendering.Html
 
             AddSelectAction(uiContainer, container.SelectAction, context);
 
+            // Revert context's value to that of outside the Container
             context.ForegroundColors = outerStyle;
             return uiContainer;
         }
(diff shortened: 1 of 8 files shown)


source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
```diff
@@ -596,11 +596,13 @@ protected static HtmlTag ContainerRender(AdaptiveContainer container, AdaptiveRe
                 .Style("flex", "1 1 100%");
             }
 
+            var outerStyle = context.ForegroundColors;
```
paulcam206: >       var outerStyle = context.ForegroundColors; [](start = 5, length = 49)

could probably use a comment here

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveCardRenderer.cs
```diff
@@ -228,7 +228,8 @@ void MediaClickCallback(object sender, AdaptiveMediaEventArgs args)
                 Config = HostConfig ?? new AdaptiveHostConfig(),
                 Resources = Resources,
                 ElementRenderers = ElementRenderers,
-                Lang = card.Lang
+                Lang = card.Lang,
+                ForegroundColors = (HostConfig == null) ? HostConfig.ContainerStyles.Default.ForegroundColors : new ContainerStylesConfig().Default.ForegroundColors
```
paulcam206: >(HostConfig == null) ? HostConfig.ContainerStyles.Default.ForegroundColors [](start = 35, length = 74)

is this backwards?

paulcam206: (and below)

---
In reply to: [257000986](https://github.com/Microsoft/AdaptiveCards/pull/2395#discussion_r257000986) [](ancestors = 257000986)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveContainerRenderer.cs
```diff
@@ -13,11 +13,13 @@ public static FrameworkElement Render(AdaptiveContainer container, AdaptiveRende
             uiContainer.Style = context.GetStyle("Adaptive.Container");
             uiContainer.SetBackgroundSource(container.BackgroundImage, context);
 
+            var outerStyle = context.ForegroundColors;
```
paulcam206: >ar outerStyle = context.ForegroundColors; [](start = 13, length = 41)

comment here too

Issue #2389, opened 2019-02-12
[.NET] HTML Media can't be tabbed


khouzam:
# Platform
* .NET HTML

A card with a Media Player should be able to tab through the media and make it play without requiring the mouse.

PR #2401, opened 2019-02-15
[.NET] Tab and Activate HTML Media

carlos-zamora:
Bugfix for #2389 

`tabindex = "0"` makes media focusable according to its position.

Changes to the Javascript in `Program.cs` just move the code to activate the media into the `activateMedia()` function. Then have that be called when the "click" event occurs of if the Spacebar is pressed (when focused on).

Also updated the event listener for inline actions. Apparently, `keycode` is deprecated ([ref](https://developer.mozilla.org/en-US/docs/Web/Events/keydown)).

# Test Card
```
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "speak": "<s>Your  meeting about \"Adaptive Card design session\"<break strength='weak'/> is starting at 12:30pm</s><s>Do you want to snooze <break strength='weak'/> or do you want to send a late notification to the attendees?</s>",
  "body": [
    {
      "type": "TextBlock",
      "text": "Adaptive Card design session",
      "size": "large",
      "weight": "bolder"
    },
    {
      "type": "TextBlock",
      "text": "Conf Room 112/3377 (10)",
      "isSubtle": true
    },
    {
      "type": "TextBlock",
      "text": "12:30 PM - 1:30 PM",
      "isSubtle": true,
      "spacing": "none"
    },
    {
      "type": "TextBlock",
      "text": "Snooze for"
    },
    {
      "type": "Input.ChoiceSet",
      "id": "snooze",
      "style": "compact",
      "value": "5",
      "choices": [
        {
          "title": "5 minutes",
          "value": "5",
          "isSelected": true
        },
        {
          "title": "15 minutes",
          "value": "15"
        },
        {
          "title": "30 minutes",
          "value": "30"
        }
      ]
    },
    {
            "type": "Media",
            "poster": "https://adaptivecards.io/content/poster-video.png",
            "sources": [
                {
                    "mimeType": "video/mp4",
                }
            ]
        }
  ],
  "actions": [
    {
      "type": "Action.Submit",
      "title": "Snooze",
      "data": {
        "x": "snooze"
      }
    },
    {
      "type": "Action.Submit",
      "title": "I'll be late",
      "data": {
        "x": "late"
      }
    }
  ]
}
```

### DIFF
--- a/source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
+++ b/source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
@@ -965,6 +965,7 @@ namespace AdaptiveCards.Rendering.Html
             var thumbnailButton = new DivTag()
                 .AddClass("ac-media-poster")
                 .Attr("role", "button")
+                .Attr("tabindex", "0")
                 .Attr("aria-label", "Play media")
                 .Attr("role", "contentinfo")
                 .Style("position", "relative")
--- a/source/dotnet/Samples/AdaptiveCards.Sample.Html/Program.cs
+++ b/source/dotnet/Samples/AdaptiveCards.Sample.Html/Program.cs
@@ -161,29 +161,44 @@ namespace AdaptiveCards.Sample.Html
         // Sample JavaScript code to make media elements work
         const mediaPosterButtons = document.getElementsByClassName('ac-media-poster');
         const allowInlinePlayback = false;
+
+        function activateMedia(button) {{
+            const mediaId = button.dataset.acMediaid;
+            const mediaPlayerContainer = document.getElementById(mediaId);
+
+            if (mediaPlayerContainer)
+            {{
+                // Hide the poster
+                button.style.display = 'none';
+
+                // Show the media player container
+                mediaPlayerContainer.style.display = '';
+
+                // Play the media
+                const mediaPlayer = document.getElementById(`${{mediaId}}-player`);
+                mediaPlayer.play();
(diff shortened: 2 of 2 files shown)


carlos-zamora:
@matthidinger @andrewleader Not sure about one thing here. Inline stuff is activated with "Enter" but the poster and HTML buttons are activated with "Spacebar". Does it make sense to have these be different?

andrewleader:
@carlos-zamora For buttons, it seems to be standard practice (on the web and in Windows apps) that BOTH "Spacebar" and "Enter" will invoke the button when it has focus.

For the "inline stuff", what are you referring to? An image with a selectAction? Or a hyperlink within text? Or something else? For those two I mentioned, behavior on the web seems to be that only "Enter" works there (Spacebar scrolls the page down).

carlos-zamora:
By "inline stuff" I'm referring to the media controls (i.e.: fullscreen, play/pause). I'll go ahead and make HTML buttons respond to "Enter" too.

andrewleader:
Cool. Looks like for media controls, YouTube responds to both enter and spacebar, so I'd follow that practice. Thanks Carlos!

carlos-zamora:
From what I found, there doesn't seem to be a way to get the id of the media controls and hook up `Spacebar` to it. They operate with `Enter` by default. Most of what I found was people creating their own controls and hooking them up (pretty straightforward).

Since there's still a straightforward way of interacting with media without a mouse, I'll close and merge. If we want the media controls to be activated with `Spacebar` instead/in addition to `Enter` (or creating our own custom controls), we should add that as a new work item.

PR #2447, opened 2019-02-21
[Android] Update Android ResourceResolver to mimic UWP and iOS behavior

almedina-ms:
While checking this PR https://github.com/Microsoft/AdaptiveCards/pull/2315 I remembered that Android does the resource resolving in a more limited way, just allowing to override the loadOnlineImage and loadDataUriImage which makes the possibilities for changing the way images are retrieved too limited. 

I removed the IDataUriLoader class and renamed IOnlineImageLoader to IResourceResolver to keep consistency with the other platforms, also the methods inside the class were renamed to keep the consistency with the class name change.

Finally, the way the resource resolvers are queried after being registered was also changed to make the code simpler and cleaner.

**How was this change tested:**

I included a Image.ResourceResolver.json file which includes an image with a uri starting with "package:". I registered a "package" Resource Resolver which grabs the requested image from the drawable folder (just as in the PR mentioned earlier)


paulcam206:
update looks good :metal:

PR #2451, opened 2019-02-22
[.NET] Update date and time input handling to only support desired formats

RebeccaAnne:


source/dotnet/Library/AdaptiveCards.Rendering.Wpf.Xceed/XceedDateInput.cs
```diff
@@ -1,26 +1,35 @@
...
 namespace AdaptiveCards.Rendering.Wpf
 {
 
     public static class XceedDateInput
     {
+        // Validate times are of the format YYYY-MM-DD. This allows us to check the string against a regular expression for
+        // this format before passing to DateTime.TryParse to ensure we support only this single format.
+        private static bool IsSupportedDateFormat(string dateString)
+        {
+            Regex TextFunctionRegex = new Regex(@"^(\d{4})-(\d{2})-(\d{2})$");
```
paulcam206: >Regex TextFunctionRegex = new Regex(@"^(\d{4})-(\d{2})-(\d{2})$"); [](start = 12, length = 66)

Not sure if it's worth the effort, but this regex might benefit perf-wise with being made a static member and being constructed with the `RegexOptions.Compiled` option

paulcam206: same in the other file...

---
In reply to: [260322301](https://github.com/Microsoft/AdaptiveCards/pull/2451#discussion_r260322301) [](ancestors = 260322301)

PR #2454, opened 2019-02-23
[Android] Fix build errors

almedina-ms:
For some reason, a merge was not made correctly and it introduced previous changes, this fixes the compilation errors


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ColumnRenderer.java
```diff
@@ -114,18 +112,6 @@ else if ((column = Column.dynamic_cast(baseCardElement)) == null)
                     context.getResources().getDisplayMetrics().widthPixels,
 					backgroundImageProperties);
```
RebeccaAnne: >					 [](start = 0, length = 5)

While you're here, could you fix these tabs?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ContainerRenderer.java
```diff
@@ -115,18 +113,6 @@ else if ((container = Container.dynamic_cast(baseCardElement)) == null)
                     context.getResources().getDisplayMetrics().widthPixels,
 					backgroundImageProperties);
```
RebeccaAnne: >					 [](start = 0, length = 5)

tabs

PR #2494, opened 2019-03-01
[UWP] Add support for custom action rendering

RebeccaAnne:
Added custom action rendering to UWP:

- Add `AdaptiveActionRendererRegistration `and `IAdaptiveActionRenderer` interface. Update existing classes for known types to implement, and rename the handlers for known types from *Parser to *Renderer for consistency with element types. Hook up action set implementation to call registered renderers.

- Refactor ShowCard code so that `IAdaptiveActionInvoker.SendActionEvent` has enough information to toggle the inline show cards. This means that if you’re overriding show card (or toggle visibility) actions, you just need to call `SendActionEvent` in your click handler (or whatever) to trigger the action just as you would for actions that call back to the host. This works by registering each show card action by its internal id (and the internal id of it's actions set) so that when `SendActionEvent` is called it can toggle the correct UIElements for that action.

- In order to make this work for the case where the host is custom rendering the entire ActionSet, I added a new method to the RenderContext to allow the host to register the potentially custom drawn inline show cards:
```
        HRESULT AddInlineShowCard([in] AdaptiveActionSet * inputValue,
                                  [in] AdaptiveShowCardAction * showCardAction,
                                  [in] Windows.UI.Xaml.UIElement * showCardUIElement);
```
This “registers” show cards so that when SendActionEvent is called it knows which cards to toggle. This is only needed if the host is custom drawing the ActionSet, in which case they become responsible for any associated show cards. [Note: Naming is AddInlineShowCard for consistency with the existing AddInputValue which serves a similar purpose]

- Put the OriginatingHostConfig on the rendered adaptive card. This is used to determine if the card was rendered with inline show actions.

- Added two new properties to IAdaptiveRenderArgs:
1. `IsInShowCard` is true when the element we're rendering is inside a show card. This is needed by our implementation to determine which style to apply to the card. 
2. `AllowAboveTitleIconPlacement` is added to handle the requirement that icons are only placed above the title if all buttons in the set have icons. This allows the renderer of a single action to know whether or not that is the case.


source/shared/cpp/ObjectModel/BaseElement.h
```diff
@@ -20,6 +20,8 @@ namespace AdaptiveSharedNamespace
         static InternalId Current();
         static constexpr unsigned int Invalid = 0;
 
+        std::size_t const Hash() const { return std::hash<int>()(m_internalId); }
```
paulcam206: >int [](start = 58, length = 3)

should this be `unsigned int`?

source/uwp/Renderer/lib/AdaptiveActionRendererRegistration.cpp
```diff
@@ -0,0 +1,44 @@
...
+    {
+        m_registration = std::make_shared<RegistrationMap>();
+        return S_OK;
+    }
+    CATCH_RETURN;
+
+    HRESULT AdaptiveActionRendererRegistration::Set(_In_ HSTRING type, _In_ IAdaptiveActionRenderer* renderer)
+    {
+        ComPtr<IAdaptiveActionRenderer> localRenderer(renderer);
+        (*m_registration)[HStringToUTF8(type)] = localRenderer;
```
paulcam206: >(*m_registration)[HStringToUTF8(type)] [](start = 8, length = 38)

thought I was reading objc for a sec there :D

source/uwp/Renderer/lib/AdaptiveActionRendererRegistration.cpp
```diff
@@ -0,0 +1,44 @@
...
+        if (found != m_registration->end())
+        {
+            found->second.CopyTo(result);
+        }
+        return S_OK;
+    }
+
+    HRESULT AdaptiveActionRendererRegistration::Remove(_In_ HSTRING type)
+    {
+        m_registration->erase(HStringToUTF8(type));
```
paulcam206: >erase [](start = 24, length = 5)

will this throw if `type` isn't already registered?

source/uwp/Renderer/lib/RenderedAdaptiveCard.h
```diff
@@ -24,43 +24,53 @@ namespace AdaptiveNamespace
...
+
+        HRESULT AddInputValue(_In_ ABI::AdaptiveNamespace::IAdaptiveInputValue* inputValue);
+        void SetFrameworkElement(_In_ ABI::Windows::UI::Xaml::IFrameworkElement* value);
+        void SetOriginatingCard(_In_ ABI::AdaptiveNamespace::IAdaptiveCard* value);
+        void SetOriginatingHostConfig(_In_ ABI::AdaptiveNamespace::IAdaptiveHostConfig* value);
+        HRESULT SendActionEvent(_In_ ABI::AdaptiveNamespace::IAdaptiveActionElement* eventArgs);
+        HRESULT SendMediaClickedEvent(_In_ ABI::AdaptiveNamespace::IAdaptiveMedia* eventArgs);
 
     private:
+        HRESULT RenderedAdaptiveCard::HandleInlineShowCardEvent(_In_ ABI::AdaptiveNamespace::IAdaptiveActionElement* actionElement);
```
paulcam206: >RenderedAdaptiveCard:: [](start = 16, length = 22)

don't need class name here

source/uwp/Renderer/lib/RenderedAdaptiveCard.cpp
```diff
@@ -124,6 +214,34 @@ namespace AdaptiveNamespace
...
+    }
+
+    HRESULT RenderedAdaptiveCard::AddInlineShowCard(_In_opt_ IAdaptiveActionSet* actionSet,
+                                                    _In_ IAdaptiveShowCardAction* showCardAction,
+                                                    _In_ ABI::Windows::UI::Xaml::IUIElement* showCardFrameworkElement) try
+    {
+        InternalId actionSetId;
+        if (actionSet != nullptr)
+        {
+            // If this action in in an actionset, store the action set id
```
paulcam206: >in [](start = 30, length = 2)

nit: typo

source/uwp/Renderer/lib/RenderedAdaptiveCard.cpp
```diff
@@ -124,6 +214,34 @@ namespace AdaptiveNamespace
...
+            ComPtr<AdaptiveNamespace::AdaptiveActionSet> actionSetImpl =
+                PeekInnards<AdaptiveNamespace::AdaptiveActionSet>(actionSet);
+            actionSetId = actionSetImpl->GetInternalId();
+        }
+
+        ComPtr<AdaptiveNamespace::AdaptiveShowCardAction> showCardImpl =
+            PeekInnards<AdaptiveNamespace::AdaptiveShowCardAction>(showCardAction);
+        InternalId showCardActionId = showCardImpl->GetInternalId();
+
+        m_showCards.emplace(std::make_pair(showCardActionId, std::make_pair(actionSetId, showCardFrameworkElement)));
```
paulcam206: >std::make_pair(showCardActionId, std::make_pair(actionSetId, showCardFrameworkElement)) [](start = 28, length = 87)

I think it's fine if you prefer this style (it's more explicit), but you should also be able to pull it off with an initializer list:

```
m_showCards.emplace({showCardActionId, {actionSetId, showCardFrameworkElement}});
```

RebeccaAnne: That approach doesn't seem to immediately compile.  Possibly i could get it to work if I futzed with it enough, but I'm going to leave it as is for now.

---
In reply to: [261749093](https://github.com/Microsoft/AdaptiveCards/pull/2494#discussion_r261749093) [](ancestors = 261749093)

paulcam206: fine by me :)

source/uwp/Renderer/lib/AdaptiveActionRendererRegistration.cpp
RebeccaAnne: Looks like no. This is ultimately an unordered_map, which according to the documentation doesn't throw on not found. Just confirmed by adding a check to the test case.

---
In reply to: [261747139](https://github.com/Microsoft/AdaptiveCards/pull/2494#discussion_r261747139) [](ancestors = 261747139)

PR #2504, opened 2019-03-05
[UWP] Element fallback support

### LINKED ISSUE #2125: [UWP] Fallback and requires properties on elements

UWP implementation of #2019

paulcam206:
Fixes #2125. Implement fallback support in UWP renderer.


samples/Tests/DeepFallback.json
```diff
@@ -0,0 +1,124 @@
+{
```
RebeccaAnne: Don't forget to update the test app :)

source/shared/cpp/ObjectModel/BaseElement.h
```diff
@@ -84,6 +84,12 @@ namespace AdaptiveSharedNamespace
         // Fallback and Requires support
         FallbackType GetFallbackType() const { return m_fallbackType; }
         std::shared_ptr<BaseElement> GetFallbackContent() const { return m_fallbackContent; }
+        void SetFallbackType(FallbackType type) { m_fallbackType = type; }
+        void SetFallbackContent(std::shared_ptr<BaseElement> element)
+        {
+            m_fallbackContent.reset();
```
RebeccaAnne: >m_fallbackContent.reset(); [](start = 12, length = 26)

Is this needed?

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -560,6 +572,9 @@ AdaptiveNamespaceStart
         [propget] HRESULT IsVisible([out, retval] boolean * value);
         [propput] HRESULT IsVisible([in] boolean value);
 
+        [propget] HRESULT FallbackType([out, retval] FallbackType * value);
+        [propget] HRESULT FallbackContent([out, retval] IAdaptiveCardElement ** content);
```
RebeccaAnne: Do we need putters for these as well?

source/uwp/Renderer/lib/AdaptiveActionElement.cpp
```diff
@@ -23,6 +24,15 @@ namespace AdaptiveNamespace
         RETURN_IF_FAILED(UTF8ToHString(sharedModel->GetSentiment(), m_sentiment.GetAddressOf()));
 
         m_internalId = sharedModel->GetInternalId();
+        m_fallbackType = MapSharedFallbackTypeToUwp(sharedModel->GetFallbackType());
```
RebeccaAnne: >MapSharedFallbackTypeToUwp [](start = 25, length = 26)

Agree that this is better, but in general this code just casts the enums and assumes they match.

RebeccaAnne:
>     HRESULT AdaptiveActionElementBase::SetSharedElementProperties(std::shared_ptr<AdaptiveSharedNamespace::BaseActionElement> sharedCardElement)

Copy the fallback content back to the shared model

---
Refers to: source/uwp/Renderer/lib/AdaptiveActionElement.cpp:97 in 6284407. [](commit_id = 62844078d2f9fefd567bf9541383be64de801959, deletion_comment = False)

RebeccaAnne:
>                                 }

Can we make this chunk of code a function or otherwise refactor so that there's only one copy?

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:964 in 6284407. [](commit_id = 62844078d2f9fefd567bf9541383be64de801959, deletion_comment = False)

RebeccaAnne:
>                                 // TODO: something

What does this TODO represent? Any reason not to combine with the None/default case for the time being?

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:972 in 6284407. [](commit_id = 62844078d2f9fefd567bf9541383be64de801959, deletion_comment = False)

RebeccaAnne:
>                     // throw exception so that parent's fallback can be invoked

Update comment, we're not throwing

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:991 in 6284407. [](commit_id = 62844078d2f9fefd567bf9541383be64de801959, deletion_comment = False)

RebeccaAnne:
>         boolean ancestorHasFallback;

Looks like this is unused

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:2763 in 6284407. [](commit_id = 62844078d2f9fefd567bf9541383be64de801959, deletion_comment = False)

RebeccaAnne:
>                 while (!renderer)

Do we support the case where if show card has something unsupported the action could fallback to a different action?

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:1464 in 6284407. [](commit_id = 62844078d2f9fefd567bf9541383be64de801959, deletion_comment = False)

source/shared/cpp/ObjectModel/BaseElement.h
paulcam206: suppose not :)

---
In reply to: [263172385](https://github.com/Microsoft/AdaptiveCards/pull/2504#discussion_r263172385) [](ancestors = 263172385)

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
paulcam206: at the moment, at least, we're only setting content as an internal detail, so we peek over to the native type. that said, I suppose in order to use fallback via the object model (which is admittedly a *little* weird), you need to be able to set these.

---
In reply to: [263172755](https://github.com/Microsoft/AdaptiveCards/pull/2504#discussion_r263172755) [](ancestors = 263172755)

paulcam206:
>     HRESULT AdaptiveActionElementBase::SetSharedElementProperties(std::shared_ptr<AdaptiveSharedNamespace::BaseActionElement> sharedCardElement)

nice catch!

---
In reply to: [470313901](https://github.com/Microsoft/AdaptiveCards/pull/2504#issuecomment-470313901) [](ancestors = 470313901)

---
Refers to: source/uwp/Renderer/lib/AdaptiveActionElement.cpp:97 in 6284407. [](commit_id = 62844078d2f9fefd567bf9541383be64de801959, deletion_comment = False)

paulcam206:
>                                 // TODO: something

yeah, I thought I was going to have to do something extra, but it turned out not to be the case.

---
In reply to: [470316547](https://github.com/Microsoft/AdaptiveCards/pull/2504#issuecomment-470316547) [](ancestors = 470316547)

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:972 in 6284407. [](commit_id = 62844078d2f9fefd567bf9541383be64de801959, deletion_comment = False)

paulcam206:
>         boolean ancestorHasFallback;

it's actually getting passed in to our `IterateOverVector` helper below

---
In reply to: [470318180](https://github.com/Microsoft/AdaptiveCards/pull/2504#issuecomment-470318180) [](ancestors = 470318180)

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:2763 in 6284407. [](commit_id = 62844078d2f9fefd567bf9541383be64de801959, deletion_comment = False)

paulcam206:
>                 while (!renderer)

hoo boy. I'm not sure. This is probably something best answered by @andrewleader or @matthidinger :)

---
In reply to: [470318921](https://github.com/Microsoft/AdaptiveCards/pull/2504#issuecomment-470318921) [](ancestors = 470318921)

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:1464 in 6284407. [](commit_id = 62844078d2f9fefd567bf9541383be64de801959, deletion_comment = False)

paulcam206:
>                                 }

yeah, sorry :)

---
In reply to: [470316170](https://github.com/Microsoft/AdaptiveCards/pull/2504#issuecomment-470316170) [](ancestors = 470316170)

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:964 in 6284407. [](commit_id = 62844078d2f9fefd567bf9541383be64de801959, deletion_comment = False)

andrewleader:
@paulcam206 I have no clue about [that while loop](https://github.com/Microsoft/AdaptiveCards/blob/9be09d5e74a1634c6dbd63ba71bfd7ceea6b1168/source/uwp/Renderer/lib/XamlBuilder.cpp#L1437), @RebeccaAnne ?

samples/Tests/DeepFallback.json
paulcam206: Done!

paulcam206:
> @paulcam206 I have no clue about [that while loop](https://github.com/Microsoft/AdaptiveCards/blob/9be09d5e74a1634c6dbd63ba71bfd7ceea6b1168/source/uwp/Renderer/lib/XamlBuilder.cpp#L1437), @RebeccaAnne ?

think there was some missing context. @RebeccaAnne originally said this:

> Do we support the case where if show card has something unsupported the action could fallback to a different action?

andrewleader:
> Do we support the case where if show card has something unsupported the action could fallback to a different action?

Ah, is the question whether we SHOULD support that? I believe we should (both an Action itself should be able to have a fallback, and the showCard card within the Action.ShowCard should be able to have a fallback).

I guess the only problem there is the showCard card having a fallback, since we didn't add the fallback property to the AdaptiveCard itself, correct?

paulcam206:
> Ah, is the question whether we SHOULD support that? I believe we should (both an Action itself should be able to have a fallback, and the showCard card within the Action.ShowCard should be able to have a fallback).

I believe showcard cards support fallback (i.e. if a showcard card has an unknown element and specifies valid fallback, the fallback will be used).

> I guess the only problem there is the showCard card having a fallback, since we didn't add the fallback property to the AdaptiveCard itself, correct?

I think that's its own distinct issue that we should probably also talk about. I think what @RebeccaAnne was asking (correct me if I'm wrong) is if a showcard action has fallback content and the card supplied to the showcard has unknown elements but no fallback strategy, should the showcard action's fallback be invoked?

andrewleader:
Ah, so I think we're talking about the following payload then...

```json
"actions": [
  "type": "Action.ShowCard",
  "title": "Click me",
  "card": {
    "type": "AdaptiveCard",
    "body": [
      {
        "type": "UnknownElement"
      }
    ]
  },
  "fallback": {
    "type": "Action.OpenUrl",
    "title": "Fallback click me",
    "url": "https://msn.com
  }
]
```

And the question is whether that ShowCard's fallback (falling back to an OpenUrl) should be used since the ShowCard Card has an unknown element in its body.

I think YES, it should engage the fallback (to the OpenUrl in this case). The error (the unknown element) bubbles up.

RebeccaAnne:
>                 XamlHelpers::CreateXamlClass<IGrid>(HStringReference(RuntimeClass_Windows_UI_Xaml_Controls_Grid));

This throws. We'd probably ideally want it to update it to return HRESULT and have the created class as an out parameter, although it's used all over this file...

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:1345 in 32619b5. [](commit_id = 32619b5fbc728904f8c73c4213f6039a4d88b3a6, deletion_comment = False)

source/uwp/Renderer/lib/XamlHelpers.h
```diff
@@ -19,6 +19,47 @@ namespace AdaptiveNamespace
             return result;
         }
 
+       template<typename T, typename TInterface, typename C>
+       static HRESULT IterateOverVector(_In_ ABI::Windows::Foundation::Collections::IVector<T*>* vector, const boolean stopOnFailure, C iterationCallback)
```
RebeccaAnne: >IterateOverVector [](start = 22, length = 17)

Hmm... having two copies of this seems fraught with peril, but it's fine for now as long as we have a plan to consolidate. Maybe make sure to log an issue to address the remainder of the exception vs hresult issues before you check in so we're sure to come back to this.

paulcam206: Good point. I've filed #2533 to track this.

---
In reply to: [264376536](https://github.com/Microsoft/AdaptiveCards/pull/2504#discussion_r264376536) [](ancestors = 264376536)

paulcam206:
>                 XamlHelpers::CreateXamlClass<IGrid>(HStringReference(RuntimeClass_Windows_UI_Xaml_Controls_Grid));

I've filed #2534 to track reviewing and fixing up our error handling strategy in XamlBuilder.cpp

---
In reply to: [471076906](https://github.com/Microsoft/AdaptiveCards/pull/2504#issuecomment-471076906) [](ancestors = 471076906)

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:1345 in ed60768. [](commit_id = ed60768c8ff16e802513aa023e162357e1ec46fa, deletion_comment = False)

source/uwp/Renderer/lib/XamlHelpers.h
paulcam206: > I think YES, it should engage the fallback (to the OpenUrl in this case). The error (the unknown element) bubbles up.

I've filed #2535 to track implementing this.

paulcam206:
> And the question is whether that ShowCard's fallback (falling back to an OpenUrl) should be used since the ShowCard Card has an unknown element in its body.
> 
> I think YES, it should engage the fallback (to the OpenUrl in this case). The error (the unknown element) bubbles up.

I've filed #2535 to track implementing this.

Issue #2359, opened 2019-02-05
[.NET] bleed property on elements


khouzam:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

.NET implementation of #2109

PR #2506, opened 2019-03-06
[NET] Add Bleed property support

almedina-ms:
Completes this work item: https://github.com/Microsoft/AdaptiveCards/issues/2359


paulcam206:
Looks like there's a test failure that needs to be investigated...

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveContainerRenderer.cs
```diff
@@ -137,8 +139,10 @@ public static void AddSeparator(AdaptiveRenderContext context, AdaptiveElement e
             Grid.SetRow(uiSep, uiContainer.RowDefinitions.Count - 1);
             uiContainer.Children.Add(uiSep);
         }
-
-        public static void ApplyPadding(Grid uiElement, AdaptiveTypedElement element, AdaptiveContainerStyle parentStyle, AdaptiveRenderContext context)
+        
+        // We have to give negative margin for this to work as expected in the case of bleed
+        // We have to give positive padding for this to work
```
paulcam206: this comment is a little confusing...

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveRenderArgs.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public AdaptiveRenderArgs Clone()
+        {
+            return new AdaptiveRenderArgs {
+                ParentStyle = ParentStyle,
+                ForegroundColors = ForegroundColors,
+                ColumnRelativePosition = ColumnRelativePosition,
+                HasParentWithPadding = HasParentWithPadding
+            };
+        }
```
paulcam206: rather than having a function do this, shouldn't we just have a copy ctor?

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
```diff
@@ -517,7 +517,9 @@ protected static HtmlTag ColumnRender(AdaptiveColumn column, AdaptiveRenderConte
                 .Style("display", "flex")
                 .Style("flex-direction", "column");
 
-            var parentContainerStyle = context.ParentStyle;
+            // var parentContainerStyle = context.ParentStyle;
```
paulcam206: >        // var parentContainerStyle = context.ParentStyle; [](start = 4, length = 58)

no longer needed

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveColumnSetRenderer.cs
```diff
@@ -12,29 +12,53 @@ public static FrameworkElement Render(AdaptiveColumnSet columnSet, AdaptiveRende
...
             foreach (var column in columnSet.Columns)
             {
+                var columnRenderArgs = elementRenderArgs.Clone();
+                if (columnSet.Columns.Count == 1)
+                {
+                    columnRenderArgs.ColumnRelativePosition = ColumnPositionEnum.Only;
+                }
+                else
+                {
+                    int index = columnSet.Columns.IndexOf(column);
```
paulcam206: >int index = columnSet.Columns.IndexOf(column); [](start = 20, length = 46)

if we need the index, maybe we should just use a regular `for` loop?

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveContainerRenderer.cs
```diff
@@ -137,27 +141,58 @@ public static void AddSeparator(AdaptiveRenderContext context, AdaptiveElement e
...
         }
-
-        public static void ApplyPadding(Grid uiElement, AdaptiveTypedElement element, AdaptiveContainerStyle parentStyle, AdaptiveRenderContext context)
+        
+        // For applying bleeding, we must know if the element has padding, so both properties are applied in the same method
+        public static bool ApplyPadding(Border border, Grid uiElement, AdaptiveCollectionElement element, AdaptiveRenderArgs parentRenderArgs, AdaptiveRenderContext context)
         {
             bool canApplyPadding = false;
 
             // AdaptiveColumn inherits from AdaptiveContainer so only one check is required for both
```
paulcam206: >// AdaptiveColumn inherits from AdaptiveContainer so only one check is required for both [](start = 12, length = 88)

is this true?

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
```diff
@@ -1761,26 +1783,60 @@ private static void ApplyBackgroundImage(AdaptiveBackgroundImage backgroundImage
             }
         }
 
-        private static void ApplyPadding(HtmlTag uiElement, AdaptiveTypedElement element, AdaptiveContainerStyle parentStyle, AdaptiveRenderContext context)
+        private static bool ApplyPadding(HtmlTag uiElement, AdaptiveCollectionElement element, AdaptiveRenderArgs parentRenderArgs, AdaptiveRenderContext context)
         {
             bool canApplyPadding = false;
 
             // AdaptiveColumn inherits from AdaptiveContainer so only one check is required for both
```
paulcam206: >// AdaptiveColumn inherits from AdaptiveContainer so only one check is required for both [](start = 12, length = 88)

is this still true?

samples/Tests/ColumnColumnSetContainer.Bleed.json
```diff
@@ -0,0 +1,315 @@
+{
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"type": "AdaptiveCard",
+	"version": "1.2",
+	"style": "default",
+	"body": [
+		{
+			"type": "ColumnSet",
+			"style": "emphasis",
+			"bleed": "true",
```
nesalang: should be boolean

samples/Tests/ColumnColumnSetContainer.Bleed.json
```diff
@@ -0,0 +1,315 @@
...
+							]
+						}
+					]
+				}
+			]
+		},
+		{
+			"type": "ColumnSet",
+			"style": "emphasis",
+			"bleed": "true",
```
nesalang: here are several other places in the file

samples/Tests/ColumnColumnSetContainer.Bleed.json
```diff
@@ -0,0 +1,315 @@
+{
```
nesalang: thanks for adding this test file.
can you break this into sevearl files?
It has several column sets and containers at top level.
since our parser doesn't tell us where it failed,
when parsing fails, it's really hard to debug it.

almedina-ms: Gilles told me to limit how many files we added to the tests as it would take more time to run the test app

---
In reply to: [264878238](https://github.com/Microsoft/AdaptiveCards/pull/2506#discussion_r264878238) [](ancestors = 264878238)

PR #2522, opened 2019-03-08
[Shared] Add RichTextBlocks, Paragraphs, and TextRun Inlines

### LINKED ISSUE #2376: [Shared] Inline text runs

Shared implementation of #1933

RebeccaAnne:
Add RichTextBlocks, Paragraphs and TextRuns to the shared model.

Added a TextElementProperties class to encapsulate the properties common to TextBlock and TextRun. My first version had this as a base class, however since Java doesn't support multiple inheritance of non-interface classes that approach didn't work for our setup. This version has TextElementProperties as a member of TextBlock and TextRun which at least allows us to share the serialization and deserialization code. 

Fixes #2376


RebeccaAnne:
> Looks great! iOS build failed, though it's not entirely clear why.

It's because i added new files. That requires some kind of sorcery on the iOs build that i can't accomplish because i don't have a mac. I'll ask @nesalang to help me sort it out.

paulcam206:
Looks good now! :shipit:

RebeccaAnne:
> Looks good now!

h/t @nesalang for updating ios for me

PR #2544, opened 2019-03-13
Updated Bleed Feature to handle nested ColumnSet

nesalang:
1. Adds logic for nested columns
2. it handle cases explained in the below.
`
             parent object
                       | 
              ColumnSetA (has padding)
                       |
              ColumnA0 | ColumnA ...
                       | 
              ColumnSetB 
                       |
******* ColumnB0 ***** |  *********** ColumnB1*************** | ************* ColumnB2
************  | ****************************** |*************************************** |
***** ContainerB0(has padding) *  | * ContainerB1(has padding) * | * ContainerB2(has padding)    

**Please note that objects have padding only if explicitly stated.**

In this example, only ContainerB0 will bleed among objects in ColumnSetB, and the bleed direction is to the leading edge of ColumnSetA only.


source/shared/cpp/ObjectModel/ColumnSet.cpp
```diff
@@ -47,11 +47,65 @@ std::shared_ptr<BaseCardElement> ColumnSetParser::Deserialize(ParseContext& cont
...
+        // Parse the element
+        auto el = Column::Deserialize(context, curJsonValue);
+        if (el != nullptr)
+        {
+            elements.push_back(el);
+            // restores the parent's bleed state
+            context.SetBleedState(previousBleedState);
+        }
+        ++currentIndex;
+    }
```
paulcam206: why not just use a regular `for` loop here since you're maintaining the index?

source/shared/cpp/ObjectModel/Enums.h
```diff
@@ -391,6 +391,15 @@ namespace AdaptiveSharedNamespace
     };
     // No mapping to string needed
 
+    enum class ContainerBleedState
+    {
+        BleedRestricted = 0,
+        BleedToLeading,
+        BleedToTrailing,
+        BleedToBothEdges,
+    };
```
paulcam206: Think this might read a bit better like this:

```
    enum class ContainerBleedType
    {
        None = 0,
        ToLeadingEdge,
        ToTrailingEdge,
        ToBothEdges,
    };
```

nesalang: thanks; will do

---
In reply to: [265290586](https://github.com/Microsoft/AdaptiveCards/pull/2544#discussion_r265290586) [](ancestors = 265290586)

source/shared/cpp/ObjectModel/ColumnSet.cpp
nesalang: sounds great. thanks

---
In reply to: [265289875](https://github.com/Microsoft/AdaptiveCards/pull/2544#discussion_r265289875) [](ancestors = 265289875)

source/shared/cpp/ObjectModel/Enums.cpp
```diff
@@ -21,7 +21,7 @@ namespace AdaptiveSharedNamespace
             {AdaptiveCardSchemaKey::BackgroundImage, "backgroundImage"},
             {AdaptiveCardSchemaKey::BackgroundImageUrl, "backgroundImageUrl"},
             {AdaptiveCardSchemaKey::BaseCardElement, "baseCardElement"},
-            {AdaptiveCardSchemaKey::Bleed, "bleed"},
+            //{AdaptiveCardSchemaKey::Bleed, "bleed"},
```
paulcam206: >//{AdaptiveCardSchemaKey::Bleed, "bleed"}, [](start = 12, length = 42)

What's this supposed to be? :)

source/shared/cpp/ObjectModel/ParseContext.cpp
```diff
@@ -261,12 +263,18 @@ namespace AdaptiveSharedNamespace
     void ParseContext::SaveContextForCollectionTypeElement(const std::shared_ptr<CollectionTypeElement>& current)
     {
         // save current style value
-        m_parentalContainerStyles.push_back(current->GetStyle());
+        if (current->GetStyle() != ContainerStyle::None)
```
paulcam206: >None [](start = 51, length = 4)

Should this be `None` or `NotSet`?

source/shared/cpp/ObjectModel/Enums.cpp
nesalang: thanks, that shouldn't be there

---
In reply to: [266114463](https://github.com/Microsoft/AdaptiveCards/pull/2544#discussion_r266114463) [](ancestors = 266114463)

source/shared/cpp/ObjectModel/ParseContext.cpp
nesalang: None is correct. container will have a None style.

---
In reply to: [266114732](https://github.com/Microsoft/AdaptiveCards/pull/2544#discussion_r266114732) [](ancestors = 266114732)

source/shared/cpp/ObjectModel/Enums.cpp
paulcam206: cool :)

---
In reply to: [266114743](https://github.com/Microsoft/AdaptiveCards/pull/2544#discussion_r266114743) [](ancestors = 266114743,266114463)

source/shared/cpp/ObjectModel/ParseContext.cpp
paulcam206: okay, sounds good :)

---
In reply to: [266114969](https://github.com/Microsoft/AdaptiveCards/pull/2544#discussion_r266114969) [](ancestors = 266114969,266114732)

samples/Tests/Nested.ColumnSet.Bleed.Complex.json
```diff
@@ -0,0 +1,110 @@
+  {
+    "type": "AdaptiveCard",
```
paulcam206: The UWP test app needs to be run against these new cards :)

nesalang: uwp doesn't have support for bleed, yet. am i missing something? i'm not sure what you mean.

---
In reply to: [266117823](https://github.com/Microsoft/AdaptiveCards/pull/2544#discussion_r266117823) [](ancestors = 266117823)

paulcam206: Just like you had to update the dotnet tests to account for the fact that it's not implemented yet, you need to run the UWP app and accept the failed results :)

---
In reply to: [266118492](https://github.com/Microsoft/AdaptiveCards/pull/2544#discussion_r266118492) [](ancestors = 266118492,266117823)

source/shared/cpp/ObjectModel/Enums.h
```diff
@@ -347,7 +347,10 @@ namespace AdaptiveSharedNamespace
         Good,
         Attention,
         Warning,
-        Accent
+        Accent,
+        // this is different from None, None is a style while NotSet isn't a style
```
RebeccaAnne: >None is a style [](start = 40, length = 15)

Are you sure "None" is a style? I'm thought we were using None to mean not-set already....

nesalang: none is a style in a sense that it has no attribute.
Not Set means value has not been initialized.

---
In reply to: [266177514](https://github.com/Microsoft/AdaptiveCards/pull/2544#discussion_r266177514) [](ancestors = 266177514)

RebeccaAnne: I'm sorry, i'm having trouble grokking this. So, we use none when the style has not been set. Saying that it "is a style" seems inaccurate (there's no host config for the "None" style, for example). If None is used when the style is not set, what is not-set for and in what case do we need to distinguish between them?

---
In reply to: [266181375](https://github.com/Microsoft/AdaptiveCards/pull/2544#discussion_r266181375) [](ancestors = 266181375,266177514)

RebeccaAnne: I think we could maybe just use None for this? Looks like the only place it's being used is as a return from GetParentalContainerStyles which is otherwise returning a value from m_parentalContainerStyles. But it also looks like you're never putting the None style into m_parentalContainterStyles, so returning None there could mean unset without ambiguity. Let me know if I'm missing something.

---
In reply to: [266523140](https://github.com/Microsoft/AdaptiveCards/pull/2544#discussion_r266523140) [](ancestors = 266523140,266181375,266177514)

paulcam206:
out of date

PR #2549, opened 2019-03-15
[UWP] Exposing user inputs through render context

### LINKED ISSUE #2532: [UWP] Expose inputs through render context
# Platform

* UWP

# Author or host
Host 

# Version of SDK
Latest master

# Issue
Some custom renderers will need the inputs from other input elements to be exposed in their render so that they can use them later on. I'm proposing exposing the inputs through the render context that is passed into the render. Currently the render context only allows addition of inputs and doesn't expose the list.

Example of where this is needed: I'm creating a custom renderer where we're calling an API on user input. The API requires other inputs which are decided by other input values in the adaptive card. Currently I have no access to these values within the custom element render.

[Link to PR](https://github.com/Microsoft/AdaptiveCards/pull/2549)

nimercer:
[Link to bug item](https://github.com/Microsoft/AdaptiveCards/issues/2532 )

Exposing the user input values from any input elements in the adaptive card through the Render Context.
This is useful for custom renderers which need to know the input values of other elements in the card as a whole.

Fixes #2532

### DIFF
--- a/source/uwp/Renderer/lib/AdaptiveRenderContext.h
+++ b/source/uwp/Renderer/lib/AdaptiveRenderContext.h
@@ -31,7 +31,7 @@ namespace AdaptiveNamespace
         IFACEMETHODIMP get_ActionRenderers(_COM_Outptr_ ABI::AdaptiveNamespace::IAdaptiveActionRendererRegistration** value) override;
         IFACEMETHODIMP get_ActionInvoker(_COM_Outptr_ ABI::AdaptiveNamespace::IAdaptiveActionInvoker** value) override;
         IFACEMETHODIMP get_MediaEventInvoker(_COM_Outptr_ ABI::AdaptiveNamespace::IAdaptiveMediaEventInvoker** value) override;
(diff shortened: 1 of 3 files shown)


source/uwp/Renderer/lib/AdaptiveRenderContext.h
```diff
@@ -31,6 +31,7 @@ namespace AdaptiveNamespace
         IFACEMETHODIMP get_ActionRenderers(_COM_Outptr_ ABI::AdaptiveNamespace::IAdaptiveActionRendererRegistration** value) override;
         IFACEMETHODIMP get_ActionInvoker(_COM_Outptr_ ABI::AdaptiveNamespace::IAdaptiveActionInvoker** value) override;
         IFACEMETHODIMP get_MediaEventInvoker(_COM_Outptr_ ABI::AdaptiveNamespace::IAdaptiveMediaEventInvoker** value) override;
+        IFACEMETHODIMP get_UserInputs(ABI::AdaptiveCards::Rendering::Uwp::IAdaptiveInputs** value) override;
```
RebeccaAnne: Please add the _COM_Outptr_ annotation here

nimercer: Done! Thanks

PR #2559, opened 2019-03-22
Refactor WPF toggle visibility code

andrewleader:
From an authoring side (constructing card programmatically using .NET library), the `ToggleVisibility.TargetElements` property was a list of objects, which is a bit confusing to a developer... they could place anything in there since it's just an object!

image

That's also tricky from a rendering side (how do you handle this non-strongly typed property).

I updated this to be strongly typed.

image

It still serializes and deserializes correctly (if there's only an ID specified, it serializes as a string). The unit tests verify this. And developers can use a string when authoring their card in C# too, thanks to implicit type conversion

image

Additionally, the code for actually toggling the visibility was contained within the renderer in a non-extensible way (meaning if a host wanted to use toggle visibility themselves, they would have to re-implment toggling the visibility themselves on the specified elements).

image

I updated this logic so that it's contained within the RenderContext, so that hosts can easily use this logic too.

image

**Tested and verified by**: All unit tests pass (and there's unit tests for both serialization and deserialization that validate both a string and object work correctly), and verified the sample payload in the WPF visualizer to confirm user functionality still works too.


dclaux:
So it looks like ToggleVisibility's effect is solely to change the Visibility property of the framework element associated with the specified Adaptive element Id. I just wanted to bring attention to a couple things:
- Does the framework element include the target element's potentially visible separator? If not, then ToggleVisibility may result in an orphaned separator
- Does setting the framework element's visibility trigger a re-layout of the card to make sure that bleeding elements bleed correctly? For instance, take the following payload:
```
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "TextBlock",
            "id": "title",
            "text": "This is a TextBlock at the top of the card",
            "wrap": true
        },
        {
            "type": "Container",
            "style": "emphasis",
            "bleed": true,
            "items": [
                {
                    "type": "TextBlock",
                    "text": "This is a bleeding container",
                    "size": "extraLarge"
                },
                {
                    "type": "ActionSet",
                    "actions": [
                        {
                            "title": "Click me",
                            "targetElements": [
                                "title"
                            ]
                        }
                    ]
                }
            ]
        }
    ]
}
```

Initially it renders like this:
image

Once the button is clicked, it should look like this:
image

andrewleader:
Really excellent feedback David. I'll try that out. Note, however, that would be a separate issue (this code is already in master and I'm just trying to improve the API surface, so fixes to how the UI behaves would be a separate pull request).

We ought to make sure all these corner cases work in all our renderers, excellent attention to detail David :)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionRenderer.cs
```diff
@@ -16,40 +16,7 @@ public static FrameworkElement Render(AdaptiveAction action, AdaptiveRenderConte
...
-                                // otherwise if we read something with the format {"elementId": <id>", "isVisible": false} or we just read the id and the element is visible
-                                else if ((targetElementIsVisible.HasValue && !targetElementIsVisible.Value) || (!targetElementIsVisible.HasValue && visibility == Visibility.Visible))
-                                {
-                                    elementFrameworkElement.Visibility = Visibility.Collapsed;
-                                }
-
-                            }
-
-                        }
+                        context.ToggleVisibility(toggleVisibilityAction.TargetElements);
```
RebeccaAnne: >ToggleVisibility [](start = 32, length = 16)

For what it's worth, in UWP i just put this under the equivalent of InvokeAction rather than having a separate function. That is, if you call InvokeAction on a toggle action it toggles the relevant elements. What do you think of that approach vs. this one?

andrewleader: One use case I was thinking of is if a host creates their own `MySuperAction` which lets you (1) toggle visibilities, (2) submit something via HTTP, and (3) toggle visibilities again after submission completed, it ought to be easy for the host to perform those. However, with that approach, it would require the host creating a temporary Action.ToggleVisibility and then calling InvokeAction on it?

RebeccaAnne: Fair enough, but if the host wants to trigger actions themselves, for example they've got custom action renderers, in the approach where you use a single function the host doesn't have to have this if (they can just call InvokeAction on all actions). BTW, does show card work with InvokeAction in dotnet? 

---
In reply to: [268849214](https://github.com/Microsoft/AdaptiveCards/pull/2559#discussion_r268849214) [](ancestors = 268849214)

andrewleader: I just discovered WPF doesn't support inline ShowCard :O it only supports popup.

But agreed, we should lump this logic into InvokeAction so hosts can just call that on all actions. I think in addition to that, we should still keep `ToggleVisibility` exposed for more custom use cases. I'll take a look at the UWP code and see how that looks and update the pull request.

andrewleader: Alright, I updated it so that all a host does is call `InvokeAction` and pass the action!

```csharp
                uiButton.Click += (sender, e) =>
                {
                    context.InvokeAction(uiButton, new AdaptiveActionEventArgs(action));
```

The renderer code then determines if it's a ToggleVisibility and handles it, otherwise it sends out the event to the host

```csharp
        public void InvokeAction(FrameworkElement ui, AdaptiveActionEventArgs args)
        {
            // ToggleVisibility is a renderer-handled action
            if (args.Action is AdaptiveToggleVisibilityAction toggleVisibilityAction)
            {
                this.ToggleVisibility(toggleVisibilityAction.TargetElements);
                return;
            }

            OnAction?.Invoke(ui, args);
        }
```

I think we should still keep `ToggleVisibility` public on the Context too, so that hosts could use that like I was describing. For example, hosts could extend validation and do something like this...

```json
{
  "type": "Input.Text",
  "validation": {
    "whenValidToggle": [
      { "id": "successText", "isVisible": true }
    ]    
```

And then they just pass that list to `context.ToggleVisibility`

RebeccaAnne: Should we log a bug for inline show cards for WPF?

---
In reply to: [268867015](https://github.com/Microsoft/AdaptiveCards/pull/2559#discussion_r268867015) [](ancestors = 268867015)

RebeccaAnne:
Could you log an issue to track this feedback (it probably applies to UWP as well)?

---
In reply to: [475806958](https://github.com/Microsoft/AdaptiveCards/pull/2559#issuecomment-475806958) [](ancestors = 475806958)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionRenderer.cs
andrewleader: 👍 Filed issue #2599

andrewleader:
Thanks Becky, created the issue to track the separator/bleed concerns for all the renderers: https://github.com/Microsoft/AdaptiveCards/issues/2600

Issue #2418, opened 2019-02-21
[.NET] Fallback and requires properties on elements


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

.NET implementation of #2019

PR #2479, opened 2019-02-26
Ignore Max, Min, and Value from AdaptiveNumberInput on Serialization

MattSFT:
proposed fix for this issue: https://github.com/Microsoft/AdaptiveCards/issues/2478

PR #2560, opened 2019-03-22
[.NET] Fallback support

carlos-zamora:
Spec: #2019 
Fixes: #2418 

# Parse-Time Changes
This is a big one. Basically, before Fallback could even be created, a few things had to be created and changed:
* Adding `AdaptiveUnknownElement`/`AdaptiveUnknownAction`:
  * Now, any elements/actions that aren't recognized are serialized into these types. We fill what we can for `AdaptiveTypedElement` (i.e.: type, fallback, id, etc.) and the rest gets thrown into `AdditionalProperties`
* Adding Requires:
  * mostly just copied over from the SharedModel
* ParseContext:
  * This is a static class used at parse-time to keep track of 2 things:
    1. Are we deserializing an element or an action? Handled by `ContextType`. Used to identify whether we create an `AdaptiveUnknownElement` or an `AdaptiveUnknownAction`
    2. ID Collision. This is really different now that there's fallback. It's implemented as close to the SharedModel version as possible. However, one of the tests on ID collision are failing because they are mis-identifying an ID collision and I can't figure out why. More details in "Known Bugs" at the end.
* Fallback:
  * Deserialization handled in `AdaptiveFallbackConverter::ParseFallback()`. Can't use the regular `ReadJson()` function because `objectID` and `internalID` need to be passed down.
  * Serialization should be handled in `AdaptiveFallbackConverter::WriteJson()` but the bug makes that not be the case.

# HTML and WPF Rendering
In general, adding a check for `AdaptiveUnknownElement/Action` in `AdaptiveElementRenderers.cs` just happened to make life easier. Now we return a null renderer, which means that in `AdaptiveRenderContext.cs`, we just look out for that and begin conducting fallback.

Rendering fallback content in an `AdaptiveUnknownElement` works. And it looks nice too.

Rendering fallback content in an `AdaptiveUnknownAction` hasn't been added. More details in "Known Bugs" below.

# How verified
* Existing fallback/requires test cards
* New unit tests

image

image

image

image

# Known Bugs
* **Render fallback content for an `AdaptiveUnknownAction`**: basically, we depend on knowning how many actions we have to be able to split the space evenly between them. Now with fallback, we need to preparse the list of actions and see how many we actually have. Take a look at `AdaptiveActionSetRenderer.cs` and how `actionsToProcess` is created. My guess is we'll have to make a few more changes to `GetActionsToProcess()`.


source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
```diff
@@ -106,42 +96,85 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
+                    var fallbackJson = jObject.GetValue("fallback");
+                    if (fallbackJson != null)
+                    {
+                        result.Fallback = AdaptiveFallbackConverter.ParseFallback(jObject.GetValue("fallback"), serializer, objectId, internalID);
+                    }
                 }
 
+                // remove id of element from ParseContext
+                if (type != typeof(AdaptiveCard))
+                    ParseContext.PopElement();
```
almedina-ms: >ParseContext.PopElement(); [](start = 20, length = 26)

surround this with {}

shalinijoshi19:
Can we start to add a "How Verified" section in our descriptions as well? thanks

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
```diff
@@ -0,0 +1,92 @@
+using System;
```
shalinijoshi19: Is there a copyright header that we need to include on OSS code?

source/dotnet/Test/AdaptiveCards.Test/SerializationTests.cs
```diff
@@ -78,9 +78,19 @@ public void TestSkippingUnknownElements()
...
-            Assert.AreEqual(1, result.Card.Body.Count);
-            Assert.AreEqual(1, result.Card.Actions.Count);
+            Assert.AreEqual(2, result.Card.Body.Count);
+            Assert.AreEqual(2, result.Card.Actions.Count);
             Assert.AreEqual(2, result.Warnings.Count);
+
+            // check first unknown element
+            var unknown_elem = (AdaptiveUnknownElement)result.Card.Body[0];
+            Assert.AreEqual(unknown_elem.Type, "IDunno");
+            Assert.AreEqual(unknown_elem.AdditionalProperties["text"], "Hello");
```
shalinijoshi19: Nit: I used to find adding messages in the assertion checks very useful when looking at logs to get more context debugging -- not something that may be useful? Just a consideration

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveRenderContext.cs
```diff
@@ -195,7 +195,21 @@ public FrameworkElement Render(AdaptiveTypedElement element)
                 return rendered;
             }
 
+            // Since no renderer exists for this element, add warning and render fallback (if available)
             Warnings.Add(new AdaptiveWarning(-1, $"No renderer for element '{element.Type}'"));
```
shalinijoshi19: what happens to these Warnings? Do they end up on cosmos or something ?

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveRenderContext.cs
```diff
@@ -195,7 +195,21 @@ public FrameworkElement Render(AdaptiveTypedElement element)
...
+                    throw new NotImplementedException("Don't know how to handle when fallback = drop");
+                }
+                else if (element.Fallback.Type == AdaptiveFallbackElement.AdaptiveFallbackType.Content && element.Fallback.Content != null)
+                {
+                    // Render fallback content
+                    return Render(element.Fallback.Content);
+                }
+            }
+
             return null;
```
shalinijoshi19: >return null; [](start = 12, length = 12)

Shuold we log and error before we return an unexpected (null) value?

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveRenderContext.cs
paulcam206: warnings are part of the output from rendering an adaptive card (see `ParseResult`). it's up to the renderer to handle them as they see fit. for our visualizers, we show the warnings alongside the rendered card to aid card authors. I'm not sure, but I'd guess that some consumers (e.g. Teams) feed the warnings into their own telemetry channels.

---
In reply to: [272391324](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r272391324) [](ancestors = 272391324)

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
paulcam206: We don't currently use a copyright header, though I think we should, according to those [guidelines](https://docs.opensource.microsoft.com/releasing/copyright-headers.html) I found a while back. I've filed #2626 to cover this work.

---
In reply to: [272373665](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r272373665) [](ancestors = 272373665)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveRenderContext.cs
shalinijoshi19: i see thanks

---
In reply to: [272392569](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r272392569) [](ancestors = 272392569,272391324)

shalinijoshi19:
@paulcam206 whats the status on this one?

source/dotnet/Test/AdaptiveCards.Test/SerializationTests.cs
paulcam206: Worth taking a look at in the future, I think. Our tests need a bit of love overall :)

---
In reply to: [272374244](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r272374244) [](ancestors = 272374244)

paulcam206:
Only thing left outstanding is final wiring so that the renderers fallback on `requires`. Hope to have that done this evening.

shalinijoshi19:
> Can we start to add a "How Verified" section in our descriptions as well? thanks

Hey Paul, any snips of cards etc that may show how things look visually in the "How Verified" section would be useful for subsequent validation (Eg bug bashes etc)

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Xml.Serialization;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace AdaptiveCards
+{
+    public class AdaptiveInternalID
```
shalinijoshi19: we should maybe start having xml style comments or descriptions on our classes/functions etc at some point for readability/maintainability

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            return new AdaptiveInternalID(CurrentInternalID);
+        }
+
+        public static AdaptiveInternalID Next()
+        {
+            CurrentInternalID++;
+
+            // handle overflow case
+            if (CurrentInternalID == AdaptiveInternalID.Invalid)
```
shalinijoshi19: Its not very intuitive to have a check against an "invalid" value and then make it "valid" by incrementing what is supposed to be invalid :( ; Seems scary. It just so happens though that the "invalid" number is just simply 0 so this will be less scary in practice and more normal. 
i would consider renaming Invalid -  to just 'defaultId' or even hard code the check against 0 directly.

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            ID = id;
+        }
+
+        public static AdaptiveInternalID Current()
+        {
+            return new AdaptiveInternalID(CurrentInternalID);
+        }
+
+        public static AdaptiveInternalID Next()
```
shalinijoshi19: is this class threadsafe? Does this need to be ?

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                CurrentInternalID++;
+            }
+
+            return Current();
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj == null || GetType() != obj.GetType())
```
shalinijoshi19: >bj == null [](start = 17, length = 10)

nit: !obj

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        public override string ToString()
+        {
+            return ID.ToString();
+        }
+    }
+
+    public class AdaptiveFallbackElement
+    {
+        public static readonly string drop = "drop";
```
shalinijoshi19: >drop [](start = 38, length = 4)

Nit: this could be a const no?

shalinijoshi19:
>                 if (new AdaptiveSchemaVersion(jObject.Value<string>("version")) > AdaptiveCard.KnownSchemaVersion)

Lots of hard coded literals in the code -- we should consider consolidating the strings in a Constants class or something instead of spewing them all over the code base

---
Refers to: source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs:48 in a782455. [](commit_id = a78245532f6d28b6c6692032740d250f53f7bc5c, deletion_comment = False)

shalinijoshi19:
>             var card = (AdaptiveCard)typedElementConverter.ReadJson(jObject.CreateReader(), objectType, existingValue, serializer);

NIt: Consider resorting to the 'as' operator for casting. [https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/as](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/as)

---
Refers to: source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs:55 in a782455. [](commit_id = a78245532f6d28b6c6692032740d250f53f7bc5c, deletion_comment = False)

source/dotnet/Library/AdaptiveCards/AdaptiveFallbackConverter.cs
```diff
@@ -0,0 +1,129 @@
...
+namespace AdaptiveCards
+{
+    public class AdaptiveFallbackConverter : JsonConverter, ILogWarnings
+    {
+        public List<AdaptiveWarning> Warnings { get; set; } = new List<AdaptiveWarning>();
+
+        public override bool CanWrite => true;
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            AdaptiveFallbackElement fallback = (AdaptiveFallbackElement)value;
```
shalinijoshi19: This could potentially throw -- let;s just start using the non-throwing 'as' operator where we can for casting.

source/dotnet/Library/AdaptiveCards/AdaptiveFeatureRegistration.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace AdaptiveCards
+{
+    public class AdaptiveFeatureRegistration
+    {
+        public AdaptiveFeatureRegistration()
+        {
+            features[adaptiveCardFeature] = "1.2";
```
shalinijoshi19: >= "1.2"; [](start = 42, length = 8)

This again feels like this should be (a) not a hard coded literal (b) configurable

source/dotnet/Library/AdaptiveCards/AdaptiveFeatureRegistration.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                throw new AdaptiveException("Not allowed to remove adaptiveCards feature");
+            }
+            else if (features.ContainsKey(feature))
+            {
+                features.Remove(feature);
+            }
+        }
+
+        private readonly string adaptiveCardFeature = "adaptiveCards";
```
shalinijoshi19: again - why not a const

source/dotnet/Library/AdaptiveCards/AdaptiveFeatureRegistration.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public void Set(string feature, string version)
+        {
+            if (feature == adaptiveCardFeature)
+            {
+                throw new AdaptiveException("Not allowed to override adaptiveCards version");
+            }
+            else
+            {
+                features[feature] = version;
```
shalinijoshi19: >    features[feature] = version; [](start = 12, length = 32)

WHat happens if you try to add the same feature twice?

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
paulcam206: a few things of note here -- Carlos adapted this directly from C++. The overflow case is where a single instance of AdaptiveCards has processed >`UINT_MAX` elements (which would take... quite some time). This code isn't correct as written as I assume that doing:

```
uint i = UINT_MAX; // or whatever it is in C#
i++;
```

will throw an exception. I'll adapt it to be more correct. :)

---
In reply to: [278316341](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278316341) [](ancestors = 278316341)

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
```diff
@@ -66,84 +55,115 @@ public static void RegisterTypedElement<T>(string typeName = null)
...
-                    throw new AdaptiveSerializationException("Required property 'type' not found on adaptive card element");
-                }
-            }
+            string typeName = GetElementTypeName(objectType, jObject);
 
             if (TypedElementTypes.Value.TryGetValue(typeName, out var type))
             {
-                if (jObject.Value<string>("id") == null)
+                string objectId = jObject.Value<string>("id");
+                if (objectId == null)
```
shalinijoshi19: >if (objectId == null) [](start = 16, length = 21)

do we want this check to be include empty/whitespace values for id? so objectId.isNullOrWhitespace() ?

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
```diff
@@ -66,84 +55,115 @@ public static void RegisterTypedElement<T>(string typeName = null)
...
                 try
                 {
                     serializer.Populate(jObject.CreateReader(), result);
                 }
-                catch (JsonSerializationException)
+                catch (JsonSerializationException) { }
+
+                // remove id of element from ParseContext
+                if (type != typeof(AdaptiveCard))
+                    ParseContext.PopElement();
```
shalinijoshi19: Should have a starting/closing brace even for single statement blocks

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
```diff
@@ -66,84 +55,115 @@ public static void RegisterTypedElement<T>(string typeName = null)
...
+                    {
+                        result = (AdaptiveTypedElement)Activator.CreateInstance(typeof(AdaptiveUnknownAction));
+                        serializer.Populate(jObject.CreateReader(), result);
+                    }
+                    catch (JsonSerializationException) { }
                 }
+                ParseContext.PopElement();
 
-                HandleAdditionalProperties(result);
+                Warnings.Add(new AdaptiveWarning(-1, $"Unknown element '{typeName}'"));
```
shalinijoshi19: >1 [](start = 50, length = 1)

What's the -1 signify?

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
```diff
@@ -66,84 +55,115 @@ public static void RegisterTypedElement<T>(string typeName = null)
...
+                }
+                else // ParseContext.Type == ParseContext.ContextType.Action
                 {
-                    return result;
+                    try
+                    {
+                        result = (AdaptiveTypedElement)Activator.CreateInstance(typeof(AdaptiveUnknownAction));
+                        serializer.Populate(jObject.CreateReader(), result);
+                    }
+                    catch (JsonSerializationException) { }
```
shalinijoshi19: >   catch (JsonSerializationException) { } [](start = 17, length = 41)

Why are we just swallowing exceptions here?

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
```diff
@@ -66,84 +55,115 @@ public static void RegisterTypedElement<T>(string typeName = null)
...
                 return result;
             }
+        }
 
-            Warnings.Add(new AdaptiveWarning(-1, $"Unknown element '{typeName}'"));
-            return null;
+        public static string GetElementTypeName(Type objectType, JObject jObject)
+        {
+            var typeName = jObject["type"]?.Value<string>() ?? jObject["@type"]?.Value<string>();
+            if (typeName == null)
```
shalinijoshi19: >ypeName == null [](start = 17, length = 15)

Nit: !typeName

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
```diff
@@ -66,84 +55,115 @@ public static void RegisterTypedElement<T>(string typeName = null)
...
+                Warnings.Add(new AdaptiveWarning(-1, $"Unknown element '{typeName}'"));
                 return result;
             }
+        }
 
-            Warnings.Add(new AdaptiveWarning(-1, $"Unknown element '{typeName}'"));
-            return null;
+        public static string GetElementTypeName(Type objectType, JObject jObject)
+        {
+            var typeName = jObject["type"]?.Value<string>() ?? jObject["@type"]?.Value<string>();
```
shalinijoshi19: Nit: let's not use var esp when it's not clear what the type is for this variable

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
```diff
@@ -66,84 +55,115 @@ public static void RegisterTypedElement<T>(string typeName = null)
...
+            if (typeName == null)
+            {
+                // Get value of this objectType's "Type" JsonProperty(Required)
+                var typeJsonPropertyRequiredValue = objectType.GetRuntimeProperty("Type")
```
shalinijoshi19: Same: Let's stay away from var in these cases where it's not obvious what the type will evaluate to be -- (it's less readable)

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
paulcam206: no on both counts. we provide no thread safety guarantees currently, and have a non-trivial amount of work to do if we end up needing it. :)

---
In reply to: [278316503](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278316503) [](ancestors = 278316503)

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
paulcam206: can't be `const` because it's also `static`

---
In reply to: [278318026](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278318026) [](ancestors = 278318026)

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
paulcam206: apparently not allowed:

```
1>AdaptiveBaseElement.cs(47,17,47,21): error CS0023: Operator '!' cannot be applied to operand of type 'object'
```

---
In reply to: [278316768](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278316768) [](ancestors = 278316768)

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
paulcam206: Agreed. I've filed #2745 to track this.

---
In reply to: [278314967](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278314967) [](ancestors = 278314967)

source/dotnet/Library/AdaptiveCards/AdaptiveBaseElement.cs
shalinijoshi19: arent const's in C#  static by nature? 

---
In reply to: [278325378](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278325378) [](ancestors = 278325378,278318026)

shalinijoshi19:
Is this still pending? 

---
In reply to: [485039280](https://github.com/Microsoft/AdaptiveCards/pull/2560#issuecomment-485039280) [](ancestors = 485039280)

paulcam206:
This is now done :)

---
In reply to: [486436276](https://github.com/Microsoft/AdaptiveCards/pull/2560#issuecomment-486436276) [](ancestors = 486436276,485039280)

paulcam206:
>                 if (new AdaptiveSchemaVersion(jObject.Value<string>("version")) > AdaptiveCard.KnownSchemaVersion)

filed #2746 to cover this

---
In reply to: [486427259](https://github.com/Microsoft/AdaptiveCards/pull/2560#issuecomment-486427259) [](ancestors = 486427259)

---
Refers to: source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs:48 in a782455. [](commit_id = a78245532f6d28b6c6692032740d250f53f7bc5c, deletion_comment = False)

paulcam206:
>             var card = (AdaptiveCard)typedElementConverter.ReadJson(jObject.CreateReader(), objectType, existingValue, serializer);

Filed #2747

---
In reply to: [486427423](https://github.com/Microsoft/AdaptiveCards/pull/2560#issuecomment-486427423) [](ancestors = 486427423)

---
Refers to: source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs:55 in a782455. [](commit_id = a78245532f6d28b6c6692032740d250f53f7bc5c, deletion_comment = False)

source/dotnet/Library/AdaptiveCards/AdaptiveFeatureRegistration.cs
paulcam206: not hardcoded makes sense. what do you mean about "configurable"?

---
In reply to: [278321123](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278321123) [](ancestors = 278321123)

source/dotnet/Library/AdaptiveCards/ParseContext.cs
```diff
@@ -0,0 +1,131 @@
...
+        private static Stack<Tuple<string, AdaptiveInternalID, bool>> IDStack = new Stack<Tuple<string, AdaptiveInternalID, bool>>();
+
+        // Push the provided state on to our ID stack
+        public static void PushElement(string idJsonProperty, AdaptiveInternalID internalId)
+        {
+            if (internalId.Equals(AdaptiveInternalID.Invalid))
+            {
+                throw new AdaptiveSerializationException($"Attemping to push an element on to the stack with an invalid ID");
+            }
+            IDStack.Push(new Tuple<string, AdaptiveInternalID, bool>(idJsonProperty, internalId, AdaptiveFallbackConverter.IsInFallback));
```
shalinijoshi19: Nit: Naming -- idStack (camelCasing for private fields )

source/dotnet/Library/AdaptiveCards/AdaptiveFeatureRegistration.cs
paulcam206: I'm a little on the fence about this one... As it stands, here's are the classes of things that we warn on:

```
    enum class WarningStatusCode
    {
        UnknownElementType = 0,
        UnknownActionElementType,
        UnknownPropertyOnElement,
        UnknownEnumValue,
        NoRendererForType,
        InteractivityNotSupported,
        MaxActionsExceeded,
        AssetLoadFailed,
        UnsupportedSchemaVersion,
        UnsupportedMediaType,
        InvalidMediaMix,
        InvalidColorFormat,
        InvalidDimensionSpecified,
        InvalidLanguage,
    };
```

I could add another one of course, but I'm not sure this would rise to the same level. It's probably worth noting that this is something that app authors would set on a renderer before rendering a parsed card. It's not unsupported, but it's probably a little unusual for folks to reuse a renderer. Certainly it would feel a bit odd to reuse a renderer while swapping out versions of a supported feature.

---
In reply to: [278322077](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278322077) [](ancestors = 278322077)

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
paulcam206: yep :)

---
In reply to: [278323682](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278323682) [](ancestors = 278323682)

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
paulcam206: hmm... I don't think we currently ban the use of whitespace for ids. we probably *should* ban it. I filed #2748 to investigate.

---
In reply to: [278323340](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278323340) [](ancestors = 278323340)

source/dotnet/Library/AdaptiveCards/AdaptiveFeatureRegistration.cs
shalinijoshi19: Would this need to change beyond 1.2 for eg? WHat happens for 1.3? or is 1.2 used as a min version in this case?

---
In reply to: [278332562](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278332562) [](ancestors = 278332562,278321123)

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveHostConfig.cs
```diff
@@ -148,7 +148,6 @@ public static AdaptiveHostConfig FromJson(string json)
             }
             catch (JsonException ex)
             {
-                Debugger.Break();
```
shalinijoshi19: Ha! nice catch!

shalinijoshi19:
>             if (string.IsNullOrEmpty(fontFamilyValue))

Again this one seems like it should be isNullOrWhiteSpace()

---
Refers to: source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveHostConfig.cs:170 in a782455. [](commit_id = a78245532f6d28b6c6692032740d250f53f7bc5c, deletion_comment = False)

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveElementRenderers.cs
```diff
@@ -30,6 +30,10 @@ public void Remove<TElement>()
 
         public Func<AdaptiveTypedElement, TContext, TUIElement> Get(Type type)
         {
+            // AdaptiveUnknownElements and AdaptiveUnknownActions don't have a renderer
+            if (type == typeof(AdaptiveUnknownElement) || type == typeof(AdaptiveUnknownAction))
+                return null;
```
shalinijoshi19: opening/closing braces here as well

shalinijoshi19:
>         // TODO: temporary warning code for fallback card. Remove when common set of error codes created and integrated.

Do we have an issue# tracking this work? If so mind doing me a favor and adding a TODO: Issue#<> reference here? tx

---
Refers to: source/dotnet/Library/AdaptiveCards/AdaptiveCardConverter.cs:14 in a782455. [](commit_id = a78245532f6d28b6c6692032740d250f53f7bc5c, deletion_comment = False)

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
paulcam206: actually, these aren't needed :)

---
In reply to: [278323956](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278323956) [](ancestors = 278323956)

source/dotnet/Library/AdaptiveCards/ParseContext.cs
```diff
@@ -0,0 +1,131 @@
...
+        // Pop the last id off our stack and perform validation 
+        public static void PopElement()
+        {
+            // about to pop an element off the stack. perform collision list maintenance and detection.
+            var idsToPop = IDStack.Peek();
+            string elementID = idsToPop.Item1;
+            AdaptiveInternalID elementInternalID = idsToPop.Item2;
+            bool isFallback = idsToPop.Item3;
+
+            if (!string.IsNullOrEmpty(elementID))
```
shalinijoshi19: >IsNullOrEmpty [](start = 24, length = 13)

IsNullOrWhitespace

source/dotnet/Library/AdaptiveCards/ParseContext.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            if (internalId.Equals(AdaptiveInternalID.Invalid))
+            {
+                throw new AdaptiveSerializationException($"Attemping to push an element on to the stack with an invalid ID");
+            }
+            IDStack.Push(new Tuple<string, AdaptiveInternalID, bool>(idJsonProperty, internalId, AdaptiveFallbackConverter.IsInFallback));
+        }
+
+        // Pop the last id off our stack and perform validation 
+        public static void PopElement()
```
shalinijoshi19: >PopElement [](start = 27, length = 10)

This makes my head reel.. Assuming we got good code coverage for this function in our unit's?

source/dotnet/Library/AdaptiveCards/ParseContext.cs
```diff
@@ -0,0 +1,131 @@
...
+                        }
+
+                        // at this point, we may or may not have a collision depending on additional entries.
+                        haveCollision = true;
+                    }
+                }
+
+                if (haveCollision)
+                {
+                    throw new AdaptiveSerializationException("Collision detected for id '" + elementID + "'");
```
shalinijoshi19: >n("Collision detected for id '" + elementID + "'"); [](start = 59, length = 51)

Who sees this message? And is this actionable for them?

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
paulcam206: we require an enum int here. unfortunately there's not a centralized one available. I've filed #2479 for this.

---
In reply to: [278323851](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278323851) [](ancestors = 278323851)

source/dotnet/Library/AdaptiveCards/ParseContext.cs
```diff
@@ -0,0 +1,131 @@
...
+                }
+
+                // no need to add an entry for this element if it's fallback (we'll add one when we parse it for non-fallback)
```
shalinijoshi19: Nit: Maybe have teh comment match the code it sits adjacent to :).. comment talks of what shouldnt happen in the fallback case but the code really is handling what happens in the non-fallback case.

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
paulcam206: doesn't compile :(

---
In reply to: [278324249](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278324249) [](ancestors = 278324249)

shalinijoshi19:
>                 return _dictionary[typeof(AdaptiveAction)];

while you are here mind updating these 3 blocks with {} as well? tx!

---
Refers to: source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveElementRenderers.cs:42 in a782455. [](commit_id = a78245532f6d28b6c6692032740d250f53f7bc5c, deletion_comment = False)

source/dotnet/Library/AdaptiveCards/ParseContext.cs
paulcam206: see previous comment on the subject :)

---
In reply to: [278345361](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278345361) [](ancestors = 278345361)

source/dotnet/Library/AdaptiveCards/ParseContext.cs
paulcam206: yes. this is an implementation of what we have in the shared model. I can give you a rundown sometime if you're interested. it's needed to solve a pretty complex problem :)

---
In reply to: [278345702](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278345702) [](ancestors = 278345702)

source/dotnet/Library/AdaptiveCards/AdaptiveFeatureRegistration.cs
shalinijoshi19: That's fine - i also noticed the API is called "Set" which shuld indicate "addorupdate" behavior by its name. (it's not for eg called "Add/Insert") so should be fine -- 

---
In reply to: [278341499](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278341499) [](ancestors = 278341499,278322077)

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveHostConfig.cs
paulcam206: found while debugging :D

---
In reply to: [278344312](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278344312) [](ancestors = 278344312)

source/dotnet/Library/AdaptiveCards/AdaptiveFallbackConverter.cs
```diff
@@ -0,0 +1,129 @@
...
+                        serializer.Populate(jObject.CreateReader(), result);
+                        ParseContext.PopElement();
+                        IsInFallback = false;
+
+                        AdaptiveFallbackElement adaptiveFallbackElement = new AdaptiveFallbackElement();
+                        adaptiveFallbackElement.Type = AdaptiveFallbackElement.AdaptiveFallbackType.Content;
+                        adaptiveFallbackElement.Content = result;
+                        return adaptiveFallbackElement;
+
+                        //return null;
```
shalinijoshi19: >    //return null; [](start = 20, length = 18)

Remove commented out code pls :) tx

source/dotnet/Library/AdaptiveCards/AdaptiveFeatureRegistration.cs
paulcam206: It needs to be manually updated. It tracks the version of the object model (now tracked in `Globals.cs`)

---
In reply to: [278343475](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278343475) [](ancestors = 278343475,278332562,278321123)

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
```diff
@@ -66,84 +55,115 @@ public static void RegisterTypedElement<T>(string typeName = null)
...
+                    .CustomAttributes.Where(a => a.AttributeType == typeof(JsonPropertyAttribute)).FirstOrDefault()?
```
shalinijoshi19: >ault()? [](start = 109, length = 7)

This is interesting; what does the FirstOrDefault()? syntax imply again? I'd assumed FirstOrDefault in the linq query would return a null if the condition wasnt satisfied in the list at all ..

source/dotnet/Library/AdaptiveCards/ParseContext.cs
shalinijoshi19: Yes pls! and ideally point me to the test case that i can debug away to glory as well when we talk about this :) thanks!

---
In reply to: [278346705](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278346705) [](ancestors = 278346705,278345702)

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
shalinijoshi19: HMm i wonder if it's because we are using a older framework version that didnt have support for isNullOrWhiteSpace()

---
In reply to: [278346453](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278346453) [](ancestors = 278346453,278324249)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveCardRenderer.cs
```diff
@@ -26,6 +26,7 @@ protected override AdaptiveSchemaVersion GetSupportedSchemaVersion()
         public AdaptiveCardRenderer(AdaptiveHostConfig hostConfig)
         {
             HostConfig = hostConfig;
+            FeatureRegistration.Set("acTest", "1.0");
```
shalinijoshi19: > FeatureRegistration.Set("acTest", "1.0"); [](start = 11, length = 42)

Wait is this test code or prod?

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveRenderContext.cs
```diff
@@ -165,46 +167,96 @@ public virtual Style GetStyle(string styleName)
...
+
+                        frameworkElementOut = rendered;
+                    }
                 }
+            }
+            catch (AdaptiveFallbackException e)
+            {
+                if (!elementHasFallback)
+                {
+                    throw e;
```
shalinijoshi19: > throw e; [](start = 19, length = 9)

I think we want just "throw" here to retain all the original stack context

source/dotnet/Samples/WPFVisualizer/MainWindow.xaml.cs
```diff
@@ -56,7 +56,8 @@ public MainWindow()
 
             Renderer = new AdaptiveCardRenderer()
             {
-                Resources = Resources
+                Resources = Resources,
+
```
shalinijoshi19: Nit: extra newline

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveRenderContext.cs
```diff
@@ -18,31 +18,81 @@ public AdaptiveRenderContext(AdaptiveHostConfig hostConfig, AdaptiveElementRende
...
+                        htmlTagOut = renderer.Invoke(element, this);
+                    }
+                }
+            }
+            catch (AdaptiveFallbackException e)
             {
-                return renderer.Invoke(element, this);
+                if (!elementHasFallback)
+                {
+                    throw e;
```
shalinijoshi19: This should be just throw to retain the original stack context in the exception that gets bubbled back up (throw e will reset the stack)

source/dotnet/Samples/WPFVisualizer/MainWindow.xaml.cs
```diff
@@ -56,7 +56,8 @@ public MainWindow()
 
             Renderer = new AdaptiveCardRenderer()
             {
-                Resources = Resources
+                Resources = Resources,
```
shalinijoshi19: So what are we even doing here ?

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElementConverter.cs
paulcam206: if the property isn't present, it instead returns the default value for the property 

---
In reply to: [278349196](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278349196) [](ancestors = 278349196)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveCardRenderer.cs
paulcam206: oops -- good catch! this was supposed to be in the sample.

---
In reply to: [278357075](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278357075) [](ancestors = 278357075)

source/dotnet/Library/AdaptiveCards/ParseContext.cs
paulcam206: The dotnet parser, unlike the shared model, uses exceptions to report errors to callers (it does still pass warnings up through a warnings collection). This particular exception is thrown for a card that has elements with duplicate `id` properties, which is considered an invalid card.

---
In reply to: [278345843](https://github.com/Microsoft/AdaptiveCards/pull/2560#discussion_r278345843) [](ancestors = 278345843)

PR #2565, opened 2019-03-26
Fix iOS build break

paulcam206:
Fix iOS break due to commit 41796ba. Not the happiest person with this fix -- it's a little unclear why the `#include "util.h"` isn't working. Don't have access to xcode currently to find a better fix, though.


paulcam206:
Superseded by #2566

PR #2576, opened 2019-03-28
[Android] Re add the IOnlineImageLoader class

almedina-ms:
This PR re adds the IOnlineImageLoader class and includes the deprecated anotations to direct users to use IResourceResolver instead as it offers more extensibility


andrewleader:
Which release did we previously remove this IOnlineImageLoader class? Did we already ship a library with it removed? Trying to understand why we're adding it back (I understand not introducing breaking changes, just wondering if there was a complaint or what the main motive behind it is).

almedina-ms:
Since 1.0.4 we have IOnlineImageLoader (1.1 also has it), I removed it during development of 1.2 as I introduced proper ResourceResolvers. Yesterday I was told I should keep it but just mark it as deprecated

---
In reply to: [477677845](https://github.com/Microsoft/AdaptiveCards/pull/2576#issuecomment-477677845) [](ancestors = 477677845)

andrewleader:
Makes sense. Seems good to me!

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/GenericImageLoaderAsync.java
```diff
@@ -15,6 +15,7 @@
 import java.net.URL;
 import java.net.URLDecoder;
 
+import io.adaptivecards.objectmodel.Paragraph;
```
paulcam206: was this intentional?

PR #2580, opened 2019-03-28
[WPF] Change ContainerStyle to nullable type

almedina-ms:
Changes the ContainerStyle in CollectionElements back to being nullable to avoid the breaking change that may arise from changing it to a not nullable type. Also, I had to create a new converter to solve the issue where it may confuse 'null' value with the first value of an enum


source/dotnet/Library/AdaptiveCards/IgnoreNullEnumConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+                // Try to read regularly
+                return base.ReadJson(reader, objectType, existingValue, serializer);
+            }
+            catch
+            {
+                // Catch invalid values and replace them with default value
+                // Add warning stating behavior
+                Warnings.Add(new AdaptiveWarning((int)WarningStatusCode.UnknownElementType, $"Value \"{reader.Value.ToString()}\" could not be converted to \"{typeof(TEnum).ToString()}\", using null instead."));
+                return null;
+            }
```
paulcam206: does this duplicate what we do in the shared model? if a card has `"style": "someRandomString"`, should it be a warning or a parse error?

source/dotnet/Test/AdaptiveCards.Test/AllPayloadTests.cs
```diff
@@ -127,6 +131,9 @@ public void TestAllTestCards()
                     "FallbackParsing",
                     "DeepFallback",
                     "EmptyFallbackCard",
+
+                    // These cards have AdpativeCards with styles on them
+                    "ColumnColumnSetContainer.Bleed.json",
```
paulcam206: >json [](start = 52, length = 4)

is this supposed to have `.json` on the end?

source/dotnet/Library/AdaptiveCards/IgnoreNullEnumConverter.cs
almedina-ms: I'm following the same pattern as the IgnoreDefaultStringEnumConverter, where a warning is also added, I assumed that's the expected behaviour

---
In reply to: [270215751](https://github.com/Microsoft/AdaptiveCards/pull/2580#discussion_r270215751) [](ancestors = 270215751)

source/dotnet/Test/AdaptiveCards.Test/AllPayloadTests.cs
almedina-ms: it shouldn't matter as the comparison made later checks if the filename contains the given string, for consistency I'll delete the ".json"

---
In reply to: [270216378](https://github.com/Microsoft/AdaptiveCards/pull/2580#discussion_r270216378) [](ancestors = 270216378)

Issue #1933, opened 2018-09-20
Inline text runs


andrewleader:
<table>
<tr>
<th>Release</th>
<th>Renderer status</th>
<th>Tasks status</th>
</tr>
<tr>
<td>1.2
<td>✔️ .NET (#2378)<br/>✔️ Android (#2380)<br/>✔️ iOS (#2379)<br/>✔️ TS (#2422)<br/>✔️ UWP (#2377)</td>
<td>✔️ Shared (#2376)<br/>✔️ Designer (#2527)</td>
</tr>
</table>

# Solves requests

* Inline text colors (#1932)
* Inline actions (#1990)
* Inline highlighting (#1934)
* Inline strikethrough on text (#2694)

# Summary
Introduce something similar to XAML's RichTextBlock, where you provide an array of runs, allowing you to do inline rich text formatting.

# Schema
**New** element, `RichTextBlock`

Property|Type|Required|Description
---|---|---|---
type|`string`|Yes|Must be `"RichTextBlock"`
inlines|`IInlineRun[]`|Yes|The inline runs to display. Required. The type is IInlineRun since we'll have future things like ImageRuns that need to be added here.
id | `string` | No, default: `null` | A unique identifier associated with the element
spacing | `string` | No, default: `default` | Controls the amount of spacing between this element and the preceding element.
separator | `boolean` | No, default: `false` | When `true`, draw a separating line at the top of the element.

**New** inline element, `TextRun` (implements `IInlineRun`).

Property|Type|Required|Description
---|---|---|---
type|`string`|Yes|Must be `"TextRun"`
text|`string`|Yes|The text to display. Does NOT support markdown.
selectAction|`IAction`|Action to invoke when this text run is clicked. Visually changes the text run into a hyperlink. `Action.ShowCard` is NOT supported.
color | `string` | No, default: `default` | Controls the color of the text. Defaults to `default`.
isSubtle | `boolean` | No, default: `false` | If `true`, displays text slightly toned down to appear less prominent.
size | `string` | No, default: `default` | Controls the size of the text.
weight | `string` | No, default: `default` | Controls the weight of the text.
italic | `boolean` | No, default: `false` | If `true`, displays the text using italic font.
strikethrough | `boolean` | No, default: `false` | If `true`, displays the text with strikethrough.
highlight | `boolean` | No, default: `false` | If `true`, displays the text highlighted.
[...]

PR #2592, opened 2019-03-29
[UWP] Add Highlighting support to inline text runs

### LINKED ISSUE #2377: [UWP] Inline text runs

UWP implementation of #1933

RebeccaAnne:
Add text highlighting to shared model and UWP. See #1933 for updated description of highlight color host config behavior.

Resolves #2377


source/shared/cpp/ObjectModel/HostConfig.h
```diff
@@ -70,6 +70,7 @@ namespace AdaptiveSharedNamespace
         std::string subtleColor;
 
         static ColorConfig Deserialize(const Json::Value& json, const ColorConfig& defaultValue);
+        void SetBaseColor(std::string baseColor);
```
paulcam206: >std::string [](start = 26, length = 11)

this should probably be `const std::string&` (and below)

Issue #2380, opened 2019-02-09
[Android] Inline text runs


khouzam:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

Android implementation of #1933

PR #2609, opened 2019-04-02
[Android] Add highlight support to inline text runs

almedina-ms:
Adds support for highlight in inline text runs actually completing this issue: https://github.com/Microsoft/AdaptiveCards/issues/2380


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextRendererUtil.java
```diff
@@ -65,6 +65,12 @@ static String getTextColor(ForegroundColor foregroundColor, HostConfig hostConfi
         return hostConfig.GetForegroundColor(containerStyle, foregroundColor, isSubtle);
     }
 
+    static String getHighlightColor(ForegroundColor foregroundColor, HostConfig hostConfig, boolean isSubtle, ContainerStyle containerStyle)
+    {
+
```
paulcam206: nit: don't need the extra newline here

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/RichTextBlockRenderer.java
```diff
@@ -107,6 +108,13 @@ else if ((textRun = TextRun.dynamic_cast(inline)) == null)
 
                 int color = getColor(TextRendererUtil.getTextColor(textRun.GetTextColor(), hostConfig, textRun.GetIsSubtle(), renderArgs.getContainerStyle()));
                 paragraph.setSpan(new ForegroundColorSpan(color), spanStart, spanEnd, Spannable.SPAN_INCLUSIVE_EXCLUSIVE);
+
+                if(textRun.GetHighlight())
```
paulcam206: >f( [](start = 17, length = 2)

nit: spacing

source/shared/cpp/ObjectModel/ParseUtil.cpp
```diff
@@ -54,6 +54,20 @@ namespace AdaptiveSharedNamespace
...
+        std::string propertyName = AdaptiveCardSchemaKeyToString(key);
+        auto propertyValue = json.get(propertyName, Json::Value());
+        if (propertyValue.empty() || !propertyValue.isString())
+        {
+            return "";
+        }
+        else
+        {
+            return propertyValue.asString();
+        }
```
paulcam206: isn't this just `ParseUtil::GetString(json, key, "", false);`?

RebeccaAnne: ParseUtil::GetString(json, key, "", false); will throw if the value at key exists but is not a string. In the case this is handling, the value at key may very well be an object.

---
In reply to: [271454369](https://github.com/Microsoft/AdaptiveCards/pull/2609#discussion_r271454369) [](ancestors = 271454369)

RebeccaAnne: (the shared model changes here are at least mostly from my highlight change that just went in, and won't show up once this is updated to the latest master)

---
In reply to: [271827206](https://github.com/Microsoft/AdaptiveCards/pull/2609#discussion_r271827206) [](ancestors = 271827206,271454369)

PR #2612, opened 2019-04-03
[Android] Add support for actions extensibility

almedina-ms:
Allows the developers to add custom renderers for android so new renderers can be added and current action renderers to be overriden.

BaseActionElementRenderer class was introduced and developers should extend it for their own custom renderers. This class includes two new listeners for developers to use
* **SelectActionOnClickListener** (Inherits from View.OnClickListener): this class allows developers to add selectAction support for their views (either a card element or action). This class doesn't support showCard actions as it's not supported by selectAction
* **ActionOnClickListener**: this class allows developers to add all actions support to their actions but the 
_ActionOnClickListener(RenderedAdaptiveCard, Context, FragmentManager, ViewGroup, BaseActionElement, ICardActionHandler, HostConfig)_ must be used if overriding the showCard action

**Examples are included in the mobile app.**

To add a new action renderer the method _void registerActionRenderer(String, IBaseActionElementRenderer)_ must be used

### How verified

In the mobile app I added three classes
* CustomRedAction (CustomRedActionParser and CustomRedActionRenderer)
* CustomGreenAction (CustomGreenActionParser and CustomGreenActionRenderer)
* ShowCardOverrideRenderer

This three classes exemplify having multiple custom action renderers and overriding the rendering for an existing action type. For testing that these implementations work, in the mobile/samples/v1.2 folder the file CustomActionRenderingTest.json was added which includes a payload that use the new classes.


shalinijoshi19:
ooohh.that;s a lot of commits :) - am assuming you will squash before commiting yes? Also generally do you not rebase -i to cleanup the commit history in your local fork before putting up for PR?

shalinijoshi19:
Can we include a "How Verified" section in teh commit descriptions as well? thanks!

samples/v1.2/Elements/ActionFallback.json
```diff
@@ -0,0 +1,150 @@
...
+	"body": [
+		{
+			"type": "ActionSet",
+			"actions":[
+				{
+					"type": "NoExist",
+					"fallback": {
+						"type": "NeitherExist",
+						"fallback": 
+						{
```
shalinijoshi19: Nit: Can you edit the IDE settings to convert the tabs to spaces pls?

paulcam206: actually, current policy for json is to use tabs :)

shalinijoshi19: Oh? I see - my bad then - any specific reason why? What are we using elsewhere?

---
In reply to: [271974436](https://github.com/Microsoft/AdaptiveCards/pull/2612#discussion_r271974436) [](ancestors = 271974436)

paulcam206:
> ooohh.that;s a lot of commits :) - am assuming you will squash before commiting yes? Also generally do you not rebase -i to cleanup the commit history in your local fork before putting up for PR?

We don't allow plain merge commits to our repo, so your options are `Squash and merge` or `Rebase and merge`. See [here](https://help.github.com/en/articles/about-pull-request-merges) for an explanation of the differences between these two GitHub options. Realistically, I don't think anyone uses the rebase option...

samples/v1.2/Elements/ActionFallback.json
paulcam206: We're using spaces everywhere else. For JSON, tabs are more accepted in the industry. Presumably because it's a data format where size is at more of a premium...

---
In reply to: [271975101](https://github.com/Microsoft/AdaptiveCards/pull/2612#discussion_r271975101) [](ancestors = 271975101,271974436)

paulcam206: You can see our whitespace policies in the `.editorconfig` file at the root of the repo :)

---
In reply to: [271975957](https://github.com/Microsoft/AdaptiveCards/pull/2612#discussion_r271975957) [](ancestors = 271975957,271975101,271974436)

shalinijoshi19: Interesting - thanks for explaining!

---
In reply to: [271975957](https://github.com/Microsoft/AdaptiveCards/pull/2612#discussion_r271975957) [](ancestors = 271975957,271975101,271974436)

source/android/adaptivecards/src/main/cpp/objectmodel_wrap.cpp
```diff
@@ -26135,6 +26251,42 @@ SWIGEXPORT jlong JNICALL Java_io_adaptivecards_objectmodel_AdaptiveCardObjectMod
...
+  
+  (void)jenv;
+  (void)jcls;
+  (void)jarg1_;
+  (void)jarg2_;
+  
+  smartarg1 = *(std::shared_ptr<  AdaptiveCards::UnknownActionParser > **)&jarg1;
+  arg1 = (AdaptiveCards::UnknownActionParser *)(smartarg1 ? smartarg1->get() : 0); 
+  
+  arg2 = (AdaptiveCards::ParseContext *)((*(std::shared_ptr<  AdaptiveCards::ParseContext > **)&jarg2) ? (*(std::shared_ptr<  AdaptiveCards::ParseContext > **)&jarg2)->get() : 0);
```
shalinijoshi19: >   [](start = 60, length = 2)

nit: is this how clang formats the file for us? (this additional spaces in general and then the fact that it's 2 spaces in the opening bracket and one when you clse seems off?

source/android/adaptivecards/src/main/cpp/objectmodel_wrap.cpp
```diff
@@ -26135,6 +26251,42 @@ SWIGEXPORT jlong JNICALL Java_io_adaptivecards_objectmodel_AdaptiveCardObjectMod
...
+  if (!arg2) {
+    SWIG_JavaThrowException(jenv, SWIG_JavaNullPointerException, "AdaptiveCards::ParseContext & reference is null");
+    return 0;
+  } 
+  if(!jarg3) {
+    SWIG_JavaThrowException(jenv, SWIG_JavaNullPointerException, "null string");
```
shalinijoshi19: >null string [](start = 66, length = 11)

our exception messages could do with some work? perhaps more specific (not sure what arg3 is here ) but perhaps include a bit more context on what "string" alludes to?

paulcam206: This is glue code generated by an automated tool called [SWIG](http://www.swig.org/) :)

---
In reply to: [271977090](https://github.com/Microsoft/AdaptiveCards/pull/2612#discussion_r271977090) [](ancestors = 271977090)

source/android/adaptivecards/src/main/cpp/objectmodel_wrap.cpp
paulcam206: because this is auto generated, we consider it exempt from clang formatting rules

---
In reply to: [271976926](https://github.com/Microsoft/AdaptiveCards/pull/2612#discussion_r271976926) [](ancestors = 271976926)

source/android/adaptivecards/src/main/cpp/objectmodel_wrap.cpp
paulcam206: general rule -- things in `source/android/adaptivecards/src/main/*.[h|cpp]` and everything under `source/android/adaptivecards/main/java/io/adaptivecards/objectmodel` are generated files

---
In reply to: [271978478](https://github.com/Microsoft/AdaptiveCards/pull/2612#discussion_r271978478) [](ancestors = 271978478,271977090)

source/android/adaptivecards/src/main/cpp/objectmodel_wrap.cpp
```diff
@@ -26135,6 +26251,42 @@ SWIGEXPORT jlong JNICALL Java_io_adaptivecards_objectmodel_AdaptiveCardObjectMod
...
+    return 0;
+  }
+  const char *arg3_pstr = (const char *)jenv->GetStringUTFChars(jarg3, 0); 
+  if (!arg3_pstr) return 0;
```
shalinijoshi19: Nit: Should we log something here in case this was null? ALso nit(hoping clang can fix this? Paul?) : can we use braces regardless if it's a single statement block or not? tx

almedina-ms: this file is generated via swig so any change to this file will be discarded the next time we run swig.

Swig is the program we use for allowing java code interact with the cpp shared model

Issue #2438, opened 2019-02-21
[.NET] ActionSet element type to render actions anywhere


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

.NET implementation of #877

PR #2614, opened 2019-04-04
[.NET][WPF][HTML] Add ActionSet element support

almedina-ms:
Adds action set support for the .NET renderers for this work item: https://github.com/Microsoft/AdaptiveCards/issues/2438

### How verified 

**Parsing**

Added two tests
* TestObjectModelActionSetElement
* TestParseActionSetElement

**Rendering**

Test apps were executed and gave this results

* WPF
image

* HTML
image


source/dotnet/Library/AdaptiveCards/AdaptiveActionSet.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <summary>
+        ///     Actions of the action set
+        /// </summary>
+        [JsonConverter(typeof(IgnoreEmptyItemsConverter<AdaptiveAction>))]
+#if !NETSTANDARD1_3
+        [XmlArray("Actions")]
+        [XmlArrayItem(ElementName = "OpenUrl", Type = typeof(AdaptiveOpenUrlAction))]
+        [XmlArrayItem(ElementName = "ShowCard", Type = typeof(AdaptiveShowCardAction))]
+        [XmlArrayItem(ElementName = "Submit", Type = typeof(AdaptiveSubmitAction))]
+        [XmlArrayItem(ElementName = "ToggleVisibility", Type = typeof(AdaptiveToggleVisibilityAction))]
```
RebeccaAnne: I'm not entirely familiar with how this works, but does this explicit list interfere with supporting the custom action scenario in this case?

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
```diff
@@ -814,5 +814,176 @@ public void BackgroundImageBackCompatWithLegacyUriType()
...
+          ""type"": ""Action.ToggleVisibility"",
+          ""targetElements"": [
+            ""test""
+          ],
+          ""title"": ""Toggle""
+        }
+      ]
+    }
+  ]
+}";
```
RebeccaAnne: [nit] Indent this to line up with the code

almedina-ms: The indentation adds spaces that make the comparisson fail

source/dotnet/Library/AdaptiveCards/AdaptiveActionSet.cs
almedina-ms: I took this part of code from the AdaptiveCard class 

just talked with Carlos, apparently there's some support for custom actions added in the falllback support (will have to check if there's actual custom action support) so when fallback PR is merged, I'll update this section

---
In reply to: [272653316](https://github.com/Microsoft/AdaptiveCards/pull/2614#discussion_r272653316) [](ancestors = 272653316)

shalinijoshi19:
@almedina-ms any reason why this is still open ?

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
RebeccaAnne: Makes sense. Could you add a comment to that effect?

---
In reply to: [273150076](https://github.com/Microsoft/AdaptiveCards/pull/2614#discussion_r273150076) [](ancestors = 273150076)

source/dotnet/Library/AdaptiveCards/AdaptiveActionSet.cs
RebeccaAnne: Sounds good. If that ends up being a separate change, could you log an issue so we don't lose track of it?

---
In reply to: [273178556](https://github.com/Microsoft/AdaptiveCards/pull/2614#discussion_r273178556) [](ancestors = 273178556,272653316)

almedina-ms: https://github.com/Microsoft/AdaptiveCards/issues/2676

PR #2625, opened 2019-04-04
[.NET] Fix Showcard behaviour (or lack of)

almedina-ms:
Add support for inline showcard in WPF and HTML renderers by implementing it.

### How verified 

Test apps were run and the correct behaviour was checked 

* WPF

image

image

* HTML

image

image


source/dotnet/Samples/AdaptiveCards.Sample.Html/Program.cs
```diff
@@ -202,7 +202,30 @@ static int Main(string[] args)
                 }}
             }});
         }}
-               
+
+        // Sample JavaScript code to test showCard action
+		const showCardActions = document.getElementsByClassName('ac-action-showCard');
```
paulcam206: >		 [](start = 0, length = 2)

nit: some sneaky tabs in here :)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/Helpers/SelectActionHelper.cs
```diff
@@ -34,19 +34,7 @@ public static FrameworkElement RenderSelectAction(this AdaptiveRenderContext con
...
-                    bool isInline = (actionsConfig.ShowCard.ActionMode == ShowCardActionMode.Inline);
-                    if (isInline && context.CardDepth == 1)
-                    {
-                        FrameworkElement uiShowCardContainer = showCardAction.CreateShowCard(context, actionsConfig);
-
-                        // Add to the list of show cards in context
-                        context.ActionShowCards.Add(new Tuple<FrameworkElement, Button>(uiShowCardContainer, uiButton));
-                    }
-                }
+                // SelectAction doesn't allow showCard actions
```
RebeccaAnne: >// SelectAction doesn't allow showCard actions [](start = 16, length = 46)

What is the behavior here after this change? Will the host still be called back to display the show card as a popup?

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveRenderContext.cs
```diff
@@ -45,7 +45,6 @@ public HtmlTag Render(AdaptiveTypedElement element)
             }
         }
 
-
         public string GetColor(AdaptiveTextColor color, bool isSubtle)
```
shalinijoshi19: Out of curiosity: Question are we not doing xml documentation for our functions/classes etc ?

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveRenderContext.cs
```diff
@@ -248,5 +254,21 @@ public void ToggleVisibility(IEnumerable<AdaptiveTargetElement> targetElements)
                 }
             }
         }
+
+        public void ToggleShowCardVisibility(Button uiAction)
+        {
+            FrameworkElement card = ActionShowCards[uiAction];
+            if (card != null)
```
shalinijoshi19: >      if (card != null) [](start = 6, length = 23)

When can this be null? If we are handling it do we actually need to log/trace out (for debugging in particular) that the input was null? On the other hand, if it's an assertion like condition then just remove the null check? Though i am observing the former pattern where we check for non-null which is great but then i am curious why it would ever be null and what do we do about it in teh code?

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/Helpers/SelectActionHelper.cs
shalinijoshi19: >llow showCard actions [](start = 41, length = 21)

Does this fact (that selectaction doesnt support showCard action) need to be propagated up (Eg via an exception ) to the caller instead of silently ignoring? Also is this a breaking change?

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveRenderContext.cs
almedina-ms: I think it should never be null, I'm just placing this in case it ever is null (but it should't happen)

---
In reply to: [272767915](https://github.com/Microsoft/AdaptiveCards/pull/2625#discussion_r272767915) [](ancestors = 272767915)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/Helpers/SelectActionHelper.cs
almedina-ms: The previous version allowed select actions to have showCard actions, that isnt allowed in any of the other renderers, this section used to set the default behaviour for inline show card which shouldn't be here

---
In reply to: [272645334](https://github.com/Microsoft/AdaptiveCards/pull/2625#discussion_r272645334) [](ancestors = 272645334)

RebeccaAnne: Show cards on select actions works in the designer, so it should probably work everywhere else. It works in UWP only if host config is set to popup, which i logged a bug for.

---
In reply to: [273187558](https://github.com/Microsoft/AdaptiveCards/pull/2625#discussion_r273187558) [](ancestors = 273187558,272645334)

RebeccaAnne: This card works in the javascript visualizer, the show card shows as a popup.
[https://adaptivecards.io/visualizer/index.html](https://adaptivecards.io/visualizer/index.html)

{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "Container",
            "selectAction": {
                "type": "Action.ShowCard",
                "card": {
                    "type": "AdaptiveCard",
                    "$schema": "[http://adaptivecards.io/schemas/adaptive-card.json](http://adaptivecards.io/schemas/adaptive-card.json)",
                    "body":[
                        {
                            "type": "TextBlock",
                            "text":"Foo"
                        }
                    ]
                }
            },
            "items": [
                {
                    "type": "TextBlock",
                    "text": "New TextBlock"
                }
            ]
        }
    ],
    "$schema": "[http://adaptivecards.io/schemas/adaptive-card.json](http://adaptivecards.io/schemas/adaptive-card.json)",
    "version": "1.0"
}

---
In reply to: [273209963](https://github.com/Microsoft/AdaptiveCards/pull/2625#discussion_r273209963) [](ancestors = 273209963,273187558,272645334)

almedina-ms: I'm following the model in UWP where I issue a warning for trying to make the select action a showcard action

---
In reply to: [272768066](https://github.com/Microsoft/AdaptiveCards/pull/2625#discussion_r272768066) [](ancestors = 272768066)

almedina-ms: I have this from the documentation

[https://adaptivecards.io/explorer/Image.html](https://adaptivecards.io/explorer/Image.html)
[https://adaptivecards.io/explorer/ColumnSet.html](https://adaptivecards.io/explorer/ColumnSet.html)
[https://adaptivecards.io/explorer/Column.html](https://adaptivecards.io/explorer/Column.html)

---
In reply to: [273210500](https://github.com/Microsoft/AdaptiveCards/pull/2625#discussion_r273210500) [](ancestors = 273210500,273209963,273187558,272645334)

source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveRenderContext.cs
almedina-ms: we have xml documentation for some methods and properties, it would be a good idea to start adding it to the majority of the methods

---
In reply to: [272767265](https://github.com/Microsoft/AdaptiveCards/pull/2625#discussion_r272767265) [](ancestors = 272767265)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/Helpers/SelectActionHelper.cs
RebeccaAnne: @matthidinger @andrewleader Should show cards work for select actions? Documentation says no, but current behavior in JS is that they work, and the designer explicitly lets you add them, although they are always popups regardless of host config settings. Behavior on UWP is that they work if host config is set to popup and are dropped (with a warning) otherwise. Not sure about the other platforms. (I logged #2631 for UWP because i assumed js was the correct behavior)

---
In reply to: [273223131](https://github.com/Microsoft/AdaptiveCards/pull/2625#discussion_r273223131) [](ancestors = 273223131,273210500,273209963,273187558,272645334)

almedina-ms: There was some discussion about this and the behaviour was agreed on not allowing show card actions and filing a warning 

---
In reply to: [273232439](https://github.com/Microsoft/AdaptiveCards/pull/2625#discussion_r273232439) [](ancestors = 273232439,273223131,273210500,273209963,273187558,272645334)

PR #2480, opened 2019-02-26
Fix Null value handling for Min, Max, and Value in AdaptiveNumberInput

MattSFT:
Fix with tests to this issue: https://github.com/Microsoft/AdaptiveCards/issues/2478

PR #2633, opened 2019-04-05
Fix Null value handling for Min, Max, and Value in AdaptiveNumberInput

paulcam206:
Fixes #2480 in `release/1.0`

* Fix Null value handling for Min, Max, and Value in AdaptiveNumberInput

* address PR comments

(cherry picked from commit a5650dc16d4c1918f18358bf888974a0df7ba66e)

### DIFF
--- a/source/dotnet/Library/AdaptiveCards/AdaptiveNumberInput.cs
+++ b/source/dotnet/Library/AdaptiveCards/AdaptiveNumberInput.cs
@@ -34,6 +34,7 @@ namespace AdaptiveCards
         /// <summary>
         ///     The initial value for the field
         /// </summary>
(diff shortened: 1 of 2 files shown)


paulcam206:
There are 3 discrepencies in the status checks for this PR:

* `AdaptiveCards-UWP-Daily-PRs` - this is failing because the `release/1.0` branch has an expired test signing certificate. This code change doesn't affect the shared model or UWP in any way, so it's safe to ignore this problem.
* `AC-Android-CI` and `AC-SharedModel-CI` - these checks never ran because they aren't triggered on builds in branches under `release`. Is also happens that both of these builds are out of a different VSO instance. However, `AdaptiveCards-SharedModel-CI` (our current VSO job) did run and succeed. It doesn't really matter, though, as these builds aren't affected by changes to the dotnet codebase.

Issue #2618, opened 2019-04-04
[.NET] minHeight property on containers


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

.NET implementation of #2293

PR #2638, opened 2019-04-08
[.NET][WPF][HTML] Add minHeight property support

almedina-ms:
Adds support for minHeight property completing this work item: https://github.com/Microsoft/AdaptiveCards/issues/2618

### How verified

The renderers were excuted and it was verified that the output looked fine

Added unit tests under AdaptiveCardApiTests for checking the property was parsed correctly and was able to be modified by a developer using the object model

* TestObjectModelMinHeight
* TestParseMinHeight

### WPF

image

The property changed and is now only for containers
image

image

### HTML

image

image

image


source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveImageRenderer.cs
```diff
@@ -25,6 +25,12 @@ public static FrameworkElement Render(AdaptiveImage image, AdaptiveRenderContext
             
             uiImage.SetHorizontalAlignment(image.HorizontalAlignment);
 
+            // This must be executed before trying to retrieve image width or height
```
RebeccaAnne: >This must be executed before trying to retrieve image width or height [](start = 15, length = 69)

Can you expand this comment to include why it has to be executed before?

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/RendererUtil.cs
```diff
@@ -0,0 +1,59 @@
...
+                    return context.RenderSelectAction(collectionElement.SelectAction, uiElement);
+                }
+            }
+            else if (element is AdaptiveImage)
+            {
+                AdaptiveImage image = element as AdaptiveImage;
+                if (image.SelectAction != null)
+                {
+                    return context.RenderSelectAction(image.SelectAction, uiElement);
+                }
```
RebeccaAnne: You could simplify this a bit by pulling the null check and the call to context.RenderSelectAction out of the if. Something like this:
if (element is AdaptiveCollectionElement)
{
    selectAction = (element as AdaptiveCollectionElement).SelectAction
}
else if (element is AdaptiveImage)
{
    selectAction = (element as AdaptiveImage).SelectAction
}

if (selectAction != null)
{
    return context.RenderSelectAction(selectAction, uiElement);
}

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
```diff
@@ -814,5 +814,157 @@ public void BackgroundImageBackCompatWithLegacyUriType()
...
+              ""minHeight"": ""300px""
+            }
+          ]
+        }
+      ],
+      ""minHeight"": ""100px""
+    }
+  ],
+  ""minHeight"": ""500px""
+}";
```
RebeccaAnne: [nit] Fix the indentation

RebeccaAnne: I see, this is because we're comparing this to what comes out of ToJson. Could you add a comment to that effect?

---
In reply to: [274046694](https://github.com/Microsoft/AdaptiveCards/pull/2638#discussion_r274046694) [](ancestors = 274046694)

shalinijoshi19:
@almedina-ms how long till this PR is committed? ANy major reasno this is still active?

PR #2664, opened 2019-04-11
Add QR sharing support to Android visualizer

andrewleader:
This pull request only adds QR sharing to the Android visualizer. There'll be a separate one for the Designer (coming later), but want to get this merged in Android earlier (Designer will need more work).

With these changes, users can go to https://adaptivecards-sharing.azurewebsites.net/designer/ and scan the QR code.

You can try it by installing the app from the Store (that's a compiled version of this branch's code) https://play.google.com/store/apps/details?id=io.adaptivecards.adaptivecardsvisualizer_master


andrewleader:
@almedina-ms can you review this QR sharing feature addition to Android? This new feature will probably help facilitate testing the Android renderer :)

source/android/mobile/src/main/java/io/adaptivecards/adaptivecardssample/RemoteClientConnection.java
```diff
@@ -0,0 +1,501 @@
...
+                "turn:ws-turn3.xirsys.com:80?transport=udp",
+                "turn:ws-turn3.xirsys.com:3478?transport=udp",
+                "turn:ws-turn3.xirsys.com:80?transport=tcp",
+                "turn:ws-turn3.xirsys.com:3478?transport=tcp",
+                "turns:ws-turn3.xirsys.com:443?transport=tcp",
+                "turns:ws-turn3.xirsys.com:5349?transport=tcp"
+            };
+        for (String turnServer : turnServers)
+        {
+            iceServers.add(PeerConnection.IceServer.builder(turnServer).setUsername("e-HcX5lqoTEzSQetY5biyT3YzM45GIl3HK4FQuo3y73xvsLAmts_gF8PylqkwBR6AAAAAFyoMKphbGVhZGVy").setPassword("d97cbeee-5827-11e9-bef6-069f0817bf63").createIceServer());
```
almedina-ms: >nServer).setUsername("e-HcX5lqoTEzSQetY5biyT3YzM45GIl3HK4FQuo3y73xvsLAmts_gF8PylqkwBR6AAAAAFyoMKphbGVhZGVy").setPassword("d97cbeee-5827-11e9-bef6-069f0817bf63") [](start = 63, length = 160)

is it ok to have this credentials like this?

andrewleader: It's for a free TURN server, not too concerned... we could always switch it to obtain the passwords somehow else, but inevitably these are always a bit vulnerable to being taken and used. Worst case is that someone uses it and consumes all the bandwidth on the TURN server, which would simply result in our QR sharing service being slow/unusable, but wouldn't charge us any money since it's free. Most connections don't use TURN anyways though. We can re-visit if we run into problems!

PR #2679, opened 2019-04-12
[Shared/UWP/Android/iOS/.NET] RichTextBlock updates

### LINKED ISSUE #2670: [RichTextBlock][UWP] Update existing richtextblock implementation to match new specification for 1.2
1.	To address the length concerns, we’re simplifying the payload
a.	Removing the concept of paragraphs, so it’s just a set of inlines
b.	Syntax sugar for auto-converting a text string into a basic TextRun
2.	For the wrap/maxLines
a.	Remove those properties
3.	For the markdown inside runs
a.	Don’t support markdown
4.	For being able to use italics within runs (since no markdown)
a.	Possibly add an “italicize” property to TextRun
5.	For fallback support/extensibility (which markdown doesn’t support)
a.	We’ll add that in 1.3 for text runs (too costly to add now)

Changes…
1.	Removed “paragraphs”. Now it directly goes to “inlines”
2.	Syntax sugar for converting a text string into a basic TextRun object (to reduce payload size when you don’t need any special formatting)
3.	Removed “wrap” and “maxLines” properties
4.	Removed markdown support from TextRun
5.	Added “italicize” (name pending) to TextRun so you can italicize text

RebeccaAnne:
This implements changes to rich text block for the shared model and UWP. This will not be merged until the corresponding iOS and Android changes are ready. This change includes the following updates:

- Remove paragraphs from RichTextBlock  (go straight to inlines)
- Remove maxLines and wrap properties from RichTextBlock 
- Remove markdown support from RichTextBlock 
- Support for specifying an inline directly as a string

This change does not currently include italic and strikethrough support.
Fixes #2670


source/uwp/Renderer/lib/TextHelpers.cpp
```diff
@@ -19,21 +19,36 @@ using namespace ABI::Windows::UI::Xaml::Controls;
...
+        RETURN_IF_FAILED(xamlTextBlock2->put_MaxLines(maxLines));
+    }
+
+    RETURN_IF_FAILED(SetHorizontalAlignment(adaptiveTextBlock, xamlTextBlock));
+
+    ComPtr<IAdaptiveTextBlock> localAdaptiveTextBlock(adaptiveTextBlock);
+    ComPtr<IAdaptiveTextElement> adaptiveTextElement;
+    RETURN_IF_FAILED(localAdaptiveTextBlock.As(&adaptiveTextElement));
+
+    if (adaptiveTextElement)
```
paulcam206: don't COM rules generally state that `S_OK` means out pointer is non-null?

source/uwp/Renderer/lib/TextHelpers.cpp
```diff
@@ -92,6 +107,14 @@ HRESULT SetXamlInlinesWithTextConfig(_In_ IAdaptiveRenderContext* renderContext,
     return S_OK;
 }
 
+HRESULT SetWrapProperties(ABI::Windows::UI::Xaml::Controls::ITextBlock* xamlTextBlock, bool wrap)
```
paulcam206: `_In_`

source/uwp/Renderer/lib/TextHelpers.cpp
```diff
@@ -328,7 +346,17 @@ HRESULT AddSingleTextInline(_In_ IAdaptiveTextElement* adaptiveTextElement,
 {
     ComPtr<ABI::Windows::UI::Xaml::Documents::IRun> run = XamlHelpers::CreateXamlClass<ABI::Windows::UI::Xaml::Documents::IRun>(
         HStringReference(RuntimeClass_Windows_UI_Xaml_Documents_Run));
-    RETURN_IF_FAILED(run->put_Text(string));
+
+    HString language;
+    RETURN_IF_FAILED(adaptiveTextElement->get_Language(language.GetAddressOf()));
+
+    DateTimeParser parser(HStringToUTF8(language.Get()));
+    auto textWithParsedDates = parser.GenerateString(HStringToUTF8(string));
```
paulcam206: nit: could be `const auto`

source/uwp/Renderer/lib/TextHelpers.h
```diff
@@ -43,42 +43,15 @@ HRESULT SetXamlInlinesWithTextConfig(_In_ ABI::AdaptiveNamespace::IAdaptiveRende
...
-    UINT32 maxLines;
-    RETURN_IF_FAILED(adaptiveTextBlock->get_MaxLines(&maxLines));
-    if (maxLines != MAXUINT32)
-    {
-        RETURN_IF_FAILED(SetMaxLines(xamlTextBlock, maxLines));
-    }
+                                     _In_ ABI::Windows::UI::Xaml::Controls::ITextBlock* xamlTextBlock);
 
+template<typename TAdaptiveType, typename TXamlTextBlockType>
+HRESULT SetHorizontalAlignment(TAdaptiveType* adaptiveTextBlock, TXamlTextBlockType* xamlTextBlock)
```
paulcam206: missing SAL

source/uwp/Renderer/lib/TextHelpers.cpp
RebeccaAnne: Yup, that's a remnant of a previous implementation, removed check.

---
In reply to: [276058184](https://github.com/Microsoft/AdaptiveCards/pull/2679#discussion_r276058184) [](ancestors = 276058184)

PR #2683, opened 2019-04-13
Removing auto-generated example test cases

thsaravana:
Found auto-generated default test cases while setting up the Android project. I don't see any need for this and hence removing them as part of housekeeping


source/android/adaptivecards/src/test/java/io/adaptivecards/ExampleUnitTest.java
```diff
@@ -1,17 +0,0 @@
-package io.adaptivecards;
-
-import org.junit.Test;
-
-import static org.junit.Assert.*;
-
-/**
```
shalinijoshi19: FYI that we will be adding unit tests soon; @almedina-ms FYI

thsaravana: Awesome! Would love to contribute.

Issue #2534, opened 2019-03-12
Review exception handling in XamlBuilder.cpp


paulcam206:
We currently have a mix of error handling strategies in XamlBuilder.cpp. We need to make sure we have consistent handling. This probably means wrapping exception-throwing methods.

PR #2684, opened 2019-04-14
[Shared][UWP] Requires support

paulcam206:
Implement `requires` support for UWP and shared model:

Shared Model:
* Introduce `FeatureRegistration` object that allows callers to add/remove features and versions, or retrieve/replace features at the collection level
* Reworks existing `BaseElement::MeetsRequirements()` method to use new collection
* Update tests

UWP:
* Introduce `IAdaptiveFeatureRegistration` and `AdaptiveFeatureRegistration` to project shared model object into UWP
* Consume feature registration at render time to trigger fallback when element requirements aren't met
* For visualizer and test app, add an `acTest` feature at version `1.0.0.0` to enable tests
* Add `Element.Requires.json` test, which leverages the `acTest` feature
* Fix found bug in fallback where falling back would skip adding a separator
* Various bits of plumbing to get everything building

Android:
* SWIG update for new interfaces


source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ObjectModelTest.cpp
```diff
@@ -1009,19 +1010,22 @@ namespace AdaptiveCardsSharedModelUnitTest
...
+            // hostProvides.clear();
+            // Assert::IsFalse(textBlock->MeetsRequirements(hostProvides));
+            // Assert::IsTrue(textBlockNoRequires->MeetsRequirements(hostProvides));
+            // hostProvides.insert({ "foobar", "1.9.9.9" });
+            // Assert::IsFalse(textBlock->MeetsRequirements(hostProvides));
+            // Assert::IsTrue(textBlockNoRequires->MeetsRequirements(hostProvides));
+            // hostProvides.clear();
+            // hostProvides.insert({ "foobar", "99" });
+            // Assert::IsTrue(textBlock->MeetsRequirements(hostProvides));
+            // Assert::IsTrue(textBlockNoRequires->MeetsRequirements(hostProvides));
```
RebeccaAnne: We should uncomment this or delete it

source/shared/cpp/ObjectModel/BaseElement.cpp
```diff
@@ -53,44 +54,44 @@ namespace AdaptiveSharedNamespace
...
-                }
             }
         }
+
         return true;
     }
 
+    std::shared_ptr<std::unordered_map<std::string, AdaptiveSharedNamespace::SemanticVersion>> BaseElement::GetRequirements() const
+    {
+        return std::make_shared<std::unordered_map<std::string, AdaptiveSharedNamespace::SemanticVersion>>(m_requires);
```
RebeccaAnne: >make_shared [](start = 20, length = 11)

Should the member itself just be a shared pointer rather than creating it here?

source/shared/cpp/ObjectModel/BaseElement.cpp
```diff
@@ -53,44 +54,44 @@ namespace AdaptiveSharedNamespace
...
+    }
+
+    void BaseElement::SetRequirements(const std::unordered_map<std::string, AdaptiveSharedNamespace::SemanticVersion>& requirements)
```
RebeccaAnne: >SetRequirements [](start = 22, length = 15)

Our other collection types just have a Get, and then the callers add things to it (like GetBody() or GetActions()). I realize those are vectors and this is a map, but i'm wondering if we should use the same approach here

samples/Tests/Element.Requires.json
```diff
@@ -0,0 +1,81 @@
...
+							"text": "✗: This textblock requires adaptiveCards 5.0",
+							"color": "attention",
+							"wrap": true,
+							"requires": {
+								"adaptiveCards": "5"
+							},
+							"fallback": {
+								"type": "TextBlock",
+								"color": "good",
+								"text": "✓:We don't support adaptiveCards 5.0",
```
RebeccaAnne: >adaptiveCards [](start = 36, length = 13)

Maybe add a case to this file for a custom require that is not met?

source/uwp/Renderer/lib/AdaptiveCardElement.cpp
```diff
@@ -130,6 +134,40 @@ namespace AdaptiveNamespace
...
+                const SemanticVersion providesVersion{HStringToUTF8(featureVersion.Get())};
+                if (providesVersion < requirementVersion)
+                {
+                    // host's provided version is too low
+                    *value = false;
+                    return S_OK;
+                }
+            }
+        }
+        return S_OK;
```
RebeccaAnne: Hmm... we have this logic in two places since it's already in the shared model. What do you think about calling GetSharedModel() and then using its version of the logic?

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -886,47 +888,73 @@ namespace AdaptiveNamespace
...
-            HString elementType;
-            RETURN_IF_FAILED(element->get_ElementTypeString(elementType.GetAddressOf()));
-            ComPtr<IAdaptiveElementRendererRegistration> elementRenderers;
-            RETURN_IF_FAILED(renderContext->get_ElementRenderers(&elementRenderers));
-            ComPtr<IAdaptiveElementRenderer> elementRenderer;
-            RETURN_IF_FAILED(elementRenderers->Get(elementType.Get(), &elementRenderer));
+
+            // Get fallback state
             ABI::AdaptiveNamespace::FallbackType elementFallback;
             element->get_FallbackType(&elementFallback);
```
RebeccaAnne: >get_FallbackType [](start = 21, length = 16)

RETURN_IF_FAILED (on put_AncestorHasFallback too)

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -886,47 +888,73 @@ namespace AdaptiveNamespace
...
-                    UINT separatorThickness;
-                    ABI::Windows::UI::Color separatorColor;
-                    GetSeparationConfigForElement(element, hostConfig.Get(), &spacing, &separatorThickness, &separatorColor, &needsSeparator);
-                    if (needsSeparator)
-                    {
-                        auto separator = CreateSeparator(renderContext, spacing, separatorThickness, separatorColor);
-                        XamlHelpers::AppendXamlElementToPanel(separator.Get(), parentPanel);
-                    }
-                }
+                AddSeparatorIfNeeded(iElement, element, hostConfig, renderContext, parentPanel);
```
RebeccaAnne: >AddSeparatorIfNeeded [](start = 16, length = 20)

This throws.... we already have a bug for fixing up HRESULTs vs throwing, right?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ObjectModelTest.cpp
paulcam206: derp. weekend development is apparently less rigorous :)

---
In reply to: [275445293](https://github.com/Microsoft/AdaptiveCards/pull/2684#discussion_r275445293) [](ancestors = 275445293)

source/shared/cpp/ObjectModel/BaseElement.cpp
paulcam206: suppose that makes sense :)

---
In reply to: [275446588](https://github.com/Microsoft/AdaptiveCards/pull/2684#discussion_r275446588) [](ancestors = 275446588)

source/uwp/Renderer/lib/XamlBuilder.cpp
paulcam206: sure do -- #2534 :)

---
In reply to: [275451172](https://github.com/Microsoft/AdaptiveCards/pull/2684#discussion_r275451172) [](ancestors = 275451172)

samples/Tests/Element.Requires.json
paulcam206: still need to update expected results

---
In reply to: [275448217](https://github.com/Microsoft/AdaptiveCards/pull/2684#discussion_r275448217) [](ancestors = 275448217)

source/uwp/Renderer/lib/AdaptiveCardElement.cpp
paulcam206: it's perhaps a bit unwieldy, if only because of how UWP is structured. that said, it's probably not any more unwieldy than duplicating logic :)



---
In reply to: [275449169](https://github.com/Microsoft/AdaptiveCards/pull/2684#discussion_r275449169) [](ancestors = 275449169)

paulcam206:
SharedModel failure is due to the agent.

/AzurePipelines run

almedina-ms:
> 

unordered_map doesn't exist in SWIG 3 so getting the Requirements in java is not possible, maybe we could add that class later

---
Refers to: source/android/adaptivecards/src/AdaptiveCardObjectModel.i:2 in 0e244fc. [](commit_id = 0e244fc5f4e2d698e759d3c061164cf29cfbefc1, deletion_comment = False)

source/shared/cpp/ObjectModel/BaseElement.h
```diff
@@ -84,23 +86,18 @@ namespace AdaptiveSharedNamespace
...
-        bool MeetsRequirements(const std::unordered_map<std::string, std::string>& hostProvides) const;
+        bool MeetsRequirements(const AdaptiveSharedNamespace::FeatureRegistration& hostProvides) const;
+        std::shared_ptr<std::unordered_map<std::string, AdaptiveSharedNamespace::SemanticVersion>> GetRequirements() const;
 
         // Misc.
         virtual void GetResourceInformation(std::vector<RemoteResourceInformation>& resourceUris);
 
     protected:
         virtual void PopulateKnownPropertiesSet();
         void SetTypeString(const std::string& type) { m_typeString = type; }
```
almedina-ms: >std::shared_ptr<std::unordered_map<std::string, AdaptiveSharedNamespace::SemanticVersion>> m_requires; [](start = 8, length = 102)

why is this a pointer to an unordered_map instead of the unordered_map by itself?

paulcam206: it was originally an `unordered_map`, but was turned into a `shared_ptr<unordered_map>` on @RebeccaAnne 's suggestion

---
In reply to: [276843946](https://github.com/Microsoft/AdaptiveCards/pull/2684#discussion_r276843946) [](ancestors = 276843946)

paulcam206:
> 

looks like it's in SWIG 4, but probably not in the pre-release version I built (see [https://github.com/swig/swig/commit/f58a6dade3b0c13626af271d3d054dc6076d8e0d](https://github.com/swig/swig/commit/f58a6dade3b0c13626af271d3d054dc6076d8e0d)). it looks like they tagged 4.0 beta recently... I'll see if I can find an official prerelease windows build...

---
In reply to: [484695699](https://github.com/Microsoft/AdaptiveCards/pull/2684#issuecomment-484695699) [](ancestors = 484695699)

---
Refers to: source/android/adaptivecards/src/AdaptiveCardObjectModel.i:2 in 0e244fc. [](commit_id = 0e244fc5f4e2d698e759d3c061164cf29cfbefc1, deletion_comment = False)

PR #2688, opened 2019-04-15
[Android] Update rich text block

almedina-ms:
Updates the RichTextBlock to remove the paragraphs and keep the same behaviour as the one in UWP


RebeccaAnne:
Can you set this pull request to merge into my branch instead of master? I think we should probably consolidate everything there and then once we're all in place we can go to master.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/RichTextBlockRenderer.java
```diff
@@ -188,49 +180,19 @@ else if ((richTextBlock = RichTextBlock.dynamic_cast(baseCardElement)) == null)
...
-            if ((i != 0) && (inlines.size() != 0))
-            {
-                textView.append(System.getProperty("line.separator"));
-            }
-
-            SpannableStringBuilder convertedString = buildSpannableParagraph(renderedCard, inlines, cardActionHandler, hostConfig, renderArgs);
-
-            textView.append(convertedString);
-        }
+        SpannableStringBuilder convertedString = buildSpannableParagraph(renderedCard, inlines, cardActionHandler, hostConfig, renderArgs);
```
RebeccaAnne: >buildSpannableParagraph [](start = 49, length = 23)

Does this change include removing markdown support for rich text blocks?

RebeccaAnne:
It also makes the code review easier so we can see which changes are part of which PRs

---
In reply to: [483372991](https://github.com/Microsoft/AdaptiveCards/pull/2688#issuecomment-483372991) [](ancestors = 483372991)

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/RichTextBlockRenderer.java
almedina-ms: now it does :D

---
In reply to: [275498255](https://github.com/Microsoft/AdaptiveCards/pull/2688#discussion_r275498255) [](ancestors = 275498255)

PR #2691, opened 2019-04-15
[Android] italic and strikethrough for RichTextBlock

almedina-ms:
Adds support for italic (already supported) and strikethrough in RichTextBlocks as in the UWP renderer


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/RichTextBlockRenderer.java
```diff
@@ -116,6 +109,12 @@ else if ((textRun = TextRun.dynamic_cast(inline)) == null)
                     paragraph.setSpan(new BackgroundColorSpan(highlightColor), spanStart, spanEnd, Spannable.SPAN_INCLUSIVE_EXCLUSIVE);
                 }
 
+                if (textRun.GetStrikethrough())
```
RebeccaAnne: Where is italic supported?

almedina-ms: the Style span formats the text to italic if required as it may return the weight "Typeface.ITALIC"

---
In reply to: [275573418](https://github.com/Microsoft/AdaptiveCards/pull/2691#discussion_r275573418) [](ancestors = 275573418)

PR #2693, opened 2019-04-15
[Android] Add Requires property support

almedina-ms:
Adds the Requires property and the methods in the CardRendererRegistration class to be able to register a FeatureRegistration objec.

### How verified

An sample was added to the mobile app and the result was as shown below

Screenshot_19700119-213609


paulcam206:
This PR should probably be targeted against its parent branch, `paulcam/featureReg` :)

paulcam206:
Looks great! Once we can reconcile this against my branch, I think we'll be good!

almedina-ms:
I have reconciled this branch but I merged master in between so maybe reconciling the base branch with master may remove the extra files

---
In reply to: [483916178](https://github.com/Microsoft/AdaptiveCards/pull/2693#issuecomment-483916178) [](ancestors = 483916178)

Issue #2109, opened 2018-10-19
bleed property on elements


andrewleader:
<table>
<tr>
<th>Release</th>
<th>Renderer status</th>
<th>Tasks status</th>
</tr>
<tr>
<td>1.2
<td>✔️ .NET (#2668)<br/>✔️ Android (#2667)<br/>✔️ iOS (#2358)<br/>✔️ TS (#2416)<br/>✔️ UWP (#2357)</td>
<td>✔️ Shared (#2356)</td>
</tr>
</table>

# Solves requests

* Bleeding background images/colors for containers #2013

# Summary
Add a `bleed` property to containers/columnSets. This allows an element to "bleed" through its parent's padding. And this makes it super easy to achieve creating headers or header dividers that go to the edge of the card.

# Schema
**New** property just on the elements where it makes sense for them to bleed to the edge of the card. Note that it can also be used within children further inside the card.

The property should be on...

* Container
* ColumnSet

The property should NOT be on...
* Column (columns will all automatically bleed, more on that later in this spec)
* TextBlock
* Image (can't create "hero" image anyways without a stretch behavior property... maybe in future?)
* Media
* FactSet
* Fact (only useful on the FactSet, facts are always displayed the same)
* ImageSet (note that for images within an ImageSet, the bleed property on the images will be ignored)
* Inputs (probably doesn't make sense having an input go to the border of the card?)

Property|Type|Required|Description
---|---|---|---
bleed | boolean, default `false` | false | Determines whether the element should bleed through its parent's padding.

# Example

Imagine a developer has the following emphasis container, and they want that to bleed to the edges of the card.

![image](https://user-images.githubusercontent.com/13246069/47187840-7a401380-d2ea-11e8-97f3-a6dff3fc307b.png)

```
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "Container",
            "style": "emphasis",
            "items": [
                {
                    "type": "TextBlock",
                    "text": "This container has a gray background that extends to the edges of the card",
                    "wrap": true
                }
            ]
        },
        {
            "type": "Container",
            "items": [
                {
                    "type": "TextBlock",
                    "text": "This other container does not have a background.",
                    "wrap": true
                }
            ]
        }
    ]
}
```

They would simply add `"bleed": "true"` to the container to remove the margins around the container...

```
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "Container",
[...]

PR #2700, opened 2019-04-16
[Android] Verify correct behavior of bleed property

almedina-ms:
Updated bleed behavior for android and consolidated some logic into functions so the padding and bleed logic would not be repeated 

### How verified 
The mobile app was run and these are the results

### Comprehensive payload
Screenshot_19700120-212829
Screenshot_19700120-212834
Screenshot_19700120-212841

### Column.Bleed
Screenshot_19700120-212846

### ColumnSet.Bleed
Screenshot_19700120-212852

### Container.Bleed
Screenshot_19700120-212857


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ContainerRenderer.java
```diff
@@ -167,5 +158,45 @@ else if ((container = Container.dynamic_cast(baseCardElement)) == null)
...
+            LinearLayout.LayoutParams layoutParams = (LinearLayout.LayoutParams) collectionElementView.getLayoutParams();
+            int marginLeft = layoutParams.leftMargin, marginRight = layoutParams.rightMargin;
+
+            ContainerBleedDirection bleedDirection = collectionElement.GetBleedDirection();
+            if (bleedDirection == ContainerBleedDirection.BleedToLeading || bleedDirection == ContainerBleedDirection.BleedToBothEdges)
+            {
+                marginLeft = -padding;
+            }
+
+            if(bleedDirection == ContainerBleedDirection.BleedToTrailing || bleedDirection == ContainerBleedDirection.BleedToBothEdges)
```
paulcam206: nit: space after `if`

dclaux:
Per the screenshots, there seems to still be a couple things not handled as expected. Namely, bleeding at the top/bottom of containers/the card itself doesn't seem to be handled. A container that bleeds should bleed not only left and right, but also top and bottom when appropriate. In the first screenshot for instance, the "Bleeding container at the top of the card" should bleed into the top padding of the card.

I am going to assume that this has been compared to the Designer's rendering; the designer voluntarily disables bleeding top and bottom in order to leave space at the top and bottom of any container onto which elements can be dragged (otherwise, it would be impossible to reorder elements.) To see final render of the card, use the "Preview mode" button. Or, use the visualizer instead.

RebeccaAnne:
We don't currently have that behavior on any of our non-javascript renderers, as we were all using the designer as a comparison point, and it's not super explicit in the spec. It does look like the last gif example shows that behavior (I assume "marginFromParent":"none" is an old way to say bleed?), but looks like we missed it. I'll log a bug to track addressing it separate of this PR.

---
In reply to: [484155715](https://github.com/Microsoft/AdaptiveCards/pull/2700#issuecomment-484155715) [](ancestors = 484155715)

RebeccaAnne:
#2702

Also, i understand the concern about needing the visualizer to be usable, but are we concerned that users won't realize that what they see in the visualizer is not what will ultimately be rendered? (Also, where is the "Preview mode" button? I can't seem to find it.)

---
In reply to: [484190751](https://github.com/Microsoft/AdaptiveCards/pull/2700#issuecomment-484190751) [](ancestors = 484190751,484155715)

dclaux:
I believe the test payload you guys used is from here: https://github.com/Microsoft/AdaptiveCards/issues/2109

There is also a screenshot of the rendered result in there; that is the reference.

dclaux:
Yes, there is a risk of potential confusion, and I have ideas for making it better. However these improvements will come later, and in general the designer should NOT be used as a reference for the final visuals (unless using the "Preview mode" feature in the 2.0 early preview) because the designer will always behave slightly differently in some cases to facilitate drag and drop.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ContainerRenderer.java
```diff
@@ -163,5 +154,45 @@ else if ((container = Container.dynamic_cast(baseCardElement)) == null)
...
+            }
+
+            layoutParams.setMargins(marginLeft, layoutParams.topMargin, marginRight, layoutParams.bottomMargin);
+            collectionElementView.setLayoutParams(layoutParams);
+        }
+    }
+
+    public static void ApplyPadding(ContainerStyle elementContainerStyle, ContainerStyle parentContainerStyle, LinearLayout collectionElementView, Context context, HostConfig hostConfig)
+    {
+        if (elementContainerStyle != parentContainerStyle)
```
nesalang: just want to let you know that you can call GetPadding() on collectionType object to see if it gets padding.

PR #2715, opened 2019-04-18
[Android] Change null returns for exceptions to avoid ambiguity

almedina-ms:
Adds AdaptiveFallbackException to signalize an element couldn't be rendered avoiding the use of null as it may be ambiguous for external developers

### How verified

The test cards we have were run and this is the output

### Deep fallback
Screenshot_19700121-222528

### ActionFallback
Screenshot_19700121-222542


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/ActionLayoutRenderer.java
```diff
@@ -126,56 +127,74 @@ else if (alignment == ActionAlignment.Center.swigValue())
...
+                                    // The element has no fallback, just clear the element so the cycle ends
+                                    fallbackElement = null;
+                                }
                             }
                         }
                     }
                 }
+                else if (renderArgs.getAncestorHasFallback())
+                {
+                    // There's an ancestor with fallback so we throw to test it
```
paulcam206: we throw to trigger it rather than test it, right?

PR #2717, opened 2019-04-18
[Android] Log warning if card element has a ShowCard select action

almedina-ms:
Adds warning log for card elements having a select action with type ShowCard


shalinijoshi19:
General: Dont forget the "How Verified" section in the description :)

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
```diff
@@ -74,113 +75,35 @@ protected static int getColor(String colorCode)
...
             {
-                ToggleVisibilityTarget target = toggleVisibilityTargetVector.get(i);
-                String elementId = target.GetElementId();
-
-                View foundView = rootView.findViewWithTag(new TagContent(elementId));
-                if(foundView != null)
-                {
-                    m_viewDictionary.put(elementId, foundView);
-                }
+                renderedCard.addWarning(new AdaptiveWarning(AdaptiveWarning.SELECT_SHOW_CARD_ACTION, "ShowCard not supported for SelectAction"));
```
shalinijoshi19: >addWarning [](start = 29, length = 10)

Cool is this coming from teh sharedmodel now?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
```diff
@@ -266,47 +207,136 @@ private Activity getActivity(Context context)
...
-                if (hostingActivity != null) {
-                    View currentFocusedView = hostingActivity.getCurrentFocus();
-                    if (currentFocusedView != null) {
-                        currentFocusedView.clearFocus();
-                    }
+                ToggleVisibilityTarget target = toggleVisibilityTargetVector.get(i);
+                String elementId = target.GetElementId();
+
+                View foundView = rootView.findViewWithTag(new TagContent(elementId));
+                if (foundView != null)
```
shalinijoshi19: do you expect to not find the target element in the rootView object? What are the cases where this is not null?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
```diff
@@ -266,47 +207,136 @@ private Activity getActivity(Context context)
...
+            {
+                View currentFocusedView = hostingActivity.getCurrentFocus();
+                if (currentFocusedView != null)
+                {
+                    currentFocusedView.clearFocus();
+                }
+            }
+
+            v.setPressed(m_invisibleCard.getVisibility() != View.VISIBLE);
+            for (int i = 0; i < m_hiddenCardsLayout.getChildCount(); ++i)
```
shalinijoshi19: >++i [](start = 69, length = 3)

Nit: Consider using the more concise foreach equivalent syntax (i think for (View child : m_hiddenCardsLayout) ) ;

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
almedina-ms: if the card creator were to toggle the visibility of an element that doesn't exist (giving an id that doesn't exist for any element in the card) then the found view will be null and as such we can't change the visibility for that view

---
In reply to: [277106481](https://github.com/Microsoft/AdaptiveCards/pull/2717#discussion_r277106481) [](ancestors = 277106481)

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
almedina-ms: Android API is weird in this kind of stuff so we have to get the elements one by one as it doesn't provide a "getChildren" method

---
In reply to: [277107266](https://github.com/Microsoft/AdaptiveCards/pull/2717#discussion_r277107266) [](ancestors = 277107266)

shalinijoshi19: Hmm you mean there is a requirement for being able to use the foreach java syntax for this collection? 

---
In reply to: [278224177](https://github.com/Microsoft/AdaptiveCards/pull/2717#discussion_r278224177) [](ancestors = 278224177,277107266)

almedina-ms: As far as I know, to use the foreach syntax you must have the collection or be able to get the collection with a getter method that returns the complete collection, in this case, Android only allows us to grab one element at a time by querying each index

[https://stackoverflow.com/questions/18668897/android-get-all-children-elements-of-a-viewgroup](https://stackoverflow.com/questions/18668897/android-get-all-children-elements-of-a-viewgroup) 

---
In reply to: [278771455](https://github.com/Microsoft/AdaptiveCards/pull/2717#discussion_r278771455) [](ancestors = 278771455,278224177,277107266)

Issue #2658, opened 2019-04-09
[Designer] Setting Image width settings incorrect behavior


matthidinger:
See gif. I set the image Width/Height to "50px" then click something else, then re-select the Image, the property sheet shows "50" instead of "50px" 

![designer](https://user-images.githubusercontent.com/1432195/55833476-7b4d4300-5acc-11e9-8db4-08286455ba80.gif)

andrewleader:
Nice ScreenToGif 😎

Issue #2497, opened 2019-03-01
[TS] Typescript renderer should not accept seconds for initial value for Input.Time


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* JavaScript

# Issue

The designer currently accepts times of the form HH:MM:SS as the initial value for an Input.Time. For consistency with the other renderers, it should only accept HH:MM.

shalinijoshi19:
@dclaux i can take a stab at this?

dclaux:
Sure, no problem.

shalinijoshi19:
This could be lower priority vis-a-vis others that are currenty open.

PR #2734, opened 2019-04-19
[TS] Updated RichTextBlock support + other fixes

dclaux:
Update highlightColors model [per spec](https://github.com/Microsoft/AdaptiveCards/issues/1933)
Fix width/height support in designer (https://github.com/Microsoft/AdaptiveCards/issues/2658)
Disallow seconds in default value of TimeInput (https://github.com/Microsoft/AdaptiveCards/issues/2497)
Opportunistic bug fix of JSON round tripping in designer


dclaux:
The bug only says seconds shouldn't be accepted in Input.Time, does the same apply to Input.Date?

RebeccaAnne:
> The bug only says seconds shouldn't be accepted in Input.Time, does the same apply to Input.Date?

Are you currently accepting time strings into Input.Date? Input.Date should only accept strings of the form YYYY-MM-DD.

dclaux:
As a matter of fact, Input.Date doesn't accept a time part in TS. My mistake, I was mixing up our DATE() function which does require the time part to properly handle time zone conversions.

PR #2743, opened 2019-04-23
Add privacy policy for UWP visualizer

andrewleader:
Windows Store is requiring a privacy policy to publish the visualizer, so had to create one for the UWP visualizer


andrewleader:
Hey @RebeccaAnne could you review this? Just adding a privacy policy for the UWP visualizer stating we don't collect any data, so that we can link to that policy and publish it to the Store :)

PR #2750, opened 2019-04-25
[TS] Always bleed columns

dclaux:
Add top/bottom padding to all columns in a set if at least one of them has top/bottom padding
Disable bleed property for Column (see comment in code, property can't be removed)
Remove bleed property on Column in designer

NOTE: Designer support is less than optimal with this change (there is extra space reserved in too many places.) This will be fixed later.

Verified manually in both visualizer and designer.

Fixes https://github.com/Microsoft/AdaptiveCards/issues/2755


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -547,6 +541,12 @@ export abstract class CardElement implements ICardObject {
         return -1;
     }
 
+    isDesignMode(): boolean {
+        var rootElement = this.getRootElement();
```
paulcam206: >var [](start = 8, length = 3)

should this be `let`?

dclaux: Should be let pretty much everywhere :) I'll change all the vars to lets later in a dedicated PR.

Issue #2643, opened 2019-04-09
[Designer] Fix showCard with selectAction behavior to being unsupported


shalinijoshi19:
# Platform

What platform is your issue or question related to? (Delete other platforms).

Designer.

# Author or host

Neither. Feature dev 

If you're an author, who are you sending cards to?

# Version of SDK

What version are you using? Ex: NuGet 1.0.2, or latest master, etc...
master

# Issue

(1) Per discussion today, at the team meeting, we will move our behavior to match what was originally decided upon ie showCard and selectAction is *NOT* supported in Adaptive Card renderers. 

This tracks the work needed in the Designer to account for the behavior change. @dclaux FYI.

shalinijoshi19:
@dclaux should this issue be closed now?

Issue #2642, opened 2019-04-09
[JavaScript] Fix showCard with selectAction behavior to being unsupported in the renderer


shalinijoshi19:
# Platform

What platform is your issue or question related to? (Delete other platforms).
* JavaScript

# Author or host

Neither. Feature dev.

If you're an author, who are you sending cards to?

# Version of SDK

master

# Issue

Wshould remove support for ShowCard and selectaction in the JS renderer.
(1) Per discussion today, at the team meeting, we will move this to match what was originally decided upon ie showCard and selectAction is *NOT* supported in Adaptive Card renderers. 
(2) In addition we will include a warning that the host would get knowing it's pop-up is being dropped though. (@dclaux FYI)

Issue #2500, opened 2019-03-04
ColumnSet allows children of any type


paulcam206:
# Platform

* JavaScript
* UWP
* Haven't tested other platforms

# Issue

This card probably shouldn't parse successfully, but does on UWP and JS (note `ColumnSet`'s direct child):

```json
{
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "ColumnSet",
      "columns": [
        {
          "type": "steve",
          "width": "auto",
          "items": [
            {
              "type": "TextBlock",
              "text": "huh?"
            }
          ]
        }
      ]
    }
  ]
}
```

shalinijoshi19:
@paulcam assigning to you for fixing

shalinijoshi19:
Ideally should be in 1.2 @Paulcam206 please pull from backlog if we start to free up with known 1.2 issue.

PR #2771, opened 2019-04-26
[TS] 1.2 design changes and bug fixes

dclaux:
Changes:
- Disallow Action.ShowCard in renderer
- Disallow Action.ShowCard in designer
- Fix overzealous fallback at the card level
- Fix Columns not falling back
- Stop accepting any type for columns
- Bump renderer current version to 1.2

Verified manually using payloads from bug reports

Addresses:
- https://github.com/Microsoft/AdaptiveCards/issues/2642
- https://github.com/Microsoft/AdaptiveCards/issues/2643
- https://github.com/Microsoft/AdaptiveCards/issues/2775
- https://github.com/Microsoft/AdaptiveCards/issues/2773
- https://github.com/Microsoft/AdaptiveCards/issues/2500
- https://github.com/Microsoft/AdaptiveCards/issues/2786


source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -1685,7 +1689,11 @@ export class ActionSetPeer extends TypedCardElementPeer<Adaptive.AdaptiveCard> {
     }
 }
 
-function createActionSelector(title: string, card: Adaptive.AdaptiveCard, defaultValue: string): ILabelAndInput<Adaptive.ChoiceSetInput> {
+function createActionSelector(
```
shalinijoshi19: Nit: Would you mind creating tracking issues for us to start adding xml-syle function/module level comments where needed? We've started tracking adding these to other source files as well. Thanks!

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -25,48 +25,60 @@ function isActionAllowed(action: Action, forbiddenActionTypes: Array<string>): b
     return true;
 }
 
+enum InstanceCreationErrorType {
+    UnknownType,
+    ForbiddenType
```
shalinijoshi19: Neat. Does this get bubbled up to the author or is this enforced host-side? Assuming it gets bubbled up, is there an equivalent mapping of error type to error message (user-friendly string) that can get bubbled up/included as well?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3794,10 +3831,12 @@ export class SubmitAction extends Action {
 }
 
 export class OpenUrlAction extends Action {
+    static readonly JsonTypeName = "Action.OpenUrl";
```
shalinijoshi19: > "Action.OpenUrl" [](start = 34, length = 17)

Nit: We should start considering adding a Constants module or class equivalent where these string literals can be housed instead of them being sprinkled all over the code (and potentially duplicated etc)

shalinijoshi19:
>                     message: "An Action.ShowCard must have its \"card\" property set to a valid AdaptiveCard object."

Maybe this could start using the JsonTypeName property as well ? 

---
Refers to: source/nodejs/adaptivecards/src/card-elements.ts:4166 in 5f0faab. [](commit_id = 5f0faabde2aa359c63107204b84b733530650668, deletion_comment = False)

source/nodejs/adaptivecards-designer/src/designer-peers.ts
dclaux: Opened https://github.com/Microsoft/AdaptiveCards/issues/2789

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5618,6 +5664,31 @@ export class ColumnSet extends StylableCardElementContainer {
     private _columns: Array<Column> = [];
     private _renderedColumns: Array<Column>;
 
+    private createColumnInstance(json: any, errors: Array<HostConfig.IValidationError>): Column {
+        return createCardObjectInstance<Column>(
+            this,
+            json,
+            [], // Forbidden types not supported for elements for now
```
shalinijoshi19: > for no [](start = 61, length = 7)

Curious: Why "for now" ? Is the concept of forbidden types a temporary idea?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5618,6 +5664,31 @@ export class ColumnSet extends StylableCardElementContainer {
...
+            (typeName: string) => {
+                return !typeName || typeName === "Column" ? new Column() : null;
+            },
+            (typeName: string, errorType: InstanceCreationErrorType) => {
+                if (errorType == InstanceCreationErrorType.UnknownType) {
+                    return {
+                        error: Enums.ValidationError.UnknownElementType,
+                        message: "Invalid element type " + typeName + " in a ColumnSet. Attempting to fall back."
```
shalinijoshi19: > fall back [](start = 101, length = 10)

Nit: "fallback" (one word) to be consistent with the usage of the term elsewhere in this same file.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Well, that's precisely what this is doing, and it's contextual to the action class, which is a better option for this particular case. I considered doing the same for element names in this PR but deferred to a later time.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5618,6 +5664,31 @@ export class ColumnSet extends StylableCardElementContainer {
...
+                    }
+                }
+                else {
+                    return {
+                        error: Enums.ValidationError.ElementTypeNotAllowed,
+                        message: "Element type " + typeName + " is not allowed in this context."
```
shalinijoshi19: Hmm we are already generalizing the concept of "invalid types" - why not consolidate this into a single handler for forbidden types? Why the special handling of UnknownType and ElementTypeNotAllowed?

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: There is no need to support it for elements right now. The underlying logic (createCardObjectInstance) supports it but there is no requirement to use it at this time for elements.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: If you only get "invalid element", do you know why it's invalid? Is it because its type is unknown? Is it because you're using an element in a context where it can't be used? Is it for another (future) reason?

Say you're coding and you write this statement: `a = b`. The compiler says "Invalid statement"; would it be better if the compiler said "Can't assign b to a because a is a Boolean and b is a string"

dclaux: Also, please remember that the JS renderer supports the designer. The designer needs to be as helpful as possible.

shalinijoshi19: I guess i am having a hard time distinguishing between "INvalid element type <> in a columnset." vs "Element type not allowed in this context" -- what is invalid and what is not allowed?

---
In reply to: [279437943](https://github.com/Microsoft/AdaptiveCards/pull/2771#discussion_r279437943) [](ancestors = 279437943)

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: That wouldn't be grammatically correct though. "To fallback" isn't a verb as far as I know.

shalinijoshi19: Sure but looks like we have been using "fallback" as a verb - in our docs, in this code file even . My comment is merely to make usage consistent.

---
In reply to: [279438647](https://github.com/Microsoft/AdaptiveCards/pull/2771#discussion_r279438647) [](ancestors = 279438647)

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Not 100% sure what you're saying. These errors get bubbled up to the developer (when they integrate the AC renderer in their own app) to help them debug and make runtime decisions. These errors are also bubble up in the designer to help the user "debug" (correct) their card. There isn't a single string associated with each error code; the string is contextual, the code is there to help the calling code make decisions.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Sorry, I should have looked at the code more carefully, which would have helped me better understand what you were saying. The messages are incorrect indeed (I copy pasted and forgot to cleanup). I'll push a change. The final messages will be:
- Unknown element type <element type>. Attempting to fall back.
- Element type <element type> isn't allowed in a ColumnSet.

source/nodejs/adaptivecards/src/card-elements.ts
shalinijoshi19: Speaking of error message to go with the error code; For eg ForbiddenType by itself isnt that helpful as an associated error string with it that could be more descriptive for eg. I think though this is a larger ticket cleanup item across all our renderers (consolidating on error codes + messages as well as tackling localization of those error messages) at some point. Doesnt need to block this PR

---
In reply to: [279439868](https://github.com/Microsoft/AdaptiveCards/pull/2771#discussion_r279439868) [](ancestors = 279439868)

dclaux: There is always a string associated with the error code. The string is for debug purposes and is in most case very descriptive of what the problem is. The error code is what allows an app to make a decision at runtime as to whether it should display the card. ForbiddenType means the parser encountered an action or element type that isn't allowed. This might have caused fallback to happen or the action/element being dropped. An app can decide that in those situations it doesn't even want to render the card.

Issue #2678, opened 2019-04-12
[Shared] Bleed properties are not updated when bleed values are changed on the object model


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* Android
* iOS
* UWP

# Issue

Bleed is calculated at parse time and will not be updated if the user updates it using the object model.

RebeccaAnne:
This is related to how we handle shared logic in the shared model.

PR #2802, opened 2019-04-30
[UWP][Shared] Support bleeding up and down in addition to left and right

### LINKED ISSUE #2703: [Shared] Elements with bleed should bleed top and bottom in addition to left and right

Shared fix of #2702

### LINKED ISSUE #2707: [iOS] Elements with bleed should bleed top and bottom in addition to left and right

iOS fix of #2702

### LINKED ISSUE #2709: [UWP] Elements with bleed should bleed top and bottom in addition to left and right

UWP fix of #2702

RebeccaAnne:
Add support for bleeding up and down in addition to left and right. Expanded the bleed enum to be a bitmask and include options for up and down. Refactored the logic we used to determine if columns should bleed left/right to determine if elements in containers/columns should bleed up/down. Updated UWP to check for the new values and bleed up and down if necessary.

Fixes #2709
Fixes #2703
Fixes #2707 
How verified: Updated shared model unit test and test app results.


source/uwp/Renderer/lib/AdaptiveColumn.cpp
```diff
@@ -37,11 +37,7 @@ namespace AdaptiveNamespace
         m_verticalAlignment =
             static_cast<ABI::AdaptiveNamespace::VerticalContentAlignment>(sharedColumn->GetVerticalContentAlignment());
         m_bleed = sharedColumn->GetBleed();
-
-        if (sharedColumn->GetCanBleed())
```
shalinijoshi19: Does this now account for the changes agreed upon yesterday ?

shalinijoshi19:
It would be great to start including a quick description summarizing the changes contained in the PR for posterity; Continue to reference (for linking logisitcs) the underlying bugs it fixes but also include a summary and a "How Verified" section in teh description. Thanks

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ObjectModelTest.cpp
```diff
@@ -1142,6 +1142,64 @@ namespace AdaptiveCardsSharedModelUnitTest
             }
         }
 
+        TEST_METHOD(ImplicitColumnTypeTest)
```
shalinijoshi19: Do we have tests that test column + bleed? Also do ensure the "corner" cases that we've identified in our various discussions are captured somewhere

shalinijoshi19:
>             {

Should we have tests that test the behavior for each of the enums you are bringing in - all/left/right etc

---
Refers to: source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ObjectModelTest.cpp:1139 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

source/shared/cpp/ObjectModel/Column.cpp
```diff
@@ -95,7 +95,8 @@ void Column::GetResourceInformation(std::vector<RemoteResourceInformation>& reso
 void Column::DeserializeChildren(ParseContext& context, const Json::Value& value)
 {
     // Parse Items
-    auto cardElements = ParseUtil::GetElementCollection(context, value, AdaptiveCardSchemaKey::Items, false);
+    auto cardElements =
+        ParseUtil::GetElementCollection<BaseCardElement>(true, context, value, AdaptiveCardSchemaKey::Items, false);
```
shalinijoshi19: NIt: Nice to include the argument names in comments /**/ style for readability

shalinijoshi19:
>         columnWidth = ParseUtil::GetValueAsString(value, AdaptiveCardSchemaKey::Size);

Q: Should we be deprecating "size" given we have a replacement/better named/explicit post-1.0 alternative? @matthidinger/@andrewleader/@dclaux? 

---
Refers to: source/shared/cpp/ObjectModel/Column.cpp:111 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

shalinijoshi19:
> std::shared_ptr<BaseCardElement> ColumnParser::Deserialize(ParseContext& context, const Json::Value& value)

What's the error contract look like in teh object model again in general? Looks like these can all potentially throw? If Deserialize() throws for eg, what's the behavior look like for the caller (assuming it's being invoked both during parsing and rendering yes?) 

---
Refers to: source/shared/cpp/ObjectModel/Column.cpp:103 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

source/shared/cpp/ObjectModel/ParseUtil.h
```diff
@@ -184,4 +187,74 @@ namespace AdaptiveSharedNamespace
         T result = jsonObject.empty() ? defaultValue : deserializer(jsonObject, defaultValue);
         return result;
     }
+
+    template<typename T>
+    std::vector<std::shared_ptr<T>> ParseUtil::GetElementCollection(bool isTopToBottomContainer,
```
shalinijoshi19: So just continuing with improving unit test coverage (if we dont already have it) let's make sure this function has good unit test coverage

source/uwp/AdaptiveCardTestApp/Expected/SourceCards/Tests.Bleed.Comprehensive.2087edd.json
```diff
@@ -0,0 +1,692 @@
...
+											"items": [
+												{
+													"type": "Container",
+													"style": "emphasis",
+													"items": [
+														{
+															"type": "TextBlock",
+															"horizontalAlignment": "Center",
+															"text": "Y"
+														}
```
shalinijoshi19: Cool - am guessing this feeds into our unit tests?

shalinijoshi19:
>         // the shared model logic.

Is this comment still applicable? Do we have a tracking issue# to track this ? If so, let's start referencing the issue# in teh TODO comments for tracking TODO: <Issue# > etc

---
Refers to: source/uwp/Renderer/lib/AdaptiveColumnSet.cpp:109 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -530,15 +530,17 @@ AdaptiveNamespaceStart
         Content
     } FallbackType;
 
-    [version(NTDDI_WIN10_RS1)]
+    [version(NTDDI_WIN10_RS1), flags]
 #ifdef ADAPTIVE_CARDS_WINDOWS
     [contract(InternalContract, 1)][internal]
 #endif
     typedef[v1_enum] enum BleedDirection {
```
shalinijoshi19: >v1_enum] [](start = 12, length = 8)

Curious: What is our versioning story in general and why do we even have this explicit typedef in place? Have we looked at the WinRT API design guidelines or gone through an API review at all all with Wiswal and Co?

shalinijoshi19:
>     HRESULT WholeItemsPanel::ArrangeOverride(Size finalSize, __RPC__out Size* returnValue)

This is a huge and complicated function and likely error prone/difficult to test. Consider refactoring at some point to make things easier - both testing-wise and maintainability/readability-wise

---
Refers to: source/uwp/Renderer/lib/WholeItemsPanel.cpp:243 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

shalinijoshi19:
>     HRESULT AdaptiveColumnSet::GetSharedModel(std::shared_ptr<AdaptiveSharedNamespace::BaseCardElement>& sharedModel) try

nit: should we annotate with noexcept here? 

---
Refers to: source/uwp/Renderer/lib/AdaptiveColumnSet.cpp:120 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

shalinijoshi19:
>                 return E_UNEXPECTED;

nit: consider using wil macro here as well RETURN_HR_IF_NULL(E_UNEXPECTED, sharedColumn) ; Also similarly lets bet consistent validating the HRESULT as well as pointer validity in extracting the sharedColumnBaseElement as well

---
Refers to: source/uwp/Renderer/lib/AdaptiveColumnSet.cpp:142 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

shalinijoshi19:
>     HRESULT AdaptiveColumnSet::get_MinHeight(_Out_ UINT32* minHeight)

Do we not validate minHeight parameters at all? no business logic around these getters/setters? 

---
Refers to: source/uwp/Renderer/lib/AdaptiveColumnSet.cpp:82 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

shalinijoshi19:
>     HRESULT AdaptiveColumnSet::RuntimeClassInitialize(const std::shared_ptr<AdaptiveSharedNamespace::ColumnSet>& sharedColumnSet) try

Probably OK to annotate with noexcept also?

---
Refers to: source/uwp/Renderer/lib/AdaptiveColumnSet.cpp:33 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

shalinijoshi19:
>     HRESULT WholeItemsPanel::GetAltText(__RPC__out HSTRING* pResult)

noexcept

---
Refers to: source/uwp/Renderer/lib/WholeItemsPanel.cpp:364 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

shalinijoshi19:
>         for (unsigned int i{}; i < m_measuredCount; ++i)

Nit Nit: let's be consistent for our initialization here; i think using auto i =0 is just fine in for loops; I can see both versions in use in this same file (with/without {} initialization.

---
Refers to: source/uwp/Renderer/lib/WholeItemsPanel.cpp:271 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

shalinijoshi19:
>             RETURN_IF_FAILED(spChildren->GetAt(i, spChild.GetAddressOf()));

Nit: In general, let's stay away from hungarian and move to the newer naming guidelines ( i think xaml was old code)

---
Refers to: source/uwp/Renderer/lib/WholeItemsPanel.cpp:274 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ObjectModelTest.cpp
RebeccaAnne: This particular case is a test for an old scenario. I touched the code for it in my refactoring so i added this test to make sure i didn't break it. The tests for column bleed are in the ContainerStyleTest.cpp file.

---
In reply to: [279799340](https://github.com/Microsoft/AdaptiveCards/pull/2802#discussion_r279799340) [](ancestors = 279799340)

RebeccaAnne: We also have the comprehensive bleed test file which tests the bleed scenarios for UWP.

---
In reply to: [280105918](https://github.com/Microsoft/AdaptiveCards/pull/2802#discussion_r280105918) [](ancestors = 280105918,279799340)

RebeccaAnne:
>             {

The tests in ContainerStyleTest test that the correct enum values are returned for various bleed scenarios. Let me know if you think something particular is missing.

---
In reply to: [487990938](https://github.com/Microsoft/AdaptiveCards/pull/2802#issuecomment-487990938) [](ancestors = 487990938)

---
Refers to: source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ObjectModelTest.cpp:1139 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

RebeccaAnne:
> std::shared_ptr<BaseCardElement> ColumnParser::Deserialize(ParseContext& context, const Json::Value& value)

This C++ shared model handles parsing json and providing an object model but not rendering. It throws on error by design. This code is consumed by our UWP, Android, and iOS code. In UWP, where the error contract is HRESULT based, the code catches all exceptions when using the shared model. I believe Android is exception based all the way up, and i'm not sure what error model iOS uses.

---
In reply to: [487995301](https://github.com/Microsoft/AdaptiveCards/pull/2802#issuecomment-487995301) [](ancestors = 487995301)

---
Refers to: source/shared/cpp/ObjectModel/Column.cpp:103 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

RebeccaAnne:
>         // the shared model logic.

We do have an issue tracking, it's #2678. I'll add the issue number to the code.

---
In reply to: [487999456](https://github.com/Microsoft/AdaptiveCards/pull/2802#issuecomment-487999456) [](ancestors = 487999456)

---
Refers to: source/uwp/Renderer/lib/AdaptiveColumnSet.cpp:109 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

RebeccaAnne:
>     HRESULT WholeItemsPanel::ArrangeOverride(Size finalSize, __RPC__out Size* returnValue)

I agree. This code is inherited from somewhere outside our team, and I'm not fully familiar with all its nuances, but it will probably need to be refactored some day.

---
In reply to: [488003938](https://github.com/Microsoft/AdaptiveCards/pull/2802#issuecomment-488003938) [](ancestors = 488003938)

---
Refers to: source/uwp/Renderer/lib/WholeItemsPanel.cpp:243 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

RebeccaAnne:
>         for (unsigned int i{}; i < m_measuredCount; ++i)

Sure. This is inherited code so the styling is not particularly consistent with other code that we've written (or apparently itself :))

---
In reply to: [488036810](https://github.com/Microsoft/AdaptiveCards/pull/2802#issuecomment-488036810) [](ancestors = 488036810)

---
Refers to: source/uwp/Renderer/lib/WholeItemsPanel.cpp:271 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

RebeccaAnne:
>             RETURN_IF_FAILED(spChildren->GetAt(i, spChild.GetAddressOf()));

We're not using Hungarian in our code. This is inherited and brought some with it. I'll update the "sp's" in this function while i'm here.

---
In reply to: [488037386](https://github.com/Microsoft/AdaptiveCards/pull/2802#issuecomment-488037386) [](ancestors = 488037386)

---
Refers to: source/uwp/Renderer/lib/WholeItemsPanel.cpp:274 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
RebeccaAnne: I believe windows API review is scoped for APIs that ship out of windows (which this does not) and thus this hasn't been formally reviewed. We did get some informal guidance from them on some things early on when we were designing the API.

---
In reply to: [279813721](https://github.com/Microsoft/AdaptiveCards/pull/2802#discussion_r279813721) [](ancestors = 279813721)

source/uwp/AdaptiveCardTestApp/Expected/SourceCards/Tests.Bleed.Comprehensive.2087edd.json
RebeccaAnne: All of the test and sample files feed into our UWP test app which tests rendering, and i believe also into .NET tests which test parsing only.

---
In reply to: [279808458](https://github.com/Microsoft/AdaptiveCards/pull/2802#discussion_r279808458) [](ancestors = 279808458)

RebeccaAnne:
>     HRESULT AdaptiveColumnSet::get_MinHeight(_Out_ UINT32* minHeight)

In general our object model doesn't contain much in the way of validation/business logic, most of which is being handled at parse time. As a result, it is a known issue in general that cards constructed by the object model directly in code (not via parsing json) may not be invalid in a number of ways.

All that said, it doesn't look like we're doing any particular validation at parse time on min height other than making sure it's a properly formatted valid number. 

---
In reply to: [488020206](https://github.com/Microsoft/AdaptiveCards/pull/2802#issuecomment-488020206) [](ancestors = 488020206)

---
Refers to: source/uwp/Renderer/lib/AdaptiveColumnSet.cpp:82 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

source/shared/cpp/ObjectModel/ParseUtil.h
RebeccaAnne: We don't call this function directly from the tests, but the logic here is well tested by the ContainerStyleTest.cpp test cases.

---
In reply to: [279807494](https://github.com/Microsoft/AdaptiveCards/pull/2802#discussion_r279807494) [](ancestors = 279807494)

RebeccaAnne:
>                 return E_UNEXPECTED;

So, we don't actually have access to WIL macros in this code, we just have our own definitions of RETURN_IF_FAILED and THROW_IF_FAILED.

Also, we should **not** be validating pointers that come out of functions that return HRESULTs. COM contract is to not return NULL on successful HRESULT, and it would add **a lot** of noise to be checking pointers out of every function call. 

This check is out of a cast, whose contract is to return NULL on failure to cast. That shouldn't happen in our code (hence the E_UNEXPECTED), but is not out of the scope of the contract.

---
In reply to: [488019869](https://github.com/Microsoft/AdaptiveCards/pull/2802#issuecomment-488019869) [](ancestors = 488019869)

---
Refers to: source/uwp/Renderer/lib/AdaptiveColumnSet.cpp:142 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

shalinijoshi19:
>     HRESULT AdaptiveColumnSet::get_MinHeight(_Out_ UINT32* minHeight)

Thanks; For the known issue do we have an actual github issue tracking this? If not could you help create one and i'll make sure we plan time for it one of the releases

---
In reply to: [488328449](https://github.com/Microsoft/AdaptiveCards/pull/2802#issuecomment-488328449) [](ancestors = 488328449,488020206)

---
Refers to: source/uwp/Renderer/lib/AdaptiveColumnSet.cpp:82 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

source/shared/cpp/ObjectModel/Column.cpp
RebeccaAnne: Personally, I tend to find it harder to read to have the comments mixed in inline like that. That said, i see that those bools are non-obvious if you're in code flow and not an IDE. Took a look at the coding guidelines, which prefer // comments over /**/ comments, and went with separating out the parameters and adding end of line comments to the bools.

---
In reply to: [279800461](https://github.com/Microsoft/AdaptiveCards/pull/2802#discussion_r279800461) [](ancestors = 279800461)

RebeccaAnne:
>     HRESULT AdaptiveColumnSet::get_MinHeight(_Out_ UINT32* minHeight)

Logged #2811. We haven't prioritized this in the past as we don't believe constructing cards locally using the object model is a common case. As a result getting the error cases of it right hasn't been a priority. That said, there may be all kinds of issues lurking if our renderers make assumptions that cards that they are getting are validated by the parser when they may not have been.

---
In reply to: [488336024](https://github.com/Microsoft/AdaptiveCards/pull/2802#issuecomment-488336024) [](ancestors = 488336024,488328449,488020206)

---
Refers to: source/uwp/Renderer/lib/AdaptiveColumnSet.cpp:82 in 96a9504. [](commit_id = 96a95045d1f40bc0ca297e499e5b6fdd8d6863de, deletion_comment = False)

source/shared/cpp/ObjectModel/ParseUtil.h
```diff
@@ -184,4 +187,74 @@ namespace AdaptiveSharedNamespace
...
+        size_t currentIndex = 0;
+        for (auto& curJsonValue : elementArray)
+        {
+            ContainerBleedDirection currentBleedState = previousBleedState;
+
+            if (currentIndex != 0)
+            {
+                // In a top to bottom container (Container, Column) only the first item can bleed up, in a left
+                // to right container (ColumnSet) only the first item can bleed left
+                currentBleedState &= isTopToBottomContainer ? ~ContainerBleedDirection::BleedUp : ~ContainerBleedDirection::BleedLeft;
```
nesalang: This line was a bit confusing because just looking at the enum name and the name of the type, it didn't readily connect that it's turning off one bit.

RebeccaAnne: I'll update the comment to say that explicitly

---
In reply to: [280529780](https://github.com/Microsoft/AdaptiveCards/pull/2802#discussion_r280529780) [](ancestors = 280529780)

source/shared/cpp/ObjectModel/Enums.h
```diff
@@ -402,11 +402,33 @@ namespace AdaptiveSharedNamespace
...
-        BleedRestricted = 0,
-        BleedToLeading,
-        BleedToTrailing,
-        BleedToBothEdges,
+        BleedRestricted = 0x0000,
+        BleedLeft = 0x0001,
+        BleedRight = 0x0010,
+        BleedUp = 0x0100,
+        BleedDown = 0x1000,
+        BleedAll = BleedLeft | BleedRight | BleedUp | BleedDown
```
almedina-ms: >BleedAll = BleedLeft | BleedRight | BleedUp | BleedDown [](start = 8, length = 55)

this sadly doesn't work in android, it looks like java doesn't allow it operators between enums 

[https://stackoverflow.com/questions/6282619/is-there-a-way-to-bitwise-or-enums-in-java](https://stackoverflow.com/questions/6282619/is-there-a-way-to-bitwise-or-enums-in-java)

almedina-ms:
Was there any change to the ParseJsonObject function? this branch doesn't build correctly in android as it can't find a template value

nesalang:
it should work now. it was same for iOS. I added fix

---
In reply to: [490656729](https://github.com/Microsoft/AdaptiveCards/pull/2802#issuecomment-490656729) [](ancestors = 490656729)

PR #2805, opened 2019-04-30
[UWP][.NET] Add logging and update DeepFallback card

### LINKED ISSUE #2790: Warnings should be logged when fallback is performed
When element fallback is performed, a warning should be logged for the renderer to consume.

paulcam206:
Fixes #2790 for UWP and .NET. Additional validation needed for iOS and Android.

While adding in fallback logging, I came across a couple of overlooked scenarios. I added the scenarios to `DeepFallback.json`, and then fixed the issues.

#### How Verified
DeepFallback.json - manual inspection of warning output


source/uwp/AdaptiveCardTestApp/Expected/cortana-skills.Tests.DeepFallback.json
```diff
@@ -1 +1 @@
-{"HostConfigHash":"a60a3b3","CardHash":"9b23448","Error":null}
\ No newline at end of file
+{"HostConfigHash":"a60a3b3","CardHash":"f7f78ea","Error":null}
```
RebeccaAnne: Maybe we should add checking for warnings to the UWP test app... Logged #2826.

Issue #2797, opened 2019-04-29
[.NET] Background image spec changes in 1.2


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

.NET fix of #2794

shalinijoshi19:
@almedina-ms assigning to you (load balancing).

Issue #2794, opened 2019-04-29
Background image spec changes in 1.2


andrewleader:
<table>
<tr>
<th>Release</th>
<th>Renderer status</th>
<th>Tasks status</th>
</tr>
<tr>
<td>1.2
<td>✔️ .NET (#2797)<br/>✔️ Android (#2798)<br/>✔️ iOS (#2799)<br/>✔️ TS (#2800)<br/>✔️ UWP (#2801)</td>
<td>✔️ Shared (#2795)<br/>✔️ Designer (#2796)</td>
</tr>
</table>

Spec changes as of 4/29/2019 (Matt, Shalini, Becky, David, Paul, Alberto, Joseph were in the meeting, Mike gave feedback on email ahead of time)

- **Rename** the `stretch` value to `cover`
  - consistent with HTML, ReactN, Flutter and seems to be the term people are converging on
- **Rename** the `mode` property to `fillMode`
  - `mode` alone was a bit ambiguous. There was no single common term in any platforms, so `fillMode` sounded best.
- **Rename** the `BackgroundImageMode` property to `ImageFillMode`
  - Removing any reference to "Background", since in the future this enum could also be useful on image itself
- **Add** support for using `verticalAlignment` and `horizontalAlignment` in the `cover` scenario
  - Previously the spec said these properties are ignored when using `cover`, but they're definitely useful! We'll support these properties.

# Auto-generated task status

- [x] Shared
- [x] Designer
- [x] .NET
- [x] Android
- [x] iOS
- [x] TS
- [x] UWP

PR #2806, opened 2019-04-30
[.NET][WPF][HTML] Update BackgroundImage implementation for spec changes 

almedina-ms:
* Updates **stretch** enum value for **cover** (BackgroundImageMode)
* Updates the enum **BackgroundImageMode** to **ImageFillMode**
* Renames the background image property **Mode** to **FillMode**
* Adds support for **verticalAlignment** and **horizontalAlignment** in the _cover_ scenario

As detailed here https://github.com/Microsoft/AdaptiveCards/issues/2794

Fixes this: https://github.com/Microsoft/AdaptiveCards/issues/2797

### Pre-check

As it updates enum and property names, the first check was for the project to compile (it does).
Also the tests were run and worked as expected though a couple had to be updated

### VerticalAlignment and HorizontalAlignment

The card Tests/BackgroundImage.VerticalAndHorizontalAlignment.json was added to check the correct alignment of the image in the containers and columns

WPF

This one's a little weird, according to the documentation, auto should use as much space as needed, so if the columnset width overflows it's "expected"? I need more info on the person who made the first column implementation
image

I removed the word background here so the horizontal alignment effect could be viewed
image

HTML
image

### All other cards

HTML
image

WPF
image

HTML
image

WPF
image

HTML
image

WPF
image

HTML
image

WPF
image

HTML
image

WPF
image

HTML
image

WPF
image

HTML
image

WPF
image

HTML
image

WPF
image


andrewleader:
Epic work Alberto!!

@matthidinger for columns, the `auto` width shouldn't overflow past the width of the card, correct?

As seen below, the column contents on the right are too large, so right now in WPF the column all exceeds

img

I believe the correct "auto" column width implementation is... the first "auto" column gets to consume as much width as it wants (but can't consume more than the remaining width), then the next "auto" width gets to consume as much remaining width available, etc...

So in that case, the right-most column's width should be set to the remaining width, causing the text to wrap

paulcam206:
@almedina-ms -- has this been updated for @andrewleader's comments above?

almedina-ms:
I'm not sure if I should address this comment here as this PR is only for the background image spec update while the column issue has been there for some time. Created an issue to fix later [https://github.com/microsoft/AdaptiveCards/issues/2862](https://github.com/microsoft/AdaptiveCards/issues/2862)

---
In reply to: [490187118](https://github.com/microsoft/AdaptiveCards/pull/2806#issuecomment-490187118) [](ancestors = 490187118)

samples/Tests/BackgroundImage.VerticalAndHorizontalAlignment.json
```diff
@@ -0,0 +1,148 @@
+{
```
paulcam206: Need to run the UWPTestApp and add reference files to this PR :)

PR #2828, opened 2019-05-01
Designer improvements

dclaux:
- Do not fall back in design mode
- Do not hide elements in design mode


andrewleader:
Ah, so this means if I have `"isVisible": false`, the element will still be displayed in the designer? It essentially ignores isVisible?

And if I have `"requires"` set to require Adaptive 50.0, it'll still display the element, essentially ignoring "requires"? 

I suppose that makes sense, because otherwise there'd be no easy way to design those isVisible elements other than accessing them from the tree hierarchy.

Does this add the "Preview mode" button at all? We'll definitely need a preview mode where people can test clicking on their Action.ToggleVisibility and whatnot!

dclaux:
Yes, correct.

No, I didn't bring the Preview mode in here with this change, but it is necessary and will address later.

PR #2836, opened 2019-05-03
[Android] Add warnings around fallback

### LINKED ISSUE #2790: Warnings should be logged when fallback is performed

paulcam206:
Fixes #2790 on Android.

#### How verified
Test card run in visualizer with debugger attached


shalinijoshi19:
Nit: A "How Verified" would be nice in the description

shalinijoshi19:
>                     throw e;

Again this should be fixed to just being 'throw' so it doesnt mess up the original stack trace

---
Refers to: source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/registration/CardRendererRegistration.java:353 in 6dff2d9. [](commit_id = 6dff2d9adb52c626c78e061b6421ab4576651daa, deletion_comment = False)

Issue #2772, opened 2019-04-26
[.NET] Action Fallback card doesn't render properly


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* .NET WPF
* .NET HTML

# Issue

The [samples\v1.2\Elements\ActionFallback.json](https://github.com/Microsoft/AdaptiveCards/blob/master/samples/v1.2/Elements/ActionFallback.json) card does not render correctly on .NET WPF visualizer. The fallback actions are not being rendered. 

UWP results [correct]:
![image](https://user-images.githubusercontent.com/9091833/56838040-87a5ff80-6831-11e9-8c16-44c525fe0334.png)

.NET WPF results [missing fallback actions]:
![image](https://user-images.githubusercontent.com/9091833/56838075-a0aeb080-6831-11e9-9fb6-06fe8eb69780.png)

shalinijoshi19:
@almedina-ms is this something you can take a stab at ?

almedina-ms:
@shalinijoshi19 yes, I can take a stab at this

PR #2837, opened 2019-05-03
[.NET][WPF] Fix Action Fallback rendering for unknown elements

almedina-ms:
The issue was that actions with unknown type were not being called for rendering and any action whose fallback ended up not being a rendereable element was  not rendering either so we were trying to insert a null children (not valid)

Also, a scrollview for the warnings was added as the warning list was cropping the payload are and the rendered card area

Fixes https://github.com/Microsoft/AdaptiveCards/issues/2772


source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionSetRenderer.cs
```diff
@@ -141,8 +145,8 @@ public static void AddActions(Grid uiContainer, IList<AdaptiveAction> actions, A
         private static List<AdaptiveAction> GetActionsToProcess(IList<AdaptiveAction> actions, int maxActions)
         {
             // only consider known actions for ActionsToProcess
-            var actionsToProcess = actions.Where(obj => obj.GetType() != typeof(AdaptiveUnknownAction)).ToList();
-            return actionsToProcess.Take(maxActions).ToList();
+            // var actionsToProcess = actions.ToList();
```
paulcam206: >// var actionsToProcess = actions.ToList(); [](start = 12, length = 43)

can remove this :)

shalinijoshi19:
@paulcam206/ @almedina-ms , is there a reason this is still open?

paulcam206:
heh -- still not sure why the bot claims there aren't any pipelines associated...

PR #2854, opened 2019-05-03
[BugBash][MinHeight] Adding additional json for minHeight testing

shalinijoshi19:
Sample that includes a background image similar to what the end-end scenario was based on the feature spec.

How verified:
- Used the sample for bug bashing minHeight


samples/v1.2/Elements/AdaptiveCard.MinHeightBackgroundImg.json
```diff
@@ -0,0 +1,7 @@
+{
+    "type": "AdaptiveCard",
```
nesalang: for json, we use tabs for indentation...

shalinijoshi19:
@paulcam206 to double check if we do have a minheight sample at this point; To add this if not.

Issue #2706, opened 2019-04-17
[Android] Elements with bleed should bleed top and bottom in addition to left and right


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

Android fix of #2702

shalinijoshi19:
@almedina-ms  / @RebeccaAnne is this required to be fixed explicitly for Android or is sharedmodel changes going to address things here?

RebeccaAnne:
There will be renderer specific work for Android once the shared model changes are in place.

PR #2867, opened 2019-05-08
[Android] Bleed all directions

almedina-ms:
Fixes this issue: https://github.com/microsoft/AdaptiveCards/issues/2706 

### How verified

The cards added or modified in v1.2/Elements were rendered and this is the result

Screenshot_2019-05-08-16-51-22

Screenshot_2019-05-08-16-51-32

Screenshot_2019-05-08-16-51-39

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ContainerRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ContainerRenderer.java
@@ -162,20 +162,31 @@ public class ContainerRenderer extends BaseCardElementRenderer
         {
             int padding = Util.dpToPixels(context, hostConfig.GetSpacing().getPaddingSpacing());
             LinearLayout.LayoutParams layoutParams = (LinearLayout.LayoutParams) collectionElementView.getLayoutParams();
-            int marginLeft = layoutParams.leftMargin, marginRight = layoutParams.rightMargin;
+            int marginLeft = layoutParams.leftMargin, marginRight = layoutParams.rightMargin, marginTop = layoutParams.topMargin, marginBottom = layoutParams.bottomMargin;
 
             ContainerBleedDirection bleedDirection = collectionElement.GetBleedDirection();
-            if (bleedDirection == ContainerBleedDirection.BleedToLeading || bleedDirection == ContainerBleedDirection.BleedToBothEdges)
+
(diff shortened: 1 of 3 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ContainerRenderer.java
```diff
@@ -162,20 +162,31 @@ public static void ApplyBleed(CollectionTypeElement collectionElement, LinearLay
         {
             int padding = Util.dpToPixels(context, hostConfig.GetSpacing().getPaddingSpacing());
             LinearLayout.LayoutParams layoutParams = (LinearLayout.LayoutParams) collectionElementView.getLayoutParams();
-            int marginLeft = layoutParams.leftMargin, marginRight = layoutParams.rightMargin;
+            int marginLeft = layoutParams.leftMargin, marginRight = layoutParams.rightMargin, marginTop = layoutParams.topMargin, marginBottom = layoutParams.bottomMargin;
 
             ContainerBleedDirection bleedDirection = collectionElement.GetBleedDirection();
-            if (bleedDirection == ContainerBleedDirection.BleedToLeading || bleedDirection == ContainerBleedDirection.BleedToBothEdges)
+
+            if ((bleedDirection.swigValue() & ContainerBleedDirection.BleedLeft.swigValue()) != ContainerBleedDirection.BleedRestricted.swigValue())
```
nesalang: >!= ContainerBleedDirection.BleedRestricted.swigValue() [](start = 93, length = 54)

What is this for?

source/shared/cpp/ObjectModel/Enums.h
```diff
@@ -402,12 +402,22 @@ namespace AdaptiveSharedNamespace
 
     enum class ContainerBleedDirection
```
RebeccaAnne: Please add a comment here explaining the issues with java and/or swig which require us to explicitly define all of these combinations.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ContainerRenderer.java
almedina-ms: That's to check that the value is not zero

---
In reply to: [282597046](https://github.com/microsoft/AdaptiveCards/pull/2867#discussion_r282597046) [](ancestors = 282597046)

PR #2870, opened 2019-05-09
[.NET] Hide all showcards before showing a new one

### LINKED ISSUE #2769: [WPF] Multiple show cards don't interact together correctly
# Platform

* .NET WPF

# Version of SDK

Latest master

# Issue

When a card has multiple ShowCards, they don't behave correctly together. Problem is illustrated with the default Adaptive Card in the [ActivityUpdate.json sample](https://github.com/Microsoft/AdaptiveCards/blob/master/samples/v1.0/Scenarios/ActivityUpdate.json).

## Problem 1

Repro steps

1. Click "Set due date" *(works correctly)*
2. Now click "Comment" *(works correctly, comment ShowCard now visible)*
3. Now click "Comment" again

**Expected:** The ShowCard disappears
**Actual:** The Comment ShowCard disappears, but the Set due date one is still there!!

WpfShowCard1

## Problem 2

Repro steps

1. Re-load the visualizer
2. Click "Comment" *(works correctly)*
3. Now click "Set due date"

**Expected:** The Comment ShowCard disappears, and the Set due date show card should appear
**Actual:** Nothing happens!! Seems to be that the Set due date show card is added underneath... if you click "Comment" again, it hides the Comment show card and reveals the Set due date one.

WpfShowCard2

paulcam206:
Fixes #2769

How verified:
Test cards and visualizer


shalinijoshi19:
Do we need new test cards for this? Curious if there had been no test coverage for this earlier? Could you update the bug bash scenario onenote with teh specific/interesting scenario here + sample links/path?

paulcam206:
> Do we need new test cards for this? Curious if there had been no test coverage for this earlier? Could you update the bug bash scenario onenote with teh specific/interesting scenario here + sample links/path?

The bug (#2769) mentions the existing card to test with (samples\v1.0\scenarios\ActivityUpdate.json). I think this was missed due to lack of interactive tests (a deficiency noted in our test onenote).

Issue #2799, opened 2019-04-29
[iOS] Background image spec changes in 1.2


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

iOS fix of #2794

shalinijoshi19:
@nesalang assigning the iOS implementation to you for this one for load balancing

shalinijoshi19:
@nesalang FYI :) . Have you had a chance to start on this at all?  tx

PR #2880, opened 2019-05-10
Jwoo/ios background image spec update

nesalang:
fixes issue #2799 

how verified:
used jsons to confirm that it correctly behaves.


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BackgroundImageLoaderAsync.java
```diff
@@ -70,13 +74,76 @@ public void draw(Canvas canvas)
...
+         */
+        private double getScaleFactorForCover(Canvas containerCanvas, Bitmap originalSizeBitmap)
+        {
+            double xScaleFactor = (double)containerCanvas.getWidth() / (double)originalSizeBitmap.getWidth();
+            double yScaleFactor = (double)containerCanvas.getHeight() / (double)originalSizeBitmap.getHeight();
+
+            return Math.max(xScaleFactor, yScaleFactor);
+        }
+
+        // TODO: Optimize the rendering as drawBitmap source rectangle can scale images automatically
```
paulcam206: >// TODO: Opti [](start = 8, length = 13)

if this is still true, we should file an issue in github for it, then refer to it here. otherwise, remove the comment :)

Issue #2850, opened 2019-05-03
[TS] RichTextBlock missing date rendering


paulcam206:
See rich text block [sample card](https://github.com/Microsoft/AdaptiveCards/blob/master/samples/v1.2/Elements/RichTextBlock.json). Typescript renderer lacks date parsing:

![image](https://user-images.githubusercontent.com/16614499/57167700-833e9100-6db3-11e9-8851-bafe45914aaf.png)

Looks correct in UWP and WPF.

dclaux:
Fixed here: https://github.com/microsoft/AdaptiveCards/pull/2905

Issue #2187, opened 2018-11-26
Card with same input id not failing as exception


paulcam206:
Fix needed for Javascript.

See #1852 for details. This has been fixed on other platforms.

See test files [Action.DuplicateIds.json](https://github.com/Microsoft/AdaptiveCards/blob/master/samples/Tests/Action.DuplicateIds.json) and [Action.NestedDuplicateIds.json](https://github.com/Microsoft/AdaptiveCards/blob/master/samples/Tests/Action.NestedDuplicateIds.json)

shalinijoshi19:
@dclaux FYI moved this to backlog

paulcam206:
@shalinijoshi19 -- are we sure we want to punt this? without a fix, you can author a card in the designer that won't render on any other platform...

dclaux:
Fix here: https://github.com/microsoft/AdaptiveCards/pull/2905

PR #2905, opened 2019-05-10
[TS] Cumulative bug fixes for RichTextBlock

dclaux:
**Description**
Fixes bugs #2912 ,  #2187 , #2850 
1. Add warning for duplicate input Ids
2. Add icon for Action.ToggleVisibility in designer
3. Renaming fontStyle property to fontType as part of #2912 

**How Verified:**


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -6579,6 +6581,27 @@ export class AdaptiveCard extends ContainerWithActions {
                 });
         }
 
+        let allInputs = this.getAllInputs();
```
RebeccaAnne: >getAllInputs [](start = 29, length = 12)

The other platforms are enforcing ID uniqueness for all elements, not just inputs. Also, there is logic for allowing ids to conflict within a given fallback hierarchy but not across it. @paulcam206

paulcam206: The full description of the algorithm with justification can be found [here](https://github.com/microsoft/AdaptiveCards/blob/master/source/shared/cpp/ObjectModel/ParseContext.cpp#L26).

---
In reply to: [283443447](https://github.com/microsoft/AdaptiveCards/pull/2905#discussion_r283443447) [](ancestors = 283443447)

dclaux: The JS renderer has always separated parse-time and post-parse validation precisely for these kinds of things.

Fallback is logic that happens at parse time. As the JS renderer parses a payload, it builds an object model representation of the parsed card; for any given element, either that element or its fallback ends up in the OM. At that point no Id validation needs to be applied. Then, once the full OM is generated, a second validation pass can be performed to validate Ids, among other things.

Splitting validation into parse-time validation and post-parse validation allows for much easier Id validation logic and plenty of other things, such as, for instance, easily verifying that a ToggleVisibility action references Ids that are actually used in the card.

paulcam206: If the JS validation is strictly post-parse, then I think we're *close* to fine. Consider this card snip, though:

```json
{
    "type": "Input.AdaptiveCard2.0.Feature",
    "fallback": {
        "type": "Input.Text",
        "id":   "uhoh"
    }
},
{
    "type": "Input.Text",
    "id":   "uhoh"
}
```

So you're using your 2.0 input element and the 2.0 designer/renderer to test. Everything looks good. The card validates fine because fallback was never invoked. You take your validated card and drop it on a 1.2 renderer and you'll hit an error because now you've got a conflict.

dclaux: That's an interesting point. How does your logic handle the below payload? From what I read it seems it will pass validation but I might be wrong:

```
{
	"type": "Container",
	"items": [
		{
			"type": "Adaptive.2.0.Feature",
			"fallback": {
				"type": "TextBlock",
				"id": "fallback",
				"text": "Nice"
			}
		},
		{
			"type": "Adaptive.2.1.Feature",
			"fallback": {
				"type": "TextBlock",
				"id": "fallback",
				"text": "Nice"
			}
		}
	]
}
```

dclaux: The point I want to make is, for the design time experience to actually make it easy for an author to know what will work and what won't, in my opinion you need to include a version selector in the designer, while in "Preview mode." As the user selects a version, the same payload is parsed and validated as it would be on a host that only supports that version. This allows for accurate warnings/errors and also to actually test fallback. And if that is what we end up doing, I am not sure we need that complex parse-time Id validation logic.

dclaux:
See latest commit with updated validation model. This is a breaking change that should be beneficial in the long run. README has been updated to explain the change.

Note this change does NOT implement the complex duplicate Id algorithm described by @paulcam206.

source/nodejs/adaptivecards/src/card-elements.ts
paulcam206: The above card in UWP visualizer:

image

It's probably worth noting that instead of "Some New Adaptive Cards Feature", it could easily be a custom element that the card author may have implemented in some but not all renderers...

dclaux: Nice! At this point, I will not be implementing this algorithm in the JS renderer because it would require slightly changing the overall way payloads are parsed, and would most likely introduce a breaking change.

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -256,9 +257,21 @@ export class CardDesignerSurface {
...
+            allErrors = allErrors.concat(this._lastParseErrors);
+
+            let validationResults = this.card.validateProperties();
+
+            for (let failure of validationResults.failures) {
+                allErrors = allErrors.concat(failure.errors);
+            }
+
+            /*
             let validationErrors = this.card.validate();
```
shalinijoshi19: Can we remove the code insead of commenting it out?

source/nodejs/adaptivecards/README.md
```diff
@@ -21,6 +21,8 @@ Please be aware of the following **breaking changes** in particular versions.
 | In version | Change description |
 |---|---|
 | **1.2** | The default `value` of an Input.Time **no longer accepts seconds**. 08:25:32 will now be treated as an invalid value and ignored; it should be replaced with 08:25. This behavior is consistent with other Adaptive Card renderers.|
+|| The `ICardObject` interface has been **REMOVED**, replaced with the `CardObject` class that both `CardElement` and `Action` extend. This change should have little to no impact on any application.|
```
shalinijoshi19: thanks! should this be in the 1.2 breaking change list as well? If so i'll make a note of it in our 1.2 release Onenote where we are tracking the breaking changes etc

source/nodejs/adaptivecards/src/enums.ts
```diff
@@ -152,7 +152,8 @@ export enum ValidationError {
     TooManyActions,
     UnknownActionType,
     UnknownElementType,
-    UnsupportedCardVersion
+    UnsupportedCardVersion,
+    DuplicateId
```
shalinijoshi19: Hmm it's too late now, but precisely for this reason (to be able to extend the enum) generally prefer to start it off wiht a default/UnsupportedCardVersion enum value rather than having it last :( ) If we'd started out with Unknown or Unsupported, we could have continued to extend the enum without worrying about breaking the flow of the type definition

shalinijoshi19: Aah n/m -- i think i may be misinterpreting the UnsupportedCardVersion to read as the enum default which it isnt ..

source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -585,25 +585,25 @@ export class FontStyleDefinition {
...
+    default: FontTypeDefinition;
+    monospace: FontTypeDefinition;
 
     constructor(obj?: any) {
-        this.default = new FontStyleDefinition();
-        this.monospace = new FontStyleDefinition("'Courier New', Courier, monospace");
+        this.default = new FontTypeDefinition();
+        this.monospace = new FontTypeDefinition("'Courier New', Courier, monospace");
 
         if (obj) {
```
shalinijoshi19: curious: does if(obj) evaluate to if (obj == null) OR if (obj === null) //strict null check?  I am assuming the latter to account for undefined here?

dclaux:
This last two commits introduces the ability to select the element that has validation errors from within the error message pane.

For instance, here we have two conflicting Ids:
image

Hover over the validation error message:
image

Now click it - the faulty element is selected and scrolled into view:
image

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
dclaux: Done

source/nodejs/adaptivecards/README.md
dclaux: It is in 1.2. I purposely not repeat 1.2 in the first column.

source/nodejs/adaptivecards/src/host-config.ts
dclaux: AFAIK it is equivalent to `if (obj !== undefined)`

The API isn't compiled with **strictNullCheck** nor **noImplicitAny**. At some point I'll make the changes, but they are non trivial (I have already looked into it.)

source/nodejs/adaptivecards/README.md
shalinijoshi19: ha no i meant the 1.2 release notes markdown file.. No matter though; I've noted it as such

---
In reply to: [284927257](https://github.com/microsoft/AdaptiveCards/pull/2905#discussion_r284927257) [](ancestors = 284927257)

source/nodejs/adaptivecards/src/host-config.ts
shalinijoshi19: tx 

---
In reply to: [284927734](https://github.com/microsoft/AdaptiveCards/pull/2905#discussion_r284927734) [](ancestors = 284927734)

PR #2918, opened 2019-05-14
Reactnative/experimental

vivekvijayakrishnan:
Below tasks are completed:-
1) Added support for HeroCard and ThumbnailCard in the experimental section.
2) Added `More Options` tab to the Visualizer.
3) Updated visualizer with experimental items


source/community/reactnative/experimental/adaptive-card-builder/AdaptiveCardBuilder.js
```diff
@@ -0,0 +1,196 @@
...
+            },
+            {
+                "type": Constants.TypeColumn,
+                "width": Constants.WidthStretch,
+                "items": []
+            }
+        ]
+    }
+
+    if (cardContent && type) {
```
pragadeeshk: Always clone JSON templates instead of manipulating on the object directly. This will help in restoring the const template if required. 

function clone(object) {
  return (JSON.parse(JSON.stringify(object)));
}

pragadeeshk: As discussed, If the template is not going to be used later, This looks fine to me

Issue #2895, opened 2019-05-10
[.NET] No warning when max actions exceeded


paulcam206:
The `Action.ToggleVisibility.json` sample card defines 6 buttons. In both the UWP and WPF visualizers, we have a 5 action cap. In UWP we throw a warning about truncating the actions. WPF appears to truncate silently.

PR #2919, opened 2019-05-14
[.NET] Add warning for max actions exceeded

almedina-ms:
Adds a warning when the amount of actions to be rendered is bigger than the max number of actions defined in the host config

Fixes https://github.com/microsoft/AdaptiveCards/issues/2895

### How verified

Visualizer was run and warning appears

### WPF

image

### HTML

image


source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCardRenderer.cs
```diff
@@ -378,6 +378,13 @@ protected static void AddActions(HtmlTag uiContainer, IList<AdaptiveAction> acti
                     }
                 }
 
+                // If the number of actions is bigger than maxActions, then log warning for it
+                if (actions.Count > actionsConfig.MaxActions)
+                {
+                    context.Warnings.Add(new AdaptiveWarning((int)AdaptiveWarning.WarningStatusCode.MaxActionsExceeded,
+                        "Some actions were not rendered due to exceeding the maximum number of actions allowed"));
```
RebeccaAnne: >"Some actions were not rendered due to exceeding the maximum number of actions allowed" [](start = 24, length = 87)

[nit] Since we use this string in two places consider defining it as a const.

Issue #2888, opened 2019-05-10
[WPF] No support for "lighter" weight in visualizer?


paulcam206:
Unclear if it's a hostconfig, implementation, or platform issue. This card:

```json
{
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "type": "AdaptiveCard",
  "version": "1.2",
  "body": [
	{
	  "type": "RichTextBlock",
	  "inlines": [
		{
		  "type": "TextRun",
		  "text": "lighter",
          "weight": "lighter"
		},
		{
		  "type": "TextRun",
		  "text": "default",
          "weight": "default"
		},
		{
		  "type": "TextRun",
		  "text": "bolder",
          "weight": "bolder"
		}
	  ]
    }
  ]
}
```

yields this output in UWP (correct):
![image](https://user-images.githubusercontent.com/16614499/57557867-792b0e00-7330-11e9-9dbd-56c90cbb7393.png)

but yields this output in WPF:
![image](https://user-images.githubusercontent.com/16614499/57557880-85af6680-7330-11e9-991c-24875306d9ac.png)

seems like `lighter` isn't working?

shalinijoshi19:
@almedina-ms / @paulcam206 does this need to be fixed for 1.2 ? It looks like a tooling bug yes?

shalinijoshi19:
@almedina-ms is already gearing this up for 1.2 PR. PUlling it back in.

PR #2923, opened 2019-05-14
[WPF] Fix lighter weight support

almedina-ms:
The issue is that in WPF only the textblocks with know font families (that have a lighter version) can render with a light weight, as we were trying to get a font family with a format similar to this "fontF1, fontF2, 'font family3'" this is not considered a valid font family and as such, a light version does not exist (also empty font families have no light weight)

Fixes https://github.com/microsoft/AdaptiveCards/issues/2888

### How verified

* IMPORTANT: To see the change you have to load a HostConfig as an empty font family has no light weight

I ran the visualizer with the sample card in the bug and the RichTextBlock.json file and these are the results

image

image


RebeccaAnne:
>IMPORTANT: To see the change you have to load a HostConfig as an empty font family has no light weight

Is this the same behavior as other platforms?

almedina-ms:
at least, in Android lighter means italic for some reason, my guess here is that the light weight doesn't exist as you have to use the light font (i.e. instead of using sans-serif you must use sans-serif-light)
i can't test uwp, but we could check what happens if an 'empty' host config is used for rendering
and iOS does work with an empty host config

---
In reply to: [492767949](https://github.com/microsoft/AdaptiveCards/pull/2923#issuecomment-492767949) [](ancestors = 492767949)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/RendererUtil.cs
```diff
@@ -11,6 +12,32 @@ namespace AdaptiveCards.Rendering.Wpf
...
+                foreach (var installedFontFamily in installedFonts.Families)
+                {
+                    if (installedFontFamily.Name == fontFamily)
+                    {
+                        return fontFamily;
+                    }
+                }
+            }
+
+            return fontFamilies[0].Trim();
```
nesalang: this doesn't make sense, if there is not matching font amonst the installed fonts, then why do we return a first font from the list? Shouldn't we return a font that's existing?

PR #2927, opened 2019-05-15
Jwoo/dot net action set fix

### LINKED ISSUE #2777: [.NET] ActionSet ShowCard
# Platform

* .NET WPF

# Version of SDK
1.2

# Issue
Json used: ActionSet.json.
When ShowCard buttons are pressed (there are two), the rendered ShowCard appears at the same position making them overlapping each other.
Capture

UWP
Capture

nesalang:
fixes #2777 

how verified:
tested with ActionSet.json and other cards that have Action.ShowCard.


source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionSetRenderer.cs
```diff
@@ -13,34 +13,40 @@ public static class AdaptiveActionSetRenderer
     {
         public static FrameworkElement Render(AdaptiveActionSet actionSet, AdaptiveRenderContext context)
         {
-            var outerActionSet = new Grid();
+            Grid outerActionSet = new Grid();
```
almedina-ms: >Grid [](start = 12, length = 4)

use var

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionSetRenderer.cs
```diff
@@ -13,34 +13,40 @@ public static class AdaptiveActionSetRenderer
...
 
             if (!context.Config.SupportsInteractivity)
                 return outerActionSet;
 
             outerActionSet.Style = context.GetStyle("Adaptive.Container");
 
             // Keep track of ContainerStyle.ForegroundColors before Container is rendered
-            var parentRenderArgs = context.RenderArgs;
-            var elementRenderArgs = new AdaptiveRenderArgs(parentRenderArgs);
+            AdaptiveRenderArgs parentRenderArgs = context.RenderArgs;
```
almedina-ms: use var here and where changed

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionSetRenderer.cs
```diff
@@ -13,34 +13,40 @@ public static class AdaptiveActionSetRenderer
...
+            AdaptiveRenderArgs elementRenderArgs = new AdaptiveRenderArgs(parentRenderArgs);
 
-            AddActions(outerActionSet, actionSet.Actions, context);
+            RenderActions(outerActionSet, actionSet.Actions, context);
 
             return outerActionSet;
         }
 
-        public static void AddActions(Grid uiContainer, IList<AdaptiveAction> actions, AdaptiveRenderContext context)
+        public static void AddRenderedActions(Grid uiContainer, IList<AdaptiveAction> actions, AdaptiveRenderContext context)
```
almedina-ms: >public static void AddRenderedActions(Grid uiContainer, IList<AdaptiveAction> actions, AdaptiveRenderContext context) [](start = 8, length = 117)

add comment for what the function does for new fuctions

you can insert them by writing this ///

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionSetRenderer.cs
```diff
@@ -138,10 +145,56 @@ public static void AddActions(Grid uiContainer, IList<AdaptiveAction> actions, A
...
         }
 
+        public static void AddActionsToRoot(AdaptiveRenderContext context)
+        {
+            // Only handle Action show cards for the main card
+            if (context.CardDepth == 1)
+            {
+                if (context.CardRoot is Grid)
+                {
+                    UIElement child = ((Grid)context.CardRoot).Children[0];
```
almedina-ms: >(Grid)context.CardRoot [](start = 39, length = 22)

cast with "as" keyword

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionSetRenderer.cs
```diff
@@ -138,10 +145,56 @@ public static void AddActions(Grid uiContainer, IList<AdaptiveAction> actions, A
             }
         }
 
+        public static void AddActionsToRoot(AdaptiveRenderContext context)
+        {
+            // Only handle Action show cards for the main card
+            if (context.CardDepth == 1)
+            {
+                if (context.CardRoot is Grid)
```
almedina-ms: >if (context.CardRoot is Grid) [](start = 16, length = 29)

you can do if (context.CardRoot is Grid variableName) and skip the casting in the line below

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionSetRenderer.cs
```diff
@@ -13,34 +13,40 @@ public static class AdaptiveActionSetRenderer
...
+            RenderActions(outerActionSet, actionSet.Actions, context);
 
             return outerActionSet;
         }
 
-        public static void AddActions(Grid uiContainer, IList<AdaptiveAction> actions, AdaptiveRenderContext context)
+        public static void AddRenderedActions(Grid uiContainer, IList<AdaptiveAction> actions, AdaptiveRenderContext context)
+        {
+            RenderActions(uiContainer, actions, context);
+            AddActionsToRoot(context);
```
almedina-ms: you could change this function call to the last lines of RenderActions to avoid calling explicitly AddActionsToRoot every time RenderActions is called

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionSetRenderer.cs
```diff
@@ -138,10 +145,56 @@ public static void AddActions(Grid uiContainer, IList<AdaptiveAction> actions, A
             }
         }
 
+        public static void AddActionsToRoot(AdaptiveRenderContext context)
```
almedina-ms: >AddActionsToRoot [](start = 27, length = 16)

this function should be renamed as it is adding the showcards or hiddencards, not the actual actions

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionSetRenderer.cs
nesalang: VS suggested to use explicit type, and I think its suggestion makes sense in this case as the change make the code clearer

---
In reply to: [284904753](https://github.com/microsoft/AdaptiveCards/pull/2927#discussion_r284904753) [](ancestors = 284904753)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionSetRenderer.cs
nesalang: it was splited by design. the order in which view for show card is added is important. render, for example, should not call AddActionsToRoot after the call to RenderActions becauase we want to add the showcard view after all buttons in their actionset are added to the root view.

---
In reply to: [284909621](https://github.com/microsoft/AdaptiveCards/pull/2927#discussion_r284909621) [](ancestors = 284909621)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveActionSetRenderer.cs
shalinijoshi19: I would typically refrain from using var for readability unless the RHS is a super complicated lambda or something that we are assigning to.. In this case though VS (public guidelines) vs one of our recommended coding guiedlines conflict and  in that case i would go with the public recommendation (ie prefer explicit type esp given that it aids in readability in this case and is not a complicated type) 

---
In reply to: [285217421](https://github.com/microsoft/AdaptiveCards/pull/2927#discussion_r285217421) [](ancestors = 285217421,284904753)

Issue #2801, opened 2019-04-29
[UWP] Background image spec changes in 1.2


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

UWP fix of #2794

shalinijoshi19:
@paulcam206, assigning this one to you.

shalinijoshi19:
@almedina might be easier if you just picked this up as well (assuming @paulcam206 hasnt already started wrking on this)

shalinijoshi19:
@paulcam206 load balancing on to you! thanks

shalinijoshi19:
@Paulcam206 i think this one is going to be a big one that we should look into if @RebeccaAnne hasnt already started looking into it; Can you pick it up please?

Issue #2697, opened 2019-04-16
[UWP] Background images on columns incorrectly increase height of the column instead of clipping


andrewleader:
# Platform

* UWP

# Version of SDK

Latest master

# Issue

I’ve been looking into an issue with the incorrect rendering of background images (in this case column backgrounds) in the UWP renderer.
I’m not really sure what’s going on, so I wanted to get your opinions on how to approach it.
 
The problem is that when setting a large image as a (column) background with the mode set to stretch/cover, then it expands the column vertically to incorporate the full resolution of the image.
This is clearly incorrect. See below.
 
For the following json:
```json
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "ColumnSet",
            "columns": [
                {
                    "type": "Column",
                    "backgroundImage": "https://www.bing.com/th?id=AMMS_925faef413dcd7b32dcd822bf3025c1b",
                    "width": 1
                },
                {
                    "type": "Column",
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "Some text"
                        }
                    ],
                    "width": 1
                }
            ]
        }
    ],
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "version": "1.1"
}
```
 
I get the following on the 1.2 designer web page (which is what I expect):

![image](https://user-images.githubusercontent.com/13246069/56225980-2e66f080-6027-11e9-80f1-cc005f70c90d.png)

 
 
But in the UWP Visualizer app I get the following:

![image](https://user-images.githubusercontent.com/13246069/56225992-332ba480-6027-11e9-891b-d2104d3d26f6.png)

 
 
I *think* that the problem is to do with the custom measure/arrange override in WholeItemsPanel, as the Image has what I believe to be the correct properties: MaxWidth & MaxHeight set to the image dimensions, Stretch set to UniformToFill.
 
If I set the Image Stetch to Uniform then the Visualizer renders as I’d expect in my case. That’s not really the correct solution though, as that would not always fill the background correctly depending on available space and image aspect ratio.
In our case the image is not behaving as you would expect it to with Stretch set to UniformToFill – the Image should be filling the available space, not defining/expanding it.
 
If I take the created XAML elements and place them inside a StackPanel instead of a WholeItemsPanel then everything renders as expected in a UWP sample app.
[...]

PR #2928, opened 2019-05-15
[UWP] Renames from Background Image spec updates

RebeccaAnne:
This includes the renames, but not the functionality changes, for #2801.

Verified test results and updated changed files.


nesalang:
cortana-skills.Tests.BackgroundImage.VerticalAndHorizontalAlignment.png shows that a background image stretches its container

RebeccaAnne:
Yeah, you're right. It's a known issue tracked by [https://github.com/microsoft/AdaptiveCards/issues/2697](https://github.com/microsoft/AdaptiveCards/issues/2697).

---
In reply to: [493519697](https://github.com/microsoft/AdaptiveCards/pull/2928#issuecomment-493519697) [](ancestors = 493519697)

source/uwp/Renderer/lib/TileControl.cpp
```diff
@@ -196,9 +196,9 @@ namespace AdaptiveNamespace
         // the container until we get the right dimensions.
         if (m_imageSize.Width != 0 && m_imageSize.Height != 0)
         {
-            switch (mode)
```
nesalang: no case for "Cover"?

RebeccaAnne: So, this is just a rename, and I'm not 100% familiar with the code here, but, i believe this is the code that handles tiling the images so cover (previously known as stretch) isn't a relevant case.  

---
In reply to: [285204074](https://github.com/microsoft/AdaptiveCards/pull/2928#discussion_r285204074) [](ancestors = 285204074)

PR #2934, opened 2019-05-16
[Android] Fix italics property in RichTextBlock

almedina-ms:
The Italic property was not implemented correctly in android, so this fixes it

Fixes [I'm fairly sure I created the issue but I can't find it]

### How verified

The visualizer app was run and this is the result (look for the third line, "Italics" string)

Screenshot_2019-05-16-15-20-34


RebeccaAnne:
In the image it looks like the "Light weight text" is also italic. Is that a bug?

almedina-ms:
yes it is, it comes from v 1.0 and I'm not sure why was it designed as it currently behaves

---
In reply to: [493515436](https://github.com/microsoft/AdaptiveCards/pull/2934#issuecomment-493515436) [](ancestors = 493515436)

RebeccaAnne:
Do we have a bug tracking? If not let's get one logged. It wouldn't block 1.2 if it's shipped in 1.0, but it would be good to sort it out at some point.

---
In reply to: [493516183](https://github.com/microsoft/AdaptiveCards/pull/2934#issuecomment-493516183) [](ancestors = 493516183,493515436)

almedina-ms:
Filed this bug [https://github.com/microsoft/AdaptiveCards/issues/2936](https://github.com/microsoft/AdaptiveCards/issues/2936) 

---
In reply to: [493527715](https://github.com/microsoft/AdaptiveCards/pull/2934#issuecomment-493527715) [](ancestors = 493527715,493516183,493515436)

Issue #2861, opened 2019-05-07
Strikethrough and italic should only be supported in Richtextblock


almedina-ms:
# Platform

What platform is your issue or question related to?

* iOS
* UWP

# Issue

Strikethrough and italic properties must only be supported in RichTextBlock; On iOS and UWP they are allowed today but shouldnt be

shalinijoshi19:
@nesalang assigning to you for fixing. Thanks

PR #2935, opened 2019-05-16
[Shared] fixes italic and strike-through being allowed in TextBlock

nesalang:
fixes issue #2861 

how verified:
TextBlock.Italic.json and TextBlock.Strikethrough.json were run, and verified that italic and strikethrough were not exercised.

RichText block json cards were also check to verify that it didn't regress.

SharedModel Unit tests were run.


shalinijoshi19:
@nesalang any reason why this is still open ? _I just went ahead and rebased with master which rekicked off CI checks. Once those complete should be good to go? _

PR #2955, opened 2019-05-20
[UWP] Fix column growing due to background image

almedina-ms:
Fixes https://github.com/microsoft/AdaptiveCards/issues/2697

The issue was fixed using TileControl to render the cover image and using the ITileBrush interface to position the image where it must be and to stretch it to fill

**There may be another issue where the background images make the card grow, this also fixes that**

### How verified 

The test app was run and the results were updated


samples/Tests/BackgroundImage.VerticalAndHorizontalAlignment.json
```diff
@@ -114,7 +114,7 @@
 				},
 				{
 					"type": "Column",
-					"width": "auto",
```
paulcam206: >th": "auto", [](start = 9, length = 12)

do we have other tests that cover the `"width": "auto"` case?

almedina-ms: yes but the result looks better with stretch width as auto overflows the card in UWP and WPF

---
In reply to: [286154178](https://github.com/microsoft/AdaptiveCards/pull/2955#discussion_r286154178) [](ancestors = 286154178)

source/uwp/Renderer/lib/TileControl.cpp
```diff
@@ -293,29 +302,47 @@ namespace AdaptiveNamespace
                 ComPtr<IUIElement> rectangleAsUIElement;
                 THROW_IF_FAILED(rectangleAsShape.As(&rectangleAsUIElement));
 
+                double originPositionX = (fillMode == BackgroundImageFillMode::Cover) ?  0.0 : (x * m_imageSize.Width) + offsetHorizontalAlignment;
+                double originPositionY = (fillMode == BackgroundImageFillMode::Cover) ? 0.0 : (y * m_imageSize.Height) + offsetVerticalAlignment;
```
paulcam206: I think I'd rather see this implemented as an if/else (same below)

samples/Tests/BackgroundImage.VerticalAndHorizontalAlignment.json
paulcam206: that's fine -- I just wanted to make sure we were getting visual verification for the `auto` case :)

---
In reply to: [286155401](https://github.com/microsoft/AdaptiveCards/pull/2955#discussion_r286155401) [](ancestors = 286155401,286154178)

PR #2520, opened 2019-03-08
[UWP] Fix for 2400 - Images not obeying parent container width constraints

barryptak:
Addresses UWP issue #2400 where images are not obeying the constraints of the parent container.
This causes problems where images within columns with set widths extend outside of the column bounds and get clipped.

Issue #2400, opened 2019-02-15
[UWP] Images ignore constraints of parent container


barryptak:
# Platform
* UWP

# Author or host
We're a host rendering cards received from Bing, etc..

# Version of SDK
1.1.2 though this applies to all UWP implementations.

# Issue
Images placed within size-constrained parents (e.g. weighted columns) ignore the parent dimensions when they have their "size" property set and will render at the requested size regardless. This is inconsistent with other renderers.

## Example Json
```json
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "ColumnSet",
            "columns": [
                {
                    "type": "Column",
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "Weight 1"
                        },
                        {
                            "type": "Image",
                            "url": "https://cdn-01.independent.ie/migration_catalog/Non-Staff/article25331906.ece/fea85/AUTOCROP/w620/No%20name",
                            "size": "Large"
                        }
                    ],
                    "width": 1
                },
                {
                    "type": "Column",
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "Weight 3"
                        }
                    ],
                    "width": 3
                }
            ]
        }
    ],
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "version": "1.0"
}
```

## Html Renderer Output
![image](https://user-images.githubusercontent.com/43549850/52850766-36301480-310c-11e9-9e17-74741e2fa5ac.png)

## UWP Renderer Output
![image](https://user-images.githubusercontent.com/43549850/52850777-434d0380-310c-11e9-8fa4-6a019b62393f.png)

## Fix
I have a fix for this (need this bug approved first before creating PR) that produces the following from the UWP renderer:
![image](https://user-images.githubusercontent.com/43549850/52850889-98891500-310c-11e9-84b7-3365642d87cf.png)

barryptak:
My fix is to modify XamlBuilder::BuildImage to replace calls to put_Width and put_Height to use put_MaxWidth and put_MaxHeight instead:

```cpp
THROW_IF_FAILED(frameworkElement->put_MaxWidth(imageSize));
THROW_IF_FAILED(frameworkElement->put_MaxHeight(imageSize));
```

If everyone is happy then I can make up a PR for this.

matthidinger:
Thanks Barry. @RebeccaAnne could you please weigh in on this one?

barryptak:
[...]

PR #2959, opened 2019-05-21
[UWP] Apply stretch property to auto sized person elements

### LINKED ISSUE #2548: [UWP] Regression: Image with explicit size and person style clipped rather than stretched
In UWP (current master), an image with explicit size and style:person ends up clipped rather than stretched (see `Tests\Image.Explicit.Size.json`):

image

paulcam206:
Fixes #2548 

The original regression occurred in 6ffc99b as a fix for #2400 (PR'd in #2520). Basically, in non-explicitly-sized-cases, we were failing to set the `stretch` property properly.

Verified with AdaptiveCardTestApp and with the original bugged card in the Visualizer. I will update this PR with a sample card covering the original bugged scenario.


nesalang:
>                 // We don't want to set a max height on the person ellipse as ellipses do not understand preserving

this comment isn't true. person image style isn't always 1:1 aspect raito, this is true in this else case because the size is set to small/medium/large, and they are squares.

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:2569 in e7ed767. [](commit_id = e7ed7677e813e8817332c6776e360a535196ccc3, deletion_comment = False)

source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -2417,11 +2417,20 @@ namespace AdaptiveNamespace
...
 
-            // Set the stretch for the ellipse - this is different to the stretch used for the image brush above.
-            // This will force the ellipse to conform to fit within the confines of its parent.
-            Stretch ellipseStretch = Stretch::Stretch_UniformToFill;
-            RETURN_IF_FAILED(ellipseAsShape->put_Stretch(ellipseStretch));
-            RETURN_IF_FAILED(backgroundEllipseAsShape->put_Stretch(ellipseStretch));
+            if (size == ABI::AdaptiveNamespace::ImageSize::None || size == ABI::AdaptiveNamespace::ImageSize::Stretch ||
+                size == ABI::AdaptiveNamespace::ImageSize::Auto || hasExplicitMeasurements)
+            {
+                THROW_IF_FAILED(ellipseAsShape->put_Stretch(imageStretch));
```
paulcam206: >THROW_IF_FAILED [](start = 16, length = 15)

should be `RETURN_IF_FAILED`

Issue #2907, opened 2019-05-11
[Shared] [1.2 API review] Rename TextBlock's fontStyle to fontType


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

Shared fix of #2906

shalinijoshi19:
@RebeccaAnne you mentioned you are working on this? OK to move this to InProgress?

Issue #2913, opened 2019-05-11
[UWP] [1.2 API review] Rename TextBlock's fontStyle to fontType


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

UWP fix of #2906

shalinijoshi19:
@RebeccaAnne is this something you are working on already (ie OK to be moved to InProgress?)

Issue #2911, opened 2019-05-11
[iOS] [1.2 API review] Rename TextBlock's fontStyle to fontType


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

iOS fix of #2906

shalinijoshi19:
@nesalang can you prioritize working on this next to unblock the corresponding .NET  PR from going in? tx

Issue #2910, opened 2019-05-11
[Android] [1.2 API review] Rename TextBlock's fontStyle to fontType


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

Android fix of #2906

PR #2960, opened 2019-05-21
[Shared][UWP][Android][iOS] FontType Spec Update

almedina-ms:
This is the base for the spec update for FontType in the shared model. It covers the basic renaming of properties so please check that everything builds correctly and then we can merge to the NET branch. When merged into the NET branch, we are going to be able to update the test results

Potentially fixes https://github.com/microsoft/AdaptiveCards/issues/2907 https://github.com/microsoft/AdaptiveCards/issues/2913 https://github.com/microsoft/AdaptiveCards/issues/2910 https://github.com/microsoft/AdaptiveCards/issues/2911


RebeccaAnne:
I had UWP and shared model in progress as well. Just merged the two of them and found a few renames that were missed here. I'll merge my changes into this branch.

Issue #2977, opened 2019-05-24
[Shared] Sentiment should be renamed to style


RebeccaAnne:
Shared model fix of #2968

Issue #2976, opened 2019-05-24
[UWP] Sentiment should be renamed to style


RebeccaAnne:
UWP fix of #2968

Issue #2974, opened 2019-05-24
[iOS] Sentiment should be renamed to style


RebeccaAnne:
iOS fix of #2968

Issue #2973, opened 2019-05-24
[Android] Sentiment should be renamed to style


RebeccaAnne:
Android fix of #2968

PR #2979, opened 2019-05-24
Rename sentiment to style

RebeccaAnne:
Rename sentiment to style. Currently includes samples, shared model, iOS, UWP, and .NET.

How verified: Run tests for all affected platforms.

Fixes: #2972 
Fixes: #2976 
Fixes: #2977
Fixes: #2974
Fixes: #2973

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/2979)


RebeccaAnne:
@nesalang Do you know why the ios build is still failing?

nesalang:
Action.ShowCard.Style.json was referred twice.

---
In reply to: [495773840](https://github.com/microsoft/AdaptiveCards/pull/2979#issuecomment-495773840) [](ancestors = 495773840)

paulcam206:
still seeing cards with `sentiment` under `source/android/mobilechatapp/src/main/assets`

there are other references to `sentiment`, but don't appear to be on the API surface, so I'm not sure how much we care...

almedina-ms:
if it's the mobilechatapp, do a Rebuild as the cards are sometimes not deleted. source/android/mobilechatapp/src/main/assets should not exist in master as it's just a copy of the samples folder so the cards can be rendered without us copying the samples explicitly to the device

---
In reply to: [496673644](https://github.com/microsoft/AdaptiveCards/pull/2979#issuecomment-496673644) [](ancestors = 496673644)

paulcam206:
ah -- that makes sense... thanks! :)

---
In reply to: [496707584](https://github.com/microsoft/AdaptiveCards/pull/2979#issuecomment-496707584) [](ancestors = 496707584,496673644)

PR #2981, opened 2019-05-24
TS: Rename action.sentiment into action.style

dclaux:
Also makes the `style` property a string for extensibility.

Hosts that want to use custom styles should declare an `.ac-pushButton.style-<custom style name>` class in their CSS file. Example:

```css
.ac-pushButton.style-myCustomStyle {
    background-color: green;
    color: white;
}
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/2981)


dclaux:
Bypassing review and unrelated checks. This is a small change with little to no risk.

Issue #2528, opened 2019-03-11
Video-control override (or adaptive sizing)


achipa:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* .NET WPF
* UWP

# Author or host

Host

# Version of SDK

Nuget uwp-v1.1.2

# Issue

By default, the Windows 10 (UWP/WPF) renderer uses the two-row Media Transport Control on Video elements:

![image](https://user-images.githubusercontent.com/822934/54121919-43fe5000-43c1-11e9-8f6a-d9491d5d7bc8.png)

However, Adaptive cards are often rendered non-fullscreen, in preset and/or limited space, so the two-line transport control can get intrusive, especially on 16:9 or wider video content, for example:

![image](https://user-images.githubusercontent.com/822934/54122408-8ecc9780-43c2-11e9-820d-f194970e92c7.png)
 
Windows *does* allow for a better presentation for use-cases like Adaptive Cards through the [isCompact](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.mediatransportcontrols.iscompact#Windows_UI_Xaml_Controls_MediaTransportControls_IsCompact) property of MediaTransportControls:

![image](https://user-images.githubusercontent.com/822934/58972842-b93caf80-877b-11e9-8d4b-d469d0cc02e3.png)

….but currently there is no way to enable this from the host side. 

Request: 
a) Expose isCompact through the hostconfig schema, and/or 
b) Enable isCompact adaptively in the renderer if the two-row control takes up more than 20% (?) of the video element.

andrewleader:
Excellent feedback! I'd say we should always use the compact style by default, since cards by nature are usually small, and then we could always add the option to enable the full size controls (which maybe we tell hosts to accomplish by extending the media renderer or something, rather than host config, since each platform might have different media control options).

achipa:
@andrewleader Certainly. While we are at the defaults - we might have another niggle that would be worth setting differently by default in the context of Adaptive Cards:

@Windows.UI.Xaml.Controls.MediaTransportControls.IsZoomButtonVisible?text=IsZoomButtonVisible (**true**)

As the media will always be rendered *within* the (normally small) card, this should be off unless there is a specific width/height setting is specified (as normally you'll not want to letterbox a video that is already of small dimensions), or just be off altogether when displayed as part of a card. As is, it's a (from a user perspective) a dead button only cluttering the card (gets extra-painful on portrait mode videos).

[...]

PR #2987, opened 2019-05-25
Use compact transport controls

achipa:
https://github.com/microsoft/AdaptiveCards/issues/2528

Uses compact transport controls for UWP rendering, which is better suited for small, card-oriented rendering sizes.

Don't show zoom button in UWP rendering, as it makes no sense in Adaptive Cards - there is no way to force an aspect ratio, so a zoom button is effectively a no-op within a card, just using up scant screen real-estate.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/2987)


paulcam206:
@shalinijoshi19 / @RebeccaAnne - do we want this for 1.2?

andrewleader:
If dev team feels comfortable with this for 1.2, I think that's the right decision and a really easy win to help our partner, HP, who needs this!

shalinijoshi19:
> @shalinijoshi19 Shalini Joshi (WINDOWS) FTE / @RebeccaAnne Rebecca Muraira FTE - do we want this for 1.2?

Thanks @paulcam206 ; Indeed sounds like an improvement here. If there are no regressions etc and since it's already been reviewed i dont mind this being part of the 1.2 payload (though it will need to be manually merged in @paulcam206 / @RebeccaAnne  since we've now forked the 1.2 branch off of master)

PR #2988, opened 2019-05-28
Reactnative/fallback - Model layer and fallback feature

muralikrish91:
*Model layer added for accessing the payload
* Fallback feature for elements added

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/2988)


source/community/reactnative/src/models/container-model.js
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+export class ColumnSetModel extends BaseContainerModel {
+    constructor(parent, payload) {
+        super(parent, payload);
+        this.type = ElementType.ColumnSet;
+        this.children = [];
+        if (payload.columns) {
+            payload.columns.forEach((item) => {
```
pragadeeshk: Can't we user ModelFactory.createGroup here too?

source/community/reactnative/src/models/container-model.js
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+export class FactSetModel extends BaseContainerModel {
+    constructor(parent, payload) {
+        super(parent, payload);
+        this.type = ElementType.FactSet;
+        this.children = [];
+        if (payload.facts) {
+            payload.facts.forEach((item) => {
```
pragadeeshk: Can't we user ModelFactory.createGroup here too?

source/community/reactnative/src/models/container-model.js
```diff
@@ -0,0 +1,155 @@
...
+
+export class ImageSetModel extends BaseContainerModel {
+    constructor(parent, payload) {
+        super(parent, payload);
+        this.type = ElementType.ImageSet;
+        this.children = [];
+        this.imageSize = payload.imageSize;
+        if (payload.images) {
+            payload.images.forEach((item) => {
+                let image = new ImageModel(this, item);
```
pragadeeshk: Can't we user ModelFactory.createGroup here too?

source/community/reactnative/src/models/model-factory.js
```diff
@@ -0,0 +1,82 @@
+import * as Models from './index'
+import { ElementType } from '../utils/enums'
+import * as Utils from '../utils/util'
+
+export class ModelFactory {
+    static createElement(parent, payload) {
```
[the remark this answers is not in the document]
muralikrish91: Yeah sure Praga, will make the changes.

source/community/reactnative/src/models/container-model.js
muralikrish91: ModelFactory.createGroup makes sense only when the groups contain a varied set of elements like in case of a container. For ColumnSet, FactSet and ImageSet, the items is going to be of a single type.This would avoid the unnecessary condition check and make the expected type clear.

source/community/reactnative/src/models/container-model.js
muralikrish91: Same reason as above.

source/community/reactnative/src/models/container-model.js
muralikrish91: Same reason as above.

source/community/reactnative/src/models/container-model.js
pragadeeshk: Got it.. Makes sense now.

PR #3017, opened 2019-06-03
Update README.md badges to point to master branch

paulcam206:
* Old badges didn't specify a branch. Some jobs pull double duty as daily builds and PR validation, so something legitimately failing in PR would show the build as failing even if master builds fine.
* Update badge url to refer to `azure-devops` instead of `vso` since the VSO designation is deprecated.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3017)


paulcam206:
@matthidinger -- can you take a look?

Issue #2986, opened 2019-05-24
[UWP] Improve host configs used by test app


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* UWP

# Issue

Change #2964 reduced our host config burden in the test app to two host configs. This issue tracks ensuring that the host configs are representative to ensure we cover all relevant cases.

Issue #2915, opened 2019-05-13
[UWP][TestApp] Test app overrides visibility leading to bad results


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* UWP

# Issue

The test app manually sets things to visible that aren't, resulting in bad results in the test app. This functionality was added to allow the test app to validate hidden show cards, but should be re-evaluated in light of the new visibility features.

PR #3018, opened 2019-06-03
[UWP][TestApp] Update host configs used by the UWP test app

RebeccaAnne:
Changes include:
- Update test app, sample host configs, and expected test results to support three host configurations:
(1) No host config, rendering against the default values 
(2) Fixed size card with interactivity turned off but no other host config changes
(3) A test host config crafted to validate alternatives to the default
- Removed the code that excluded the "WeatherLarge" card. This card renders consistently for me, and i will be working on reliability in a future change to ensure that all cards do.
- Removed code that manually shows hidden grids. This was designed to allow us to test show cards, but causes issues now that we have other hidden elements in cards. We'll need to find another way to test show cards, ideally as part of a holistic approach to testing interactivity.

How verified: Ran test app

Fixes: #2986
Fixes: #2915

Note: CodeFlow will be overwhelmed by the test files changes. The test file changes are in commit 4, so take a look at commit 3 to see everything except the test file changes.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3018)


source/shared/cpp/ObjectModel/HostConfig.h
```diff
@@ -132,9 +132,9 @@ namespace AdaptiveSharedNamespace
 
     struct ImageSizesConfig
     {
-        unsigned int smallSize = 80;
-        unsigned int mediumSize = 120;
-        unsigned int largeSize = 180;
+        unsigned int smallSize = 40;
```
paulcam206: >unsigned int smallSize = 40; [](start = 8, length = 28)

was it a bug that these were the old values? should this be considered a breaking change for hosts that don't specify image sizes in their config?

RebeccaAnne: I shouldn't have made this change. I did it locally because the sizes that were in here didn't "look good" to me, but the existing values are the documented defaults and i should have reverted before sending the PR. I'll undo this and update the affected test files.

---
In reply to: [290044941](https://github.com/microsoft/AdaptiveCards/pull/3018#discussion_r290044941) [](ancestors = 290044941)

paulcam206: hooray for code review! :)

---
In reply to: [290054705](https://github.com/microsoft/AdaptiveCards/pull/3018#discussion_r290054705) [](ancestors = 290054705,290044941)

Issue #3032, opened 2019-06-06
Adaptive Cards v1.2.0 throwing NullPointerException while rendering


desaibijon:
# Platform

Android

# Author or host

Host

# Version of SDK

io.adaptivecards:adaptivecards-android:1.2.0

# Issue

I just updated the adaptive cards version on my native app from _v1.1.0_ to _v1.2.0_. After making the necessary changes around _AdaptiveCardRenderer.render_ and _BaseCardElementRenderer.render_, when I try to run the app, I get a **NullPointerException** in the **AdaptiveCardRenderer.getInstance.render()** method. It complains about **AdaptiveCards::FeatureRegistration** const & reference being null. Below is the snippet of the log:

`java.lang.RuntimeException: Unable to start activity ComponentInfo{com.dynamicsignal.adaptivecards/com.dynamicsignal.adaptivecards.AdaptiveCardActivity}: java.lang.NullPointerException: AdaptiveCards::FeatureRegistration const & reference is null
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2678)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2743)
        at android.app.ActivityThread.-wrap12(ActivityThread.java)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1490)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6165)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:888)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:778)
     Caused by: java.lang.NullPointerException: AdaptiveCards::FeatureRegistration const & reference is null
        at io.adaptivecards.objectmodel.AdaptiveCardObjectModelJNI.BaseElement_MeetsRequirements(Native Method)
        at io.adaptivecards.objectmodel.BaseElement.MeetsRequirements(BaseElement.java:125)
        at io.adaptivecards.renderer.registration.CardRendererRegistration.render(CardRendererRegistration.java:280)
        at io.adaptivecards.renderer.AdaptiveCardRenderer.internalRender(AdaptiveCardRenderer.java:156)
[...]

PR #3038, opened 2019-06-07
[Android] Fix missing feature registration issue

almedina-ms:
Fixes https://github.com/microsoft/AdaptiveCards/issues/3032

Adds a validation check to verify that a FeatureRegistration object was added before checking if the element meetsRequirements

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3038)


shalinijoshi19:
@almedina-ms any reason why this is still open?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/registration/CardRendererRegistration.java
```diff
@@ -277,7 +277,7 @@ else if(tag instanceof AdaptiveCard)
                     throw new AdaptiveFallbackException(cardElement);
                 }
 
-                if (!cardElement.MeetsRequirements(featureRegistration))
+                if ((featureRegistration != null) && (!cardElement.MeetsRequirements(featureRegistration)))
```
shalinijoshi19: Can we add a unit test for this in the suite you are adding?

shalinijoshi19:
>                                     throw new InternalError("Unable to convert BaseElement to BaseCardElement object model.");

Do we have documentation around what specific errors maybe thrown from specific functions that bubble up to the caller or otherwise in general say what InternalError implies and what cases it can be thrown in ? if not can you file a doc task to address this? thanks

---
Refers to: source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/registration/CardRendererRegistration.java:306 in 95486c2. [](commit_id = 95486c2f8112e019c101990632fa9aa41c476a8a, deletion_comment = False)

PR #3046, opened 2019-06-10
[UWP][TestApp] Usability Improvements

### LINKED ISSUE #3045: [UWP][TestApp] Update the test app to better communicate what caused the card to fail
# Platform

* UWP

# Issue

Today the "Failed/source changed" results category doesn't specify whether it was the image or the json round trip that failed. This issue tracks making that clearer, and possibly additional refactoring of the implementation/UI for tracking test results.

RebeccaAnne:
Several improvements to the UWP test app.

- More information in test results. Specifically, calling out the results of original comparison, json round tripping and image comparison. In addition, added a colored border to call out whether the image comparison failed:

image

- Ability to accept originals, jsons, or images independently, both per card and for the entire list:

image
image

- As a result of those changes, we no longer need all of the combinations of tabs on the side. Simplified that list:

image

- To support these changes, changed the TestStatus from an enum to class, and updated the code accordingly.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3046)

Fixes #3045

### DIFF
--- a/source/uwp/AdaptiveCardTestApp/Views/TestResultView.xaml.cs
+++ b/source/uwp/AdaptiveCardTestApp/Views/TestResultView.xaml.cs
@@ -25,47 +25,46 @@ namespace AdaptiveCardTestApp.Views
             {
                 _currModel = model;
 
+                ButtonUpdateImage.Visibility = Visibility.Collapsed;
+                ButtonUpdateJson.Visibility = Visibility.Collapsed;
+                ButtonUpdateOriginals.Visibility = Visibility.Collapsed;
+                ButtonUpdateAll.Visibility = Visibility.Collapsed;
+
                 if (model.Status.NewCard)
                 {
                     // For a new card, only let them update "all"
-                    ButtonUpdateImage.Visibility = Visibility.Collapsed;
-                    ButtonUpdateJson.Visibility = Visibility.Collapsed;
-                    ButtonUpdateOriginals.Visibility = Visibility.Collapsed;
                     ButtonUpdateAll.Visibility = Visibility.Visible;
 
                 }
-                if (model.Status.MatchedViaError)
+                else if (model.Status.MatchedViaError)
                 {
                     // If they matched via error nothing to update except possibly the original
-                    ButtonUpdateImage.Visibility = Visibility.Collapsed;
-                    ButtonUpdateJson.Visibility = Visibility.Collapsed;
-                    ButtonUpdateAll.Visibility = Visibility.Collapsed;
                     ButtonUpdateOriginals.Visibility = model.Status.OriginalMatched ? Visibility.Collapsed : Visibility.Visible;
                 }
                 else
                 {
                     // Provide the appropriate buttons to update as needed
-                    uint buttonsShown = 3;
-                    if (model.Status.ImageMatched)
+                    uint buttonsShown = 0;
+                    if (!model.Status.ImageMatched)
                     {
-                        ButtonUpdateImage.Visibility = Visibility.Collapsed;
-                        buttonsShown--;
+                        ButtonUpdateImage.Visibility = Visibility.Visible;
+                        buttonsShown++;
                     }
-                    if (model.Status.JsonRoundTripMatched)
+                    if (!model.Status.JsonRoundTripMatched)
                     {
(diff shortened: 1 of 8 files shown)


source/uwp/UWPTestLibrary/TestResultViewModel.cs
```diff
@@ -317,41 +446,70 @@ private static string GetStoredSourceFileName(string name, string hash)
...
-        /// Visual did NOT match, but source files (payload and host config) changed, so changes are possibly expected
-        /// </summary>
-        FailedButSourceWasChanged,
-
-        /// <summary>
-        /// New test, no existing Expected
-        /// </summary>
-        New
+        // The error results match (either both pass or both fail with the same result)
+        public bool MatchedViaError { get; set; } = false;
```
andrewleader: NIT: Use the 3 slashes comments like `/// <summary>`, those will be picked up by IntelliSense so that when you're accessing these properties elsewhere, their summaries appear :)

source/uwp/AdaptiveCardTestApp/Views/TestResultView.xaml.cs
```diff
@@ -25,30 +25,47 @@ private void TestResultView_DataContextChanged(FrameworkElement sender, DataCont
...
 
-                if (model.Status == TestStatus.Passed)
+                if (model.Status.NewCard)
                 {
-                    ButtonSaveAsExpected.Visibility = Visibility.Collapsed;
+                    // For a new card, only let them update "all"
+                    ButtonUpdateImage.Visibility = Visibility.Collapsed;
+                    ButtonUpdateJson.Visibility = Visibility.Collapsed;
+                    ButtonUpdateOriginals.Visibility = Visibility.Collapsed;
+                    ButtonUpdateAll.Visibility = Visibility.Visible;
```
andrewleader: One thing I like to do for UIs like this is first set all the buttons to Collapsed, and then inside each logic branch, all you have to worry about is showing buttons, rather than both showing and hiding. Makes it a little cleaner (and reduces the chance of forgetting to collapse a button). But just a small nit :)

source/uwp/AdaptiveCardTestApp/Views/TestResultView.xaml.cs
```diff
@@ -25,30 +25,47 @@ private void TestResultView_DataContextChanged(FrameworkElement sender, DataCont
...
+
+                }
+                if (model.Status.MatchedViaError)
```
andrewleader: >if (model.Status.MatchedViaError) [](start = 16, length = 33)

Should this be an `else if`? I'm not sure, but seems odd having the two if statements next to each other?

source/uwp/AdaptiveCardTestApp/Views/TestResultView.xaml.cs
```diff
@@ -60,8 +77,30 @@ private void TestResultView_DataContextChanged(FrameworkElement sender, DataCont
 
         private async void ButtonSaveAsExpected_Click(object sender, RoutedEventArgs e)
         {
+            bool updateOriginals = false;
+            bool updateImage = false;
+            bool updateJson = false;
+
+            Button buttonSender = sender as Button;
+            if (buttonSender.Name == "ButtonUpdateOriginals")
```
andrewleader: >if (buttonSender.Name == "ButtonUpdateOriginals") [](start = 12, length = 49)

Instead of doing string comparison, you should be able to do the following... compare the reference

```
if (sender == ButtonUpdateOriginals)
```

That way it'll also be compiled and checked in case we ever rename the buttons :) The following could also work if the reference comparison provides a warning

```
if (buttonSender.Name == ButtonUpdateOriginals.Name)
```

source/uwp/AdaptiveCardTestApp/Views/TestResultView.xaml.cs
RebeccaAnne: Sure

---
In reply to: [292200213](https://github.com/microsoft/AdaptiveCards/pull/3046#discussion_r292200213) [](ancestors = 292200213)

Issue #3064, opened 2019-06-14
[Android][Shared Model] Richtextblock crashes when Serializing


almedina-ms:
# Platform

* Android
* Shared

# Version of SDK

master

# Issue

Following the debugger I got to TextElementProperties line 54 

under function `Json::Value TextElementProperties::SerializeToJsonValue(Json::Value& root) const`

`root[AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::Text)] = GetText();`

when this line is executed the process crashes throwing a "LogicError" from json.cpp and the message is 

"in Json::Value::resolveReference(key, end): requires objectValue"

PR #3067, opened 2019-06-14
[Android] Add android unit tests for all elements

almedina-ms:
Adds tests for element properties for all read-only elements up to v1.2 for Android except for RichTextBlock as the Serialize method crashes (see https://github.com/microsoft/AdaptiveCards/issues/3064).

This set of tests tries to avoid regresions in the shared model

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3067)


source/android/adaptivecards/src/androidTest/java/io/adaptivecards/objectmodel/TextBlockPropertiesTest.java
```diff
@@ -0,0 +1,332 @@
...
+            castedElement = (T) baseElement;
+        }
+        else if ((castedElement = type.dynamic_cast(baseElement)) == null)
+        {
+            throw new InternalError("Unable to convert BaseCardElement to TextBlock object model.");
+        }
+
+        return castedElement;
+    }
+    */
```
almedina-ms: remove comments

almedina-ms:
Take a look at X lang for alternatives to SWIG

PR #3068, opened 2019-06-14
Spec tooling and specs

andrewleader:
Added tooling so that we can generate specs from our schema, so that our spec'ing process starts with updating the schema first, leading to less errors and mistakes in the end :)

In `schemas\src`, add your new properties/elements to the schema

UpdateSchema

### 3. Generate the updated spec

If you followed the setup instructions, simply save the file and the spec will be updated! Add any necessary rendering details to the spec.

UpdatedSpec

## Review spec

Submit a pull request for your updated spec/schema. The proposed schema in the proposal becomes outdated and should be replaced with a link to the PR.

image

I'll share more details about this at Monday's meeting. Lots of details about the new proposed spec'ing process are captured here: https://github.com/microsoft/AdaptiveCards/wiki/Project-Management

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3068)


shalinijoshi19:
this is AWESOME! thanks @andrewleader

.vscode/settings.json
```diff
@@ -2,5 +2,22 @@
...
+	"runOnSave.statusMessageTimeout": 18000,
+	"runOnSave.commands": [
+		{
+			"match": ".*schemas\\\\src\\\\.*\\.json$",
+			"command": "powershell cd ${workspaceFolder}\\source\\nodejs\\spec-generator; .\\generate.cmd",
+			"runningStatusMessage": "Updating specs...",
+			"finishStatusMessage": "Updated specs!"
+		}
+	]
+	// "emeraldwalk.runonsave": {
```
matthidinger: Do you still want this commentd out here?

source/nodejs/spec-generator/generate.cmd
```diff
@@ -0,0 +1,16 @@
+IF EXIST "dist\spec-generator-script.0.5.0.0.min.js" (
```
matthidinger: So I'm wondering if there's any way to have node run this logic? If we write this as a .cmd then our Mac and Linux friends will be unable to generate specs. npm has a way of declaring "bin" functions which are designed to be executed like command line scripts, except x-plat. Take a look here for an example: 

https://github.com/matthidinger/marked-schema/blob/master/bin/marked-schema

.vscode/settings.json
andrewleader: Whoops yeah I'll remove that. As for the cmd/script for Mac/Linux, that can be addressed in the future, forward progress is needed!

source/nodejs/spec-generator/generate.cmd
andrewleader: That's a great future improvement, but we need to make progress now. That marked schema example doesn't **build** the node js project. That's the key commands in the current script. The current script checks if a pre-built version of the script exists and if not builds it on-demand.

I'm sure it's possible to make that work, but we don't have time right now (or I don't).

Issue #3104, opened 2019-06-25
iOS child of issue 3099


nesalang:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* iOS

# Version of SDK

1.2

What version are you using? Ex: NuGet 1.0.2, or latest master, etc...

# Issue

This is a child of #3099

PR #3105, opened 2019-06-25
fixes issue #3104

nesalang:
Fixes issue #3104 

there were two issues;

1. logic for disabling playIcon was properly adjusted after resource resolver changes.
2. the new resource resolver changes require new instances of UIImageViews, as UIImageView is a container of UIImage, and can't be duped. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3105)

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRMediaRenderer.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRMediaRenderer.mm
@@ -50,10 +50,10 @@
     ACRContentHoldingUIView *contentholdingview = nil;
 
     // if poster is available, restrict the image size to the width of superview, and adjust the height accordingly
-    if(img) {
+    if (img) {
         view = [[UIImageView alloc] initWithImage:img];
 
-        if(img.size.width > 0) {
+        if (img.size.width > 0) {
             heightToWidthRatio = img.size.height / img.size.width;
         }
         contentholdingview = [[ACRContentHoldingUIView alloc] init];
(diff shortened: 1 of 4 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRMediaRenderer.mm
```diff
@@ -85,44 +85,44 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
...
     UIImageView *playIconImageView = nil;
-    BOOL drawDefaultPlayIcon = YES;
+    BOOL hideDefaultPlayIcon = NO;
 
-    if(!playIconImage) {
-        playIconImageView  = [rootView getImageView:@"playIconImage"];
-    }else {
+    if (!playIconImage) {
+        NSNumber *number = [NSNumber numberWithUnsignedLongLong:(unsigned long long)elem.get()];
+        NSString *key = [NSString stringWithFormat:@"%@_%@", [number stringValue], @"playIcon" ];
```
bviglietta: >stringWithFormat:@"%@_%@", [number stringValue], @"playIcon" [](start = 34, length = 60)

It isn't necessary to create an NSNumber here. You can just pass the number right into the string formatter: `... stringWithFormat:@"%llu_%@", (unsigned long long)elem.get(), ...`

You could actually also eliminate the other string format entry altogether if you wanted: `... stringWithFormat:@"%llu_playIcon", (unsigned long long)elem.get()]` Although if you prefer keeping the "playIcon" part separate for clarity, I have no objection.

nesalang: thanks!

---
In reply to: [299690929](https://github.com/microsoft/AdaptiveCards/pull/3105#discussion_r299690929) [](ancestors = 299690929)

PR #3106, opened 2019-06-25
Added adaptivecards-fabric

dfisenko:
this package contains a set of input elements extensions to override regular adaptivecards elements with Office Fabric UI components.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3106)


source/nodejs/adaptivecards-fabric/src/Components/Inputs/InputTimeFabric.tsx
```diff
@@ -17,8 +17,8 @@ export class InputTimeFabric extends Shared.ReactInputElement {
         this.id = AC.getStringValue(json.id);
         this.value = json.value;
         this.placeholder = AC.getStringValue(json.placeholder);
-        this.min = AC.getStringValue(json.min);
-        this.max = AC.getStringValue(json.max);
+        this.min = json.min ? AC.getStringValue(json.min) : undefined;
```
dclaux: getStringValue's implementation is
```typescript
export function getStringValue(obj: any, defaultValue: string = undefined): string {
    return obj ? obj.toString() : defaultValue;
}
```
So you shouldn't have to do the test yourself.

dfisenko: got it, will update

Issue #2712, opened 2019-04-17
[UWP] Don't re-use custom AdaptiveImage renderer for background images


andrewleader:
Using the AdaptiveImage renderer to render background images causes strange things. As a host, if I override rendering of AdaptiveImage, I also override rendering of background images, which probably wasn't my intention.

![image](https://user-images.githubusercontent.com/13246069/56326426-ce597280-612a-11e9-8e88-d67d9d61d797.png)

At the working group meeting today (with Matt, David, Andrew, Bill, Alberto, Shalini, Becky, and Paul), consensus was that we shouldn't be doing this.

When a need arises for overriding background image rendering, we'll solve that with a separate API

shalinijoshi19:
@almedina-ms assigning to you for load balancing; Also @RebeccaAnne / @andrewleader moving this out of 1.2 to make space for some of the other breaking/pressing bugs

shalinijoshi19:
The work that this bug needs in XamlBuilder is being tracked by issue #3026

shalinijoshi19:
@paulcam206 assigning to you; Do you just want to use this as a test for the rework you are doing as part of #2534 now?

RebeccaAnne:
See also #1605

PR #3113, opened 2019-06-27
[UWP] Use default Image renderer for background

almedina-ms:
Changes the ApplyBackgroundToRoot implementation to always use the default AdaptiveImageRenderer

This fixes: https://github.com/microsoft/AdaptiveCards/issues/2712

### Test 1

The test app was run and no card failed to render correctly

### Test 2

I added a custom mock Image renderer which only draws a cat image for all image elements, so the ActivityUpdate card looks like this  

image

The cards AdaptiveCard.BackgroundImage.FillMode.json and Column.BackgroundImage.json yielded this results

image

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3113)

### DIFF
--- a/source/uwp/Renderer/lib/XamlBuilder.cpp
+++ b/source/uwp/Renderer/lib/XamlBuilder.cpp
@@ -11,6 +11,7 @@
 #include "AdaptiveFeatureRegistration.h"
 #include "AdaptiveHostConfig.h"
 #include "AdaptiveImage.h"
+#include "AdaptiveImageRenderer.h"
 #include "AdaptiveRenderArgs.h"
 #include "AdaptiveShowCardAction.h"
 #include "AdaptiveTextRun.h"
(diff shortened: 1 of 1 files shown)


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -487,9 +488,10 @@ namespace AdaptiveNamespace
         ComPtr<IAdaptiveElementRendererRegistration> elementRenderers;
         THROW_IF_FAILED(renderContext->get_ElementRenderers(&elementRenderers));
 
+        // For background image we use the default image renderer instead of custom image renderers
         ComPtr<IAdaptiveElementRenderer> elementRenderer;
-        THROW_IF_FAILED(elementRenderers->Get(HStringReference(L"Image").Get(), &elementRenderer));
-
+        THROW_IF_FAILED(MakeAndInitialize<AdaptiveNamespace::AdaptiveImageRenderer>(&elementRenderer));
```
RebeccaAnne: >THROW_IF_FAILED(MakeAndInitialize<AdaptiveNamespace::AdaptiveImageRenderer>(&elementRenderer)); [](start = 8, length = 95)

The problem with this is that the image renderer, unlike the other renderers, maintains a member which is the current xamlbuilder. The XamlBuilder, in turn, has members tracking things related to tracking the image loading for the Async case. (m_enableXamlImageHandling, m_listeners, m_imageLoadTracker, etc.). These won't be set properly on a new instance, and as a result i think the async call won't wait on completion of the images rendered here. I don't have a solution here, and that image async code could probably stand to be revisited, but we probably shouldn't make this change without making sure we understand the impact.

almedina-ms:
Blocked by #4593

PR #3123, opened 2019-06-30
Switch to using textContent over innerText

rcasto:
I'm rendering cards in a Node environment, using a simulated browser context via [jsdom](https://github.com/jsdom/jsdom) and the innerText property seems to be an unsupported, due to the need for layout:
https://github.com/jsdom/jsdom/issues/1245

There is a way to workaround it if innerText is used for a specific reason, demonstrated at the link below:
https://runkit.com/rcasto/5d18030b25184700154cb642

It may just generally be more performant to use textContent instead of innerText though, if possible:
https://jsperf.com/innerhtml-vs-innertext/17

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3123)


paulcam206:
I find nothing wrong with this change. My only concern would be support in older browsers. @matthidinger / @dclaux -- how far back are we targeting?

dclaux:
The renderer makes use of the date input type, which isn't supported in any version of IE. So I think this change is perfectly acceptable, especially considering the considerable perf improvement it brings.

Thanks!

paulcam206:
Looks like this one needs to be redone -- we've since moved to using `main` as our main branch, and the code has skewed quite a bit since then. So sorry @rcasto for not getting this merged in. Please feel free to resubmit and I'll make sure the change gets in this time!

PR #3126, opened 2019-07-01
removed play icon from hostconfig

nesalang:
fixes ios ci issue
the ios unit test pull down media icon from https://www.shareicon.net/ for each test.
lately, we seemed to hit their rate limit, and download speed from unit tests was severely limited.
removed the url from unit test host config.
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3126)

### DIFF
--- a/source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
+++ b/source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
@@ -246,7 +246,6 @@
 		"spacing": 5
 	},
 	"media": {
-		"playButton" :"https://www.shareicon.net/download/64x64/2017/04/08/883361_media_512x512.png",
 		"allowInlinePlayback" : true
 	}
 }


source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
```diff
@@ -246,7 +246,6 @@
 		"spacing": 5
 	},
 	"media": {
-		"playButton" :"https://www.shareicon.net/download/64x64/2017/04/08/883361_media_512x512.png",
```
paulcam206: should we perhaps host a play button icon on adaptivecards.io that you can reference? seems like it's worth having a custom `playButton` in our tests...

shalinijoshi19: Alternatively can we get this image added as part of the test artifacts/resources that are referenced locally and dont require an external call over the wire? 

---
In reply to: [299165267](https://github.com/microsoft/AdaptiveCards/pull/3126#discussion_r299165267) [](ancestors = 299165267)

paulcam206: only concern would be if there's a runtime difference between pulling from a URL vs. local disk :)

---
In reply to: [299171295](https://github.com/microsoft/AdaptiveCards/pull/3126#discussion_r299171295) [](ancestors = 299171295,299165267)

nesalang: both approaches look good to me. i can add resource resolver to make references to a local bundle. 

I think that Paul's concern is legitimate one. I think that there is a value in testing imaging loading through network as they are the most frequently used mode of transport.  

I will create an issue to track this issue and for discussion.
[https://github.com/microsoft/AdaptiveCards/issues/3127](https://github.com/microsoft/AdaptiveCards/issues/3127)


---
In reply to: [299206526](https://github.com/microsoft/AdaptiveCards/pull/3126#discussion_r299206526) [](ancestors = 299206526,299171295,299165267)

PR #3129, opened 2019-07-01
[TS] Turn on built in input validation

RebeccaAnne:
Turned on built in input validation on the javascript renderer. 

How Verified: Built visualizer with the change and confirmed that cards with the new input validation properties work as expected.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3129)


shalinijoshi19:
Are there changes and additional flags you found you needed to enable for the visualizer? or is the additional flag that David was talking about something that applies only to the Designer?

RebeccaAnne:
This was sufficient to turn it on in the renderer, so it worked in the visualizer, but i don't believe this gets us designer support.

---
In reply to: [507812063](https://github.com/microsoft/AdaptiveCards/pull/3129#issuecomment-507812063) [](ancestors = 507812063)

PR #3133, opened 2019-07-02
Publish fabric

matthidinger:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3133)


source/nodejs/adaptivecards-fabric/README.md
```diff
@@ -20,6 +23,8 @@ This package extends standard AdaptiveCards package replacing regular input cont
 npm install adaptivecards-fabric
 ```
 
+**NOTE**: you must also install the [`adaptivecards`](https://npmjs.com/package/adaptivecards) library if not already in your project.
```
dfisenko: all peer dependencies are required. 
- office-ui-fabric-react
- react
- react-dom

shalinijoshi19:
Hey @Matt Hidinger/@Denis Fisenko, what is the release story we are looking for for this specific package going forward? Do we publish it everytime we do a new AdaptiveCards release? Do you own defining the release cycles for this package over on thefabric UI side? It doesnt look like it needs to be coupled or needed to be owned by us on AdaptiveCards yes? Is this  a larger conversation we should have separately? Looking forward to playing more with this! Tx

shalinijoshi19:

source/nodejs/adaptivecards-fabric/src/Components/Inputs/InputToggleFabric.tsx
```diff
@@ -14,6 +14,7 @@ export class InputToggleFabric extends Shared.ReactInputElement {
     public parse = (json: any, errors?: AC.IValidationError[]) => {
         this.id = AC.getStringValue(json.id);
         this.valueInternal = json.value;
+        this.defaultValue = this.valueInternal;
```
dclaux: There are a number of things that need to be changed - I am sorry I didn't spot them earlier, I reviewed this too fast:
- Your custom inputs extend the AC.Input class. Your `parse` method **MUST** first call `super.parse()` which already parses `id` and `defaultValue`. Please don't reimplement these behaviors.
- In your `internalRender` implementation, you have to initialize the value of your actual Fabric input control to the `defaultValue` property
- Your current internalRender implementation does this: `this.inputControlContainerElement.style.display = "block"` **Don't do that**; you mustn't change the layout of any element you do not own (e.g create yourself) otherwise you will interfere with the layout of the card.
- Your inputs _should_ also implement `toJSON()` (please look at how it's implemented for one of the built-in inputs). I say they should because technically toJSON is only used by the AC designer which doesn't use the Fabric extensions.

shalinijoshi19: @dclaux, trying to understand a bit more -- are these guidelines in place for some of these do's/don't's you point out? I am assuming most of these points are more guidelines and wouldn't actually cause functionality/runtime errors? Or would they? (trying to see if we need CI's in place or some other automation that can catch some of these violations eventually)

dfisenko: imo, would be really nice to have these requirements documented somewhere. I think only **style.display** potentially can cause some rendering issues, others are mostly guidelines.

dclaux: @shalinijoshi19 these are more than guidelines, this is how to correctly extend the renderer. When we add new features, such as validation on inputs, or new properties in general, we don't want nor can we update everyone's custom input implementation. So the renderer provides the core behavior and extensions just need to implement the value add they're bringing, in this case a better looking, more functional control.

Extending the renderer should be better documented. It is unfortunate that it is not at this point, but the least I can do right now is point folks in the right direction.

dclaux: @dfisenko the general rule is do not modify any element not created by you. That is especially true when it comes to layout, but it is also true for styles. You can however set the layout and styles of your custom element whichever way you desire/makes sense.

The main role of the renderer is to take care of layout and styles as defined in the Host Configuration. Overriding the styles of any element that was created under you can result in the renderer not rendering properly. While that is ok for as long as the extensions are yours and your are the only one using them, in this case the extensions are meant to be used by other developers.

I hope that makes sense.

Technically the proper way to extend the renderer is already documented here: https://docs.microsoft.com/en-us/adaptive-cards/sdk/rendering-cards/javascript/extensibility
That being said, this documentation could probably be more extensive.

source/nodejs/adaptivecards-fabric/src/utils/shared.tsx
```diff
@@ -32,20 +32,25 @@ export abstract class ReactCardElementContainer extends AC.Container {
...
+        element.style.width = "100%";
+        return element;
     }
 
     protected handleChange = (event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) => {
-        this.valueInternal = newValue;
+        this._value = newValue;
+    }
+
+    get value(): any {
```
dclaux: `value` is supposed to be of type **string**. The underlying code expects that.

source/nodejs/adaptivecards-fabric/src/components/inputs/InputToggleFabric.tsx
```diff
@@ -12,9 +12,8 @@ export class InputToggleFabric extends Shared.ReactInputElement {
     private valueOff: string;
 
     public parse = (json: any, errors?: AC.IValidationError[]) => {
-        this.id = AC.getStringValue(json.id);
-        this.valueInternal = json.value;
-        this.defaultValue = this.valueInternal;
+        super.parse(json, errors);
+        this.value = AC.getStringValue(json.value, "");
```
dclaux: Should be `this.value = this.defaultValue;`

source/nodejs/adaptivecards-fabric/src/components/inputs/InputTimeFabric.tsx
```diff
@@ -13,9 +13,8 @@ export class InputTimeFabric extends Shared.ReactInputElement {
     private max: string;
 
     public parse = (json: any, errors?: AC.IValidationError[]) => {
-        this.id = AC.getStringValue(json.id);
-        this.valueInternal = json.value;
-        this.defaultValue = this.valueInternal;
+        super.parse(json, errors);
+        this.value = json.value;
```
dclaux: Should be `this.value = this.defaultValue;`

source/nodejs/adaptivecards-fabric/src/components/inputs/InputTextFabric.tsx
```diff
@@ -20,20 +20,29 @@ export class InputTextFabric extends Shared.ReactInputElement {
     public getJsonTypeName = (): string => "Input.Text";
 
     public parse = (json: any, errors?: any) => {
-        this.valueInternal = json.value;
-        this.defaultValue = this.valueInternal;
-        this.id = AC.getStringValue(json.id);
+        super.parse(json, errors);
+        this.value = json.value;
```
dclaux: `this.value = this.defaukltValue`

source/nodejs/adaptivecards-fabric/src/components/inputs/InputNumberFabric.tsx
```diff
@@ -13,9 +13,8 @@ export class InputNumberFabric extends Shared.ReactInputElement {
     private placeholder: string;
 
     public parse = (json: any, errors?: AC.IValidationError[]) => {
-        this.valueInternal = json.value;
-        this.defaultValue = this.valueInternal;
-        this.id = AC.getStringValue(json.id);
+        super.parse(json, errors);
+        this.value = AC.getStringValue(json.value, "");
```
dclaux: `this.value = this.defaultValue;`

source/nodejs/adaptivecards-fabric/src/components/inputs/InputDateFabric.tsx
```diff
@@ -20,15 +20,24 @@ export class InputDateFabric extends Shared.ReactInputElement {
...
+        AC.setProperty(result, "placeholder", this.placeholder);
+
+        return result;
+    }
+
     private parseDates = (json: any, errors?: AC.IValidationError[]) => {
-        const dateString = AC.getStringValue(json.value);
-        this.valueInternal = dateString;
-        this.defaultValue = this.valueInternal;
+        const dateString = AC.getStringValue(json.value, "");
```
dclaux: `const dateString = this.defaultValue;`

source/nodejs/adaptivecards-fabric/src/components/inputs/InputChoiceSetFabric.tsx
```diff
@@ -20,9 +20,8 @@ export class InputChoiceSetFabric extends Shared.ReactInputElement {
     private selectedValues: string[] = [];
 
     public parse = (json: any, errors?: AC.IValidationError[]) => {
-        this.id = json.id ? AC.getStringValue(json.id) : null;
-        this.valueInternal = json.value;
-        this.defaultValue = this.valueInternal;
+        super.parse(json, errors);
+        this.value = AC.getStringValue(json.value, "");
```
dclaux: `this.value = this.defaultValue;`

source/nodejs/adaptivecards-fabric/src/components/inputs/InputChoiceSetFabric.tsx
```diff
@@ -31,13 +30,31 @@ export class InputChoiceSetFabric extends Shared.ReactInputElement {
...
             this.createExpandedView()
     )
 
     public getJsonTypeName = (): string => "Input.ChoiceSet";
 
+    public toJSON = () => {
+        let result = super.toJSON();
+
+        if (this.choices.length > 0) {
+            AC.setProperty(result, "choices", this.choices);
```
dclaux: Use `AC.setArrayProperty` instead. setProperty won't work as desired.

source/nodejs/adaptivecards-fabric/src/components/inputs/InputChoiceSetFabric.tsx
```diff
@@ -20,9 +20,8 @@ export class InputChoiceSetFabric extends Shared.ReactInputElement {
     private selectedValues: string[] = [];
 
     public parse = (json: any, errors?: AC.IValidationError[]) => {
-        this.id = json.id ? AC.getStringValue(json.id) : null;
-        this.valueInternal = json.value;
-        this.defaultValue = this.valueInternal;
+        super.parse(json, errors);
+        this.value = AC.getStringValue(json.value, "");
         this.selectedValues = this.defaultValueToArray(this.value);
         this.choices = json.choices;
```
dclaux: Ideally you'd use the Choice class that's already part of the library. The built-in ChoiceSet input parses choices like this, which I suggest you copy. Again that is because Choice extends SerializableObject which handles (de)serialization in a specific way that works well with the designer:

```typescript
        this.choices = [];

        if (Array.isArray(json["choices"])) {
            for (let jsonChoice of json["choices"]) {
                let choice = new Choice();
                choice.parse(jsonChoice);

                this.choices.push(choice);
            }
        }
```

source/nodejs/adaptivecards-fabric/src/components/inputs/InputChoiceSetFabric.tsx
```diff
@@ -20,9 +20,8 @@ export class InputChoiceSetFabric extends Shared.ReactInputElement {
     private selectedValues: string[] = [];
 
     public parse = (json: any, errors?: AC.IValidationError[]) => {
-        this.id = json.id ? AC.getStringValue(json.id) : null;
```
dclaux: You do not appear to be parsing/handling the `wrap` and `placeholder`  properties.

source/nodejs/adaptivecards-fabric/src/components/inputs/InputTextFabric.tsx
```diff
@@ -20,20 +20,29 @@ export class InputTextFabric extends Shared.ReactInputElement {
     public getJsonTypeName = (): string => "Input.Text";
 
     public parse = (json: any, errors?: any) => {
-        this.valueInternal = json.value;
```
dclaux: You don't appear to be parsing/handling the `inlineAction` and `style` properties.

source/nodejs/adaptivecards-fabric/src/components/inputs/InputToggleFabric.tsx
```diff
@@ -12,9 +12,8 @@ export class InputToggleFabric extends Shared.ReactInputElement {
...
         this.valueOn = AC.getStringValue(json.valueOn);
         this.valueOff = AC.getStringValue(json.valueOff);
         this.title = AC.getStringValue(json.title);
```
dclaux: You don't appear to be parsing/handling the `wrap` property.

source/nodejs/adaptivecards-fabric/src/components/inputs/InputChoiceSetFabric.tsx
dfisenko: wrap is 1.2 feature which is not supported by this extension.

source/nodejs/adaptivecards-fabric/src/utils/shared.tsx
dfisenko: I was following this example : https://docs.microsoft.com/en-us/adaptive-cards/sdk/rendering-cards/javascript/extensibility
here value is number. imo it makes more sense to have this as any, since not value are strings. For example input.number

source/nodejs/adaptivecards-fabric/src/components/inputs/InputToggleFabric.tsx
dfisenko: wrap is not supported in this version

source/nodejs/adaptivecards-fabric/src/components/inputs/InputTextFabric.tsx
dfisenko: inlineAction is not supported by this version. adding style

source/nodejs/adaptivecards-fabric/src/components/inputs/InputToggleFabric.tsx
dclaux: You mean in the Fabric control?

source/nodejs/adaptivecards-fabric/src/components/inputs/InputTextFabric.tsx
dclaux: Not sure what you mean. inlineAction is definitely supported in 1.2, which has shipped. inlineAction isn't a native capability of the input control, it needs to be handled explicitly.

dfisenko:
Screenshot. Top input element has an inlineAction associated 

image

dclaux:
Looks good. I assume button is aligned to the bottom when the input is multiline?

dfisenko:
yes, button is aligned to the bottom
image

PR #3135, opened 2019-07-02
[TS] Update package versions

paulcam206:
Updates for security fixes

Verified by building and running site

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3135)


shalinijoshi19:
@matthidinger let's wait until this gets committed from @paulcampbell before checking in the template branch into master if that's ok -- should fix a bunch of security alerts/flags i was seeing against that branch

Issue #2891, opened 2019-05-10
[.NET] Per action type styling for action sentiments


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* .NET WPF

# Issue

In .NET sentiment styles are specific to the type of action. You can (and must if you want all buttons to behave the same) set the "Adaptive.Action.Submit.Positive" style and the "Adaptive.Action.OpenUrl.Positive" style separately. In UWP, for example, you set a single "Adaptive.Action.Positive" style that applies to all action types. Per the spec, the latter appears to be the correct behavior:

From #861:
"A renderer must output the appropriate styling name on the action, so that hosts can natively style the buttons. For example, a default button might get a style name of Action, but an emphasized button might get a style name of Action.Positive (or however the native platform styling naming convention works)"

shalinijoshi19:
@RebeccaAnne, good catch. FYI I've approved for fix but moved it out of 1.2 for post release fixing

PR #3138, opened 2019-07-02
[.NET] Remove action type from style

almedina-ms:
Fixes https://github.com/microsoft/AdaptiveCards/issues/2891

Changes the style name to be used for style from Adaptive.Action.<Style>.<Sentiment> to Adaptive.Action.<Style>following the given specification 

## How verified

The change is for removing the specification of which type of action is being rendered so I tested if the same style would be used for two different types of actions and it yielded this results 

image

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3138)


paulcam206:
> Changes the style name to be used for style from Adaptive.Action.<Style>. to Adaptive.Action.<Style>following the given specification

@almedina-ms: did you mean "from Adaptive.<Style> to Adaptive.Action.<Style>"?

PR #3153, opened 2019-07-08
[UWP] Fix debug build

RebeccaAnne:
Building UWP locally (x86/debug) I ran into the following error:
"'/ZI' and '/guard:cf' command-line options are incompatible"
Set DebugInformationFormat to ProgramDatabase (/Zi instead of /ZI) to avoid this conflict. 

Additionally, hit two warnings about unreachable code (now treated as errors) and fixed them.

How verified: UWP debug now builds correctly. No new test failures.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3153)


paulcam206:
doh! thanks for catching this. I modified the CI job to include debug builds (previously disabled due to the number of agents required).

PR #3156, opened 2019-07-08
[.NET] Port WPF renderer to work for .NETCore3.0

almedina-ms:
Adds a .NET Core 3.0 project so the WPF renderer works with .NET Core, also copies the Image Renderer sample to target the .NET Core project and to test that the rendering works as expected 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3156)


source/dotnet/AdaptiveCards.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27004.2005
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29025.244
```
paulcam206: we're not ready to support VS2019 at this time :(

almedina-ms: According to this page we should use .net core 3.0 with the latest preview of Visual Studio, I could try Visual Studio 17 to check if it works too there

https://dotnet.microsoft.com/download/dotnet-core/3.0

almedina-ms: But according to this one we must use 2019

> Prerequisites for .NET Core 3.0 Preview 3
> 
> .NET Core 3.0 Preview 3 has the same prerequisites as other versions of .NET Core. However, if you want to use Visual Studio to create .NET Core 3.0 projects, you must use the Visual Studio 2019. Visual Studio 2019 can be installed side-by-side with other versions of Visual Studio without conflict.

https://docs.microsoft.com/en-us/dotnet/core/windows-prerequisites?tabs=netcore2x

shalinijoshi19: @paulcam206, @almedina-ms what happened around this one? Are we ready to roll up to VS 2019? Let's try to close on this one!

source/dotnet/global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "3.0.100-preview6-012264"
```
shalinijoshi19: >3.0.100-preview6-012264" [](start = 16, length = 24)

is this still supposed to be preview?

almedina-ms: The package that was just released is 3.0.100 (aka v3.0.0), if we want to build against it we must wait for the build agents to support it as the MSBuild they have is 16.2 and 3.0.100 requires 16.3 
https://dotnet.microsoft.com/download/dotnet-core/3.0

preview 6 is the latest version for the SDK we can use as of this time

almedina-ms: This is the error I got when trying to use 3.0.100
Error: Version 3.0.100 of the .NET Core SDK requires at least version 16.3.0 of MSBuild. The current available version of MSBuild is 16.2.37902.0.

shalinijoshi19: ok thanks -- can we loop in the ES folks + Fabian/Daniel from WPF team and make sure we can get an ETA on when RTM release would be supported on the build agents? @paulcam206 FYI.

---
In reply to: [328866285](https://github.com/microsoft/AdaptiveCards/pull/3156#discussion_r328866285) [](ancestors = 328866285)

source/dotnet/Library/AdaptiveCards.Rendering.Wpf.Core/AdaptiveCards.Rendering.Wpf.Core.csproj
```diff
@@ -0,0 +1,34 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+
+  <PropertyGroup>
+    <OutputType>Library</OutputType>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+	  <UseWPF>true</UseWPF>
```
paulcam206: >	  < [](start = 0, length = 4)

there's a rogue tab in here :)

Issue #3184, opened 2019-07-10
[Android] Value from toggle input not being retrieved


almedina-ms:
# Platform

* Android

# Version of SDK

master

# Issue

Action.Submit is throwing errors when clicking

![Screenshot_2019-07-10-16-29-09](https://user-images.githubusercontent.com/35784165/61009533-be1af300-a327-11e9-9341-8636d7e3eb56.png)

PR #3193, opened 2019-07-11
[Android] Add validation for empty data in submit actions

almedina-ms:
Fixes https://github.com/microsoft/AdaptiveCards/issues/3184

Android is actually retrieving the values correctly, the issue has to do with the value returned by the empty data being "null\n" (value retrieved by the shared model) so once the JSONObject tries to parse that string it throws an exception thus creating a misleading message

This is how the app works now

Screenshot_2019-07-11-11-57-59

### DIFF
--- a/source/android/mobilechatapp/src/main/java/com/example/mobilechatapp/MainActivity.java
+++ b/source/android/mobilechatapp/src/main/java/com/example/mobilechatapp/MainActivity.java
@@ -263,11 +263,15 @@ public class MainActivity extends AppCompatActivity implements ICardActionHandle
         this.runOnUiThread(new RunnableExtended(this, text, duration));
     }
 
-    private void onSubmit(BaseActionElement actionElement, RenderedAdaptiveCard renderedAdaptiveCard) {
+    private void onSubmit(BaseActionElement actionElement, RenderedAdaptiveCard renderedAdaptiveCard)
+    {
         SubmitAction submitAction = null;
-        if (actionElement instanceof SubmitAction) {
+        if (actionElement instanceof SubmitAction)
+        {
             submitAction = (SubmitAction) actionElement;
-        } else if ((submitAction = SubmitAction.dynamic_cast(actionElement)) == null) {
+        }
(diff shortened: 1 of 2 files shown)


source/android/mobilechatapp/src/main/java/com/example/mobilechatapp/MainActivity.java
```diff
@@ -263,22 +263,32 @@ public void run()
...
         String data = submitAction.GetDataJson();
         Map<String, String> keyValueMap = renderedAdaptiveCard.getInputs();
         if (!data.isEmpty())
         {
-            try {
-                JSONObject object = new JSONObject(data);
+            try
+            {
+                JSONObject object = new JSONObject();
+                if (!data.equals("null\n"))
```
RebeccaAnne: >null\n [](start = 34, length = 6)

Where is this string coming from? Is getting this string the expected behavior or should the shared model be doing something better? If we  have to do this, third parties would do, and it seems like a bad experience.

source/android/mobilechatapp/src/main/java/com/example/mobilechatapp/MainActivity.java
```diff
@@ -263,22 +263,32 @@ public void run()
...
 
         String data = submitAction.GetDataJson();
         Map<String, String> keyValueMap = renderedAdaptiveCard.getInputs();
         if (!data.isEmpty())
         {
-            try {
-                JSONObject object = new JSONObject(data);
+            try
+            {
+                JSONObject object = new JSONObject();
```
paulcam206: > = new JSONObject(); [](start = 33, length = 20)

maybe do this in an `else` case below to avoid an unnecessary allocation? (also in the other file)

source/android/mobilechatapp/src/main/java/com/example/mobilechatapp/MainActivity.java
almedina-ms: that string is coming from the shared model

PR #3226, opened 2019-07-12
[Shared Model][UWP] Add input validation to the object model

RebeccaAnne:
Add the P0 properties for input validation to the object model for the shared model and UWP. This change currently includes object model definitions only, and no functionality beyond serializing/deserializing and property management.

How verified: Object model tests at shared model and UWP layers. New sample files added including these properties added to the test app.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3226)


source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -543,6 +543,16 @@ AdaptiveNamespaceStart
         All = Left | Right | Up | Down
     } BleedDirection;
 
+    [version(NTDDI_WIN10_RS1), flags]
+#ifdef ADAPTIVE_CARDS_WINDOWS
+    [contract(InternalContract, 1)][internal]
+#endif
+    typedef[v1_enum] enum InputNecessityIndicators {
```
paulcam206: >InputNecessityIndicators [](start = 26, length = 24)

Code looks good. I was hoping that we'd have another round of spec review once the showcard stuff was ironed out. I don't like this name, and wonder if we can't come up with something better. That said, this PR shouldn't be blocked on that... Please let me know once this branch is merged and I'll file a bug. :)

Issue #3192, opened 2019-07-11
For "column" component, the UI is different when set "width" = "auto" in IOS platform


FelixZhang-yong:
# Platform

* iOS

# Author or host

No

# Version of SDK

IOS v1.2.0

# Issue

For "column" component, the UI is different when set "width" = "auto" in IOS platform
For the same payload
```
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "Container",
      "items": [
        {
          "type": "ColumnSet",
          "columns": [
            {
              "type": "Column",
              "width": "auto",
              "items": [
                
              ]
            },
            {
              "type": "Column",
              "width": "auto",
              "items": [
                {
                  "type": "TextBlock",
                  "size": "medium",
                  "weight": "bolder",
                  "text": "title"
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

Other platform rendered as
<img width="420" alt="屏幕快照 2019-07-11 上午10 51 46" src="https://user-images.githubusercontent.com/52741892/61020276-b878e800-a3cf-11e9-8d59-132b5e394793.png">
But in IOS, it is rendered as 
<img width="329" alt="屏幕快照 2019-07-11 上午11 26 04" src="https://user-images.githubusercontent.com/52741892/61020286-c4fd4080-a3cf-11e9-9d9c-5665f836e8eb.png">

nesalang:
Hello @FelixZhang-yong 
Thanks for the feedback.
I tried your json, I could reproduce the issue, and this inconsistency is caused by the empty column. Is there a need for this use case?

FelixZhang-yong:
I intended to simplify the issue with previous payload. Below is the sample payload for the use case we meet. We want to show an image in the left and place name & title on the right of the image.
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "Container",
      "items": [
        {
          "type": "ColumnSet",
          "columns": [
            {
              "type": "Column",
              "width": "auto",
              "items": [
                {
                  "type": "Image",
                  "size": "small",
                  "style": "person",
                  "url": "https://pbs.twimg.com/profile_images/3647943215/d7f12830b3c17a5a9e4afcc370e3a37e_400x400.jpeg"
                }
              ]
            },
            {
              "type": "Column",
              "width": "auto",
              "items": [
                {
                  "type": "TextBlock",
                  "size": "medium",
                  "weight": "bolder",
[...]

PR #3232, opened 2019-07-15
Jwoo/ios auto priority update

nesalang:
fixes issue #3192 

when there are two columns, and both have width "auto", the last column stretches in other renderers.
This PR adjust hugging priories to mimic the behavior.

How verified:
verified with tests cards given by customer and existing cards.

 ##### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3232)


shalinijoshi19:
@almedina-ms can you take a stab at resolving the conflicts and merging this in since @nesalang is OOF this week?  Tx

Issue #3209, opened 2019-07-12
[Android] Not all renderers support fallback on Column


andrewleader:
**Automatically generated task. Any updates to the title and body will be overwritten. Do NOT modify the labels.**

Android fix of #3021

andrewleader:
Android has been untested, someone needs to confirm if Android supports fallback on column, and if not needs to be fixed!

PR #3233, opened 2019-07-16
[Android] Fix column fallback

almedina-ms:
Fixes https://github.com/microsoft/AdaptiveCards/issues/3209

The issue was that column was being "rendered" even when the contents couldn't be rendered, so we only got an empty column.

## How verified

The mobile app was run and, using the Column.Fallback file, the result was the following:

Screenshot_2019-07-15-17-59-23

As can be seen, the first column is changed for a column with the textblock "Fallback" and the second column is dropped

Meanwhile, the previously working DeepFallback card was revised so the dropped container spacing wouldn't appear at all after rendering

image

Screenshot_2019-07-15-17-59-36

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3233)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/registration/CardRendererRegistration.java
```diff
@@ -263,103 +263,7 @@ else if(tag instanceof AdaptiveCard)
...
-                }
-                else
-                {
-                    // The element doesn't have a fallback, so the element can't be rendered and it's dropped
-                    renderedCard.addWarning(new AdaptiveWarning(AdaptiveWarning.UNKNOWN_ELEMENT_TYPE,"Unsupported card element type: " + cardElement.GetElementTypeString()));
-                    continue;
-                }
-
-            }
+            renderFallback(renderedCard, context, fragmentManager, cardElement, layout, cardActionHandler, hostConfig, renderArgs, featureRegistration);
```
nesalang: nit: this method name is not discriptive enough to tell what the method does.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ColumnSetRenderer.java
```diff
@@ -96,7 +95,17 @@ else if ((columnSet = ColumnSet.dynamic_cast(baseCardElement)) == null)
             RenderArgs columnRenderArgs = new RenderArgs(renderArgs);
             columnRenderArgs.setContainerStyle(styleForThis);
 
-            columnRenderer.render(renderedCard, context, fragmentManager, layout, column, cardActionHandler, hostConfig, columnRenderArgs);
+            FeatureRegistration featureRegistration = CardRendererRegistration.getInstance().getFeatureRegistration();
+
+            CardRendererRegistration.getInstance().renderElementAndPerformFallback(renderedCard,
+                                                                  context,
```
RebeccaAnne: >                                                                   [](start = 0, length = 66)

[nit] line up this spacing?

Issue #2569, opened 2019-03-27
.NET WPF doesn't render a simple card


paulcam206:
.NET WPF (haven't tested HTML yet) fails to render this card:

```json
{
	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
	"type": "AdaptiveCard",
	"version": "1.2",
	"body": [
		{
			"type": "TextBlock",
			"id": "3",
			"text": "Container default style"
		}
	]
}
```

exception message:
```
System.ArgumentException
  HResult=0x80070057
  Message='3' is not a valid value for property 'Name'.
  Source=WindowsBase
  StackTrace:
   at System.Windows.DependencyObject.SetValueCommon(DependencyProperty dp, Object value, PropertyMetadata metadata, Boolean coerceWithDeferredReference, Boolean coerceWithCurrentValue, OperationType operationType, Boolean isInternal)
   at System.Windows.DependencyObject.SetValue(DependencyProperty dp, Object value)
   at AdaptiveCards.Rendering.Wpf.AdaptiveRenderContext.Render(AdaptiveTypedElement element)
   at AdaptiveCards.Rendering.Wpf.AdaptiveContainerRenderer.AddContainerElements(Grid uiContainer, IList`1 elements, AdaptiveRenderContext context)
   at AdaptiveCards.Rendering.Wpf.AdaptiveCardRenderer.RenderAdaptiveCardWrapper(AdaptiveCard card, AdaptiveRenderContext context)
   at AdaptiveCards.Rendering.AdaptiveElementRenderers`2.<>c__DisplayClass1_0`1.<Set>b__0(AdaptiveTypedElement typedElement, TContext tContext)
   at AdaptiveCards.Rendering.Wpf.AdaptiveRenderContext.Render(AdaptiveTypedElement element)
   at AdaptiveCards.Rendering.Wpf.AdaptiveCardRenderer.RenderCard(AdaptiveCard card)
```
Aside from the exception itself, it's noteworthy that we're not converting the `System.ArgumentException` to an `AdaptiveException`.

paulcam206:
card renders as expected in designer...

shalinijoshi19:
@paulcam206 does this need to be fixed for 1.2? If so we should get to this soon

paulcam206:
[...]

PR #3242, opened 2019-07-17
[WPF][Draft] Fix validation for id

almedina-ms:
Fixes https://github.com/microsoft/AdaptiveCards/issues/2569

## Issue

WPF doesn't accept names composed of only numbers or special characters so we have to find a different option for it

## Proposed solution

Create a dictionary to match the id and the rendered element, this allows setting an element by id without any limitation and performing the search faster as the ui tree must not be transversed

## How verified 

This card was created to test the behaviour of toggling the visibility of the card and the card rendered successfully

```
{
	"type": "AdaptiveCard",
	"version": "1.2",
	"body": [
		{
			"type": "TextBlock",
			"id": "3",
			"text": "Container default style",
            "height": "stretch"
		}
	],
	"actions": [
		{
			"title": "Also Toggle!",
			"targetElements": [
				{
					"elementId": "3"
				}
			]
		}
	]
}
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3242)


source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveRenderContext.cs
```diff
@@ -142,6 +145,9 @@ public SolidColorBrush GetColorBrush(string color)
             }
         }
 
+        // This dictionary exists for setting the ids to elements without messing with the Name property to avoid crashes with weird ids
+        public IDictionary<string, FrameworkElement> RendererElementsWithId = new Dictionary<string, FrameworkElement>();
```
andrewleader: Do we need to make this public? I worry about future breaking API changes if we change this structure, like maybe we decide it's a dictionary of `<string, UIElement>` or something in the future.

I believe the only use cases for hosts using this is they need to toggle visibility of an element, which we already have a public toggle visibility method I believe?

shalinijoshi19:
@almedina-ms can you schedule some time to work through this one and get folks together to work through this one ? There seem to be multiple id validation conversations happening looks like and it would be worthwhile to have a consolidated onenote section _for starters, and try to jot down our current and proposed id story and the verification that we are proposing go with it across all platforms and go from there to trying to close this out.  @paulcam206 FYI.

PR #3246, opened 2019-07-19
Create telemetry.md

utsavlathia:
Created the Telemetry Dev spec for Input.Rating 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3246)


shalinijoshi19:
Curious about the file name? Why "Create DevSpecs.md" ? This should be more like "TelemetryDesign.md" or something

utsavlathia:
> Curious about the file name? Why "Create DevSpecs.md" ? This should be more like "TelemetryDesign.md" or something

Yes, that was a mistake on my part in making the subfolder and md file, fixed now 👍

specs/DevSpecs/telemetry.md
```diff
@@ -0,0 +1,187 @@
+# Input.Rating Telemetry Design Specification 
+ 
+## Abstract 
+This design specification covers telemetry integration into the `Input.Rating` control, which will be included in a future release of Adaptive Cards. Telemetry integration is currently slated only for the JavaScript platform; however, this developer specification will be general enough to cover all platforms supported by Adaptive Cards. This telemetry design will focus on sending telemetry via the 1DS platform, but the design is flexible enough to allow for possible integration of other telemetry platforms in the future. See more at the GitHub issue [3058](https://github.com/microsoft/AdaptiveCards/issues/3058).
+ 
+## 1. Feature Dependencies 
+### 1.1 Dependencies on other features 
+Because telemetry will be baked into the `Input.Rating` control as a result of the implementation of this design, this feature is inherently dependent on the `Input.Rating` control. Externally, this feature is dependent on the functionality of the 1DS SDK.
```
shalinijoshi19: I wouldn't call Input rating control a dependency perse here. 1DS on the other hand is.

shalinijoshi19: for sending data up through the Microsoft telemetry pipeline

specs/DevSpecs/telemetry.md
```diff
@@ -0,0 +1,187 @@
...
+### 1.1 Dependencies on other features 
+Because telemetry will be baked into the `Input.Rating` control as a result of the implementation of this design, this feature is inherently dependent on the `Input.Rating` control. Externally, this feature is dependent on the functionality of the 1DS SDK. 
+ 
+### 1.2 Other features' dependencies on this feature 
+There are currently no other existing features that directly depend on this feature. However, future telemetry integration for other elements will be based on the design presented here. 
+ 
+## 2. Platform-Specific Considerations 
+UWP, JS, and .NET are all currently supported by 1DS, allowing for relatively seamless integration of telemetry for these platforms. Mobile platforms (ie., Android and iOS) are not fully supported by 1DS, so future communication and cooperation with the 1DS team to achieve support for these platforms is necessary. 
+ 
+## 3. Performance Implications
```
shalinijoshi19: There may be some memory/binary size etc footprint analysis that we may need to do once we start to integrate with partners like Outlook/Cortana and any expectations around framework binary sizes on different platforms etc. But that will be tested and analyzed at a later stage as part of integration and perf testing.

specs/DevSpecs/telemetry.md
```diff
@@ -0,0 +1,187 @@
...
+UWP, JS, and .NET are all currently supported by 1DS, allowing for relatively seamless integration of telemetry for these platforms. Mobile platforms (ie., Android and iOS) are not fully supported by 1DS, so future communication and cooperation with the 1DS team to achieve support for these platforms is necessary. 
+ 
+## 3. Performance Implications 
+> “Perf-penalty in terms of runtime perf on foreground threads should not be affected; all our log event API calls are async. Background threads would kick in when there's time to upload. Total CPU time to upload would depend how much data you logged. In UTC mode - the perf penalty vs. trace logging would be near-zero. About the same.” - Max Golovanov (1DS) 
+
+We will look further into how 1DS affects startup time and memory usage with comparison tests. The real-world perf numbers would depend on telemetry volume per minute. The performance implications shouldn’t be major, but will exist.  
+ 
+ 
+## 4. Backwards Compatibility Concerns  
+If the renderer is not up to date, we will not be able to receive telemetry, which is not a problem because ‘Input.Rating’ would not work correctly without the up-to-date renderer.
```
shalinijoshi19: >ut.Rating’ would not work correctly without the up-to-date renderer.   [](start = 112, length = 70)

Hmm that's not quite accurate. Input.Rating is expected to have a fallback story in place if one is not explicitly specified by the author.. But specifically the requirement here is to ensure that if the renderer with telemetry enabled is not able to be constructed on older hosts, that failure should not result in breaking the functionality of the controls.

specs/DevSpecs/telemetry.md
```diff
@@ -0,0 +1,187 @@
...
+ 
+    logWarn(message: string): void; 
+ 
+    logErr(message: string): void; 
+ 
+    logVerbose(message: string): void; 
+ 
+    logInfo(message: string): void; 
+ 
+    logEvent(event: string, eventSourceName: string, correlationID: string, valueSet?: object): void;
```
shalinijoshi19: should this be nullable too? and we generate it if not's specified explicitly?

specs/DevSpecs/telemetry.md
```diff
@@ -0,0 +1,187 @@
...
+    logWarn(message: string): void; 
+ 
+    logErr(message: string): void; 
+ 
+    logVerbose(message: string): void; 
+ 
+    logInfo(message: string): void; 
+ 
+    logEvent(event: string, eventSourceName: string, correlationID: string, valueSet?: object): void; 
+
```
shalinijoshi19: We also exposing a way to specify/configure different telemetry providers right.. Should be up here too included as part of the interface

specs/DevSpecs/telemetry.md
```diff
@@ -0,0 +1,187 @@
...
+  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json", 
+  "version": "1.0" 
+} 
+ 
+``` 
+Expected Values: 
+ 
+| Property                     | Expected Value | 
+|------------------------------|----------------| 
+| CorrelationID                | any            |
```
shalinijoshi19: >any [](start = 33, length = 3)

what does 'any' imply? is that the type or the actual value?

specs/DevSpecs/telemetry.md
```diff
@@ -0,0 +1,187 @@
...
+ 
+    logVerbose(message: string): void; 
+ 
+    logInfo(message: string): void; 
+ 
+    logEvent(event: string, eventSourceName: string, correlationID: string, valueSet?: object): void; 
+ 
+} 
+ 
+enum TelemetryProvider {
```
shalinijoshi19: >TelemetryProvider [](start = 5, length = 17)

We should introduce/reuse namespaces here that may be in use for different platforms -- maybe namespace AdaptiveCards{export IACLogger..} similarly for the telemetryProvider

specs/DevSpecs/telemetry.md
```diff
@@ -0,0 +1,187 @@
...
+    logVerbose(message: string): void; 
+ 
+    logInfo(message: string): void; 
+ 
+    logEvent(event: string, eventSourceName: string, correlationID: string, valueSet?: object): void; 
+ 
+} 
+ 
+enum TelemetryProvider { 
+    OneDataStrategy,
```
shalinijoshi19: >OneDataStrategy [](start = 4, length = 15)

Maybe Microsoft1DS ?

specs/DevSpecs/telemetry.md
```diff
@@ -0,0 +1,187 @@
...
+ 
+    logInfo(message: string): void; 
+ 
+    logEvent(event: string, eventSourceName: string, correlationID: string, valueSet?: object): void; 
+ 
+} 
+ 
+enum TelemetryProvider { 
+    OneDataStrategy, 
+    File
```
shalinijoshi19: Console also

specs/DevSpecs/telemetry.md
```diff
@@ -0,0 +1,187 @@
...
+ 
+    public static getLogger(): IACLogger { 
+        if (instance) { 
+            return instance; 
+        } else { 
+            // throw error/exception 
+        } 
+    } 
+ 
+    public static setLogger(...providers: TelemetryProvider[]): void {
```
shalinijoshi19: by default we should add console provider i think? Maybe expose a InitializeDefaultProviders() or something

specs/DevSpecs/telemetry.md
```diff
@@ -0,0 +1,226 @@
...
+ 
+ 
+## 4. Backwards Compatibility Concerns  
+There are no backwards compatibility concerns at this time.
+ 
+## 5. Security Impact 
+Security should be considered in the situation of protecting the token and telemetry pipeline from malicious parties. Although Adaptive Cards is open source, we do not plan on exposing the telemetry token to the public. Thus, distribution of Adaptive Cards versions with telemetry should be limited to internal use cases only (within Microsoft Corp. only). This may change upon further discussion with the 1DS team.
+ 
+## 6. New APIs 
+This feature will introduce the IACLogger, which provides the following interface (see [section 9.1](#91-aclogger-telemetry-wrapper) for examples and implementation details):
```
shalinijoshi19: Open Issues (notes from Frank + Shalini discussion):
Business Logic: 
0. Where do we want telemetry enabled? Host side or authoring side? 
Ans: **Both**; implementation priority for host/rendering higher than authoring. For eg, for authoring the business justification for adding telemetry in, is to make data-driven decisions on per-platform authoring SDK adoption

1. Do we want telemetry enabled by default? 
Ans: Yes (for Frank+Shalini); We need to provide customers (developers) with an intentional and explicit opt-out story for telemetry in general.

2. Should there be default trace providers? 
Ans: Yes;(For Frank + Shalini); If we are calling an API we should have *some* sense of that API having been invoked at some point in the workflow; This becomes esp important in cases where we are flighting APIs/experimenting.

3. If so, which ones should our default providers be? Console? 1DS? File? 
Ans: Ideally 1DS (something that leaves the machine to be able to be processed by the product team); Otherwise the whole notion of telemetry to be able to make and feed into business decisions becomes pointless.

4. Once enabled how does a customer (Host app) turn it off if they choose to change their minds
Ans: Same configuration switch/API that would/could enable it. Nothing fancy.

**Implementation Open Questions:**

paulcam206:
closing out old PRs

PR #3250, opened 2019-07-19
Remove backgroundImage from ColumnSet

### LINKED ISSUE #3249: Docs/schema incorrectly says ColumnSet has backgroundImage property
I just noticed this today, our docs say that ColumnSet has a `backgroundImage` property added in 1.2, however none of the renderers (JS, UWP, .NET) support that and I couldn't find anywhere where we ever claimed that we SHOULD add that in 1.2... I believe it was an unintentional addition to the schema and should be removed

image

andrewleader:
The property wasn't supposed to be on there and it was an accidental addition to the schema which none of the renderers support

* Removed the `backgroundImage` property from the schema
* Updated the `Input.Toggle.value` description since it had a new typo and was also worded very confusingly
* Generated the new `1.2.1` schema

Part of fixing issue #3249, but we need to also pull this change down to the 1.2 branch and then update the site before that's fixed. But fixing in master is the first step

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3250)

### DIFF
--- a/schemas/1.2.1/adaptive-card.json
+++ b/schemas/1.2.1/adaptive-card.json
@@ -1,0 +1,2483 @@
+{
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "id": "http://adaptivecards.io/schemas/adaptive-card.json",
+  "definitions": {
+    "Action.OpenUrl": {
+      "description": "When invoked, show the given url either by launching it in an external web browser or showing within an embedded web browser.",
+      "properties": {
+        "type": {
+          "enum": [
+            "Action.OpenUrl"
+          ],
+          "description": "Must be `Action.OpenUrl`"
+        },
+        "url": {
+          "type": "string",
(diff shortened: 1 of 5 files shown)


shalinijoshi19:
Should this be included in 1.2.1? If so could you add the 1.2.1 label etc ? Tx!

andrewleader:
I don't believe schema updates impact the renderer SDKs, so when we "ship" out this website update is irrelevant of when the renderers ship out 1.2.1 (and the reality is we should "ship" out this schema website update ASAP since it's just going to cause confusion when people see that you can use backgroundImage on ColumnSet and you actually can't)

matthidinger:
Thanks Andrew! I'll get the site published with this fix and some other nice updates this week

PR #3256, opened 2019-07-25
improve designer docs to not hard code cdn versions

matthidinger:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3256)


source/nodejs/adaptivecards-designer/README.md
```diff
@@ -227,5 +232,13 @@ For advanced configuration of the designer use the following APIs.
 			]
 		}
 	);
+
+	/* Modify the Element toolbox (BEFORE designer attached) */
```
shalinijoshi19: Curious - what does "after/before designer attached" imply?

source/nodejs/adaptivecards-designer-app/src/app.ts
```diff
@@ -3,6 +3,7 @@
 import * as monaco from "monaco-editor";
 import * as markdownit from "markdown-it";
 import * as ACDesigner from "adaptivecards-designer";
+import * as Adaptive from "adaptivecards";
```
shalinijoshi19: why this import? Don't see it being used here?

shalinijoshi19:
Could you elaborate on why we need the changes ? Is it for CDG compliance purposes? If so, could you add a quick mention of this in teh description please? thanks

source/nodejs/adaptivecards-designer/README.md
dclaux: It means before the designer instance's `attachTo` method is called.

dclaux:
@shalinijoshi19 at the very least it forces developers to actually decide what version to use, and thus avoids that we have a bunch of developers all using 0.7.0 simply because they copy/pasted.

shalinijoshi19:
Thanks @dclaux makes sense. Ask is to include this in the description for clarity. @matthidinger, fyi there is a PR template that should be auto-populated when you spin up a new PR now. 

---
In reply to: [528451200](https://github.com/microsoft/AdaptiveCards/pull/3256#issuecomment-528451200) [](ancestors = 528451200)

dclaux:
I am honestly not sure why that would be needed. Doing a simple copy/paste will simply not work, and developers will be forced to specify a version, which they will know to do by reading the comment in the code. Not sure how telling them "we're doing this to force you to pick a version" would improve the experience?

Or are you referring to "attaching" and would like it rephrased to something like "do this before/after you call designer.attachTo" instead of "do this before/after attaching the designer"?

shalinijoshi19:
> I am honestly not sure why that would be needed. Doing a simple copy/paste will simply not work, and developers will be forced to specify a version, which they will know to do by reading the comment in the code. Not sure how telling them "we're doing this to force you to pick a version" would improve the experience?
> 
> Or are you referring to "attaching" and would like it rephrased to something like "do this before/after you call designer.attachTo" instead of "do this before/after attaching the designer"?

Sorry? I am simply asking for the description of the PR to be more descriptive following code review/PR hygiene; Codeflow i think broke the comment conversation here (my bad) but this is wrt change for the uninitiated. So that comment you had above explaining the "good user experience" and avoiding accidentally using older card versions.

dclaux:
Ah, I misunderstood, sorry. Makes sense.

shalinijoshi19:
@matthidinger, is this just a stale PR at this point that can be closed out?

matthidinger:
I'm going to merge this into master and fully run through the designer and prepare for a publish with the updated docs

PR #3268, opened 2019-07-29
[.NET] Add manual versioned references to System.Net.Http 4.3.1

paulcam206:
These version revs had to be manually added (outside of VS), which is why they were missed.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3268)


RebeccaAnne:
Changes look fine. Could you update the description to include what problem this is fixing?

PR #3281, opened 2019-07-31
[Android] Fix fallback again

almedina-ms:
Adds the null check in a consequent check as it started failing again

This is needed for 1.2.1

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3281)


shalinijoshi19:
Was this a merge issue or did something else break the feature?  Do.we have unit tests to catch this regression scenario? If not, can we add one? Tx.for catching

PR #3289, opened 2019-07-31
Update custom props

almedina-ms:
Updates version

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3289)


shalinijoshi19:
Thanks @almedina-ms;   any other artifacts for the other releases that need updating similar to this one for other platforms?

PR #3292, opened 2019-08-01
[Android] Add options page for setting custom elements

almedina-ms:
Adds an extra tab for adding custom features to the mobile app so now the app by default renders all cards with the default parameters (no registrations made) but you are allowed to register and unregister stuff during runtime

Also, all custom objects were removed from the MainActivity java file to make it easier to read

This is the new tab

Screenshot_2019-08-01-17-56-22

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3292)


source/android/mobile/samples/v1.0/CustomElement.json
```diff
@@ -0,0 +1,11 @@
...
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"type": "AdaptiveCard",
+	"version": "1.0",
+	"body": [
+		{
+            "type": "blah",
+            "secret": "This is a custom element"
+		}
+	]
+}
```
paulcam206: inconsistent whitespace here

PR #3325, opened 2019-08-09
[.NET] Get .Net Standard 1.3 back in the build

### LINKED ISSUE #3323: Netstandard1.3 DLL from the latest nuget is not correctly SNK signed
# Platform

* .NET HTML
* .NET WPF
* UWP

# Author or host

Host

# Version of SDK

[Nuget 1.2.1](https://www.nuget.org/packages/AdaptiveCards/1.2.1)

# Issue

I spent all day today struggling with an issue when loading the AdaptiveCards netstandard1.3 library into my project (indirectly, via a bunch of complicated Assembly.Load() / AppDomain stuff that is only semi-relevant). Every time, my program would fail at startup with the exception:

```
FileLoadException: Could not load file or assembly 'AdaptiveCards, Version=1.2.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. Strong name validation failed. (Exception from HRESULT: 0x8013141A)
```

After much frustration I discovered that the netstandard1.3 dll was different from the other two in terms of strong name signing, in that it was only delay-signed but not "fully" validated. This is what was causing my runtime to fail to load them. Validating them manually with Sn.exe shows this:

```
C:\WINDOWS\system32>Sn -v "C:\Users\lostromb\Desktop\adaptivecards\net452\AdaptiveCards.dll"

Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Assembly 'C:\Users\lostromb\Desktop\adaptivecards\net452\AdaptiveCards.dll' is valid

C:\WINDOWS\system32>Sn -v "C:\Users\lostromb\Desktop\adaptivecards\netstandard2.0\AdaptiveCards.dll"

Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Assembly 'C:\Users\lostromb\Desktop\adaptivecards\netstandard2.0\AdaptiveCards.dll' is valid

C:\WINDOWS\system32>Sn -v "C:\Users\lostromb\Desktop\adaptivecards\netstandard1.3\AdaptiveCards.dll"

Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

C:\Users\lostromb\Desktop\adaptivecards\netstandard1.3\AdaptiveCards.dll is a delay-signed or test-signed assembly
```

Of course one option is to upgrade my project to one of the other runtimes, but my project is plugin-based and I'm trying to maintain the broadest level of compatability which is why I'd prefer to target the lower version. Would it be possible to fix this simple issue in the next release?

paulcam206:
Fixes #3323.
This reverts commit 3c63421357a55b25f3cc055aa9f28b8eb54b4be7.
This reverts commit 1974570054959ccf03135692d3f1260b3475547d.

Originally I moved us to `MarkedNet` rather than `Microsoft.MarkedNet` due to a transient dependency on a vulnerable version of `System.Net.Http`. Unfortunately, `MarkedNet` isn't published with `netstandard1.3` support. However, after making this change, I found out that we can have our cake and eat it too by forcing an explicit `System.Net.Http` dependency on an updated version.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3325)


source/dotnet/Samples/AdaptiveCards.Sample.Html/Program.cs
```diff
@@ -147,6 +147,7 @@ static int Main(string[] args)
                     }
                     catch (Exception err)
                     {
+                        Debugger.Break();
```
almedina-ms: >Debugger.Break(); [](start = 24, length = 17)

wasn't this line just removed in another PR?

RebeccaAnne:
Could you add a little more context for this change to the description?

shalinijoshi19:
If we are reverting the dependence on the new version of marked.net does that mean we'll be in violation again of the compliance gates?

source/dotnet/Samples/AdaptiveCards.Sample.Html/Program.cs
paulcam206: it was removed in one of the PRs I'm reverting. I'll get rid of it again :)

---
In reply to: [312569440](https://github.com/microsoft/AdaptiveCards/pull/3325#discussion_r312569440) [](ancestors = 312569440)

paulcam206:
I've updated the description, which should answer the question -- let me know if it doesn't :)

---
In reply to: [519997845](https://github.com/microsoft/AdaptiveCards/pull/3325#issuecomment-519997845) [](ancestors = 519997845)

shalinijoshi19:
> I've updated the description, which should answer the question -- let me know if it doesn't :)
> 
> In reply to: [519997845](https://github.com/microsoft/AdaptiveCards/pull/3325#issuecomment-519997845) [](ancestors = 519997845)

Thanks! so we are continuing though to depend on the Msft-forked version of MarkedNet? This is fine for now, maybe even required (looks like markednet isn't signed?) but i am wondering if we should also get a PR out/request out to get the original markednet to generate signed binaries?

PR #3137, opened 2019-07-02
Add support for custom typefaces in the Android renderer

patgoley:
This is a possible implementation for adding custom Typeface support to the Android renderer. See issue #271. I've not yet updated the test suite to cover this feature, but if you're open to the idea I certainly can. Thanks!

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3137)

PR #3328, opened 2019-08-09
[Android] Custom typeface support

almedina-ms:
Adds sample for https://github.com/microsoft/AdaptiveCards/pull/3137

## What was done??

I added a new option in the "Options" tab in the mobile app that when checked registers a custom font to the renderer

## How verified 

The application was run and this is the result when the font is registered

Screenshot_2019-08-09-16-23-44

Screenshot_2019-08-09-16-24-04

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3328)


source/android/mobile/licenses/OFL.txt
```diff
@@ -0,0 +1,93 @@
+Copyright 2018 The Beth Ellen Project Authors (https://github.com/googlefonts/BethEllen)
+
+This Font Software is licensed under the SIL Open Font License, Version 1.1.
+This license is copied below, and is also available with a FAQ at:
```
paulcam206: this feels like a reasonable license, but it's not on the list of generally approved licenses and I'm not a lawyer :)

are there fonts that are MIT/BSD licensed we could use? if not, we might have to got through an approval process...

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextRendererUtil.java
```diff
@@ -62,12 +62,12 @@ static Typeface getTextFormat(HostConfig hostConfig, FontType type)
         {
             typeface = Typeface.MONOSPACE;
         }
-        else if (customTypefaces.containsKey(fontFamily)) {
+        else if (customTypefaces.containsKey(fontFamily))
```
almedina-ms, line 65: hey, this is just a test

source/android/mobile/licenses/OFL.txt
almedina-ms: we went through the approval process and didn't receive a negative answer from the reviewers, so I assume it succeeded

PR #3342, opened 2019-08-15
[Shared][UWP] Markdown support for non-latin chars (>0x80)

### LINKED ISSUE #2992: non-latin characters breaks markdown on UWP renderer
# Platform

* UWP

# Author or host

Author

# Version of SDK

Latest master (~1.2.0-beta1)

# Issue

Whitespace can break (valid) markdown styling when there are non-latin characters in the string.

For example, the following elements
```
{
	"type": "TextBlock",
	"text": "以前の製品のリンクで検索",
	"size": "large"
},
{
	"type": "TextBlock",
	"text": "**以前の製品のリンクで検索**",
	"size": "large"
},
{
	"type": "TextBlock",
	"text": "How about **以前の製品のリンクで検索**",
	"size": "large"
},
{
	"type": "TextBlock",
	"text": "How about **Search**",
	"size": "large"
},
{
	"type": "TextBlock",
	"text": "How about**以前の製品のリンクで検索**",
	"size": "large"
},
{
	"type": "TextBlock",
	"text": "It's OK!\rClick [以前の製品のリンクで検索](https://www.microsoft.com)\rClick [以前の製品のリンクで検索](https://www.microsoft.com)",
	"size": "large"
},
{
	"type": "TextBlock",
	"text": "1. Click [以前の製品のリンクで検索](https://www.microsoft.com)\r2. Click [以前の製品のリンクで検索](https://www.microsoft.com)",
	"size": "large"
},
{
	"type": "TextBlock",
	"text": "It's not OK!\r1. Click [以前の製品のリンクで検索](https://www.microsoft.com)\r2. Click [以前の製品のリンクで検索](https://www.microsoft.com)",
	"size": "large"
},

```
…when rendered in the UWP visualizer, becomes...

image

The only obvious difference I could notice is markdown failing if there was whitespace in front of it (for some reason, it does not manifest with latin/ASCII characters). The issue is not limited to bold (styling), hyperlinks and lists are susceptible, too. 

I did a quick check and the JS visualizer at adaptivecards.io DOES render these more to expectation:

image

### LINKED ISSUE #3337: [UWP] Mixing lists and links in markdown breaks rendering
A combination of leading text, lists, and links renders incorrectly in UWP (fine in .NET WPF and JS).

```json
{
  "version": "1.0",
  "type": "AdaptiveCard",
  "body": [
    {
      "type": "TextBlock",
      "text": "Leading text\r1. Click [link1](https://adaptivecards.io/)\r2. Click [link2](https://adaptivecards.io/)"
    }
  ]
}
```

Renders in UWP like this:
image
(note that if you omit the leading text, it renders correctly)

For reference, here's how it looks in .NET WPF:
image

paulcam206:
Fixes #2992
Fixes #3337

Ultimately, our broken markdown parsing behavior with non-latin characters comes down to our use of CRT character class functions (i.e. `isalpha()`, `isspace()`, `ispunct()`, and `isalnum()`). These functions were written for use with 7-bit ASCII. Their behavior is officially undefined when a `char`'s high bit is set (though they will fire `assert`s in debug with MSVCRT). It probably goes without saying, but UTF-8 characters often have the high bit set, leading to incorrect markdown parsing in the shared model.

When I first realized what the problem was, my initial inclination was to look for a different portable C++ markdown parser that could either completely suit our needs or at least be easily modified to suit our needs (we currently have our own parser because we need to be able to disable some functionality, e.g. tables). I found one promising library, but in prototyping its use, it became clear that it didn't do what we needed. Other libraries I looked at had problematic licensing, problematic packaging, or problematic dependencies. 

Another option, perhaps, would be to move to UTF-8-aware character classification functions. The recommended option, ICU, is complex, and would likely be overkill for our scenario.

So I opted for the option of treating high-bit UTF-8 characters as non-space non-punctuation characters (there's [precedent](https://github.com/vmg/sundown#unicode-character-handling) for this approach). While I was at it, I did some cleanup of our markdown parsing code. It's probably worth noting that we use standard signed `char`s in our codebase, so high-bit characters are `<0`.

Once I had unit tests generating correct markdown for our test scenarios (if you're wondering, the unit test payloads come straight from #2992), I tested in UWP. This surfaced a new issue I filed as #3337. The gist of the problem is that the shared model hands HTML to the UWP renderer, which then uses an XML parser to iterate over the content and generate XAML elements. Unfortunately, a string like `foo\r1. bar\r2. baz` spit out something like `<p>foo</p><ol start="1"><li>bar</li><li>baz</li></ol>`, which has two toplevel nodes, which is invalid XML. I fixed this by wrapping shared model HTML output with a `<root>` tag, which we then strip back out for content. This still left a visual problem where there's no line break after the leading `foo` paragraph, so I added extra logic to put linebreaks after paragraphs (but only if there's more content after them).

Verified with unit tests and UWP test app.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3342)


nesalang:
I see; so your approach is that since delimiters types with '+' sign won't happen between UTF-8 encoding bytes size bigger than 1; therefore adding the check for the sign bit is sufficient.

source/shared/cpp/ObjectModel/MarkDownBlockParser.cpp
```diff
@@ -655,7 +676,7 @@ void OrderedListParser::Match(std::stringstream& stream)
 {
     // used to capture digit char
     std::string number_string = "";
-    if (isdigit(stream.peek()))
+    if (stream.peek() > 0 && isdigit(stream.peek()))
```
RebeccaAnne: >stream.peek() > 0 [](start = 8, length = 17)

Do you think it might make this easier to read if we put the 0 comparison check behind a helpfully named inline function? (IsExtendedCharacter or something?) There's a lot of places where we use this. We could even go so far as to have our own is digit function (and others) that includes that check so in the logic we're just checking for digit and we don't have the cognitive overhead about thinking about this issue at every check.

source/uwp/Renderer/lib/TextHelpers.cpp
```diff
@@ -177,7 +177,8 @@ HRESULT SetXamlInlines(_In_ ABI::AdaptiveNamespace::IAdaptiveTextElement* adapti
     if (markdownParser.HasHtmlTags())
     {
         HString htmlHString;
-        UTF8ToHString(htmlString, htmlHString.GetAddressOf());
+        // wrap html in <root></root> in case there's more than one toplevel element.
```
RebeccaAnne: >// wrap html in <root></root> in case there's more than one toplevel element. [](start = 8, length = 77)

Do we want to do this in the markdown parser itself? (or does that break other platforms?)

Also, out of curiosity, did you ever hunt down whether this issue had always existed or was new?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/MarkDownUnitTest.cpp
```diff
@@ -552,22 +604,70 @@ namespace AdaptiveCardsSharedModelUnitTest
...
+
+        TEST_METHOD(NonLatinCharacters_BoldSentence)
+        {
+            MarkDownParser parser("How about **以前の製品のリンクで検索**");
+            Assert::AreEqual<std::string>("<p>How about <strong>以前の製品のリンクで検索</strong></p>", parser.TransformToHtml());
+        }
+
+        TEST_METHOD(NonLatinCharacters_BoldSentenceNoWhitespace)
+        {
+            MarkDownParser parser("How about**以前の製品のリンクで検索**");
```
nesalang: would be nice to have (a) test(s) that have non-Latin characters appear before emphasis opening delimiters and / or after emphasis closing delimiters.

source/shared/cpp/ObjectModel/MarkDownBlockParser.cpp
paulcam206: I was a bit on the fence about it, but I'll take another look. One complicating factor is EOF, which is `0xFF`... Is that an extended character? I didn't know how to answer, so I ended up doing nothing.

---
In reply to: [314840442](https://github.com/microsoft/AdaptiveCards/pull/3342#discussion_r314840442) [](ancestors = 314840442)

source/uwp/Renderer/lib/TextHelpers.cpp
paulcam206: I think it might break other platforms -- particularly since I think Android (and possibly iOS) use HTML parsers to handle markdown output.

AFAIK this behavior has always existed. I filed #3337 to describe the behavior using latin chars...

---
In reply to: [314841203](https://github.com/microsoft/AdaptiveCards/pull/3342#discussion_r314841203) [](ancestors = 314841203)

shalinijoshi19:
This is great thanks for fixing Paul. Can you also loop in Axel on this review, see what his take is? I am a little nervous about this change (thanks for including the precedent link which makes me slightly less nervous); What was the barrier to use the ms-icu library again?

paulcam206:
> This is great thanks for fixing Paul. Can you also loop in Axel on this review, see what his take is? I am a little nervous about this change (thanks for including the precedent link which makes me slightly less nervous); What was the barrier to use the ms-icu library again?

Digging a *bit* deeper, I looked at the CommonMark spec over the weekend. For context, CommonMark is an attempt by some of the early adopters of markdown (reddit, github, stackexchange, et. al) to place create a proper spec on the markdown language. Previously, it was a loose set of rules largely based on the original markdown implementation (including bugs). CommonMark mentions of whitespace:

> A whitespace character is a space (`U+0020`), tab (`U+0009`), newline (`U+000A`), line tabulation (`U+000B`), form feed (`U+000C`), or carriage return (`U+000D`).
>
> Whitespace is a sequence of one or more whitespace characters.

and Unicode whitespace:

> A Unicode whitespace character is any code point in the Unicode Zs general category, or a tab (`U+0009`), carriage return (`U+000D`), newline (`U+000A`), or form feed (`U+000C`).
>
> Unicode whitespace is a sequence of one or more Unicode whitespace characters.

Thus, "whitespace" is ASCII whitespace, and "Unicode whitespace" is a superset of ASCII including [Unicode Zs characters](https://en.wikipedia.org/wiki/Unicode_character_property#General_Category).

The only place where Unicode whitespace is referenced in the spec is in [Emphasis and Strong Emphasis](https://spec.commonmark.org/0.29/#emphasis-and-strong-emphasis), where these rules are defined for determining whether a encountered `*`, `**`, `_`, or `__` are delimiter runs denoting emphasis:

> A left-flanking delimiter run is a delimiter run that is (1) not followed by Unicode whitespace, and either (2a) not followed by a punctuation character, or (2b) followed by a punctuation character and preceded by Unicode whitespace or a punctuation character. For purposes of this definition, the beginning and the end of the line count as Unicode whitespace.
> 
> A right-flanking delimiter run is a delimiter run that is (1) not preceded by Unicode whitespace, and either (2a) not preceded by a punctuation character, or (2b) preceded by a punctuation character and followed by Unicode whitespace or a punctuation character. For purposes of this definition, the beginning and the end of the line count as Unicode whitespace.

So our implementation will be non-conforming in these specific instances. Given that handling Unicode space will require more smarts around UTF-8, I've filed #3348 to track this work.

PR #3354, opened 2019-08-19
Fixes TextInput InlineAction with show card crashes 

### LINKED ISSUE #3295: [iOS] TextInput InlineAction with show card crashes
# Platform

* iOS

# Version of SDK

master

# Issue

If you add an inline action to a text input of type show card and click on it, the process crashes. ShowCard actions should not be allowed as inline actions and they shouldn't be tried to be rendered at all

nesalang:
Fixes #3295 

### Summary of Issue
if show card is added as QuickReply action, adaptive cards crashes.
although  the current schema doesn't support showcard action, it should not crash.

### what's implemented
- show card issue is fixed
- missing toggle visibility action is added
- ACRTargetBuilderDirector manages building and configuring targets for buttons and gesture handler.  as new kind of target will be added and existing targets become more complicated to build,
  this module will help ease the burden of maintaining and implementing new targets.
  To add a new target, the class for the target with an initializer should be added
  then target builder for the new target should be added. target builder should inherit
  ACRTargetBuilder which has two methods, build and build: forButton
      they both build targets, but build: forButton designed to provide place to add target to
      button, and configure the button
  then the new builder is added to _builders; _builders is an array of dictionary
  each entry in the array is a capability
  a capability indicates available targets for each entry.
  ACRAction is associated with actions in adaptive card
  ACRSelectAction is associated with select action in adaptive card
  ACRQuickReply is associated with input.inline

  if more arguments are needed for building target than provided by the ACRITargetBuilder
  protocol, parameters should be added via this module, director, to maintain clean and unified
  interface.

  target builders are singletons since they do exactly same works regardless of renderer
  instances. however builders' director can have different configuration per rendering, and having
  single instance per renderer is preferred. for these reasons, ACRView owns the director, and
  clients who wish to use the director's service should access the service through two functions
  in UtiliOS.h, buildTarget and buildTargetForButton

### How verified
target being added did not have selector for show card action; thus the crash.
tested using json and made sure showcard action doesn't cause crash.
tested using json with toggle visibility action and made sure that toggle visibility works

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3354)


shalinijoshi19:
Thanks for fixing. @Nesalang, how hard is it to add a unit test to simulate the crash scenario? Can you get one added?

shalinijoshi19:
Thanks for fixing the description; Can you also fix the title of the PR to be a bit more descriptive? [Issue #3295] Fix the crash in ..etc

nesalang:
it requires user interaction, so it would be hard. there is a way to test user interactions, but we have to record exact steps. Since the files location can change, the recording the steps will be brittle.

---
In reply to: [523593500](https://github.com/microsoft/AdaptiveCards/pull/3354#issuecomment-523593500) [](ancestors = 523593500)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRInputRenderer.mm
```diff
@@ -217,20 +218,30 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
...
-            case ActionType::ShowCard:
-                [button addTarget:target action:@selector(toggleVisibilityOfShowCard) forControlEvents:UIControlEventTouchUpInside];
-                break;
+
             case ActionType::Submit:
                 [txtInput setReturnKeyType:UIReturnKeySend];
                 quickReplyView.target = target;
             case ActionType::OpenUrl:
                 [button addTarget:target action:@selector(send:) forControlEvents:UIControlEventTouchUpInside];
+                [viewGroup addTarget:target];
```
shalinijoshi19: nit: enclose within opening/ending curly braces

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRInputRenderer.mm
```diff
@@ -217,20 +218,30 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
         ACRAggregateTarget *target = [[ACRAggregateTarget alloc] initWithActionElement:acoAction rootView:rootView];
 
         switch (action->GetElementType()) {
-            case ActionType::ShowCard:
-                [button addTarget:target action:@selector(toggleVisibilityOfShowCard) forControlEvents:UIControlEventTouchUpInside];
-                break;
+
             case ActionType::Submit:
                 [txtInput setReturnKeyType:UIReturnKeySend];
                 quickReplyView.target = target;
```
shalinijoshi19: same: consistency issue here around the usage of curly braces (we should use them like you do in the new block of code you added); Is there an editorconfig/clang-formatting rule or something that we can enable to catch and auto-fix these syntax rules? @paulcam206?

paulcam206: filed #3381 to track this ask

---
In reply to: [316344102](https://github.com/microsoft/AdaptiveCards/pull/3354#discussion_r316344102) [](ancestors = 316344102)

bviglietta:
> //  Util.mm

I don't know why we need to put the file name inside each file, but if we're going to do it, it should be the right file name.

Same for the header file.

---
Refers to: source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/UtiliOS.mm:3 in 8547a4c. [](commit_id = 8547a4c1a15a105e1137cca64a0673febfa55a52, deletion_comment = False)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRErrors.h
```diff
@@ -19,6 +19,13 @@ typedef NS_ENUM(NSUInteger, ACRInputError)
     ACRInputErrorGreaterThanMaxDate,
 };
 
+typedef NS_ENUM(NSUInteger, ACRRenderingError)
```
bviglietta: >ACRRenderingError [](start = 28, length = 17)

Minor: I would call this ACRRenderingStatus or ACRRenderingResult, since they aren't all errors.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTargetBuilderDirector.mm
```diff
@@ -0,0 +1,268 @@
...
+        self.adaptiveHostConfig = adaptiveHostConfig;
+
+        NSNumber *openUrl = [NSNumber numberWithInt:static_cast<int>(ActionType::OpenUrl)];
+        NSNumber *submit = [NSNumber numberWithInt:static_cast<int>(ActionType::Submit)];
+        NSNumber *showcard = [NSNumber numberWithInt:static_cast<int>(ActionType::ShowCard)];
+        NSNumber *toggle = [NSNumber numberWithInt:static_cast<int>(ActionType::ToggleVisibility)];
+
+        // target capabilities list supported events and corresponding target builders
+        _builders = @[
+            // ACRAction
```
bviglietta: >// ACRAction [](start = 12, length = 12)

Kind of fragile to assume the enum values like this. If someone changes the enum without updating this code, it will continue to build but won't work correctly.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTargetBuilderDirector.mm
```diff
@@ -0,0 +1,268 @@
...
+        return nil;
+    }
+
+    ACRTargetBuilder *builder = [self getBuilder:action];
+    return [builder build:action director:self];
+}
+
+- (NSObject *)build:(std::shared_ptr<BaseActionElement> const &)action
+          forButton:(UIButton *)button {
+    if (![self checkAgainstMyCapability:action]) {
```
bviglietta: >![self checkAgainstMyCapability:action] [](start = 8, length = 39)

This is a very confusing way of writing this logic. It looks like `checkAgainstMyCapability` is returning a boolean, but it's actually returning an enum where 0 (== ACROk) happens to be the success value.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTargetBuilderDirector.mm
bviglietta: There's no need for this to be an array, is there? Each BuilderDirector is used for just one capability, so we only need to keep track of the builders for that one capability, not for all capabilities.

---
In reply to: [321506664](https://github.com/microsoft/AdaptiveCards/pull/3354#discussion_r321506664) [](ancestors = 321506664)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRErrors.h
nesalang: thanks; make sense

---
In reply to: [321498435](https://github.com/microsoft/AdaptiveCards/pull/3354#discussion_r321498435) [](ancestors = 321498435)

nesalang:
> //  Util.mm

thansk

---
In reply to: [528600878](https://github.com/microsoft/AdaptiveCards/pull/3354#issuecomment-528600878) [](ancestors = 528600878)

---
Refers to: source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/Util.mm:3 in 8547a4c. [](commit_id = 8547a4c1a15a105e1137cca64a0673febfa55a52, deletion_comment = False)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTargetBuilderDirector.mm
nesalang: I agree with you on the fragility. are you suggesting to have different classes of BuilderDirector? would that be too many different types of directors when their difference is only a capability.

---
In reply to: [321509166](https://github.com/microsoft/AdaptiveCards/pull/3354#discussion_r321509166) [](ancestors = 321509166,321506664)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTargetBuilderDirector.mm
nesalang: thanks; 

---
In reply to: [321507661](https://github.com/microsoft/AdaptiveCards/pull/3354#discussion_r321507661) [](ancestors = 321507661)

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTargetBuilderDirector.mm
bviglietta: No, not different classes. Just each instance only maintains info for its own set of builders.


```
switch (_builderCapability)
{
    case ACRAction:
        _builders = @{ openUrl : openUrlBuilder, submit : submitBuilder, etc. };
        break;
    case ACRSelectAction:
        _builders = // etc
}
```

---
In reply to: [321511766](https://github.com/microsoft/AdaptiveCards/pull/3354#discussion_r321511766) [](ancestors = 321511766,321509166,321506664)

nesalang: got it thanks

---
In reply to: [321519397](https://github.com/microsoft/AdaptiveCards/pull/3354#discussion_r321519397) [](ancestors = 321519397,321511766,321509166,321506664)

PR #3375, opened 2019-08-21
[UWP] Fix bugs in custom warnings

### LINKED ISSUE #3373: [UWP][Custom Rendering] AdaptiveWarning doesn't include putters, making it impossible to set status code and message
# Platform

* UWP

# Details
AdaptiveWarning is defined as follows:
```
    runtimeclass AdaptiveWarning
    {
        AdaptiveWarning();

        WarningStatusCode StatusCode { get; };
        String Message { get; };
    };
```

Because there are no putters, custom parsers and renderers can not specify the nature of their warnings.

### LINKED ISSUE #3374: [UWP][Custom Rendering] UWP loses warnings when converting to the shared model
# Platform

* UWP

# Details

The method that converts UWP warnings to shared model warnings is defined as follows
```
HRESULT AdaptiveWarningsToSharedWarnings(
    _In_  IVector<AdaptiveWarning*>* adaptiveWarnings,
   std::vector<std::shared_ptr<AdaptiveCardParseWarning>> sharedWarnings)
```

Due to the fact that the sharedWarnings vector is passed by value, the vector passed in by the calling function is not updated and all warnings are lost.

RebeccaAnne:
## Related Issues
Fixes #3374, Fixes #3373

## Description
Fixes three issues related to setting warnings and errors from a custom parsers and renders:
- Made the status and message properties read/write rather than readonly so that custom parsers and renderers can set them
- Added CustomWarning and CustomError types to the status code enums to allow custom clients to indicate that their error or warning doesn't fall into one of our existing categories
- Fixed an issue where a vector was passed by value when it should have been passed by reference, leading to the vector not being updated as expected.

## How Verified
Verified a custom parser could set warnings that were propagated to the visualizer using local code and manual testing. Automated and checked in custom rendering testing that will cover this scenario coming soon to a test app near you via #2720.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3375)

### DIFF
--- a/source/shared/cpp/ObjectModel/Enums.h
+++ b/source/shared/cpp/ObjectModel/Enums.h
@@ -371,7 +371,8 @@ namespace AdaptiveSharedNamespace
         RequiredPropertyMissing,
         InvalidPropertyValue,
         UnsupportedParserOverride,
-        IdCollision
+        IdCollision,
+        CustomError,
     };
     // No mapping to string needed
 
@@ -391,6 +392,7 @@ namespace AdaptiveSharedNamespace
         InvalidColorFormat,
         InvalidDimensionSpecified,
         InvalidLanguage,
+        CustomWarning,
     };
     // No mapping to string needed
 
--- a/source/android/adaptivecards/src/main/cpp/objectmodel_wrap.cpp
+++ b/source/android/adaptivecards/src/main/cpp/objectmodel_wrap.cpp
@@ -19608,6 +19608,24 @@ SWIGEXPORT jstring JNICALL Java_io_adaptivecards_objectmodel_AdaptiveCardObjectM
 }
 
 
+SWIGEXPORT jlong JNICALL Java_io_adaptivecards_objectmodel_AdaptiveCardObjectModelJNI_AdaptiveCard_1GetInternalId(JNIEnv *jenv, jclass jcls, jlong jarg1, jobject jarg1_) {
+  jlong jresult = 0 ;
+  AdaptiveCards::AdaptiveCard *arg1 = (AdaptiveCards::AdaptiveCard *) 0 ;
+  std::shared_ptr< AdaptiveCards::AdaptiveCard const > *smartarg1 = 0 ;
+  AdaptiveCards::InternalId result;
+  
+  (void)jenv;
+  (void)jcls;
+  (void)jarg1_;
+  
+  smartarg1 = *(std::shared_ptr< const AdaptiveCards::AdaptiveCard > **)&jarg1;
+  arg1 = (AdaptiveCards::AdaptiveCard *)(smartarg1 ? smartarg1->get() : 0); 
+  result = ((AdaptiveCards::AdaptiveCard const *)arg1)->GetInternalId();
+  *(AdaptiveCards::InternalId **)&jresult = new AdaptiveCards::InternalId((const AdaptiveCards::InternalId &)result); 
+  return jresult;
+}
+
+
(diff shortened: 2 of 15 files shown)


source/shared/cpp/ObjectModel/Enums.h
```diff
@@ -371,7 +371,8 @@ namespace AdaptiveSharedNamespace
         RequiredPropertyMissing,
         InvalidPropertyValue,
         UnsupportedParserOverride,
-        IdCollision
+        IdCollision,
+        CustomError,
```
paulcam206: do enum changes require a re-swig?

shalinijoshi19: >CustomError [](start = 8, length = 11)

Nit: CustomError sounds odd ? Would Default be better? (Is this similar to an E_FAIL?

source/shared/cpp/ObjectModel/Enums.h
```diff
@@ -391,6 +392,7 @@ namespace AdaptiveSharedNamespace
         InvalidColorFormat,
         InvalidDimensionSpecified,
         InvalidLanguage,
+        CustomWarning,
```
shalinijoshi19: >CustomWarning [](start = 8, length = 13)

Similarly "CustomWarning" seems an odd enumeraiton value -- having a notion of Default seems more intuitive?

source/shared/cpp/ObjectModel/Enums.h
RebeccaAnne: Indeed.

---
In reply to: [316819721](https://github.com/microsoft/AdaptiveCards/pull/3375#discussion_r316819721) [](ancestors = 316819721)

RebeccaAnne: This basically means that the error is not one of the known errors returned by the parser/renderer, and is a "custom" error provided by, for example, a custom parser. I chose "Custom" thinking of consistency with the Custom enum values in CardElementType and ActionType.

As implemented, it's not really "Default" in any meaningful way (the error object doesn't default to that value, for example). Default error or warning code doesn't necessarily seem less odd to me (what does a "default" warning mean to someone consuming it?), but perhaps "Other" or "Unknown"? I think i still like "Custom" best, as it seems to me to convey the information that this isn't a known type but has been explicitly customized (i.e., there's likely something interesting in the message). What do you think?

Thinking about defaults though, i wonder if the constructor should require an error/warning status code to be passed in to ensure a meaningful object here...

---
In reply to: [316942726](https://github.com/microsoft/AdaptiveCards/pull/3375#discussion_r316942726) [](ancestors = 316942726)

paulcam206: FWIW, I'm in agreement on `CustomX` here for consistency reasons. Implementing a "custom" renderer and a "custom" parser feels like it might require "custom" errors and warnings.

---
In reply to: [317235188](https://github.com/microsoft/AdaptiveCards/pull/3375#discussion_r317235188) [](ancestors = 317235188,316942726)

RebeccaAnne: I did update the constructor to require status code and string. Shared model already does this and as we don't actually have a default value, requiring one to be passed in makes sense.

I'm going to stick with CustomX for now. @shalinijoshi19 feel free to follow up when you're back if you have further concerns.

---
In reply to: [317236272](https://github.com/microsoft/AdaptiveCards/pull/3375#discussion_r317236272) [](ancestors = 317236272,317235188,316942726)

PR #3380, opened 2019-08-22
[UWP] Fix leak in IsBackgroundImageValid

### LINKED ISSUE #3372: [UWP] Memory leak in IsBackgroundImageValid
https://github.com/microsoft/AdaptiveCards/blob/10dd6f373de16f845539967759454235bf36b9d8/source/uwp/Renderer/lib/Util.cpp#L1313-L1328

`url` gets leaked because it's an `HSTRING`, not an `HString`

paulcam206:
Fixes #3372

## Description
We were accidentally using an `HSTRING*` (bare pointer) rather than an `HString` (WRL smart ptr) in `IsBackgroundImageValid()`, and weren't calling the appropriate deallocator. Fix is to move to `HString`.

## How Verified
Verified with UWP test app and debugger stepping

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3380)


shalinijoshi19:
> HRESULT JsonObjectToJsonCpp(_In_ ABI::Windows::Data::Json::IJsonObject* jsonObject, _Out_ Json::Value* jsonCppValue)

nit: Unrelated but something to think about and address -- In general (zero)initialize out parameters first thing. refer osgwiki ([https://osgwiki.com/wiki/Cpp_Core_Guidelines#OUT:_Output_Parameters](https://osgwiki.com/wiki/Cpp_Core_Guidelines#OUT:_Output_Parameters)) .. 

---
Refers to: source/uwp/Renderer/lib/Util.cpp:1290 in fa4f17a. [](commit_id = fa4f17aba6bb77f857b428516c05fb79a7a28ac5, deletion_comment = False)

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -1312,18 +1312,14 @@ HRESULT ProjectedElementTypeToHString(ABI::AdaptiveNamespace::ElementType projec
 
 HRESULT IsBackgroundImageValid(_In_ ABI::AdaptiveNamespace::IAdaptiveBackgroundImage* backgroundImageElement, _Out_ BOOL* isValid)
```
shalinijoshi19: Curious -- why an error contract here for Utils?

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -1312,18 +1312,14 @@ HRESULT ProjectedElementTypeToHString(ABI::AdaptiveNamespace::ElementType projec
...
     {
-        HSTRING url;
-        THROW_IF_FAILED(backgroundImage->get_Url(&url));
-        if (url != NULL)
-        {
-            *isValid = TRUE;
-            return S_OK;
-        }
+        HString url;
+        THROW_IF_FAILED(backgroundImage->get_Url(url.GetAddressOf()));
```
shalinijoshi19: >THROW_IF_FAILED [](start = 8, length = 15)

hmm we are mixing exceptions wtih suposedly error contracts here.. RETURN_IF_FAILED()

shalinijoshi19:
Do we have unit tests that exercise the code path that we could potentially turn into perf tests? If not, let's add them?

source/uwp/Renderer/lib/Util.cpp
paulcam206: in general we have a mix because a lot of these helpers were originally written as throwing. however, it became apparent that in some scenarios exceptions weren't sufficient (IIRC, there was a problem where thrown exceptions wouldn't cross our ABI boundaries correctly or something). #2534 covers reconciling our error handling regimes. that said, I did find at least one place where an exception from here wouldn't get properly handled, so I'll make this `RETURN_IF_FAILED()`.

---
In reply to: [316950627](https://github.com/microsoft/AdaptiveCards/pull/3380#discussion_r316950627) [](ancestors = 316950627)

paulcam206:
> HRESULT JsonObjectToJsonCpp(_In_ ABI::Windows::Data::Json::IJsonObject* jsonObject, _Out_ Json::Value* jsonCppValue)

interesting -- I didn't realize that we were using this guidance for non-public-facing APIs. the old rules said that if the function is a public API, it should zero out params, but if it was an internal helper, it was explicitly preferred *not* to initialize since you control the callers and can define the contract however you see fit. I filed #3384 to track.

---
In reply to: [524142178](https://github.com/microsoft/AdaptiveCards/pull/3380#issuecomment-524142178) [](ancestors = 524142178)

---
Refers to: source/uwp/Renderer/lib/Util.cpp:1290 in fa4f17a. [](commit_id = fa4f17aba6bb77f857b428516c05fb79a7a28ac5, deletion_comment = False)

source/uwp/Renderer/lib/Util.cpp
paulcam206: I dunno. @RebeccaAnne?

---
In reply to: [316949926](https://github.com/microsoft/AdaptiveCards/pull/3380#discussion_r316949926) [](ancestors = 316949926)

paulcam206:
The `Card` object model test hits this codepath. We don't currently have any tracking of these tests for perf.

---
In reply to: [524143619](https://github.com/microsoft/AdaptiveCards/pull/3380#issuecomment-524143619) [](ancestors = 524143619)

source/uwp/Renderer/lib/Util.cpp
RebeccaAnne: I tend to think all our UWP code should be error contract, as that's both what we need to provide at our API boundary, and what most of the things we consume present to us as WinRT APIs used from WRL (Xaml, Collection types, our own WinRT object model, etc.).

One hiccup is that shared model (and std:: classes used by it, and by us internally) are all exception contracts. Even so, as most of what we're doing here is WinRT WRL code, and the shared model stuff is mostly localized to the object model classes, i think error contracts throughout and wrapping the throwers makes the most sense.  (as called out in #2534)

That said, i don't know enough to have an opinion about whether that will still be the right answer if/when we move to C++/WinRT. (a brief web search looks like the C++/WinRT projection is exception based, so that probably changes my position here depending on when and to what extent we move over to C++/WinRT)

---
In reply to: [317215774](https://github.com/microsoft/AdaptiveCards/pull/3380#discussion_r317215774) [](ancestors = 317215774,316949926)

paulcam206:
@shalinijoshi19 - can you take another look? thanks!

shalinijoshi19:
Great thanks! Let's file a separate issue here so we can start tracking potential perf scenarios? @RebeccaAnne for FYI

---
In reply to: [524387679](https://github.com/microsoft/AdaptiveCards/pull/3380#issuecomment-524387679) [](ancestors = 524387679,524143619)

source/uwp/Renderer/lib/Util.cpp
shalinijoshi19: Lets follow up here on best practices etc in case we have doubts with JonWis & API guidance crew please

---
In reply to: [317223062](https://github.com/microsoft/AdaptiveCards/pull/3380#discussion_r317223062) [](ancestors = 317223062,317215774,316949926)

shalinijoshi19:
Looks good - but let's file tracking issues and/or follow up on the error contract related issues offline as well. Also nit but i updated the description to have a clear Description + How Verified title in the PR description :) (the content was there just added the headings.)

---
In reply to: [525857790](https://github.com/microsoft/AdaptiveCards/pull/3380#issuecomment-525857790) [](ancestors = 525857790)

PR #3403, opened 2019-08-28
Implemented min and max value range test for Input.Date

### LINKED ISSUE #3174: [UWP] Visualizer crashes with erroneous values on Input.Date
# Platform

* UWP

# Issue

The UWP visualizer crashes when the JSON has faulty min and max values on `Input.Date` (when max comes before min). The visualizer initially renders the card, but when clicking the button to expand the date picker, the application crashes.

```
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "Input.Date",
            "title": "New Input.Toggle",
            "id": "asdf",
            "min": "2017-08-01",
            "max": "2017-07-08"
        }
    ],
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json"
}
```

# Result:

image
Clicking the calendar icon will cause a crash.

nesalang:
## Related Issue
Fixes #3174 

## Description
XML will throws at run time when min date value is larger than max date value for Input.Date, and this will be an unhandled exception.
PR ensures that if min and max are valid date, min is less than max.
if either of them are invalid dates, then only the valid date is set.

## How Verified
1. Ran the test app to guard against regression.
2. test json was used with invalid mix of dates to test.
3. screenshot

Screenshot (10)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3403)


nesalang:
azpn /run

nesalang:
azpl /run

Issue #3420, opened 2019-09-04
[UWP][Fallback Requires] Requirement comparison is case sensitive


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* UWP

# Details

The following card will drop it's text block because the casing on the requirement is "AdaptiveCards" rather than the expected "adaptiveCards" camel casing. This text block is not dropped on .NET
```
{
  "version": "1.2",
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "type": "AdaptiveCard",
  "body": [
    {
      "type": "TextBlock",
      "text": "This textblock requires adaptiveCards 1.0",
      "color": "Good",
      "wrap": true,
      "requires": {
        "AdaptiveCards": "1.0"
      }
    }
  ]
}
```

shalinijoshi19:
@almedina-ms FYI assigning to you for fixing.

Issue #3418, opened 2019-09-04
[UWP][Fallback Requirements] A element that fails its requirements reports "No Renderer found"


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* UWP

# Details

For the following card which requires a future version of adaptive cards:
```
{
  "version": "1.2",
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "type": "AdaptiveCard",
  "body": [
    {
      "type": "TextBlock",
      "text": "This textblock requires adaptiveCards 5.0",
      "color": "Attention",
      "wrap": true,
      "requires": {
        "adaptiveCards": "5.0"
      }
    }
  ]
}
```

The UWP renderer reports:
![image](https://user-images.githubusercontent.com/9091833/64288768-bd61a000-cf17-11e9-8420-4f305c31c6a7.png)

This is an inaccurate and confusing warning message as there **is** a renderer for TextBlock.

jonmill:
Closing as fallback is not used

PR #3421, opened 2019-09-04
[UWP] Change MeetsRequirements method to Requirements property

### LINKED ISSUE #3409: [UWP][Fallback Requirements] Change requirements handling from a function to a property
# Platform

* UWP

# Details

Currently ```IAdaptiveCardElement``` has a ```MeetsRequirements()``` method on it to determine whether the element meets its requirements. For our known types, a single implementation exists in the base class that is used by all elements. For custom types, however, defining the requirement handling this way requires custom type developers to provide their own implementation of ```MeetsRequirements``` for their type.

To address this, we should change the ```MeetsRequirements``` function to a ```Requirements``` property which simply contains the requirements. We can then run our logic on that property to determine if it meets the requirements, removing the need from third party developers to understand and re-implement that logic in their own types.

RebeccaAnne:
## Related Issue
Fixes #3409 

## Description
Changed the MeetsRequirements function on IAdaptiveCardElement to a Requirements property. This has two benefits:

1. This removes the burden on custom element renderers needing to implement the MeetsRequirements function. There's no expectation that this method would be different on different elements, so this doesn't need to be a per-element implementable function.

2. Allows callers to set element requirements programmatically in the case where they are building the object model. In the previous API design, there was no way for a caller to add a requirement to an element using the object model API.

## How Verified
Object model and functional UWP unit tests (opened #3418 and #3420 for existing issues). Ran the test app.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3421)

### DIFF
--- a/source/uwp/Renderer/lib/AdaptiveRequirement.h
+++ b/source/uwp/Renderer/lib/AdaptiveRequirement.h
@@ -1,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#pragma once
+
+#include "AdaptiveCards.Rendering.Uwp.h"
+
+namespace AdaptiveNamespace
+{
+    class AdaptiveRequirement
+        : public Microsoft::WRL::RuntimeClass<Microsoft::WRL::RuntimeClassFlags<Microsoft::WRL::RuntimeClassType::WinRtClassicComMix>,
+                                              Microsoft::WRL::Implements<ABI::AdaptiveNamespace::IAdaptiveRequirement>,
+                                              Microsoft::WRL::FtmBase>
+    {
+        AdaptiveRuntime(AdaptiveRequirement);
+
+    public:
+        HRESULT RuntimeClassInitialize();
+        HRESULT RuntimeClassInitialize(HSTRING name, HSTRING version);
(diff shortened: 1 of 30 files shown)


source/uwp/Renderer/lib/AdaptiveRequirement.h
```diff
@@ -0,0 +1,44 @@
...
+    class AdaptiveRequirement
+        : public Microsoft::WRL::RuntimeClass<Microsoft::WRL::RuntimeClassFlags<Microsoft::WRL::RuntimeClassType::WinRtClassicComMix>,
+                                              Microsoft::WRL::Implements<ABI::AdaptiveNamespace::IAdaptiveRequirement>,
+                                              Microsoft::WRL::FtmBase>
+    {
+        AdaptiveRuntime(AdaptiveRequirement);
+
+    public:
+        HRESULT RuntimeClassInitialize();
+        HRESULT RuntimeClassInitialize(HSTRING name, HSTRING version);
```
paulcam206: >(HSTRING name, HSTRING version); [](start = 38, length = 32)

missing SAL

PR #3423, opened 2019-09-05
Fix ActionSet description

v-kydela:
## Description
Just fixing a typo

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3423)

PR #3424, opened 2019-09-05
Fix ActionSet description in 1.2.0

v-kydela:
## Related Issue
https://github.com/microsoft/AdaptiveCards/pull/3423

## Description
Backwards propagation of this fix

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3424)


shalinijoshi19:
Out of curiosity is this backport really needed @v-kydela / @dclaux/ @andrewleader ?

dclaux:
Why not? It's a benign change that only increases the quality of the schema.

Here is what I told Kyle:
- Harmless changes such as the one you made (description) should probably be applied to all older versions provided the definition existed in the older versions
- Important changes (type, required, etc., essentially anything that can generate a break in a client that would validate an incoming payload against the schema) should probably not be applied to older versions

v-kydela:
@dclaux 
@shalinijoshi19 

Please see my comment here: https://github.com/microsoft/AdaptiveCards/pull/3423#issuecomment-558400067

> Sorry for the trouble. I'm not sure why "Generate Specs" isn't passing, but I'm guessing it's comparing the schema's descriptions to some other file somewhere and noticing they don't match. I can't see the details of the error, but I presume you can. Can you help figure out why this typo fix can't be merged?

andrewleader:
@shalinijoshi19 the Generate Specs task is failing since this is a pull request from an external fork (not a branch within our repro).

The Generate Specs task attempts to switch to the branch that the changes are coming from and then update the specs, but it can't switch to the external branch, so it fails (I never thought of that scenario). I don't know of a good way to solve that. The purpose of this task is to ensure that the correct dependent files (the spec files) were produced if the source schema was changed.

@v-kydela could you update the [ActionSet.md spec file](https://github.com/v-kydela/AdaptiveCards/blob/patch-2/specs/elements/ActionSet.md) with your updated description? It needs to be updated in two places, at the top table summary, and also further down in the detailed property description.

After that, it should be safe to merge (and maybe we disable the Generate Specs tasks for external pull requests or something, idk...)

v-kydela:
Thank you @andrewleader 

I have made the change in this branch and the other branch: https://github.com/microsoft/AdaptiveCards/pull/3423

However, I believe the only reason I made this branch to begin with is because I believed that the other one had already been merged. I hadn't expected this to be drawn out so long. Should we close this PR without merging and then apply all the changes from both branches to `patch-1`, or should both branches be merged?

paulcam206:
looks like this is fixed!

PR #3426, opened 2019-09-06
[UWP][.NET][Shared] Update to VS 2019

### LINKED ISSUE #3194: [UWP,X64, VS2019] solution fails to compile with a certificate mismatch error for the 2 store apps
# Platform

* UWP

# Version of SDK

latest master with VS2019

# Issue

Trying to compile the VS solution fails with the following errors in VS2019; Confirmed that this isnt an issue with VS2017

Severity	Code	Description	Project	File	Line	Suppression State
Error		Certificate does not match supplied signing thumbprint: 7DF59BF73D990455B63DFF7735D301D0BB319CD2	AdaptiveCardTestApp			
Error		Certificate does not match supplied signing thumbprint: BD1E51E915EF811D813E0C6C4143C44144367328	AdaptiveCardVisualizer

paulcam206:
## Related Issue
Fixes #3194

## Description
Get our UWP, Shared Model, and .NET solutions building in VS2019:
* Upgrade platform from old VS2017 v141 toolchain to new hotness v142
* Some interproject dependencies didn't survive the automated upgrade process, so those were re-added
* VS2019 doesn't install temporary certificates, unlike VS2017. This means that the CI builds will fail with a mismatched package thumbprint if you attempt to build an appx without either 
  * running a script to add the cert to the store
  * removing the thumbprint from the solution (thereby bypassing signing)

Because we don't need signing integrated into our VS builds (it's handled as a separate task outside of VS), and because the signing issues are around non-shipping projects (UWPUnitTests, AdaptiveCardsTestApp, et al), I removed the thumbprints from the projects.

## How Verified
Built locally w/ VS2019 and built in CI using our VS2019 pool

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3426)


paulcam206:
Note that I manually ran jobs to validate UWP, .NET, and Shared Model, but our CI jobs haven't been pointed to the new agent pool yet, so those jobs will fail here.

RebeccaAnne:
I'm not familiar enough with Visual Studio to entirely make sense of these changes :). Could you update the description with a little more information about the changes that were required?

source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
```diff
@@ -42,13 +42,13 @@
     <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
     <AppContainerApplication>true</AppContainerApplication>
     <ApplicationType>Windows Store</ApplicationType>
-    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
-    <WindowsTargetPlatformMinVersion>10.0.16299.0</WindowsTargetPlatformMinVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformMinVersion>10.0.15063.0</WindowsTargetPlatformMinVersion>
```
paulcam206: >.0.15063.0 [](start = 39, length = 10)

I should probably back out these Windows Platform version changes...

PR #3429, opened 2019-09-09
[UWP] Ensure correct parser registrations are passed to custom parsers

### LINKED ISSUE #3427: [UWP][Custom Parsing] Element Registration passed to IAdaptiveElementParser.FromJson is missing custom types
# Platform

* UWP

# Details

Custom parser implementers implement the following interface:

```
IAdaptiveCardElement FromJson(
    JsonObject inputJson, 
    AdaptiveElementParserRegistration elementParsers, 
    AdaptiveActionParserRegistration actionParsers, 
    IList<AdaptiveWarning> warnings);
```

When called back, the elementParsers and actionParsers registrations should include all registered parsers, including custom types. In actuality, the registration only contains the known schema types, preventing a custom parser from properly handling contained elements of other custom types.

RebeccaAnne:
## Related Issue
Fixes #3427

### Problem
In order to store custom parsers from UWP callers in the shared model, a wrapper class (```SharedModelElementParser```) is used to implement the shared model parser base class (```BaseCardElementParser```). This wrapper class holds a pointer to the UWP level parser registration for the wrapped type (either action or element).

The bug was that when this wrapper class was called back, instead of using the existing UWP level parser registration, it created a new one to pass to the UWP custom parser. This new registration did not include the UWP level parser registrations, and so it appeared to the callback as if the custom registrations were missing.

### Fixes
- Registered the element and action parsers with the shared model under a known guid. This allows the callback in the UWP interface level to find the correctly populated UWP parser registrations from the shared model registration. The new methods ```GetAdaptiveActionParserRegistrationFromSharedModel``` and ```GetAdaptiveElementParserRegistrationFromSharedModel``` use this to retrieve the appropriate registration from the callback.

- Removed the UWP ```AdaptiveElementParserRegistration```/```AdaptiveActionParserRegistration         RuntimeClassInitialize``` methods that take the shared model registration as input. The existence of this method implied that the UWP interface could be initialized based on the shared model. That would require more code to walk the shared model registration and get out the corresponding UWP parsers, which does not exist and would add more complexity.

- Changed the ```m_parserRegistration``` in ```SharedModelElementParser``` and ```SharedModelActionParser``` to a weak reference. While white boarding the current design i realized that the fact that the shared model registration has a reference to the UWP registration and the UWP registration has a reference to the shared model registration creates a circular reference. Added a WeakRef to break the cycle.

## How Verified
Updated existing ```ElementParserRegistraton_CustomElementTest``` and ```ActionParserRegistraton_CustomActionTest``` unit tests to verify that the custom renderers are in the renderer registrations passed to the custom parsing callback.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3429)

### DIFF
--- a/source/uwp/Renderer/lib/AdaptiveElementParserRegistration.cpp
+++ b/source/uwp/Renderer/lib/AdaptiveElementParserRegistration.cpp
@@ -33,8 +33,6 @@ using namespace ABI::Windows::UI;
 
 namespace AdaptiveNamespace
 {
-    const char* c_upwElementParserRegistration = "447C3D76-CAAD-405F-B929-E3201F1537AB";
-
     AdaptiveElementParserRegistration::AdaptiveElementParserRegistration() {}
 
     HRESULT AdaptiveElementParserRegistration::RuntimeClassInitialize() noexcept try
@@ -47,7 +45,7 @@ namespace AdaptiveNamespace
 
         // Register this (UWP) registration with a well known guid string in the shared model
         // registration so we can get it back again
-        m_sharedParserRegistration->AddParser(c_upwElementParserRegistration, std::make_shared<SharedModelElementParser>(this));
+        m_sharedParserRegistration->AddParser(c_uwpElementParserRegistration, std::make_shared<SharedModelElementParser>(this));
 
         m_isInitializing = false;
         return S_OK;
@@ -150,10 +148,6 @@ namespace AdaptiveNamespace
 
     HRESULT SharedModelElementParser::GetAdaptiveParserRegistration(_COM_Outptr_ IAdaptiveElementParserRegistration** elementParserRegistration)
     {
-        ComPtr<IAdaptiveElementParserRegistration> parserRegistration;
-        RETURN_IF_FAILED(m_parserRegistration.As(&parserRegistration));
-        RETURN_IF_FAILED(parserRegistration.CopyTo(elementParserRegistration));
-
-        return S_OK;
+        return m_parserRegistration.CopyTo<IAdaptiveElementParserRegistration>(elementParserRegistration);
     }
 }
--- a/source/uwp/Renderer/lib/AdaptiveElementParserRegistration.cpp
+++ b/source/uwp/Renderer/lib/AdaptiveElementParserRegistration.cpp
@@ -33,24 +33,22 @@ using namespace ABI::Windows::UI;
 
 namespace AdaptiveNamespace
 {
+    const char* c_upwElementParserRegistration = "447C3D76-CAAD-405F-B929-E3201F1537AB";
+
     AdaptiveElementParserRegistration::AdaptiveElementParserRegistration() {}
 
     HRESULT AdaptiveElementParserRegistration::RuntimeClassInitialize() noexcept try
     {
-        std::shared_ptr<ElementParserRegistration> sharedParserRegistration = std::make_shared<ElementParserRegistration>();
-        RuntimeClassInitialize(sharedParserRegistration);
-        return S_OK;
-    }
-    CATCH_RETURN;
-
-    HRESULT AdaptiveElementParserRegistration::RuntimeClassInitialize(
-        std::shared_ptr<AdaptiveSharedNamespace::ElementParserRegistration> sharedParserRegistration) noexcept try
-    {
         m_registration = std::make_shared<RegistrationMap>();
-        m_sharedParserRegistration = sharedParserRegistration;
+        m_sharedParserRegistration = std::make_shared<ElementParserRegistration>();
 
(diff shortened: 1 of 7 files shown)


source/uwp/Renderer/lib/AdaptiveElementParserRegistration.cpp
```diff
@@ -33,24 +33,22 @@ using namespace ABI::Windows::UI;
 
 namespace AdaptiveNamespace
 {
+    const char* c_upwElementParserRegistration = "447C3D76-CAAD-405F-B929-E3201F1537AB";
```
paulcam206: >c_upwElementParserRegistration [](start = 16, length = 30)

should this be `c_uwpElementParserRegistration`?

paulcam206: >const char* [](start = 4, length = 11)

instead of declaring in the header and defining here, should it be a `constexpr` in the header?

source/uwp/Renderer/lib/AdaptiveElementParserRegistration.cpp
```diff
@@ -143,4 +147,13 @@ namespace AdaptiveNamespace
         std::shared_ptr<CustomElementWrapper> elementWrapper = std::make_shared<CustomElementWrapper>(cardElement.Get());
         return elementWrapper;
     }
+
+    HRESULT SharedModelElementParser::GetAdaptiveParserRegistration(_COM_Outptr_ IAdaptiveElementParserRegistration** elementParserRegistration)
+    {
+        ComPtr<IAdaptiveElementParserRegistration> parserRegistration;
+        RETURN_IF_FAILED(m_parserRegistration.As(&parserRegistration));
+        RETURN_IF_FAILED(parserRegistration.CopyTo(elementParserRegistration));
```
paulcam206: not positive, so disregard if incorrect, but can't we just do:

```
RETURN_IF_FAILED(m_parserRegistration.CopyTo(IID_PPV_ARGS(elementParserRegistration)));
```

?

source/uwp/Renderer/lib/AdaptiveElementParserRegistration.cpp
RebeccaAnne: Sorry, i'm missing something here. Looks like it is that?

---
In reply to: [322488406](https://github.com/microsoft/AdaptiveCards/pull/3429#discussion_r322488406) [](ancestors = 322488406)

RebeccaAnne: Indeed. Updated.

---
In reply to: [322488612](https://github.com/microsoft/AdaptiveCards/pull/3429#discussion_r322488612) [](ancestors = 322488612)

paulcam206: `c_upw` vs. `c_uwp` :)

---
In reply to: [322490525](https://github.com/microsoft/AdaptiveCards/pull/3429#discussion_r322490525) [](ancestors = 322490525,322488406)

source/uwp/Renderer/lib/AdaptiveElementParserRegistration.cpp
RebeccaAnne: Doesn't quite seem to work, as IID_PPV_ARGS gives you void** and we need IInspectable** here.

HRESULT Microsoft::WRL::WeakRef::CopyTo(IWeakReference **) noexcept const': cannot convert argument 2 from 'void **' to 'IInspectable **'

---
In reply to: [322489870](https://github.com/microsoft/AdaptiveCards/pull/3429#discussion_r322489870) [](ancestors = 322489870)

RebeccaAnne: Okay, fooling around with this a bit i got us down to:

return m_parserRegistration.CopyTo<IAdaptiveElementParserRegistration>(elementParserRegistration);

which seems better.

---
In reply to: [322496738](https://github.com/microsoft/AdaptiveCards/pull/3429#discussion_r322496738) [](ancestors = 322496738,322489870)

source/uwp/Renderer/lib/AdaptiveElementParserRegistration.cpp
RebeccaAnne: Ha, i looked at it several times and didn't see that. Fixed!

---
In reply to: [322490525](https://github.com/microsoft/AdaptiveCards/pull/3429#discussion_r322490525) [](ancestors = 322490525,322488406)

PR #3434, opened 2019-09-10
[SharedModel] Make requirements case insensitive

### LINKED ISSUE #3420: [UWP][Fallback Requires] Requirement comparison is case sensitive

almedina-ms:
## Related Issue
Fixes #3420 

## Description
Makes the shared model's (Android, iOS and UWP) requirement comparisson for element rendering work with case insensitive requirements, this means that "AdaptiveCards" and "adaptivecards" are the same requirement making it easier for users to correctly render their cards even if they got a small mistake when authoring their cards

## How Verified
The __RequiresCustomFeatureTest__ test was modified to check that two requirements written with mismatching capitalization were being matched correctly

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3434)


source/shared/cpp/ObjectModel/FeatureRegistration.h
```diff
@@ -3,11 +3,28 @@
 #pragma once
 
 #include "pch.h"
+#include "ParseUtil.h"
 
 namespace AdaptiveSharedNamespace
 {
     constexpr const char* const c_adaptiveCardsFeature = "adaptiveCards";
 
+    struct KeyHash
```
paulcam206: >KeyHash [](start = 11, length = 7)

`KeyHash` and `KeyEquals` should probably be named `CaseInsensitiveKeyHash` and `CaseInsensitiveKeyEquals`, respectively

paulcam206: (otherwise, looks great!)

---
In reply to: [322983007](https://github.com/microsoft/AdaptiveCards/pull/3434#discussion_r322983007) [](ancestors = 322983007)

PR #3438, opened 2019-09-11
[.NET] Relax Newtonsoft.Json version requirements

### LINKED ISSUE #3413: Relax the minimum version number of Newtonsoft
Adaptive card library nuget package requires atleast 12.0.2 version of Newtonsoft. This forces users of the library (such as Azure Bot Framework) to also upgrade the library.
Since Adaptive card does not really have a strong requirement for any feature/bug fix from 12.0.2 it would be better if it requires the minimum version of Newtonsoft that it really needs - this way, a larger group of applications can use the library.

paulcam206:
## Related Issue
Fixes #3413 

## Description
Per original bug, in 1.2 we introduced a hard dependency on `Newtonsoft.Json` >= 12.0.2. This isn't strictly necessary, as we don't rely on 12.X features. Here we set the minimum to 11.0.2, which is the version being targeted by BotFramework. To facilitate this, the WPFVisualizer project has been migrated away from a `packages.config` file to `<PackageReference />` entries in the csproj.

## How Verified
VS/CI build

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3438)

### DIFF
--- a/source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCards.Rendering.Html.csproj
+++ b/source/dotnet/Library/AdaptiveCards.Rendering.Html/AdaptiveCards.Rendering.Html.csproj
@@ -18,7 +18,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.MarkedNet" Version="1.0.13" />
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
     <PackageReference Include="System.Reflection" Version="4.3.0" />
   </ItemGroup>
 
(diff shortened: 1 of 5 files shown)


shalinijoshi19:
> <?xml version="1.0" encoding="utf-8"?>

curious - why had we been using pacakges.config here again? 

---
Refers to: source/dotnet/Samples/WPFVisualizer/packages.config:1 in 93df157. [](commit_id = 93df1578e388ff35a20bf174a3989a2c9bced08d, deletion_comment = True)

paulcam206:
> <?xml version="1.0" encoding="utf-8"?>

Not sure really. Looks like it's been old-style packages.config since its [inception](https://github.com/microsoft/AdaptiveCards/commits/master/source/dotnet/Samples/WPFVisualizer/packages.config)

---
In reply to: [530580826](https://github.com/microsoft/AdaptiveCards/pull/3438#issuecomment-530580826) [](ancestors = 530580826)

---
Refers to: source/dotnet/Samples/WPFVisualizer/packages.config:1 in 93df157. [](commit_id = 93df1578e388ff35a20bf174a3989a2c9bced08d, deletion_comment = True)

PR #3455, opened 2019-09-16
fix xml serialization

### LINKED ISSUE #3454: [.NET] Serialization is broken for XML
# Platform
* .NET 

# Author or host
Author

# Version of SDK
1.2.1

# Details
Additional classes and properties added since 1.0 have not been properly modeled to support XML serialization.  In particular

* AdaptiveRichTextBlock
* TextRun

Xml Serialization is extremely useful as it allows usage of XML based templating engines such as Razor to data bind your data to adaptive cards.

tomlm:
## Related Issue
Fixed #3454 Xml Serialization was broken between 1.0 and 1.2.1

## Description
Additional classes and properties added since 1.0 have not been properly modeled to support XML serialization. In particular

* AdaptiveRichTextBlock
* TextRun

Fixes are 
* apply proper XML attributes 
* Add nullable properties backing properties
* Properly model AdditionalProperties via serializabledictionary
* Fix all of the unit tests to verify all elements can be serialized via XML

## How Verified
1. unit tests were expanded to include all scenarios in all versions of the scenarios folder
2. Unit tests were expanded to verify roundtrip serialization of both Xml and JSON

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3455)


source/dotnet/Library/AdaptiveCards/SerializableDictionary.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Collections.Generic;
+using System.Xml.Serialization;
+
+namespace AdaptiveCards
+{
+#if !NETSTANDARD1_3
+    [XmlRoot("dictionary")]
+    public class SerializableDictionary<TKey, TValue> : Dictionary<TKey, TValue>, IXmlSerializable
+    {
+        public SerializableDictionary()
```
shalinijoshi19: Thanks for doing this Tom; This is missing the Msft copyright header if you could add it in.. the newer guidelines need just a trimmed down version per [https://docs.opensource.microsoft.com/releasing/copyright-headers.html](https://docs.opensource.microsoft.com/releasing/copyright-headers.html)

shalinijoshi19:
THANKS a ton for finding && fixing this Tom! Am trying to understand how we missed this and didnt catch it-- did you discover this as part of the upgrade to 1.2?

tomlm:
> THANKS a ton for finding && fixing this Tom! Am trying to understand how we missed this and didnt catch it-- did you discover this as part of the upgrade to 1.2?

Basically the xml serialization unit tests were driven by making sure that all samples/v1/*.json could be round-tripped between json and xml.  When new element types were created the samples files ended up in folders which the unit tests weren't aware of.

I changed the unit tests to take all versions in the samples folder.

source/dotnet/Library/AdaptiveCards/AdaptiveTypedElement.cs
```diff
@@ -26,10 +28,14 @@ public abstract class AdaptiveTypedElement
...
         [JsonExtensionData]
-#if !NETSTANDARD1_3
-        [XmlIgnore]
-#endif
+#if NETSTANDARD1_3
         public IDictionary<string, object> AdditionalProperties { get; set; } = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+#else
+        // Dictionary<> is not supported with XmlSerialization because Dictionary is not serializable, SerializableDictionary<> is
+        [XmlElement]
+        public SerializableDictionary<string, object> AdditionalProperties { get; set; } = new SerializableDictionary<string, object>(StringComparer.OrdinalIgnoreCase);
```
matthidinger: I'm just guessing here but wanted to throw this out before I closed the tab: Would changing this property type to `IDictionary<string, object>` fix the breaking change report? The fact that it happens to be implemented by a `SerializableDictionary` doesn't seem super relevant. Presumably this should have had a private setter anyway (typically people would just add/remove/clear the dictionary, not assign one wholesale)

shalinijoshi19: So, no because we end up in the same boat -- The type info is significant because that's what ends up getting reflected on in the call to XmlSerializer() .. @tomlm sent you a mail offline around removing this change from the release/1.2 branch since the change in and of itself blocks upgrade to 1.2.3 by Cortana skills ..

shalinijoshi19: I am evaluating using the DataContractSerializer approach here as well since that would mean not needing the custom implmentation of the SerializableDictionary but for now am just going to roll this change back from release/1.2

PR #3457, opened 2019-09-16
[UWP] Clean up XamlBuilder

paulcam206:
## Related Issue
Related to #2534 (around reconciling our error-handling in UWP)

## Description
This (rather large) change is staging to allow us to draw cleaner lines around our API boundaries. Specifically, we need to have clear delineation between `XamlBuilder`, various element renderers, and various common helper functions. Here's what changed:

* Rework precompiled header to speed up compile times
* Move element rendering (i.e. `IAdaptiveCardElement -> IUIElement`) functions to actual rendering classes
  * Removes most `static XamlBuilder::Build*` functions
  * I left image handling in `XamlBuilder` for now (due to its lifetime requirements)
* Move helper functions out of `XamlBuilder` (these don't need state, so they're just namespaced):
  * Action-related helpers go to `ActionHelpers.[cpp|h]`
  * Common helpers go to `XamlHelpers.[cpp|h]`
  * Some helpers were only used by one element renderer, so they're located proximal to the caller (e.g. `TryInsertResourceToResourceDictionaries` becomes a static method in `AdaptiveCardRendererComponent.cpp`)
* While this isn't the entirety of our move to a consolidated error-handling regime, I did take the opportunity to clarify the behavior of some methods via `noexcept` where it was obvious how they should behave.

I should note that I expect and welcome input on this new structure. I'm sure to have missed additional opportunities to refactor (especially given the scope).

## How Verified
TestApp clean. Some spot checking of cards in the visualizer.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3457)


source/uwp/Renderer/lib/AdaptiveActionSet.cpp
```diff
@@ -3,32 +3,26 @@
...
     {
         std::shared_ptr<AdaptiveCards::ActionSet> ActionSet = std::make_shared<AdaptiveCards::ActionSet>();
         return RuntimeClassInitialize(ActionSet);
-    } CATCH_RETURN;
+    }
+    CATCH_RETURN;
 
-    _Use_decl_annotations_
-    HRESULT AdaptiveActionSet::RuntimeClassInitialize(const std::shared_ptr<AdaptiveCards::ActionSet>& sharedActionSet) try
+    _Use_decl_annotations_ HRESULT AdaptiveActionSet::RuntimeClassInitialize(const std::shared_ptr<AdaptiveCards::ActionSet>& sharedActionSet)
```
RebeccaAnne: >_Use_decl_annotations_ [](start = 4, length = 22)

While you're here, maybe remove these _Use_decl_annotations_ here.

source/uwp/Renderer/lib/AdaptiveCardRendererComponent.cpp
```diff
@@ -138,7 +134,7 @@ namespace AdaptiveNamespace
 
             if (m_explicitDimensions)
             {
-                RETURN_IF_FAILED(m_xamlBuilder->SetFixedDimensions(m_desiredWidth, m_desiredHeight));
+                m_xamlBuilder->SetFixedDimensions(m_desiredWidth, m_desiredHeight);
```
RebeccaAnne: >m_xamlBuilder->SetFixedDimensions(m_desiredWidth, m_desiredHeight); [](start = 16, length = 67)

I see that this method doesn't have any error cases, so changing the return to void is probably a perfectly reasonable thing to do, but i'll admit I feel slightly uncomfortable about it. I guess it's mostly about expecting to see consistent error checking on all methods at the callers (in a hypothetical world were we're consistently HRESULT based), and theoretically about the need to remember to update all callers if this ever gets error conditions in the future (admittedly unlikely, and internal so not really a huge problem). No need to change it necessarily, but I'd be interested if you have any particular thoughts on this, and/or if we have any coding conventions guidance.

source/uwp/Renderer/lib/AdaptiveCardRendererComponent.cpp
```diff
@@ -300,6 +296,31 @@ namespace AdaptiveNamespace
         m_actionSentimentResourceDictionary = actionSentimentResourceDictionary;
     }
 
+    static HRESULT TryInsertResourceToResourceDictionaries(_In_ IResourceDictionary* resourceDictionary,
```
RebeccaAnne: >TryInsertResourceToResourceDictionaries [](start = 19, length = 39)

Look like this is only used to update m_actionSentimentResourceDictionary on this class. Consider making this a private class method TryInsertResourceIntoActionSentimentDictionary that doesn't need the resource dictionary passed in.

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
```diff
@@ -42,4 +73,176 @@ namespace AdaptiveNamespace
             jsonObject, elementParserRegistration, actionParserRegistration, adaptiveWarnings, element);
     }
     CATCH_RETURN;
+
+    static std::vector<std::string> GetChoiceSetValueVector(_In_ IAdaptiveChoiceSetInput* adaptiveChoiceSetInput)
```
RebeccaAnne: >GetChoiceSetValueVector [](start = 36, length = 23)

Should these also be on the renderer class? I realize there's no state on these renderer classes, but these methods should probably conceptually be treated the same as the _Build*ChoiceSetInput methods.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -48,4 +67,687 @@ namespace AdaptiveNamespace
             jsonObject, elementParserRegistration, actionParserRegistration, adaptiveWarnings, element);
     }
     CATCH_RETURN;
+
+    HRESULT XamlBuilder::BuildImage(_In_ IAdaptiveCardElement* adaptiveCardElement,
+                                    _In_ IAdaptiveRenderContext* renderContext,
+                                    _In_ IAdaptiveRenderArgs* renderArgs,
+                                    _COM_Outptr_ IUIElement** imageControl)
```
RebeccaAnne: This makes sense, but is a little tricky. Let's stick a comment on this explaining why we're defining the XamlBuilder method in the ImageRenderer code file.

source/uwp/Renderer/lib/AdaptiveActionSetRenderer.cpp
```diff
@@ -4,15 +4,20 @@
 
 #include "AdaptiveActionSet.h"
 #include "AdaptiveActionSetRenderer.h"
+#include "AdaptiveShowCardActionRenderer.h"
```
RebeccaAnne: Is this needed?

source/uwp/Renderer/lib/AdaptiveTimeInputRenderer.cpp
```diff
@@ -22,12 +22,54 @@ namespace AdaptiveNamespace
...
+        RETURN_IF_FAILED(timePickerAsFrameworkElement->put_HorizontalAlignment(ABI::Windows::UI::Xaml::HorizontalAlignment_Stretch));
+        RETURN_IF_FAILED(timePickerAsFrameworkElement->put_VerticalAlignment(ABI::Windows::UI::Xaml::VerticalAlignment_Top));
+
+        RETURN_IF_FAILED(XamlHelpers::SetStyleFromResourceDictionary(renderContext,
+                                                                     L"Adaptive.Input.Time",
+                                                                     timePickerAsFrameworkElement.Get()));
+
+        ComPtr<IAdaptiveCardElement> cardElement(adaptiveCardElement);
+        ComPtr<IAdaptiveTimeInput> adaptiveTimeInput;
+        THROW_IF_FAILED(cardElement.As(&adaptiveTimeInput));
```
RebeccaAnne: >THROW_IF_FAILED [](start = 8, length = 15)

I assume this is copy paste, but there are some THROW_IF_FAILEDs in here that should probably be updated to RETURN_IF_FAILED

RebeccaAnne:
>     static HRESULT ApplyMarginToXamlElement(_In_ IAdaptiveHostConfig* hostConfig, _In_ IFrameworkElement* element) noexcept

Should this go to XamlHelpers as well?

---
Refers to: source/uwp/Renderer/lib/XamlBuilder.cpp:218 in 0d9b13b. [](commit_id = 0d9b13b0e225ac20f96a8ff1e1589f57586ab38b, deletion_comment = False)

source/uwp/Renderer/lib/XamlHelpers.h
```diff
@@ -28,8 +28,83 @@ namespace AdaptiveNamespace::XamlHelpers
         return solidColorBrushAsBrush;
     }
 
+    void WireButtonClickToAction(_In_ ABI::Windows::UI::Xaml::Controls::IButton* button,
```
RebeccaAnne: >WireButtonClickToAction [](start = 9, length = 23)

What's the scope of ActionHelpers? Should this go there?

RebeccaAnne: Also, potentially:
WrapInTouchTarget
HandleSelectAction
BuildActions
BuildActionSetHelper

---
In reply to: [325846545](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r325846545) [](ancestors = 325846545)

RebeccaAnne:
Overall this looks great, love to see the giant pile of code in XamlBuilders sorted out a bit.

source/uwp/Renderer/lib/XamlBuilder.h
```diff
@@ -147,17 +65,12 @@ namespace AdaptiveNamespace
...
-                              _In_ ABI::AdaptiveNamespace::IAdaptiveRenderContext* renderContext,
-                              _In_ ABI::AdaptiveNamespace::IAdaptiveRenderArgs* renderArgs,
-                              Microsoft::WRL::ComPtr<XamlBuilder> xamlBuilder,
-                              _Outptr_ ABI::Windows::UI::Xaml::Controls::IPanel** bodyElementContainer);
-
-        static void ApplyBackgroundToRoot(_In_ ABI::Windows::UI::Xaml::Controls::IPanel* rootPanel,
-                                          _In_ ABI::AdaptiveNamespace::IAdaptiveBackgroundImage* backgroundImage,
-                                          _In_ ABI::AdaptiveNamespace::IAdaptiveRenderContext* renderContext,
-                                          _In_ ABI::AdaptiveNamespace::IAdaptiveRenderArgs* renderArgs);
+        static HRESULT CreateRootCardElement(_In_ ABI::AdaptiveNamespace::IAdaptiveCard* adaptiveCard,
```
RebeccaAnne: >CreateRootCardElement [](start = 23, length = 21)

Is the end game that XamlBuilder ceases to exist altogether (assuming we eventually untangle the image renderer issue)? Wondering whether these top level functions (this and BuildPanelChildren) ultimately belong in XamlHelpers as well (whether or not that's current feasible).

RebeccaAnne: Or maybe even on AdaptiveCardRendereComponent (as they do seem more substantial than mere "helpers")

---
In reply to: [325849538](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r325849538) [](ancestors = 325849538)

nesalang: I like that idea

---
In reply to: [325849965](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r325849965) [](ancestors = 325849965,325849538)

nesalang:
since master is changed to use vs2019, can you update this branch to use vs2019 also? I want to use this branch for my bug, and I have removed vs2017 due to the lack of storage. It's not big deal, i just wanted to check in case you're going to rebase to master. thanks!

paulcam206:
> since master is changed to use vs2019, can you update this branch to use vs2019 also? I want to use this branch for my bug, and I have removed vs2017 due to the lack of storage. It's not big deal, i just wanted to check in case you're going to rebase to master. thanks!

just merged master in. was going to rebase to master, but don't want to lose the PR feedback :)

nesalang:
Thanks, really appreciate it.

---
In reply to: [532937408](https://github.com/microsoft/AdaptiveCards/pull/3457#issuecomment-532937408) [](ancestors = 532937408)

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -0,0 +1,595 @@
...
+                              _In_ IAdaptiveRenderContext* renderContext,
+                              ABI::AdaptiveNamespace::ContainerStyle containerStyle,
+                              _In_ ABI::AdaptiveNamespace::IAdaptiveHostConfig* hostConfig,
+                              bool allActionsHaveIcons,
+                              _In_ IButton* button)
+    {
+        HString title;
+        THROW_IF_FAILED(action->get_Title(title.GetAddressOf()));
+
+        HSTRING iconUrl;
```
almedina-ms: >HSTRING [](start = 8, length = 7)

This should be a HString

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -0,0 +1,595 @@
...
+                                                   buttonText](IInspectable* /*sender*/, IRoutedEventArgs * /*args*/) -> HRESULT {
+                        ComPtr<IFrameworkElement> buttonTextAsFrameworkElement;
+                        RETURN_IF_FAILED(buttonText.As(&buttonTextAsFrameworkElement));
+
+                        return SetMatchingHeight(buttonIconAsFrameworkElement.Get(), buttonTextAsFrameworkElement.Get());
+                    }).Get(),
+                    &eventToken));
+
+                // Only add spacing when the icon must be located at the left of the title
+                UINT spacingSize;
```
almedina-ms: >UINT [](start = 16, length = 4)

should we normalize to use UINT32?

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -0,0 +1,595 @@
...
+
+        ComPtr<IResourceDictionary> resourceDictionary;
+        RETURN_IF_FAILED(renderContext->get_OverrideStyles(&resourceDictionary));
+        ComPtr<IStyle> styleToApply;
+
+        ComPtr<AdaptiveNamespace::AdaptiveRenderContext> contextImpl =
+            PeekInnards<AdaptiveNamespace::AdaptiveRenderContext>(renderContext);
+
+        if ((SUCCEEDED(WindowsCompareStringOrdinal(actionSentiment.Get(), HStringReference(L"default").Get(), &isSentimentDefault)) &&
+             (isSentimentDefault == 0)) ||
```
almedina-ms: >  [](start = 12, length = 1)

nit: extra space

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -0,0 +1,595 @@
...
+                RETURN_IF_FAILED(buttonFrameworkElement->put_HorizontalAlignment(ABI::Windows::UI::Xaml::HorizontalAlignment_Right));
+                break;
+            case ABI::AdaptiveNamespace::ActionAlignment::Stretch:
+                RETURN_IF_FAILED(buttonFrameworkElement->put_HorizontalAlignment(ABI::Windows::UI::Xaml::HorizontalAlignment_Stretch));
+                break;
+            }
+        }
+
+        ABI::AdaptiveNamespace::ContainerStyle containerStyle;
+        renderArgs->get_ContainerStyle(&containerStyle);
```
almedina-ms: >renderArgs->get_ContainerStyle(&containerStyle); [](start = 8, length = 48)

missing RETURN_IF_FAILED

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.h
```diff
@@ -28,6 +26,16 @@ namespace AdaptiveNamespace
                                 _In_ ABI::AdaptiveNamespace::IAdaptiveActionParserRegistration* actionParsers,
                                 _In_ ABI::Windows::Foundation::Collections::IVector<ABI::AdaptiveNamespace::AdaptiveWarning*>* adaptiveWarnings,
                                 _COM_Outptr_ ABI::AdaptiveNamespace::IAdaptiveCardElement** element) noexcept override;
+
+    private:
+        HRESULT _BuildExpandedChoiceSetInput(_In_ ABI::AdaptiveNamespace::IAdaptiveRenderContext* renderContext,
```
almedina-ms: >_ [](start = 16, length = 1)

why do these methods start with _ ? Is it to avoid build errors?

source/uwp/Renderer/lib/AdaptiveCardRendererComponent.cpp
paulcam206: I don't see any explicit guidance in cppcoreguidelines, but at least in codebases I've been in, it's actually preferable to have a void return type if at all possible (obviously, it depends on what the function is actually *doing*). This alleviates the need for callers to anticipate a failure condition that may never happen. It feels a little weirder here since we're simply doing a `RETURN_IF_FAILED`, but there have been problems in the past with people writing error-handling code that never runs.

and then maybe someday it does run. That can be a real problem.

In code that exists in a tight contract (e.g. COM interface impls), it makes sense to pass `HRESULT`s if for no reason other than you might not have a chance to change in the future. With COM you also have no guarantee that code won't run between your function returning and the caller getting the return value (think proxy/stubs). However, when you control the caller and the callee and have no promises to keep with a contract, I think it's better to just rename the function if you need to change the return type. Going from "this function can never fail" to "this function might fail" feels like a pretty fundamental change in functionality.

---
In reply to: [325828953](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r325828953) [](ancestors = 325828953)

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
paulcam206: yeah, probably :D

---
In reply to: [325832661](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r325832661) [](ancestors = 325832661)

source/uwp/Renderer/lib/AdaptiveActionSetRenderer.cpp
paulcam206: nope... looks like some of the other ones are also unneeded. Probably snuck in while I was trying to get things building :)

---
In reply to: [325838868](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r325838868) [](ancestors = 325838868)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -48,4 +67,687 @@ namespace AdaptiveNamespace
...
+    {
+        ComPtr<IAdaptiveCardElement> cardElement(adaptiveCardElement);
+        ComPtr<IAdaptiveImage> adaptiveImage;
+        RETURN_IF_FAILED(cardElement.As(&adaptiveImage));
+
+        ComPtr<IAdaptiveHostConfig> hostConfig;
+        RETURN_IF_FAILED(renderContext->get_HostConfig(&hostConfig));
+
+        HSTRING url;
```
almedina-ms: >HSTRING [](start = 8, length = 7)

HString

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -48,4 +67,687 @@ namespace AdaptiveNamespace
...
+            RETURN_IF_FAILED(hostConfig->get_Image(&imageConfig));
+            RETURN_IF_FAILED(imageConfig->get_ImageSize(&size));
+        }
+
+        ABI::AdaptiveNamespace::ImageStyle imageStyle;
+        RETURN_IF_FAILED(adaptiveImage->get_Style(&imageStyle));
+        ComPtr<IAdaptiveCardResourceResolvers> resourceResolvers;
+        RETURN_IF_FAILED(renderContext->get_ResourceResolvers(&resourceResolvers));
+
+        HSTRING backgroundColor;
```
almedina-ms: >HSTRING [](start = 8, length = 7)

HString

source/uwp/Renderer/lib/AdaptiveImageSetRenderer.cpp
```diff
@@ -22,12 +23,86 @@ namespace AdaptiveNamespace
...
+
+            XamlHelpers::IterateOverVector<ABI::AdaptiveNamespace::AdaptiveImage, ABI::AdaptiveNamespace::IAdaptiveImage>(
+                images.Get(),
+                [imageSize, xamlGrid, renderContext, childRenderArgs, imageRenderer, imageSetConfig](IAdaptiveImage* adaptiveImage) {
+                    ComPtr<IUIElement> uiImage;
+                    ComPtr<IAdaptiveImage> localAdaptiveImage(adaptiveImage);
+                    RETURN_IF_FAILED(localAdaptiveImage->put_Size(imageSize));
+
+                    ComPtr<IAdaptiveCardElement> adaptiveElementImage;
+                    localAdaptiveImage.As(&adaptiveElementImage);
```
almedina-ms: >localAdaptiveImage.As(&adaptiveElementImage); [](start = 20, length = 45)

missing RETURN_IF_FAILED

source/uwp/Renderer/lib/AdaptiveMediaRenderer.cpp
```diff
@@ -18,12 +25,146 @@ namespace AdaptiveNamespace
...
+        }
+
+        ComPtr<IUIElement> mediaPanelAsUIElement;
+        RETURN_IF_FAILED(mediaPanel.As(&mediaPanelAsUIElement));
+
+        ComPtr<IButtonBase> touchTargetAsButtonBase;
+        RETURN_IF_FAILED(touchTargetUIElement.As(&touchTargetAsButtonBase));
+
+        // Take a reference to the mime type string for the lambda
+        HSTRING lambdaMimeType;
```
almedina-ms: >HSTRING [](start = 8, length = 7)

HString

source/uwp/Renderer/lib/AdaptiveShowCardActionRenderer.cpp
```diff
@@ -39,4 +41,69 @@ namespace AdaptiveNamespace
...
+        RETURN_IF_FAILED(showCardActionConfig->get_ActionMode(&showCardActionMode));
+
+        // Set the top margin
+        ComPtr<IFrameworkElement> showCardFrameworkElement;
+        RETURN_IF_FAILED(localUiShowCard.As(&showCardFrameworkElement));
+
+        UINT32 inlineTopMargin;
+        RETURN_IF_FAILED(showCardActionConfig->get_InlineTopMargin(&inlineTopMargin));
+
+        double sideMargin = (double)padding * -1;
```
almedina-ms: >(double)padding * -1; [](start = 28, length = 21)

nit: this could be changed for (double)(-padding)

but it would be better to use static_cast<double>(-padding)

source/uwp/Renderer/lib/AdaptiveShowCardActionRenderer.h
```diff
@@ -26,6 +26,12 @@ namespace AdaptiveNamespace
                                 _In_ ABI::AdaptiveNamespace::IAdaptiveActionParserRegistration* actionParserRegistration,
                                 _In_ ABI::Windows::Foundation::Collections::IVector<ABI::AdaptiveNamespace::AdaptiveWarning*>* adaptiveWarnings,
                                 _COM_Outptr_ ABI::AdaptiveNamespace::IAdaptiveActionElement** element) noexcept override;
+
+        static HRESULT BuildShowCard(_In_ ABI::AdaptiveNamespace::IAdaptiveCard* showCard,
+                                     _In_ ABI::AdaptiveNamespace::IAdaptiveRenderContext* renderContext,
+                                     _In_ ABI::AdaptiveNamespace::IAdaptiveRenderArgs* renderArgs,
+                                     bool isBottomActionBar,
```
almedina-ms: >bool [](start = 37, length = 4)

should we try to consolidate bool, boolean, BOOL and BOOLEAN? I understand some types are required by some APIs but we have 4 ways of representing true and false

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
paulcam206: most of this code is bulk copy/pasted, but the `HSTRING` -> `HString` errors should be all fixed now :)

---
In reply to: [326386971](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r326386971) [](ancestors = 326386971)

source/uwp/Renderer/lib/ActionHelpers.cpp
paulcam206: this was moved from its original location, so I don't want to mess with it here. that said, we should probably take a look at our `UINT`/`UINT32` usage

---
In reply to: [326325155](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r326325155) [](ancestors = 326325155)

source/uwp/Renderer/lib/ActionHelpers.cpp
paulcam206: this is how `clang-format` spits it out. it's technically correct, though, because of the extra `(` nesting

---
In reply to: [326325757](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r326325757) [](ancestors = 326325757)

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.h
paulcam206: old habits die hard. it's to denote them as private methods. I'll fix since we don't use this convention here :)

---
In reply to: [326334977](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r326334977) [](ancestors = 326334977)

source/uwp/Renderer/lib/XamlBuilder.h
paulcam206: that's kinda how I've been thinking about it. after this PR, XamlBuilder's sole reason for existing is pretty much for lifetime management reasons. if we find a way to get rid of that last reason (imaging stuff like you said), the remaining functionality should be moved closer to its use (i.e. `XamlBuilder::BuildImage` -> `AdaptiveImageRenderer::Render`, `XamlBuilder::BuildPanelChildren` -> `XamlHelpers::BuildPanelChildren`)

---
In reply to: [325849965](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r325849965) [](ancestors = 325849965,325849538)

source/uwp/Renderer/lib/AdaptiveShowCardActionRenderer.cpp
paulcam206: `padding` is unsigned, so it has to be `-static_cast<double>(padding)`, but your point stands :)

---
In reply to: [326399165](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r326399165) [](ancestors = 326399165)

source/uwp/Renderer/lib/AdaptiveShowCardActionRenderer.h
paulcam206: this has been tricky for a while. I think the basics are:

* whenever possible, use `bool`
* WinRT contracts require `boolean` (not positive, but guessing this is to deal with the cross-language nature of WinRT)
* BOOL and BOOLEAN are holdovers from Win32 (and Win16 before it), which originated in a time before `bool` was a real type in C. they should only be used as required to interact with non-WinRT Windows APIs. 

of course, like any style guidance, there's room for interpretation. you probably wouldn't introduce the only use of `bool` in a file that uses `BOOL` everywhere (unless, I guess, you can confidently switch everything to `bool`)

---
In reply to: [326399601](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r326399601) [](ancestors = 326399601)

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -0,0 +1,1014 @@
...
+            // Define the alignment for the button contents
+            ComPtr<IStackPanel> buttonContentsStackPanel =
+                XamlHelpers::CreateXamlClass<IStackPanel>(HStringReference(RuntimeClass_Windows_UI_Xaml_Controls_StackPanel));
+
+            // Create image and add it to the button
+            ComPtr<IAdaptiveImage> adaptiveImage;
+            THROW_IF_FAILED(MakeAndInitialize<AdaptiveImage>(&adaptiveImage));
+
+            adaptiveImage->put_Url(iconUrl.Get());
+            adaptiveImage->put_HorizontalAlignment(ABI::AdaptiveNamespace::HAlignment::Center);
```
RebeccaAnne: I realize this is copy paste, but since this line changed with the HString fix I now notice we're missing error checks on these calls.

RebeccaAnne:
>         IFACEMETHODIMP put_Spacing(_In_ ABI::AdaptiveNamespace::Spacing spacing)

Possibly out of scope of this change, but there are more annotations in this file that could use some fixing. (This one should be removed, for example)

---
Refers to: source/uwp/Renderer/lib/AdaptiveActionSet.h:34 in e0a6fcb. [](commit_id = e0a6fcb8f22c79736064d75938b8c15b79f05a89, deletion_comment = False)

source/uwp/Renderer/lib/AdaptiveActionSetRenderer.cpp
```diff
@@ -2,44 +2,64 @@
...
         return AdaptiveNamespace::FromJson<AdaptiveNamespace::AdaptiveActionSet, AdaptiveSharedNamespace::ActionSet, AdaptiveSharedNamespace::ActionSetParser>(
             jsonObject, elementParserRegistration, actionParserRegistration, adaptiveWarnings, element);
     }
     CATCH_RETURN;
+
+    HRESULT AdaptiveActionSetRenderer::Render(_In_ IAdaptiveCardElement* adaptiveCardElement,
+                                              _In_ IAdaptiveRenderContext* renderContext,
+                                              _In_ IAdaptiveRenderArgs* renderArgs,
+                                              _COM_Outptr_ IUIElement** actionSetControl) noexcept
+    try
```
RebeccaAnne: >try [](start = 4, length = 3)

I assume this is automated, but any idea why sometimes these trys are on their own line and sometimes they're on the line with the noexcept?

paulcam206: I was manually fixing some of these up, as I thought it was a quirk in my setup. Probably not, though, so I'll re-format everything :)

---
In reply to: [327215358](https://github.com/microsoft/AdaptiveCards/pull/3457#discussion_r327215358) [](ancestors = 327215358)

paulcam206:
>         IFACEMETHODIMP put_Spacing(_In_ ABI::AdaptiveNamespace::Spacing spacing)

Filed #3483

---
In reply to: [534180541](https://github.com/microsoft/AdaptiveCards/pull/3457#issuecomment-534180541) [](ancestors = 534180541)

---
Refers to: source/uwp/Renderer/lib/AdaptiveActionSet.h:34 in e0a6fcb. [](commit_id = e0a6fcb8f22c79736064d75938b8c15b79f05a89, deletion_comment = False)

Issue #3425, opened 2019-09-06
[UWP][Memory] [Rendered Cards Leaking Memory]


andycb:
# Platform

* UWP

# Version of SDK

AdaptiveCards.Redering.Uwp  V1.2.2

# Details

When redering an adaptive card, it appears that its memory is not released. The bellow code sample loads an adaptive card, displays it on the UI before removing it and forcing a garbage collection. After 50 itteraltions, an additional 90MB of memory is being used by the test app.

![image](https://user-images.githubusercontent.com/4417762/64429002-f3845880-d0ac-11e9-8ccd-8e768a062e20.png)

```csharp
var renderer = new AdaptiveCardRenderer();
[...]

PR #3507, opened 2019-10-02
[UWP] Fix circular reference in ImageOpened lambda for Auto sized image handling

RebeccaAnne:
## Related Issue
Fixed as part of the investigation for #3425. This seems to be the largest part of the leak, but there is still a leak in that scenario after this fix.

## Description
The imageOpened lambdas in SetAutoSize were taking strong references to the image or ellipse item and its parent. This lead to a circular reference which caused these items to leak. Updated the lambdas to take weak references instead to break the circular reference and avoid the leak.

## How Verified
Confirmed no new failures in the UWP test app. Manual verification that the memory leak is gone in a local scenario where it had previously repro'd.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3507)

### DIFF
--- a/source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
+++ b/source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
@@ -697,17 +697,33 @@ namespace AdaptiveNamespace
                 THROW_IF_FAILED(ellipseAsUIElement->put_Visibility(Visibility::Visibility_Collapsed));
                 // Handle ImageOpened event so we can check the imageSource's size to determine if it fits in its parent
                 EventRegistrationToken eventToken;
-                ComPtr<IInspectable> strongParentElement(parentElement);
+                ComPtr<IInspectable> localParentElement(parentElement);
+
+                // Take weak references to the ellipse and parent to avoid circular references between this lambda and
+                // its parents (Parent->Ellipse->ImageBrush->Lambda->(Parent and Ellipse))
+                WeakRef weakParent;
+                THROW_IF_FAILED(localParentElement.AsWeak(&weakParent));
+
+                WeakRef weakEllipse;
+                THROW_IF_FAILED(ellipseAsUIElement.AsWeak(&weakEllipse));
                 THROW_IF_FAILED(brushAsImageBrush->add_ImageOpened(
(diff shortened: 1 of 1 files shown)


source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -697,17 +697,33 @@ namespace AdaptiveNamespace
...
+                // Take weak references to the ellipse and parent to avoid circular references between this lambda and
+                // its parents (Parent->Ellipse->ImageBrush->Lambda->(Parent and Ellipse))
+                WeakRef weakParent;
+                THROW_IF_FAILED(localParentElement.AsWeak(&weakParent));
+
+                WeakRef weakEllipse;
+                THROW_IF_FAILED(ellipseAsUIElement.AsWeak(&weakEllipse));
                 THROW_IF_FAILED(brushAsImageBrush->add_ImageOpened(
-                    Callback<IRoutedEventHandler>([ellipseAsUIElement, ellipseAsFrameworkElement, imageSourceAsBitmap, strongParentElement, isVisible](
+                    Callback<IRoutedEventHandler>([weakEllipse, imageSourceAsBitmap, weakParent, isVisible](
```
paulcam206: >Callback [](start = 20, length = 8)

Hmm... perhaps we should audit all of our uses of `Callback` lambdas? e.g. `ArrangeButtonContent()` in ActionHelpers.cpp looks like it might have the same issue...

paulcam206: (of course, you may have already done this)

---
In reply to: [331661884](https://github.com/microsoft/AdaptiveCards/pull/3507#discussion_r331661884) [](ancestors = 331661884)

RebeccaAnne: Good point. Logged #3511 to follow up.

---
In reply to: [331663216](https://github.com/microsoft/AdaptiveCards/pull/3507#discussion_r331663216) [](ancestors = 331663216,331661884)

Issue #3260, opened 2019-07-26
[Designer] Version Restriction


jeffhoang:
# Platform
* JavaScript

# Version of SDK
1.1

# Version of Adaptive Cards Designer
0.7.1

# Issue / Feature Request

Is there a way to restrict the adaptive cards designer to only support a certain adapative cards version?  We're still on version 1.1 and don't want our developers to use features beyond this.

matthidinger:
Thanks for the feedback. This is something that is being worked on, hopefully we'll have it finished it shortly. In the mean time, can you try this code to at least restrict items from the toolbar?

```js
let designer = new ACDesigner.CardDesigner(hostContainers);
Adaptive.AdaptiveCard.elementTypeRegistry.unregisterType("RichTextBlock");
ACDesigner.CardDesignerSurface.cardElementPeerRegistry.unregisterPeer(Adaptive.RichTextBlock);
 
Adaptive.AdaptiveCard.elementTypeRegistry.unregisterType("ActionSet");
ACDesigner.CardDesignerSurface.cardElementPeerRegistry.unregisterPeer(Adaptive.ActionSet);
 
designer.attachTo(document.getElementById("designerRootHost"));
 
Adaptive.AdaptiveCard.actionTypeRegistry.unregisterType("Action.ToggleVisibility");
    ACDesigner.CardDesignerSurface.actionPeerRegistry.unregisterPeer(Adaptive.ToggleVisibilityAction);
```

shalinijoshi19:
@dclaux, can this be closed now?

dclaux:
@shalinijoshi19  no, there is still a lot to do to actually make the designer version aware. What Matt suggests above is a way to de-register elements, but that doesn't fully solve the problem.

shalinijoshi19:
Thanks; Do we have a quick one-pager/spec consolidating the versioning requirements we are targeting  for the designer?

dclaux:
We don't. By the time I have a PR for the feature, I will open an issue with a description of what it entails. But it's fairly simple:
- Designer has a version selector
- Cards are parsed per the selected version (e.g. don't parse the "minHeight" property if selected version isn't 1.2 or greater)
- Cards are serialized to JSON per the selected version  (e.g. don't serialize the "minHeight" property if selected version isn't 1.2 or greater)
- Property sheet only shows properties that existed in the selected version
- Toolbox only shows elements that were available in selected version

shalinijoshi19:
[...]

Issue #3048, opened 2019-06-11
Designer doesn't issue warnings for unknown enums


andrewleader:
If a card has an unknown enum, like `"size": "medium2"`, the Designer doesn't say there's any warnings or issues about this

shalinijoshi19:
@dclaux this looks like something we should fix in the designer? Thanks!

PR #3508, opened 2019-10-02
[Designer] Full version picker support + Designer UI improvements (including enhanced data binding)

dclaux:
## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/3260
Fixes https://github.com/microsoft/AdaptiveCards/issues/3515
Fixes https://github.com/microsoft/AdaptiveCards/issues/3048
Fixes https://github.com/microsoft/AdaptiveCards/issues/3901
Fixes https://github.com/microsoft/AdaptiveCards/issues/3926
Fixes https://github.com/microsoft/AdaptiveCards/issues/3972
Fixes https://github.com/microsoft/AdaptiveCards/issues/3973

## Try it
All the changes in this PR are live at https://adaptivecardsci.z5.web.core.windows.net/pr/3508/designer/?preview=1

## Adaptive Expressions
This PR drops its custom expression parser/evaluator implementation and moves to Adaptive Expressions instead, a more complete and capable implementation shared with the Bot Framework.

**Important:**
This PR is currently dependent on daily builds of the adaptive-expressions library. These builds are sourced from https://botbuilder.myget.org/F/botbuilder-v4-js-daily/npm/

To build, you first have to update your npm registry:
`npm config set registry https://botbuilder.myget.org/F/botbuilder-v4-js-daily/npm/`

Then run `npm install`

**Notes**:
- The JS build checks **will not pass** while the code uses the daily builds.
- To revert your npm registry to the default run: `npm config set registry https://registry.npmjs.org/`

## Full version picker support
This PR enables full support for the version picker in the designer. It includes a massive refactor of the JS renderer, without which this feature, although doable, would have been a big hack. While this PR is big, it is split into logical commits that can be reviewed and made sense of individually and sequentially.

With this PR:
- The designer now parses cards per the selected version (e.g. don't parse the "minHeight" property if selected version isn't 1.2 or greater)
- The designer now serializes cards to JSON per the selected version  (e.g. don't serialize the "minHeight" property if selected version isn't 1.2 or greater)
- The designer's property sheet only shows properties that exist in the selected version
- The designer's toolbox only shows elements that are available in the selected version
- The designer's action type picker only shows actions that are available in the selected version

The main purpose of the refactor is to move the renderer to a schema model, where properties are described in terms of type and schema version in which they were introduced (among other things) using "property definition" objects which are themselves listed into a collection (a "schema".) Classes in the renderer's AC model then declare members that are mapped to the corresponding property definition. As an example, here is how the base **CardObject** class defines its schema and maps members:

```typescript
export abstract class CardObject extends SerializableObject {
    //#region Schema

    static readonly typeNameProperty = new StringProperty(
        Versions.v1_0,
        "type",
        undefined,
        undefined,
        undefined,
        (sender: object) => {
            return (<CardObject>sender).getJsonTypeName()
        });
    static readonly idProperty = new StringProperty(Versions.v1_0, "id");
    static readonly requiresProperty = new SerializableObjectProperty(
        Versions.v1_2,
        "requires",
        HostCapabilities);

    protected getSchemaKey(): string {
        return this.getJsonTypeName();
    }

    @property(CardObject.idProperty)
    id?: string;

    @property(CardObject.requiresProperty)
    get requires(): HostCapabilities {
        return this.getValue(CardObject.requiresProperty);
    }

    //#endregion
    
    ...
}
```

Note the use of the `@property` decorator.

Other notable things:
- The code has been reorganized to some extent into separate files. Although ideally the code would be split into even more files, webpack (at least the way we have it setup) doesn't properly handle circular references which the library relies upon. While it might be possible to remove some of those, this work isn't necessary and is beyond the scope of this work
- The library is now compiled using the **strictNullChecks** and **noImplicitAny** TypeScript compiler flags
- Beyond the version picker, the refactor allows for better parse error reporting, such as invalid enum values and more
- As a result of this refactor, the JS renderer is now able to parse and serialize from/to specific versions of the AC schema
- The refactor **doesn't touch the Fabric extension**, which is now **incompatible with the new codebase** and will need to be updated.
- Some breaking changes have been introduced. They are listed below and also included in the **README.md** file
- A **LOT** of effort has been put into trying to keep the size of the library as small as possible with these changes. In the end, the size of the minified library **has still grown by 7 KB**. Note that if we ever switch to compiling to ES6 (currently we compile to ES5) the library will instantly shed some 30+ KB because TypeScript will not have to emit emulation code for ES6 native features

### Improved data binding UI
This PR introduces changes in the way data binding works in the designer. Namely:
- The **Data Structure** toolbox has been removed (it is actually only hidden by default, but its future is uncertain)
- A new **Bind...** command has been added to the **TextBlock** selection rectangle. Clicking it pops up a field picker allowing the user to select which data field the `text` property of the TextBlock is bound to
- A select few additional properties (such as `Image.Url`) can also be bound using the field picker from the **Element Properties**
- Elements that are bound to data now render the actual field values rather than the binding expression, even in design mode. However, repeating elements only repeat in "Preview Mode".

image

image

image

### Improved data structure and sample data support
This PR enhances the way a host can provide a sample data structure as a JSON payload, using the new `IData` interface and the new `FieldDefinition.parse` static method, as shown below. With this enhancement, a host can provide a sample data value for each field in the sample data structure, instead of relying on an actual sample data JSON payload, which in turn makes it possible for that host to hide the **Sample Data Editor** pane and still provide actual data for bindings.

```typescript
let sampleDataStructure: ACDesigner.IData = {
    valueType: "Object",
    fields: [
        {
            name: "firstName",
            displayName: "First name",
            valueType: "String",
            sampleValue: "John"
        },
        {
            name: "lastName",
            displayName: "Last name",
            valueType: "String",
            sampleValue: "Doe"
        },
        {
            name: "age",
            displayName: "Age",
            valueType: "Number",
            sampleValue: 36
        },
        {
            name: "isMarried",
            displayName: "Married",
            valueType: "Boolean",
            sampleValue: false
        },
        {
            name: "address",
            displayName: "Address",
            valueType: "Object",
            fields: [
                {
                    name: "street",
                    displayName: "Street",
                    valueType: "String",
                    sampleValue: "1234 555th Ave NE"
                },
                {
                    name: "city",
                    displayName: "City",
                    valueType: "String",
                    sampleValue: "Redmond"
                },
                {
                    name: "state",
                    displayName: "State",
                    valueType: "String",
                    sampleValue: "WA"
                },
                {
                    name: "countryOrRegion",
                    displayName: "Country/region",
                    valueType: "String",
                    sampleValue: "USA"
                }
            ]
        },
        {
            name: "children",
            displayName: "Children",
            valueType: "Array",
            itemType: {
                valueType: "Object",
                fields: [
                    {
                        name: "firstName",
                        displayName: "First name",
                        valueType: "String",
                        sampleValue: "Jennifer"
                    },
                    {
                        name: "lastName",
                        displayName: "Last name",
                        valueType: "String",
                        sampleValue: "Doe"
                    },
                    {
                        name: "age",
                        displayName: "Age",
                        valueType: "Number",
                        sampleValue: 13
                    }
                ]
            }
        }
    ]
};

designer.dataStructure = ACDesigner.FieldDefinition.parse(sampleDataStructure);
```

## Enhanced sample picker dialog
The built-in sample picker dialog has been improved to show previews of the available cards. This makes it a lot easier for users to pick a sample that matches the design they are after:

image

## Breaking changes
The below changes are also mentioned in **README.md**:
- `ColumnSet.getCount()` has been **REMOVED**. Use `ColumnSet.getItemCount()` instead.
- The `isNullOrEmpty(value: string): boolean` function has been **REMOVED**. Use `if (!stringValue)` instead.
- The library is now compiled with the `noImplicitAny` flag. As a result, anything that can be undefined/not set now has the `undefined` value. All uses of `null` have been removed.
- The following global setting statics have been moved from the AdaptiveCard class to the new GlobalSettings class: `useAdvancedTextBlockTruncation`, `useAdvancedCardBottomTruncation`, `useMarkdownInRadioButtonAndCheckbox`, `allowMarkForTextHighlighting`, `alwaysBleedSeparators`, `enableFullJsonRoundTrip`, `useBuiltInInputValidation`, `displayInputValidationErrors`
- `CardElement.getForbiddenElementTypes()` has been **REMOVED**
- The signature of **CardElement.getForbiddenActionTypes()** has changed to `getForbiddenActionTypes(): CardObjectType<Action>[]` with **CardObjectType** defined as `type CardObjectType<T extends CardObject> = { new(): T }`
- The `AdaptiveCard.onParseError` event has been **REMOVED**. Parse errors are now collected into the `SerializationContext.errors` property.
- The `AdaptiveCard.onParseElement` and `AdaptiveCard.onParseAction` events (both static and instance versions) have been **REMOVED**. Use `SerializationContext.onParseElement` and `SerializationContext.onParseAction` instead.
- The `createActionInstance` and `createElementInstance` global functions have been **REMOVED** and replaced with the `parseAction` and `parseElement` methods of the `SerializationContext` class.
- A new base class, `SerializableObject`, has been introduced. It implements core serialization and deserialization behaviors and pretty much all objects handled by the library (including `Action` and `CardElement` extend it.
- **Serialization** and **deserialization** are now done using a "schema" model, where each property of a serializable object is defined as a `PropertyDefinition` instance. Class members that expose those properties are identified via the use of the `@property` decorator. The PropertyDefinition model makes it possible to associate version information with pretty much eveything, which allows the library to (de)serialize to/from any version of the Adaptive Card schema.
- The signature of the `CardObject.parse` method has changed to `parse(source: any, context?: SerializationContext)`. The context object includes a `targetVersion` property, which tells the API which version of the Adaptive Card schema to honor when deserializing the source object.
- A new protected `internalParse` method has been introduced on **SerializableObject**. When implementing custom actions or elements, have your class override `internalParse` rather than `parse`.
- `SerializableObject.toJSON` now accepts a **targetVersion** parameter. When a targetVersion value is specified, the object is serialized to that particular version of the Adaptive Card schema. If no targetVersion is specified, the object is serialized to the latest version of the Adaptive Card schema supported by the library.
- A new protected `internalToJSON` method has been introduced on **SerializableObject**. When implementing custom actions or elements, have your class override `internalToJSON` rather than `toJSON`.
- `AdaptiveCard.elementTypeRegistry` and `AdaptiveCard.actionTypeRegistry` have been **REMOVED**. They are replaced with `GlobalRegistry.elements` and `GlobalRegistry.actions`.
- The `TypeRegistry<T>`, `ElementTypeRegistry` and `ActionTypeRegistry` classes have been **REMOVED**, replaced with the single `CardObjectRegistry<T>` class.
- The global `setProperty`, `setNumberProperty`, `setEnumProperty` and `setArrayProperty` functions have been **REMOVED**. They are replaced by `SerializationContext` instance methods `serializeValue`, `serializeNumber`, `serializeEnum` and `serializeArray`.
- The global `getStringValue`, `getNumberValue`, `getBoolValue` and `getEnumValue` functions have been renamed into `parseString`, `parseNumber`, `parseBool` and `parseEnum`.
- The global `parseHostConfigEnum` function is no longer exported.
- The `ValidationError` enum has been renamed into `ValidationEvent`.
- The `IValidationError` interface has been renamed into `IValidationLogEntry`. It has a new required `phase` field of type `ValidationPhase` and its `error` field has been renamed into `event`.

## How Verified
Verified manually using adaptivecards-designer-app

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3508)


source/nodejs/adaptivecards/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
 	"extends": "../tsconfig.json",
 	"compilerOptions": {
-		"outDir": "./lib"
+        "outDir": "./lib",
```
paulcam206: >        [](start = 1, length = 7)

are we doing spaces or tabs?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -218,12 +219,239 @@ export class ValidationResults {
...
+export class StringPropertyDefinition extends TypedPropertyDefinition<string> {
+    parse(value: any): any {
+        let parsedValue = Utils.getStringValue(value, this.defaultValue);
+        let isUndefined = parsedValue === undefined || (parsedValue === "" && this.treatEmptyAsUndefined);
+
+        if (!isUndefined && this.regEx !== undefined) {
+            let matches = this.regEx.exec(value);
+
+            if (!matches) {
+                // TODO: Log parse error
```
paulcam206: >// TODO: Log parse error [](start = 16, length = 24)

We should use a tracking bug here (unless we're not actually going to log an error here, in which case the comment should be removed)

source/nodejs/adaptivecards/tsconfig.json
dclaux: VS Code is configured for spaces.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Most (probably all) TODOs are addressed in later commits. These were placeholders.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -218,12 +219,239 @@ export class ValidationResults {
...
+        readonly treatEmptyAsUndefined: boolean = true,
+        readonly regEx?: RegExp,
+        readonly defaultValue?: string) {
+        super(targetVersion, name, defaultValue);
+    }
+}
+
+export class BooleanPropertyDefinition extends TypedPropertyDefinition<boolean> {
+    parse(value: any): any {
+        return Utils.getBoolValue(value, this.defaultValue);;
```
paulcam206: >; [](start = 60, length = 1)

nit: extra `;` here :)

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -218,12 +219,239 @@ export class ValidationResults {
...
+    targetVersion: Shared.TargetVersion;
+    value: string;
+}
+
+export class ValueSetPropertyDefinition extends TypedPropertyDefinition<string> {
+    parse(value: any): any {
+        let parsedValue = Utils.getStringValue(value, this.defaultValue);
+
+        for (let value of this.values) {
+            if (parsedValue.toLowerCase() === value.value) {
```
paulcam206: >parsedValue.toLowerCase() [](start = 16, length = 25)

might be a little more efficient to calculate the lowercase version once. perhaps 

```
let parsedValue = Utils.getStringValue(value, this.defaultValue).toLowerCase();
```

source/nodejs/adaptivecards/src/card-elements.ts
paulcam206: I see. Should I just review the end product rather than per-commit?

---
In reply to: [331170192](https://github.com/microsoft/AdaptiveCards/pull/3508#discussion_r331170192) [](ancestors = 331170192)

source/nodejs/adaptivecards/tsconfig.json
paulcam206: We're supposed to use tabs for json files generally according to our `.editorconfig`. You can configure VS Code to use the editor config with [this plugin](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig).

Regardless, we shouldn't mix tabs and spaces :)

---
In reply to: [331169850](https://github.com/microsoft/AdaptiveCards/pull/3508#discussion_r331169850) [](ancestors = 331169850)

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: This is outdated. BTW I don't know if your using CodeFlow or GitHub for this review, but GitHub has an "outdated" indicator that makes skipping over these easier:
image

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: No, you will still need to review every commit, sorry. But there will be multiple instances of outdated changes. GitHub shows an "outdated" tag if that helps.

source/nodejs/adaptivecards/tsconfig.json
dclaux: Last time I used tabs in the code I was asked to switch to spaces. Now that I have spaces, I should use tabs :)

I don't know how to configure VS Code to use tabs or spaces depending on the file type, and I also do not know about your guidelines...

paulcam206: Our guidelines are encoded in our `.editorconfig`:

https://github.com/microsoft/AdaptiveCards/blob/16172639b4aa70c7461e7b2b52824ae6d5c68d25/.editorconfig#L3-L14

If you configure the plugin, it will do The Right Thing.

dclaux: Ah I guess the plugin you pointed me to is the way to configure make VS Code use different settings for different file types. I will consider that in the future if you can point to what the per-file type guidelines are.

dclaux: Sorry, I posted my last comment before I saw yours. I'll look into it.

paulcam206: All you have to do is install the plugin. When you open a file, it will search parent directories until it finds a `.editorconfig` file, which it loads and uses to configure itself. We already have a `.editorconfig` in place.

dclaux: I just did it, and I can see it works.

dclaux:
Last commit (merge) also addresses review feedback so far.

source/nodejs/adaptivecards/src/card-elements.ts
paulcam206: unfortunately, that's not what I see:

image

paulcam206:
> export abstract class CardElement extends CardObject {

shouldn't all of the properties here be marked public?

---
Refers to: source/nodejs/adaptivecards/src/card-elements.ts:18 in 5da9b08. [](commit_id = 5da9b08aed29b7dae69da7e9ce8f025e912733fd, deletion_comment = False)

paulcam206:
>     protected createPlaceholderElement(): HTMLElement {

I think `protected` stuff should come before `private`, generally speaking

---
Refers to: source/nodejs/adaptivecards/src/card-elements.ts:203 in 5da9b08. [](commit_id = 5da9b08aed29b7dae69da7e9ce8f025e912733fd, deletion_comment = False)

source/nodejs/adaptivecards/src/shared.ts
```diff
@@ -14,27 +25,27 @@ export interface ISeparationDefinition {
...
 }
 
 export type Dictionary<T> = { [key: string]: T };
 
 export class StringWithSubstitutions {
 	private _isProcessed: boolean = false;
-	private _original: string = null;
-    private _processed: string = null;
+	private _original?: string;
+    private _processed?: string;
```
paulcam206: mixed tabs and spaces in this file

source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -44,109 +40,44 @@ export function getBoolValue(value: any, defaultValue: boolean): boolean {
...
-export function getEnumValue(targetEnum: { [s: number]: string }, name: string, defaultValue: number): number {
-    if (isNullOrEmpty(name)) {
-        return defaultValue;
-    }
-
-    for (var key in targetEnum) {
-        let isValueProperty = parseInt(key, 10) >= 0
+export function getEnumValueByName(enumType: { [s: number]: string }, name: string) : number | undefined {
+    for (let key in enumType) {
+        let keyAsNumber = parseInt(key, 10);
```
paulcam206: >let [](start = 8, length = 3)

could be `const` instead of `let`

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Weird...

source/nodejs/adaptivecards/src/utils.ts
dclaux: Sure, but that's a nit and if I do it here I would have to do it everywhere. Will keep the lets for the time being.

source/nodejs/adaptivecards/src/shared.ts
dclaux: Will fix

dclaux:
@paulcam206 not sure why but there are two of your comments I can't respond to inline. So here goes:
- `export abstract class CardElement extends CardObject {`: Shouldn't all of the properties here be marked public?
  - Not sure what you are actually asking. Does "here" refer to the CardElement class only? The entire library?
  - I assume by "marked public" you mean "use the `public` keyword"? If so, the library omits the public keyword throughout, by design.
- I think `protected` stuff should come before `private`, generally speaking
  - I don't know where that rule comes from, but the library attempts to declare things in this order: private statics, public statics, private fields, private methods, private properties, protected fields, protected methods, protected properties, public fields, constructor, public methods, public properties. All serializable classes declare schema-related fields/methods/properties at the very top of the class so it's easy to find. I like that order and I do not think there is a good reason to change it.

source/nodejs/adaptivecards/src/shared.ts
dclaux: Addressed with https://github.com/microsoft/AdaptiveCards/pull/3508/commits/34f9d1bc0147914f4dedf54a1367f3ab2e08e4b7

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Addressed with https://github.com/microsoft/AdaptiveCards/pull/3508/commits/5da9b08aed29b7dae69da7e9ce8f025e912733fd

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -1509,36 +1532,48 @@ export class AdaptiveCardPeer extends TypedCardElementPeer<Adaptive.AdaptiveCard
...
+                new PeerCommand(
+                    {
+                        name: "Add an action",
+                        iconClass: "acd-icon-bolt",
+                        execute: (command: PeerCommand, clickedElement: HTMLElement) => {
+                            let popupMenu = new Controls.PopupMenu();
 
-                            popupMenu.items.add(menuItem);
-                        }
+                            for (let typeRegistration of availableActions) {
```
nesalang: nit: indentation

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -1509,36 +1532,48 @@ export class AdaptiveCardPeer extends TypedCardElementPeer<Adaptive.AdaptiveCard
...
+                                let menuItem = new Controls.DropDownItem(i.toString(), typeRegistration.typeName);
```
nesalang: where did i come from?

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -317,29 +321,39 @@ export class CardDesigner {
...
-                for (let message of errorMessages) {
-                    errorPane.appendChild(this.renderErrorPaneElement("[Error] " + message));
-                }
-            }
-
-            if (validationResults.failures.length > 0) {
-                for (let failure of validationResults.failures) {
-                    for (let error of failure.errors) {
-                        errorPane.appendChild(this.renderErrorPaneElement("[" + failure.cardObject.getJsonTypeName() + "] " + error.message, failure.cardObject));
+                for (let entry of dedupedEntries) {
```
nesalang: nit: consider adding line 342-353 at line 337 after removing ln 337

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -350,16 +364,22 @@ export class CardDesigner {
                 errorPane.classList.add("acd-hidden");
             }
         };
+        this._designerSurface.onStartDrag = (sender: Designer.CardDesignerSurface) => {
+            this._startDragPayload = JSON.parse(this.getCurrentCardEditorPayload());
+        };
+        this._designerSurface.onEndDrag = (sender: Designer.CardDesignerSurface, wasCancelled: boolean) => {
+            if (wasCancelled) {
+                this.setCardPayload(this._startDragPayload, false);
```
nesalang: nit: is this statement hit when wasCancelled is set true?
nit: adding a hint of comments would be helpful.

source/nodejs/adaptivecards-designer/src/designer-peers.ts
dclaux: Not sure I understand the question?

source/nodejs/adaptivecards-designer/src/card-designer.ts
dclaux: This is an event handler. It will be called with wasCancelled either set to true or false.

nesalang: i was asking this to make sure when "wasCancelled"" is set true, it means it's cancelled; if not, having a comment woudl be helpful; just a suggestion...

---
In reply to: [333249001](https://github.com/microsoft/AdaptiveCards/pull/3508#discussion_r333249001) [](ancestors = 333249001)

source/nodejs/adaptivecards-designer/src/designer-peers.ts
nesalang: it's my fault. i should have set the comment at the exact spot. 

---
In reply to: [333671882](https://github.com/microsoft/AdaptiveCards/pull/3508#discussion_r333671882) [](ancestors = 333671882)

dclaux: Fixed with https://github.com/microsoft/AdaptiveCards/pull/3508/commits/3d3a1b8338432d304982196d0c5d7f5f77c2b59a

source/nodejs/adaptivecards/src/shared.ts
```diff
@@ -2,170 +2,183 @@
...
+                    if (contentType === ContentTypes.applicationJson) {
+                        valueForReplace = JSON.stringify(valueForReplace);
+                        valueForReplace = valueForReplace.slice(1, -1);
+                    }
+                    else if (contentType === ContentTypes.applicationXWwwFormUrlencoded) {
+                        valueForReplace = encodeURIComponent(valueForReplace);
+                    }
+
+                    this._processed = (<string>this._processed).replace(matches[0], valueForReplace);
+                }
```
nesalang: nit: consider moving these lines to LN67

source/nodejs/adaptivecards/src/shared.ts
```diff
@@ -2,170 +2,183 @@
...
+            };
```
nesalang: >; [](start = 13, length = 1)

nit: seems that `;`  is not needed.

nesalang:
>                 if (this.maxLines && this.maxLines > 0) {

nit: `if (this.maxLines > 0)`

---
Refers to: source/nodejs/adaptivecards/src/card-elements.ts:972 in b8c32c8. [](commit_id = b8c32c82cb8298b595377a0b0bac11843b465b2a, deletion_comment = False)

nesalang:
>                 if (this.maxLines && this.maxLines > 0) {

nit: consistency with LN 974 consider changing from `&& this.maxLines` to && this.maxLines > 0`

---
Refers to: source/nodejs/adaptivecards/src/card-elements.ts:972 in b8c32c8. [](commit_id = b8c32c82cb8298b595377a0b0bac11843b465b2a, deletion_comment = False)

source/nodejs/adaptivecards/src/registry.ts
```diff
@@ -0,0 +1,68 @@
...
+            }
+
+            this._items.push(registrationInfo);
+        }
+    }
+
+    unregister(typeName: string) {
+        for (let i = 0; i < this._items.length; i++) {
+            if (this._items[i].typeName === typeName) {
+                this._items.splice(i, 1);
```
nesalang: nit: consider using a dictionary.

source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -132,7 +140,7 @@ export class FactTitleDefinition extends FactTextDefinition {
 
         if (obj) {
             this.maxWidth = obj["maxWidth"] != null ? obj["maxWidth"] : this.maxWidth;
-			this.weight = Utils.parseHostConfigEnum(Enums.TextWeight, obj["weight"], Enums.TextWeight.Bolder);
+			this.weight = parseHostConfigEnum(Enums.TextWeight, obj["weight"], Enums.TextWeight.Bolder);
```
nesalang: nit: tab

source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -3,10 +3,18 @@
 import * as Enums from "./enums";
 import * as Utils from "./utils";
 import * as Shared from "./shared";
+import { HostCapabilities } from "./host-capabilities";
 
-export interface IValidationError {
-    error: Enums.ValidationError,
-    message: string;
+function parseHostConfigEnum(targetEnum: { [s: number]: string }, value: string | number, defaultValue: any): any {
```
nesalang: >any): any [](start = 104, length = 9)

`number` seems to be better choice since this function returns number, so it seems we can assume deafultValue to be number.

source/nodejs/adaptivecards/src/registry.ts
dclaux: Done here: https://github.com/microsoft/AdaptiveCards/pull/3508/commits/6f4fec115d80a095a9051d14fa8d288f4f0b7a14

source/nodejs/adaptivecards/src/host-config.ts
dclaux: Fixed: https://github.com/microsoft/AdaptiveCards/pull/3508/commits/6f4fec115d80a095a9051d14fa8d288f4f0b7a14

source/nodejs/adaptivecards/src/host-config.ts
dclaux: Fixed: https://github.com/microsoft/AdaptiveCards/pull/3508/commits/6f4fec115d80a095a9051d14fa8d288f4f0b7a14

source/nodejs/adaptivecards/src/serializable-object.ts
```diff
@@ -0,0 +1,284 @@
...
+    protected populateSchema(schema: SerializableObjectSchema) {
+        // Do nothing in base implementation
+    }
+
+    protected getValue(property: PropertyDefinition) {
+        return this._propertyBag.hasOwnProperty(property.name) ? this._propertyBag[property.name] : property.defaultValue;
+    }
+
+    protected setValue(property: PropertyDefinition, value: any) {
+        if (value === undefined) {
```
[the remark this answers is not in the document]
dclaux: Yes. Fixed: https://github.com/microsoft/AdaptiveCards/pull/3508/commits/431880c3068575ae57938a1a169857dcea238b86

source/nodejs/adaptivecards/src/shared.ts
dclaux: Done: https://github.com/microsoft/AdaptiveCards/pull/3508/commits/431880c3068575ae57938a1a169857dcea238b86

source/nodejs/adaptivecards/src/shared.ts
dclaux: Done: https://github.com/microsoft/AdaptiveCards/pull/3508/commits/431880c3068575ae57938a1a169857dcea238b86

source/nodejs/adaptivecards-designer/src/card-designer.ts
dclaux: I find it more legible to keep as is (e.g. first build deduped array, then loop through it.)

source/nodejs/adaptivecards-designer-app/src/app.ts
```diff
@@ -10,14 +10,10 @@ import "./app.css";
 // import "adaptivecards-designer/dist/adaptivecards-defaulthost.css";
 
 window.onload = function() {
-    // Uncomment to enabled preview features such as data binding
-    /*
```
shalinijoshi19: Sorry for jumping on this late @dclaux. But are we now enabling preview features by default on master with this uncommenting?

source/nodejs/adaptivecards-designer/src/card-designer.ts
dclaux: Seems pretty self explanatory to me.. What else could `wasCancelle` mean in an `endDrag` event handler? Comment would state what is already quite obvious IMO.

source/nodejs/adaptivecards-designer-app/src/app.ts
dclaux: Good catch - those are set to true so I can test. We need to decide what to do: the version picker should be enabled because this PR makes it a finished (as opposed to "preview") feature. Other flags should remain false I suppose.

But I'm not sure these changes should go to master just yet. Sent you a mail this morning about that.

shalinijoshi19: Agreed -- let's revert this to what it was until we can figure out an approach offline then? _The code itself landing in master is fine IMO; For publishing we could then discuss the environment - eg vnext perhaps before prod etc -- I am a bit hesitant exposing preview features as a default experience on our tooling lest we confuse our customers so thanks for starting that thread to kick start hte discussion! Thanks again!

dclaux: I am actually not sure the code landing in master is fine. Please refer to my email sent this morning, and let's take this offline.

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -260,52 +260,52 @@ export class CardDesignerSurface {
     private renderCard() {
         this._cardHost.innerHTML = "";
 
-        if (this.card) {
-            if (this.onCardValidated) {
-                this.onCardValidated(this._lastParseErrors, this.card.validateProperties());
+        if (this.onCardValidated) {
```
shalinijoshi19: Unrelated but should probable be addressed; Input Validation as a feature is not expected to land until our next big release (1.3); we had enabled it in master thinking it was closer to ship but since it's not can we disable the feature again please so we don't end up confusing customers? Thanks!

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -260,52 +260,52 @@ export class CardDesignerSurface {
...
-                if (Shared.GlobalSettings.enableDataBindingSupport) {
-                    let cardPayload = this.card.toJSON();
+        let cardToRender: Adaptive.AdaptiveCard = this.card;
 
-                    try {
-                        let template = new ACData.Template(cardPayload);
-            
-                        let context = new ACData.EvaluationContext();
-                        context.$root = this.sampleData;
+        if (this.isPreviewMode) {
```
shalinijoshi19: Trying to understand this one - so today previewmode is designervnext is it? how are you choosing to define "preview" in the codebase today?

dclaux: This is unrelated to "preview features." The designer can operate in "design mode" and "preview mode"; this flag says whether the designer is in "preview mode."

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
dclaux: There is no change to what's already in master here. Input validation (as currently implemented) is and has been disabled by default in JS.

viperfx:
Just wanted to say, out of all these cool features, I am looking forward to the version picker the most! As someone who targets Microsoft Teams exclusively, it would be really nice to get a version picker so I am not disappointed when a feature in Adaptive Cards is not available in Team :)  

Hope this comes out soon!

shalinijoshi19:
@dclaux are you able to verify these changes on the PR website also please? Thanks

shalinijoshi19:
@dclaux what is the status of this change again?

shalinijoshi19:
@matthidinger / @dclaux wondering about next steps for this PR?

shalinijoshi19:
@matthidinger looks like the code up on vnext.adaptivecards.io needs to be refreshed with the latest version of these changes from this branch. 
@dclaux:  This is available as an alpha package already being used by Msft flow. It does need testing/bashing. @nesalang can we try to slot this in as part of our bug bash scenarios in the next month or so? (this month is focused on Mobile so let's target Feb's bash for this)

nesalang:
> @matthidinger looks like the code up on vnext.adaptivecards.io needs to be refreshed with the latest version of these changes from this branch.
> @dclaux: This is available as an alpha package already being used by Msft flow. It does need testing/bashing. @nesalang can we try to slot this in as part of our bug bash scenarios in the next month or so? (this month is focused on Mobile so let's target Feb's bash for this)

sure

dclaux:
Styling has been slightly updated:
image

paulcam206:
@dclaux -- can you fix up the merge conflicts here?

paulcam206:
Looks like the temporary CI build fix for pulling the unpublished package works fine!

@shalinijoshi19 - FYI

Issue #3461, opened 2019-09-17
[Shared] ActionSet::m_orientation unused


paulcam206:
Looks like `ActionSet`'s `m_orientation` member is initialized, but not referenced (introduced in 393d1b0a5ac8dc62c1b750c8f643848753b31048).

Declared:
https://github.com/microsoft/AdaptiveCards/blob/302b0d8a397ef46fce1f80c73ff73253ba05397e/source/shared/cpp/ObjectModel/ActionSet.h#L12-L29

Only use:
https://github.com/microsoft/AdaptiveCards/blob/302b0d8a397ef46fce1f80c73ff73253ba05397e/source/shared/cpp/ObjectModel/ActionSet.cpp#L11-L20

paulcam206:
Assumption is that there's either some missing feature work here, or we can just get rid of the unused member...

dclaux:
ActionSet has an orientation property in JS allowing card authors to pick horizontal vs. vertical layout, overriding HostConfig. In many cases, horizontal orientation might be what works best at the card level, but not within a column.

paulcam206:
Makes sense. As implemented here, it's a set-only property :)

shalinijoshi19:
@Paulcam206 what is the proposal here with this issue?

RebeccaAnne:
@shalinjoshi19 The story here is that this member variable exists but is not hooked up to anything. It would in theory be in support of an orientation property on ActionSet. That is not an officially documented feature, but per David's comment it already exists on JavaScript. 

The proposal would be to either document and implement orientation for ActionSet on all non-JavaScript platforms, or otherwise to remove this stray variable that's not doing anything.

paulcam206:
This code has been removed for now. However, I'm leaving this issue open to track the possibility of eventually adding this functionality back in :)

PR #3520, opened 2019-10-11
[Shared] Introduce clang-cl build support

paulcam206:
## Description
VS 2019 provides support for something called `clang-cl`, which is a wrapper for `clang` that adapts `cl`-style command line arguments to `clang`-style command args. This allows traditional msbuild-based projects to (relatively) trivially use `clang` as their compiler. It's interesting to us because it allows us to integrate `clang` builds directly into our VS workflows (Android builds the shared model using `clang` via `cmake`) and make use of their analysis tooling.

In this delta, we enable `clang`-based builds via a new configuration (named `Clang-Debug`), and fix up issues exposed by `clang-tidy`:

* Add missing initialization
* Add missing `virtual` dtor declarations
* Add or fix missing or out-of-order class ctor member initializations
* Remove ignored `const` declarations on some return types
* Make `PopulateKnownPropertiesSet` no longer `virtual`
   * The warning from `clang` was that `PopulateKnownPropertiesSet` was a `virtual` method that is called from inside a ctor. The reason that's a warning is that the base class's call won't actually go to the overridden implementation (if one exists) due to how class initialization works (less derived ctors run first, so the overridden implementations aren't known yet). This is actually the behavior we want -- we were really only using `virtual` and `override` in this case to sort of imply a contract with derived classes. There's really no reason to do it this way, though, so it's probably best to structure the code how it's intended to run.
* Exempt jsoncpp from build tests

## How Verified
Unit tests

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3520)


source/shared/cpp/ObjectModel/ActionSet.h
```diff
@@ -23,9 +23,8 @@ namespace AdaptiveCards
         const std::vector<std::shared_ptr<BaseActionElement>>& GetActions() const;
 
     private:
-        void PopulateKnownPropertiesSet() override;
+        void PopulateKnownPropertiesSet();
         std::vector<std::shared_ptr<AdaptiveCards::BaseActionElement>> m_actions;
-        ActionsOrientation m_orientation;
```
RebeccaAnne: >m_orientation [](start = 27, length = 13)

Are we calling this a fix to #3461? Or is that bug now tracking the possibility of adding orientation to the action set element?

paulcam206: I'll mark 3461 as related :)

---
In reply to: [335689678](https://github.com/microsoft/AdaptiveCards/pull/3520#discussion_r335689678) [](ancestors = 335689678)

PR #3528, opened 2019-10-18
[UWP] Add accessible names to all action buttons

### LINKED ISSUE #3527: [UWP] Accessible names missing from some action buttons
In UWP, buttons don't have an accessible name. This isn't a problem for plain buttons, as accessibility tools are smart enough to infer that the text of the button is the name. However, for buttons that have an `iconUrl`, the tools can't infer because there are too many children. We should probably set the accessible name on all of our buttons.

paulcam206:
## Related Issue
Fixes #3527

## Description
While investigating alternatives for a partner team, I noticed that ActionSet buttons lack appropriate accessibility information, particularly when an `iconUrl` is present. Most accessibility tools will decide that if there's a single text child of a button, that the text child must be the name. However, if there's >1 child, it gives up. Consider this:

```json
{
    "type": "AdaptiveCard",
    "version": "1.1",
    "body": [
        {
            "type": "ActionSet",
            "actions": [
                {
                    "type": "Action.Submit",
                    "title": "action!",
                    "iconUrl": "data:image/png;base64,[...]="
                }
            ]
        }
    ]
}
```

Accessibility Insights will show this as not having a name. Similarly, Narrator will read it simply as "button".

image

The fix is to explicitly add the action title as the accessible name for the button control when it's built.

## How Verified
Verified with Accessibility Insights and Narrator

image

Narrator now says the name before saying button (e.g. before it would just say "button", now it might say something like "Microsoft To Do - button").

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3528)

### DIFF
--- a/source/uwp/Renderer/lib/ActionHelpers.cpp
+++ b/source/uwp/Renderer/lib/ActionHelpers.cpp
@@ -57,6 +57,12 @@ namespace AdaptiveNamespace::ActionHelpers
         THROW_IF_FAILED(action->get_IconUrl(iconUrl.GetAddressOf()));
 
         ComPtr<IButton> localButton(button);
+        ComPtr<IAutomationPropertiesStatics> automationProperties;
+        THROW_IF_FAILED(GetActivationFactory(
+            HStringReference(RuntimeClass_Windows_UI_Xaml_Automation_AutomationProperties).Get(), &automationProperties));
+        ComPtr<IDependencyObject> buttonAsDependencyObject;
+        THROW_IF_FAILED(localButton.As(&buttonAsDependencyObject));
+        THROW_IF_FAILED(automationProperties->SetName(buttonAsDependencyObject.Get(), title.Get()));
 
         // Check if the button has an iconUrl
         if (iconUrl != nullptr)


shalinijoshi19:
Do these need to be added to our sample cards that the Visualizer uses?

paulcam206:
> Do these need to be added to our sample cards that the Visualizer uses?

We already have cards with actions among our samples. What would probably be worthwhile in the future, though, would be to add accessibility tests.

shalinijoshi19:
So today the Accessibility tests are essentially the Visualizer + the pre-baked samples they come with. It might make sense therefore to just make sure we tweak those to include this sample as well

---
In reply to: [543987716](https://github.com/microsoft/AdaptiveCards/pull/3528#issuecomment-543987716) [](ancestors = 543987716)

shalinijoshi19:
@almedina-ms another instance of failing .NET PR builds

RebeccaAnne:
I think we probably want a better solution than that. The samples there are meant to be complete scenarios demonstrating the usages of cards to people exploring the platform. I don't think we want to pollute that with lower quality test case cards for all the accessibility scenarios we'll end up coming across. We probably need an approach that lets us define accessibility test cards independently of the scenario cards. 

---
In reply to: [544000775](https://github.com/microsoft/AdaptiveCards/pull/3528#issuecomment-544000775) [](ancestors = 544000775,543987716)

PR #3530, opened 2019-10-18
User/almedina ms/update net core version

almedina-ms:
## Related Issue
Builds are not working

## Description
For some reason version 3.0.100 was not being used in the builds so I'm updating the SDK version

## How Verified
This is what the test branch uses and it works

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3530)


source/dotnet/global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "3.0.100-preview6-012264"
+    "version": "3.0.100"
```
shalinijoshi19: Thanks. so now all our .NET builds want core or was it just that specific package you'd created? Unsure why the entire pipeline wants to install Core?

PR #3539, opened 2019-10-28
[Tools] Update scss-to-json dependencies

paulcam206:
## Description
Update to latest `node-sass` package to pick up downstream dependency updates.

## How Verified
`npm exec test`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3539)


shalinijoshi19:
@paulcam206 does this need to land for 1.2.4?  Where is this coming from?

paulcam206:
Strictly speaking we don't need this for 1.2.4. This requirement is coming from component governance. The dependencies being updated here are for a tool that isn't shipped by our release pipelines. However, rather than simply dismissing the warnings, it seemed better to fix them. This issue is marked as 1.2.4 because we're on the cusp of needing to renew our certs.

shalinijoshi19:
> Strictly speaking we don't need this for 1.2.4. This requirement is coming from component governance. The dependencies being updated here are for a tool that isn't shipped by our release pipelines. However, rather than simply dismissing the warnings, it seemed better to fix them. This issue is marked as 1.2.4 because we're on the cusp of needing to renew our certs.

I see .. So i've approved it but looks like there arent any noted JS-side releases going up today.. will this still be relevant to the publish/release pipeline?

shalinijoshi19:
@matthidinger / @dclaux for PR request

shalinijoshi19:
Since we aren't pushing any JS updates with 1.2.4 at this point, let's push this separately next month?

matthidinger:
This is a standalone tool written by a former WebChat dev to help work with SCSS/HostConfig. I'm actually not sure if it's ever been shipped anywhere. I think it was just parked in the `tools` folder to share the source code for folks. We could consider deprecating/removing it if it's causing you headache or issues?

PR #3540, opened 2019-10-28
Site compliance fixes

matthidinger:
## Description

Site compliance scan showed 2 issues:

1. Missing HTTP header `X-Content-Type-Options`
2. Non-HTTPs URLs for Microsoft properties.  I believe most of these were from the Sample payloads themselves, which render into HTML `<a>` and `` tags that should be `https://` and not `http://`

## How Verified

Built the site and will reattempt a scan.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3540)


shalinijoshi19:
Looks like an iOS test failure - likely related to the url change? @nesalang to confirm: this is the error: 

ADCIOSVisualizerTests
  testRemoteResouceInformation, (([[testStrings objectAtIndex:index++] isEqualToString:info.url.absoluteString]) is true) failed
  /Users/runner/runners/2.159.2/work/1/s/source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizerTests/ADCIOSVisualizerTests.mm:89
  ```

nesalang:
i'm looking at it.

PR #3546, opened 2019-10-29
Jwoo/dot net backup 1

nesalang:
## Description
To support concurrency during the parsing, this change commented definitions in ParseContext apis
This PR also include a unit test testing the concurrency in parsing.
Due to the change id collision detection won't work.
- json payload test will fail

## How Verified
1. created a new unit test to test concurrent parsing

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3546)


shalinijoshi19:
Nit: Can you formalize the description a little bit more ? Is this to unblock Bot Service ? Is this a temporary change?

source/dotnet/Library/AdaptiveCards/ParseContext.cs
```diff
@@ -1,7 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using Newtonsoft.Json;
```
shalinijoshi19: Looks like this file is missing our copyright header..could you add pls

shalinijoshi19:

source/dotnet/Library/AdaptiveCards/ParseContext.cs
```diff
@@ -13,6 +13,7 @@ public static void Initialize()
             idStack = new Stack<Tuple<string, AdaptiveInternalID, bool>>();
         }
 
+        private static Mutex mut = new Mutex();
```
shalinijoshi19: Why a mutex here instead of a simple Monitor-based lock (theC# lock statement effectively) ? A mutex is backed by kernel objects (so slower/heavier) typically to synchronize access for shared objects across processes.  Also i thought we needed to revisit the implementation here to make needing the lock unnecessary ? @Paulcam206? Not having to keep shared state would be ideal of course wrt perf and complexity but for hte short term is there a way to not even introduce any synch constructs and just turn off the usage of ParseContext?

source/dotnet/Library/AdaptiveCards/ParseContext.cs
nesalang: sure

---
In reply to: [340256403](https://github.com/microsoft/AdaptiveCards/pull/3546#discussion_r340256403) [](ancestors = 340256403)

source/dotnet/Library/AdaptiveCards/ParseContext.cs
nesalang: sure, i will replace it with a monitor.



---
In reply to: [340260783](https://github.com/microsoft/AdaptiveCards/pull/3546#discussion_r340260783) [](ancestors = 340260783)

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
```diff
@@ -1245,5 +1247,161 @@ public void TestParsingRichTextBlockWithInvalidInlineType()
...
+        {
+            // card with invalid inline type
+            var adaptiveCard =
+@"{
+	""$schema"": ""http://adaptivecards.io/schemas/adaptive-card.json"",
+	""type"": ""AdaptiveCard"",
+	""version"": ""1.0"",
+	""body"": [
+		{
+			""type"": ""TextBlock"",
```
paulcam206: >			 [](start = 0, length = 3)

looks like we've got tabs here

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
```diff
@@ -1245,5 +1247,161 @@ public void TestParsingRichTextBlockWithInvalidInlineType()
...
+            {
+                var payload = adaptiveCard;
+                Thread job = new Thread(() => RenderCardTask(payload));
+                jobs.Add(job);
+                job.Start();
+            }
+
+            foreach (var job in jobs)
+            {
+                job.Join();
```
shalinijoshi19: >job.Join(); [](start = 16, length = 11)

IF there is any unhandled exception anywhere in the workflow here there is a potential for thread.join to just block and hang the entire test process; Spinning up threads manually is also not typically needed now wiht the TPL.. Spin up tasks instead via the tasks parallel library (TPL) - look this up [https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/index](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/index)

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
```diff
@@ -5,12 +5,14 @@
...
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Newtonsoft.Json;
+using System.Threading;
 
 namespace AdaptiveCards.Test
 {
     [TestClass]
     public class AdaptiveCardApiTests
     {
+        List<Exception> exceptions = new List<Exception>();
```
shalinijoshi19: You can do away wiht the manual lock + List<> by switching to a ConcurrentBag<> which is thread safe and optimized for r/w from the calling thread.. I believe you dont care about the order of the exceptions in the list just that it should be non-empty - [https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/data-structures-for-parallel-programming](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/data-structures-for-parallel-programming)

source/dotnet/Library/AdaptiveCards/ParseContext.cs
shalinijoshi19: >mut [](start = 29, length = 3)

Nit Nit: Probably remember to name your variables well.

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
nesalang: thanks

---
In reply to: [340284336](https://github.com/microsoft/AdaptiveCards/pull/3546#discussion_r340284336) [](ancestors = 340284336)

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
nesalang: thanks

---
In reply to: [340304521](https://github.com/microsoft/AdaptiveCards/pull/3546#discussion_r340304521) [](ancestors = 340304521)

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
nesalang: exceptions are handled at LN1253-1263, I will look into TPL

---
In reply to: [340304111](https://github.com/microsoft/AdaptiveCards/pull/3546#discussion_r340304111) [](ancestors = 340304111)

source/dotnet/Library/AdaptiveCards/ParseContext.cs
```diff
@@ -1,7 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using Newtonsoft.Json;
+using System.Threading;
```
shalinijoshi19: We can problably remove this

nesalang: thanks

---
In reply to: [340381126](https://github.com/microsoft/AdaptiveCards/pull/3546#discussion_r340381126) [](ancestors = 340381126)

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
```diff
@@ -1245,5 +1248,151 @@ public void TestParsingRichTextBlockWithInvalidInlineType()
...
+                tasks.Add(task);
+            }
+
+            try
+            {
+                Task.WaitAll(tasks.ToArray());
+            }
+            catch
+            {
+                Assert.Fail();
```
shalinijoshi19: Nit: Should have a message explainng the assertion failure here for debugging

nesalang: thanks; will add it.

---
In reply to: [340389525](https://github.com/microsoft/AdaptiveCards/pull/3546#discussion_r340389525) [](ancestors = 340389525)

source/dotnet/Library/AdaptiveCards/ParseContext.cs
```diff
@@ -24,26 +24,36 @@ public enum ContextType { Element, Action };
         // Push the provided state on to our ID stack
         public static void PushElement(string idJsonProperty, AdaptiveInternalID internalId)
         {
+// re-enable when proper fix is checked-in
+#if USEFALLBACK
```
shalinijoshi19: >USEFALLBACK [](start = 4, length = 11)

Where is this being configured again?

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
```diff
@@ -5,6 +5,9 @@
 using System.Collections;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Newtonsoft.Json;
+using System.Threading;
```
shalinijoshi19: >Threading [](start = 13, length = 9)

We no longer need to include System.Threading at this point

shalinijoshi19:
@arturl is added to the review. #Closed

shalinijoshi19:
Should update the description to reflect latest changes we are making to clarify a stop gap measure to be able to spin up a private package for the Bot framework team

nesalang:
o.k.

---
In reply to: [547692997](https://github.com/microsoft/AdaptiveCards/pull/3546#issuecomment-547692997) [](ancestors = 547692997)

source/dotnet/Library/AdaptiveCards/ParseContext.cs
nesalang: yes, instead of commenting it out the define; i just didn't add it since we won't be debugging it.; but i will addi it.

---
In reply to: [340390172](https://github.com/microsoft/AdaptiveCards/pull/3546#discussion_r340390172) [](ancestors = 340390172)

paulcam206:
we don't need this any more, right?

PR #3547, opened 2019-10-29
[Android] Fix action icon being shown outside the rendered button

### LINKED ISSUE #3543: [Android][Button Icon Rendering] [Button icons placed on the left of the title are pushed out of the button, icon size not honored]
# Platform

* Android

# Author or host

Host

# Version of SDK

1.2.0

# Details

- Upon configuring the HostConfig to place the button icons to the left of the button title, the icons are pushed out of the bounds of the button.

- Icon size for the button icon as mentioned in the HostConfig is not honoured.

button_icon

Note the following observations : 
 

- For the case that there is only a single button the icon is infact rendered within the bounds of the button.

- For the case of multiple buttons, only the icon in the last button is rendered within the bounds of the button.

- The size of the icon is always rendered at the very small size as shown in the screenshot above irrespective of the value defined in the host config json.

Expecting the result to be similar to what is rendered on web, as shown in the image below
desktop_btns

Please find the following in the attached zip
1. Json for the card shown in the screenshot
2. HostConfig used for the card in the screenshot.
3. The screenshot itself.

[button_icon_render_issue.zip](https://github.com/microsoft/AdaptiveCards/files/3783314/button_icon_render_issue.zip)

almedina-ms:
## Related Issue
Fixes #3543 

## Description
Fixes the location of the icon when rendered in the left side of an action. Removes the OnLayoutChangedListener as using the compound padding is more than enough.

## How Verified
Manual verification was done using the given test file and an extra card to simulate if the behaviour happens when the buttons are bigger than the screen width

Screenshot_20191029-175107_AdaptiveCards Visualizer

Screenshot_20191029-175114_AdaptiveCards Visualizer

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3547)


shalinijoshi19:
Can we include the supplied json file in our own test samples for testing ?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/action/ActionElementRenderer.java
```diff
@@ -69,12 +69,16 @@ public static ActionElementRenderer getInstance()
     {
         private IconPlacement m_iconPlacement;
         private long m_iconSize;
+        private long m_padding;
+        private Context m_context;
 
-        protected ActionElementRendererIconImageLoaderAsync(RenderedAdaptiveCard renderedCard, View containerView, String imageBaseUrl, IconPlacement iconPlacement, long iconSize)
+        protected ActionElementRendererIconImageLoaderAsync(RenderedAdaptiveCard renderedCard, View containerView, String imageBaseUrl, IconPlacement iconPlacement, long iconSize, long padding, Context context)
```
shalinijoshi19: >ActionElementRendererIconImageLoaderAsync [](start = 18, length = 41)

Hmm what is the case for having a protected constructor for this private class?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
```diff
@@ -35,43 +35,6 @@
 
 public abstract class BaseActionElementRenderer implements IBaseActionElementRenderer
 {
-    protected static class ButtonOnLayoutChangedListener implements View.OnLayoutChangeListener
```
shalinijoshi19: >ButtonOnLayoutChangedListener [](start = 27, length = 29)

Is removing this a breaking contract change? Is this exposed today as part of the SDK API ? How does this relate to our extensibility story today ?

almedina-ms: this is not exposed, this listener was in charge of resizing and relocating the image when the button was rendered but it was not needed at all

shalinijoshi19: ok thanks the fact tthat this is public threw me off a bit

---
In reply to: [342855406](https://github.com/microsoft/AdaptiveCards/pull/3547#discussion_r342855406) [](ancestors = 342855406)

shalinijoshi19:
Signing off pending rest of the comments around sample additions for testing addressed.. thanks

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
shalinijoshi19: @Almedina-ms let's try to get this checked in today

PR #3557, opened 2019-10-31
[UWP] Update expected testapp results

paulcam206:
## Related Issue
Looks like we need to update our expected results in the UWP test app after #3540 (which moved everything from `http` to `https`).

## Description
Ran test app, accepted JSON changes, and spot checked results.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3557)


shalinijoshi19:
Thanks for catching + fixing @paulcam206 ; I am assuming the reason these didn't appear in the original PR to begin with is because we aren't running the UWP test in CI yet right? @RebeccaAnne FYI using this to bump up the backlog item around running the test app in CI that you are driving :) .

paulcam206:
I noticed this while testing my OpenUrl button changes :)

paulcam206:
and yes, @shalinijoshi19 -- not noticed in PR because no testapp execution in CI :)

PR #3558, opened 2019-10-31
[.NET][Core3.0] Split NETCore 3.0 to another solution

almedina-ms:
## Related Issue
Fixes intermitent build breaks

## Description
Splits the .NET solution into the original .NET solution and a specific solution for .NET Core 3.0 to avoid any future build breaks

## How Verified
Through modifying and running the build pipelines

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3558)


shalinijoshi19:
Per the offline thread, is this a stopgap mitigation at this point to recover the .NET builds ? If so lets update the description and then file a github issue tot rack the permanent solution.. Does it affect the release for this months patch from release/1,2?

almedina-ms:
> 
> 
> Per the offline thread, is this a stopgap mitigation at this point to recover the .NET builds ? If so lets update the description and then file a github issue tot rack the permanent solution.. Does it affect the release for this months patch from release/1,2?

With this PR I'm returning the .NET build to the state previous to the addition of the .NETCore 3.0 library, in the code as in the pipeline

This is the issue that I filed for this: https://github.com/microsoft/AdaptiveCards/issues/3560 

For the release patch, as long as this PR works, it shouldn't affect the patch as the code would be the same and I haven't touched the Release pipeline

nesalang:
i approved it given that this is a revert to th verified original. have you explored including .netcore 3.0 library installation as a build task?

---
In reply to: [548623712](https://github.com/microsoft/AdaptiveCards/pull/3558#issuecomment-548623712) [](ancestors = 548623712)

Issue #3564, opened 2019-11-01
[TS] Column default width is now 'auto' instead of 'stretch'


susanshen96:
# Platform

* Typescript

# Version of SDK

1.2.3

# Details

We've been seeing a regression in how some columns are being rendered on cards. It seems that the default width, when no value is provided by the developer, is now 'auto' instead of 'stretch', which is causing some 1.0 cards to break in terms of previous styling. I've included a screenshot of a sample card that exhibits this behavior. 

In Microsoft Teams, the card in question rendered as such, with the 1.0 version. 
![image](https://user-images.githubusercontent.com/10697383/68055697-80632100-fcae-11e9-9911-2e163d3e307f.png)

Now, with the 1.2 version, the card renders like so. 
![image](https://user-images.githubusercontent.com/10697383/68055684-750ff580-fcae-11e9-9a95-6dd75efdbabd.png)

shalinijoshi19:
@dclaux / @matthidinger  FYI moved to triage-investigate as David has picked this up for investigating (_we update the label and it will automatically move to the right projectboard column_)

Issue #3563, opened 2019-11-01
[TS] Empty fact title shows up as 'Title' instead of a blank


susanshen96:
# Platform

* Typescript

# Version of SDK

1.2.3

# Details
When a title is not provided for Fact, the default value is now 'Title' instead of a blank. 

I'm using Microsoft Teams, and this is how the card used to render, with the 1.0 SDK. 
![image](https://user-images.githubusercontent.com/10697383/68055377-c23f9780-fcad-11e9-9b1a-733f9f7a199b.png)

With the update to 1.2, this is what the card looks like.
![image](https://user-images.githubusercontent.com/10697383/68055481-ef8c4580-fcad-11e9-887c-7ab7bc6e2f6d.png)

shalinijoshi19:
@dclaux / @matthidinger FYI moved to triage-investigate as David has picked this up for investigating (we update the label and it will automatically move to the right projectboard column)

dclaux:
No investigation is needed, this is a regression and it needs to be fixed.

shalinijoshi19:
> No investigation is needed, this is a regression and it needs to be fixed.

Great. Thanks for confirming; In that case I just simply add the Triage-Approved for Fix label and the automation takes care of the rest of the workflow..

PR #3571, opened 2019-11-05
[TS] Regression fixes

dclaux:
## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/3564
Fixes https://github.com/microsoft/AdaptiveCards/issues/3563

## Description
- Reverted to **stretch** as the default column width.
- Only render the default **"Title"** value for a Fact's title in the designer

These fixes impact Teams and are somewhat urgent. A new npm package should be generated once they are merged.

## How Verified
Verified manually in both adaptivecards-designer-app and in adaptivecards-visualizer

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3571)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -296,15 +296,15 @@ export abstract class CardElement extends CardObject {
             if (AdaptiveCard.alwaysBleedSeparators && renderedSeparator && this.separatorOrientation == Enums.Orientation.Horizontal) {
                 // Adjust separator's margins if the option to always bleed separators is turned on
                 let parentContainer = this.getParentContainer();
-    
+
                 if (parentContainer && parentContainer.getEffectivePadding()) {
                     let parentPhysicalPadding = this.hostConfig.paddingDefinitionToSpacingDefinition(parentContainer.getEffectivePadding());
-    
+
                     renderedSeparator.style.marginLeft = "-" + parentPhysicalPadding.left + "px";
                     renderedSeparator.style.marginRight = "-" + parentPhysicalPadding.right + "px";
                 }
             }
-    
+
             return renderedSeparator;
     }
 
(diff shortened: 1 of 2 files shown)


shalinijoshi19:
are the desktio team consuming the JS renderer off of our npm package or direct from source?

source/nodejs/package-lock.json
```diff
@@ -5757,20 +5754,17 @@
...
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "console-control-strings": {
           "version": "1.1.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
```
shalinijoshi19: Question: Are these intentional changes?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5843,9 +5829,9 @@ export class Column extends Container {
         return Enums.Orientation.Vertical;
     }
 
-    width: ColumnWidth = "auto";
+    width: ColumnWidth = "stretch";
```
shalinijoshi19: Question: Is this default behavior documented someplace where we will need to update our documentation as well with this change?

shalinijoshi19:
@RebeccaAnne , @paulcam206 , @nesalang could one of you take a look at this PR real quick? SHould be a quick review. tx

source/nodejs/adaptivecards/src/card-elements.ts
RebeccaAnne: Looks like the default is not doc'd here: [https://adaptivecards.io/explorer/Column.html](https://adaptivecards.io/explorer/Column.html). It probably should be.

Looks like shared model is defaulting to stretch as well.

---
In reply to: [342851467](https://github.com/microsoft/AdaptiveCards/pull/3571#discussion_r342851467) [](ancestors = 342851467)

Issue #3580, opened 2019-11-07
[TS] Column width no longer accepts decimals as valid values


susanshen96:
# Platform

* Typescript

# Version of SDK

1.2.3 

# Details

On Teams, we've discovered that an app that has been using adaptive cards has been setting the width of a column to 2.6, and the card now rendering differently than what it used to. It seems that the SDK renderer no longer accepts decimal values as valid values for column widths, and is treating it as invalid and setting it to the default width, which is now "stretch". Screenshots of how the card used to look like vs. now are included, along with the JSON payload used to generate this card. 

**1.0:** 
![image](https://user-images.githubusercontent.com/10697383/68433156-bef35280-016a-11ea-9663-627b968ee4bb.png)

**1.2:** 
![image](https://user-images.githubusercontent.com/10697383/68433141-b569ea80-016a-11ea-989f-fdc8d16d04ff.png)

**JSON payload:** 
`{
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "ColumnSet",
            "columns": [
                {
                    "type": "Column",
                    "items": [
                        {
                            "type": "Image",
                            "url": "http://adaptivecards.io/content/cats/1.png",
                            "altText": "Creative",
                            "horizontalAlignment": "Center"
                        }
                    ],
                    "width": 2.6
                },
                {
                    "type": "Column",
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "Susan Shen sent praise to",
                            "spacing": "None",
                            "wrap": true,
                            "isSubtle": true
                        },
                        {
                            "type": "TextBlock",
                            "text": "New TextBlock"
                        },
                        {
                            "type": "TextBlock",
                            "text": "New TextBlock"
                        }
                    ],
                    "width": 6,
                    "spacing": "Medium"
                }
            ]
        }
    ]
}`

PR #3586, opened 2019-11-08
[TS] Fix weighted column width regression

dclaux:
## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/3580

## Description
Accept decimal numbers for column width

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3586)


source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -75,10 +75,10 @@ export class PropertySheetCategory {
                 let header = new Adaptive.TextBlock();
```
shalinijoshi19: Could we test this out in the actual website designer as well while you are at it? Thanks!

dclaux: It is the exact same thing. Also the designer is used in third party applications I can't test - if I can't rely on testing the designer component on its own (but I can), I think we have a problem.

PR #3555, opened 2019-10-31
[UWP] Use custom button to show OpenUrl actions as links

paulcam206:
Fixes #3562

## Description
OpenUrl actions should appear as a link to accessibility. Unfortunately, to make that work, we need to create a new type of button. Unfortunately, to do that in a reasonable way, we need to use C++/CX or C++/WinRT rather than going directly against the ABI. C++/CX is a non-starter for us, but C++/WinRT is just right. 

## How Verified
* [Accessibility Insights](https://accessibilityinsights.io/)
* UWP visualizer
* UWP TestApp

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3555)

PR #3593, opened 2019-11-11
[UWP] OpenUrl selectActions appear as links to accessibility

### LINKED ISSUE #3591: [UWP][Accessibility] SelectActions should be exposed to accessibility as links
Currently, elements with a `selectAction` are shown to accessibility as `button`, but they should be of `link` type.

paulcam206:
## Related Issue
Fixes #3591
Related #3555 

## Description
There was another scenario where an action was being reported as a `button` rather than a `link` to accessibility. The fix is to instantiate the appropriate `LinkButton` for `selectAction`s of type `Action.OpenUrl`. I also looked for other instances where this would be appropriate, but found none.

## How Verified
Unit tests, Visualizer+AccInsights, and TestApp

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3593)

### DIFF
--- a/source/uwp/Renderer/lib/XamlBuilder.cpp
+++ b/source/uwp/Renderer/lib/XamlBuilder.cpp
@@ -1847,20 +1847,12 @@ namespace AdaptiveNamespace
         ABI::AdaptiveNamespace::ActionType actionType;
         RETURN_IF_FAILED(action->get_ActionType(&actionType));
 
-        // now construct an appropriate button for the action type
         ComPtr<IButton> button;
-        if (actionType == ABI::AdaptiveNamespace::ActionType_OpenUrl)
+        try
         {
-            // OpenUrl buttons should appear as links for accessibility purposes, so we use our custom LinkButton.
-            auto linkButton = winrt::make<LinkButton>();
-            button = linkButton.as<IButton>().detach();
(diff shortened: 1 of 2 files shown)


source/uwp/Renderer/lib/XamlBuilder.cpp
```diff
@@ -1847,20 +1847,12 @@ namespace AdaptiveNamespace
         ABI::AdaptiveNamespace::ActionType actionType;
         RETURN_IF_FAILED(action->get_ActionType(&actionType));
 
-        // now construct an appropriate button for the action type
```
shalinijoshi19: Do we actually have a unit test that you added or had failed earlier ? Or are we checking for no regressions wiht the existing unit tests at this point? Latter if fine (probably update teh HowVerified to clarify on the unit test bit)

paulcam206: I just ran existing tests. Unfortunately, we don't have a great way to perform a unit test for this type of change at the moment.

paulcam206:
Note -- above commit is cherry-pick to `master` :)

PR #3594, opened 2019-11-12
[Android] Fix spacing in ChoiceSets with wrappjng

### LINKED ISSUE #3554: [Android][Element Rendering] [Spacing within Choice sets not rendered]
**this needs to be fixed for radio button as well**

# Platform
* Android
# Author or host
Host
# Version of SDK
1.2.0
# Details
There is no spacing between the elements in a choice set, for an "expanded" style of the choice set, for both checkboxes and radio buttons. The issue is illustrated in the image below
choice_without_wrap
However upon setting "wrap" : true, the spacing shows up
choice_with_wrap
The attached zip contains the card jsons and screenshots
[choice_set_spacing.zip](https://github.com/microsoft/AdaptiveCards/files/3792740/choice_set_spacing.zip)

almedina-ms:
## Related Issue
Fixes #3554 

## Description
When having a choice set with wrap set to false, the options appear one after the other with no spacing between them, while wrapping options appear with the previously mentioned spacing

## How Verified
The test cases were used in the sample application and it yielded this result. As can be seen the spacing is kept while the text is truncated as expected.

Screenshot_20191111-213920_AdaptiveCards Visualizer

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3594)


shalinijoshi19:
@Almedina-ms this needs to be made against release/1.2 bits to ensure it lands in the 1.2.5 branch - can you send out a PR for that as well? (either a CP after you commit this to master OR manually editing the 1.2 version of the code - whichever is easier). Thanks

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/ChoiceSetInputRenderer.java
```diff
@@ -76,6 +76,7 @@ public View renderCheckBoxSet(
         List<CheckBox> checkBoxList = new Vector<CheckBox>();
         ChoiceInputVector choiceInputVector = choiceSetInput.GetChoices();
         long size = choiceInputVector.size();
```
shalinijoshi19: Do we have a sample card that you can check in to teh samples folder for the bug bash testing. Also can we add a unit test for this?

Issue #3595, opened 2019-11-12
[UWP] Symbols incorrectly published for UWP.


nimercer:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* UWP

# Author or host

Host: Cortana UWP.

# Version of SDK

Currently upgrading to 1.2.4

# Details
The symbols for the UWP renderer are uploaded automatically to the nuget symbol server using the -Symbol flag on nuget pack. However, this server only supports managed code so the symbols can never be found.

Instead we should pack them inside the regular package.

shalinijoshi19:
This is fixed already.

PR #3596, opened 2019-11-12
Remove symbols nuget package signing reference

nimercer:
#3595  Related Issue
[UWP] Symbols incorrectly published for UWP.

## Description
The UWP renderer is native code which is unsupported by the nuget symbol server. Currently the release process nuget packs with the Symbols flag to generate a separate symbols.nupkg that is uploaded to that server.

If we remove that from the pipeline, this config file still expects to find it. So we need to remove this reference.

## How Verified
1. n/a.
2. n/a
3. n/a, need to make this change before changing the pipeline.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3596)


shalinijoshi19:
Thanks for fixing this @nimercer !

PR #3607, opened 2019-11-13
Adding /profile linker option

### LINKED ISSUE #3606: [UWP] Renderer binaries are not static analysis friendly
**What platform is your feature request related to? (Delete other platforms).**
* UWP

**Is your feature request related to a problem? Please describe.**
Our compliance process requires binaries to be built with /profile to generate additional context for mandatory static analysis. 

**Describe the solution you'd like**
Add the /profile linker option. This should not change our dll from its existing state [Documented here ](https://docs.microsoft.com/en-us/cpp/build/reference/profile-performance-tools-profiler?view=vs-2019)

Note: The documentation specifies that the program image will be affected by a "relocation section"- however this is caused by the implied /FIXED:NO flag. /FIXED:NO is the default for msbuild, so we were generating that anyway.

**Describe alternatives you've considered**
I've thought about pairing a local build of this PDB to avoid affecting the build system, but this would be a bit of an overhead for updating in the future.

nimercer:
For this issue: https://github.com/microsoft/AdaptiveCards/issues/3606

This flag is needed for mandatory static analysis as part of our process.

## Related Issue
Fixes #3606 

## Description
Our compliance process requires binaries to be built with /profile to generate additional context for mandatory static analysis.

This should not change our dll from its existing state [Documented here](https://docs.microsoft.com/en-us/cpp/build/reference/profile-performance-tools-profiler?view=vs-2019)

Note: The documentation specifies that the program image will be affected by a "relocation section"- however this is caused by the implied /FIXED:NO flag. /FIXED:NO is the default for msbuild, so we were generating that anyway.

## How Verified
No unit testing. I built locally with this flag and passed it through the static analysis tool and where before it had errors it now passes.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3607)

### DIFF
--- a/source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
+++ b/source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
@@ -93,6 +93,7 @@
         user32.lib;
         %(AdditionalDependencies);
       </AdditionalDependencies>
+      <AdditionalOptions>/profile %(AdditionalOptions)</AdditionalOptions>
     </Link>
     <CustomBuildStep>
       <Command>mdmerge -partial -i "$(OutDir)$(TargetName).winmd" -o "$(OutDir)Output" -metadata_dir "$(WindowsSDK_MetadataFoundationPath)" -metadata_dir "$(WindowsSDK_MetadataPathVersioned)\Windows.Foundation.UniversalApiContract\8.0.0.0" &amp;&amp; copy /y "$(OutDir)Output\*" "$(OutDir)"</Command>


paulcam206:
Interesting. It looks like the SharedModel tests are failing with a privilege problem. **However**, it doesn't appear to be related to this change. A manually-queued job against `master` had the [same problem](https://microsoft.visualstudio.com/AdaptiveCards/_build/results?buildId=19598956&view=results).

Failing test code in question:
https://github.com/microsoft/AdaptiveCards/blob/db318f581699f0e7e158aab1a40c325afb134c70/source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/DateAndTimeUnitTest.cpp#L19-L32

Failure:
```
[18:42:41] Exception: E06D7363.?AVCSEException@CppUnitTestFramework@VisualStudio@Microsoft@@
  X AdaptiveCardParseWarningTest
  Error Message:
   Exception Code: C0000096
  Stack Trace:
  	at AdaptiveCardsSharedModelUnitTest::EnableTimeZoneChange() in E:\BA\98\s\source\shared\cpp\AdaptiveCardsSharedModel\AdaptiveCardsSharedModelUnitTest\DateAndTimeUnitTest.cpp:line 32
	at AdaptiveCardsSharedModelUnitTest::InitializeTimeZone() in E:\BA\98\s\source\shared\cpp\AdaptiveCardsSharedModel\AdaptiveCardsSharedModelUnitTest\DateAndTimeUnitTest.cpp:line 60
```

I'm guessing this must be a change in the execution environment for the tests in CI (timezone adjusting privs might no longer be granted?). I'm still looking at this, but don't think the failure should hold up this PR.

shalinijoshi19:
Thanks for investigating Paul; Is this something that is likely going to cause failures in the release pipeline job also? 

---
In reply to: [553551998](https://github.com/microsoft/AdaptiveCards/pull/3607#issuecomment-553551998) [](ancestors = 553551998)

paulcam206:
shouldn't be a problem -- the UWP release pipeline doesn't build the SharedModel project :)

---
In reply to: [553557217](https://github.com/microsoft/AdaptiveCards/pull/3607#issuecomment-553557217) [](ancestors = 553557217,553551998)

shalinijoshi19:
> shouldn't be a problem -- the UWP release pipeline doesn't build the SharedModel project :)
> 
> In reply to: [553557217](https://github.com/microsoft/AdaptiveCards/pull/3607#issuecomment-553557217) [](ancestors = 553557217,553551998)

aaah good save - question back to you: Should it or why doesn't it already today (longer term)?

paulcam206:
building and testing sharedmodel during the UWP pipeline can be added, but I think it makes more sense to have that change as part of our push for using proper release pipelines. 

---
In reply to: [553558283](https://github.com/microsoft/AdaptiveCards/pull/3607#issuecomment-553558283) [](ancestors = 553558283)

PR #3611, opened 2019-11-13
[SharedModel] Skip date/time tests outside of Pacific TZ

### LINKED ISSUE #3609: [SharedModel] DateAndTime tests failing in CI
Presumably this is due to a change in the test execution environment, as neither the test code nor the implementation have changed. See discussion here: https://github.com/microsoft/AdaptiveCards/pull/3607#issuecomment-553551998

paulcam206:
## Related Issue
Fixes #3609

## Description
It's a bad idea to dynamically modify the timezone of the build machines where these unit tests sometimes run. To unblock us for now, I've added logic to skip running these specific tests when the hosting machine isn't in the Pacific timezone.

## How Verified
Unit tests & debugger

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3611)


source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/DateAndTimeUnitTest.cpp
```diff
@@ -11,82 +11,37 @@
...
-            {
-                Assert::Fail(L"Could not change the timezone");
-            }
-            changedTime = true;
-        }
-    }
+        // Make sure we only run our tests in Pacific timezone
+        TIME_ZONE_INFORMATION tz{};
+        GetTimeZoneInformation(&tz);
+        fShouldRunTests = (tz.Bias == 480l);
```
shalinijoshi19: >480l [](start = 38, length = 4)

Hmm does the bias here account for both PDT and PST? Do we need to check the StandardBias and DaylightBias fields separately?

shalinijoshi19: Also i remembered on the Rome codebase we'd been working with std::chrono for most/all time-related operations --apparently it has timezone helpers? [https://en.cppreference.com/w/cpp/chrono/current_zone](https://en.cppreference.com/w/cpp/chrono/current_zone) ? Maybe? 

---
In reply to: [346021077](https://github.com/microsoft/AdaptiveCards/pull/3611#discussion_r346021077) [](ancestors = 346021077)

paulcam206: looks like `std::chrono::current_zone` is a C++20 feature, so it's probably not something we want to take a dependency on just yet :)

---
In reply to: [346023811](https://github.com/microsoft/AdaptiveCards/pull/3611#discussion_r346023811) [](ancestors = 346023811,346021077)

paulcam206: also, we don't have to check the `StandardBias` or `DaylightBias` fields as they are intended to be added to `Bias` to get actual offset from UTC (i.e. for Pacific time, `StandardBias` is `0` and `DaylightBias` is `-60`

---
In reply to: [346035923](https://github.com/microsoft/AdaptiveCards/pull/3611#discussion_r346035923) [](ancestors = 346035923,346023811,346021077)

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/DateAndTimeUnitTest.cpp
```diff
@@ -11,82 +11,30 @@
...
-            bool timeZoneChanged = SetTimeZoneInformation(&tzSeattle);
-            EnableTimeZoneChange(false);
-            if (!timeZoneChanged)
-            {
-                Assert::Fail(L"Could not change the timezone");
-            }
-            changedTime = true;
-        }
+// Use this macro to effectively skip a test when run outside of the Pacific timezone
+#define SKIP_TEST_IF_NEEDED()                                           \
```
RebeccaAnne: >SKIP_TEST_IF_NEEDED [](start = 8, length = 19)

I'd put something in the name of this macro to indicate that it's related to date time stuff. SKIP_TEST_OUTSIDE_PACIFIC_TIME? Or something.

PR #3615, opened 2019-11-15
[Android] Add empty options support for ChoiceSet

### LINKED ISSUE #3605: [Android][Rendering Discrepancy between Android and web] [For Choice sets with compact style, when no option is explicitly pre-selected, first option gets pre-selected]
# Platform
* Android
# Author or host
Host
# Version of SDK
1.2
# Details
For choice sets of the type shown below in the screenshot, even if no value is specified to be pre selected, the first option shows up pre selected.
Screenshot 2019-11-13 at 3 49 27 PM
On the other hand web shows nothing
Screenshot 2019-11-13 at 3 50 25 PM

Please find the json for the card in the attached zip
[choice_set_pre_selected.zip](https://github.com/microsoft/AdaptiveCards/files/3840574/choice_set_pre_selected.zip)

almedina-ms:
## Related Issue
Fixes #3605 

## Description
Overrides the base behaviour of the ArrayAdapter to handle default empty values in such a way where the empty option is the default value but it's not a selectable value afterwards

## How Verified
The scenario was tested manually using the attached payload yielding this result

On the first image we can see that the empty value is "selected"
Screenshot_20191115-131748_AdaptiveCards Visualizer

On the second image we can see that the empty value is not an option
Screenshot_20191115-131754_AdaptiveCards Visualizer

On the third image we can see the selected non-empty element
Screenshot_20191115-131805_AdaptiveCards Visualizer

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3615)

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/ChoiceSetInputRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/ChoiceSetInputRenderer.java
@@ -49,26 +49,15 @@ import java.util.Vector;
 
 class ItemSelectedListener implements AdapterView.OnItemSelectedListener
 {
-    public ItemSelectedListener(ComboBoxInputHandler comboBoxInputHandler, boolean hasEmptyDefault, Vector<String> titles, ArrayAdapter arrayAdapter)
+    public ItemSelectedListener(ComboBoxInputHandler comboBoxInputHandler)
     {
         m_InputHandler = comboBoxInputHandler;
-        m_hasEmptyDefault = hasEmptyDefault;
-        m_arrayAdapter = arrayAdapter;
-        m_titles = titles;
     }
 
     @Override
     public void onItemSelected(AdapterView<?> parent, View view, int position, long id)
     {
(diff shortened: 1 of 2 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/ChoiceSetInputRenderer.java
```diff
@@ -46,6 +47,28 @@
 import java.util.List;
 import java.util.Vector;
 
+class ItemSelectedListener implements AdapterView.OnItemSelectedListener
```
RebeccaAnne: >class ItemSelectedListener implements AdapterView.OnItemSelectedListene [](start = 0, length = 71)

Could you add comments to this class explaining why it is needed and what it is doing for future reference? Thanks!

almedina-ms: that class was not needed XD

Issue #203, opened 2017-04-05
Need way to associate TextBlock with Input (or add Label)


matthidinger:

shalinijoshi19:
Dup of #3081

PR #3618, opened 2019-11-15
Input Labels Spec Documents

RebeccaAnne:
## Related Issue
#203

## Description
Design spec for input labels. Includes:
 - InputLabels.md, which discusses design options for input labels
 - Updated Input.Json file for proposed design, and generated downstream Input.*.md & adaptive-card.json files. 
 - Updated Input.*.md file Renderer section with relevant details
- Sample file in Input.Label.json and generated UWP test results

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3618)


specs/DesignDiscussions/InputLabels.md
```diff
@@ -0,0 +1,279 @@
...
+			"title": "OK"
+		}
+	]
+}
+```
+This has the benefit of being simple and concise, but doesn’t immediately allow for formatting on the input label. A solution to that would be to add support for TextBlock and/or RichTextBlock labels to add formatting support. This is similar to the approach in Xaml:
+
+```json
+{
+	"type": "Input.Text",
```
RebeccaAnne: >	 [](start = 0, length = 1)

Tabs vs. Spaces

specs/DesignDiscussions/InputLabels.md
```diff
@@ -0,0 +1,279 @@
...
+        "type": "TextBlock",
+        "text": "Please enter your name:",
+        "weight": "Bolder",
+        "color": "Good"
+    },
+    "placeholder": "Last, First"
+}
+```
+
+We should also consider which item types are valid as labels. Probably just TextBlock and RichTextBlock make sense, but as Xaml supports other types we should consider whether we want to as well.
```
RebeccaAnne: Discussion in room: Probably fine with text types only, but will follow up with customers once we have a prototype.

specs/DesignDiscussions/InputLabels.md
```diff
@@ -0,0 +1,279 @@
...
+
+``` json
+{
+	"type": "AdaptiveCard",
+	"version": "1.0",
+	"body": [
+		{
+			"type": "Input.Text",
+			"id": "nameInput",
+			"label": "Please enter your name:",
```
RebeccaAnne: >label [](start = 4, length = 5)

Comment in the room from Matt: This has a bad back compat story in that the label would be *lost* on earlier versions. That said, the templating library may provide a compatibility layer in the future.

Fallback doesn't work for this scenario, but we could use requires. Even requires doesn't help pre 1.2.

RebeccaAnne: A new element type (LabeledInput) would make the fallback story better.

---
In reply to: [347039377](https://github.com/microsoft/AdaptiveCards/pull/3618#discussion_r347039377) [](ancestors = 347039377)

specs/DesignDiscussions/InputLabels.md
```diff
@@ -0,0 +1,279 @@
...
+```json
+{
+	"type": "Input.Text",
+	"id": "nameInput",
+	"label": "Please enter your name:",
+	"labelPosition": "Side",
+	"placeholder": "Last, First"
+}
+```
+![img](assets/InputLabels/LabelPositionSide.png)
```
RebeccaAnne: >LabelPositionSide [](start = 26, length = 17)

Matt: We would also need a way to group inputs so that their labels have the same width and the inputs line up.

RebeccaAnne:
Matt: What about other inputs? Consider how this works against non Input.Text inputs.

specs/DesignDiscussions/InputLabels.md
```diff
@@ -0,0 +1,279 @@
...
+    "actions": [
+        {
+            "type": "Action.Submit",
+            "title": "Submit"
+        }
+    ]
+}
+```
+
+![img](assets/InputLabels/Placeholders.png)
```
RebeccaAnne: Alberto: Formatted placeholders likely wouldn't work

RebeccaAnne:
@matthidinger Added a section with examples of all input types. I don't think there are any special considerations for other types, but let me know if you have further concerns.

---
In reply to: [554560109](https://github.com/microsoft/AdaptiveCards/pull/3618#issuecomment-554560109) [](ancestors = 554560109)

specs/DesignDiscussions/InputLabels.md
RebeccaAnne: @paulcam206 @matthidinger Added back compat sections to each option. I realized that we don't necessarily need a new requires type, the author could just require adaptive cards 1.3. 

Also added a third option to capture @paulcam206's idea of using a new type which includes both the input and the label.

---
In reply to: [347039935](https://github.com/microsoft/AdaptiveCards/pull/3618#discussion_r347039935) [](ancestors = 347039935,347039377)

specs/DesignDiscussions/InputLabels.md
RebeccaAnne: Added a note about this to the doc.

---
In reply to: [347042572](https://github.com/microsoft/AdaptiveCards/pull/3618#discussion_r347042572) [](ancestors = 347042572)

specs/DesignDiscussions/InputLabels.md
```diff
@@ -0,0 +1,279 @@
...
+	"actions": [
+		{
+			"type": "Action.Submit",
+			"title": "OK"
+		}
+	]
+}
+```
+
+![img](assets/InputLabels/InputCard.png)
```
shalinijoshi19: This image doesnt render? (Results in a 404 when clicked on also)

RebeccaAnne: If you're just looking at the rich diff in the Files Changed tab, the images don't resolve. Try clicking the "..." all the way on the right and then choosing "View file". Let me know if that doesn't work for you.

---
In reply to: [352410831](https://github.com/microsoft/AdaptiveCards/pull/3618#discussion_r352410831) [](ancestors = 352410831)

specs/DesignDiscussions/InputLabels.md
RebeccaAnne: Added comments to the spec about this case.

---
In reply to: [347043525](https://github.com/microsoft/AdaptiveCards/pull/3618#discussion_r347043525) [](ancestors = 347043525)

PR #3639, opened 2019-11-22
Adding Sample App's Workspace File that uses AdaptiveCards iOS package (cocoapods)

nesalang:
* ultimately, iOS's bug bashing requirement will be just downloading the sample app from app store.
* this change is one of the small steps that will get us there.
* this change repurposes existing sample app's project file to create the same sample app but uses adaptive cards cocoapods.
* it can be run from dev box, or build agents.
* by running the workspace file in this PR, sample app will be built using the latest cocoapods.
* podspec file can be modified to include any versions of adaptive cards pod.
* convenience script will be followed to assist updating the pod from azure build servers.
## **usage for dev machine** (usage for build pipeline is coming)
* locate directory of `Podfile`
* cd into the directory.
* `pod install`
*  (becomes required step only if one wants to test with pre-release pods) run script (in development) to update pod with ones from build servers.
*  `open ADCIOSVisualizer.xcworkspace/ `

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3639)


shalinijoshi19:
@nesalang is this still needed for the appcenter publishing instead of cocoapod? Also the iOS builds seem to be failing?

PR #3640, opened 2019-11-24
in-built markdown processing for NodeJS adaptive cards

### LINKED ISSUE #910: JavaScript library does not natively support markdown, and third-party libs are too permissive
This sample has a lot of markdown syntax that shouldn't be supported according to the docs.
```json
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "TextBlock",
            "text": "## Markdown Headers should not be supported"
        },
        {
            "type": "TextBlock",
            "text": "___"
        },
        {
            "type": "TextBlock",
            "text": "Neither should horizontal rules"
        },
        {
            "type": "TextBlock",
            "text": "Some of my spaces are being eaten up by the         markdown parser"
        },
        {
            "type": "TextBlock",
            "text": "~~Strikethrough should not be supported.~~"
        },
        {
            "type": "TextBlock",
            "text": ">Quote blocks should be ignored"
        },
        {
            "type": "TextBlock",
            "text": "Markdown | Tables | Work\n:--- | :---: | ---:\n*And* | change | **horizontal alignment**\n1 | 2 | 3"
        },
        {
            "type": "TextBlock",
            "text": "    This one I don't really understand"
        },
        {
            "type": "TextBlock",
            "text": "````\nBut Seems to be a shortcut for code blocks, which should be ignored."
        }
    ]
}
```

Renders as:

image

bhuvana-guna:
## Related Issue
Fixes #910 

## Description
**Built-in simple markdown processing for NodeJS Adaptive Cards**

Adaptive card implementation in NodeJS now sees if the markdown-it library is included, and if so, uses that to process markdown. To remove the dependency from markdown-it and also directly enable the users to use markdown, we have added a simple markdown processing inside the NodeJS adaptive cards. It also gives the user the option to extend and customize. 

If the current markdown-it users want to continue using the same, they would just have to call markdown-it's render() function inside the onProcessMardown event. We have added the example code in README. 

Other users who are using their own markdown logic or other 3rd party libraries won't need to change anything.

## How Verified

Added the changes in example.html
InBuiltMarkdownProcessingAC

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3640)


bhuvana-guna:
This is also available as a separate npm - https://www.npmjs.com/package/js-markdown-formatter . Users who want just the basic markdown formatting options can make use of this.

bhuvana-guna:
@dclaux @matthidinger Can you please review?

dclaux:
Thanks very much for your submission!

I am going to review your code, can't guarantee that I'll be able to finish until next week as I'm leaving in the afternoon.

My number 1 feedback, which applies to basically all individual comments I'll put in the review is this: the entire AC library is written in TypeScript for it can be very strongly typed, allowing errors to be detected at compile time and allowing developers to benefit from "IntelliSense" when coding against it. A future version of the library (currently in this PR) is actually compiled with the strictNullCheck and noImplicitAny compiler directives to go even further when it comes to strong typing.

I think this Markdown formatter needs to be strongly typed as well in order to meet the bar. Currently it isn't, which incidentally makes it harder for me to review.

Beyond that, it would really be great if you could add comments to help understand the logic you used. I know, I'm guilty of not commenting myself, and I feel a bit bad asking for something I haven't really done so far... but this just shows that I also need to do better in that area, which I will.

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -0,0 +1,182 @@
...
+	render() {
+        
+		for (var i = 0; i <= this.styleTypes.length - 1; i++) {		
+			this.text = this.parseText(this.text, this.styleTypes[i], this.styles[i], this.patterns[i], this.patternTypes[i]);
+		}
+
+		return this.text;
+	}
+
+	parseText = (text, type, styles, pattern, patternType) => {
```
dclaux: - Should be `private`
- All parameters must be typed
- Return value must be typed
- Why this inconsistent method declaration syntax? Should be
```typescript
private parseText(text, ...): string { ... }
```
- Given that this method is called like this:
```typescript
this.parseText(this.text, this.styleTypes[i], this.styles[i], this.patterns[i], this.patternTypes[i]);
```
why does it need to be passed all these parameters? They can all go and be replaced with a simple `index: number`.

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -0,0 +1,182 @@
...
+				pattern = regex;
+			}
+            this.patterns[i] = new RegExp(pattern, 'gim');
+			this.styles[i] = MarkdownFormatter.MD_FORMATTER_CONFIG[i].styles;
+			this.styleTypes[i] = MarkdownFormatter.MD_FORMATTER_CONFIG[i].type;
+			this.patternTypes[i] = patternType;
+		}
+	}
+
+	render() {
```
dclaux: - Return value must be typed

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -0,0 +1,182 @@
...
+	parseText = (text, type, styles, pattern, patternType) => {
+        let first=true;
+        let parsed;
+		while ((parsed = pattern.exec(text)) !== null) {
+			if(parsed[1] === undefined){
+				continue;
+            } 
+
+            //For lists.
+            if(patternType === "start-end"){
```
dclaux: The "start-end" pattern is supposedly (or at least seems to be) generic, e.g. able to handle anything where the start and end markers are different. But in reality, is it completely tied to rendering lists, and the generation of `<li>` elements is hard coded. Maybe this pattern type should be renamed into "list"?

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -0,0 +1,182 @@
...
+            }
+
+		}
+		return text;
+    }
+    
+}
+
+//formats the markdown in the text passed.
+export function formatText(text: string): string {
```
dclaux: This should be a static method of the MarkdownFormatter class

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -0,0 +1,182 @@
...
+    let mdformatter = new MarkdownFormatter(text);
+    return mdformatter.render();
+}
+
+//adds custom regex to the format config.
+export function setCustomMarkdownRegex(regexArray: any[]): void {
```
dclaux: This isn't necessary if the model I suggest is used

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -0,0 +1,182 @@
...
+    text:string;
+    patterns:any[] = [];
+    styles:any[] = [];
+    patternTypes:any[] = [];
+    styleTypes:any[] = [];
+
+    static regexArray:any[];
+    static SPECIAL_CHAR_REGEX = new RegExp("[^|a-z\\\\s\\d]", 'gi');
+    
+	static MD_FORMATTER_CONFIG=[
```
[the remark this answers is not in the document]
bhuvana-guna: I have changed it according to this. But had to make the configurations as an array rather than JSON as we would have to overwrite the rule if the user passes the same type, say different pattern to format italic. Also, if it was JSON I would have to call the initialize logic mentioning every JSON key. It made sense to use an array. Please review the new code and let me know if it is fine.

bhuvana-guna:
@dclaux Thanks so much for providing a detailed review. I have made the changes. Also compiled with the strictNullCheck and noImplicitAny.  Please review.

source/nodejs/adaptivecards/src/markdown-formatter.ts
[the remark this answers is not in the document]
dclaux: Thanks for making the changes.

I still believe that `configuration` should be declared as I suggested. Looking at your update, I can see that:
- There is still a `type` property, which is of type string
- That `type` property is used solely to allow the developer to replace the built-in rules, if the custom rules they pass have a type name that matches
- The only way for the developer to know what type names to use is either through documentation or by looking up the code

At the very least, `type` should be strongly typed, e.g.:
```typescript
export type BuiltInRuleTypeName = "orderedList" | "unorderedList" | ...;

export interface IFormatterRule {
    type?: BuiltInRuleTupeName;
    ...
}
```

But I do believe that's still less than optimal. With the approach I suggest, a developer can replace the rule for say `unorderedList` by doing this:
```typescript
MarkdownFormatter.configuration.unorderedList = { ... };
```

That's a super easy, strongly typed and discoverable way to replace a built-in rule IMO.

And adding custom rules is as easy as setting the `customRules` field:
```typescript
MarkdownFormatter.configuration.customRules = [
    {
        patternType: "start-end",
        ...
    }
]
```

Alternatively, `IFormatterConfiguration` could be declared like this:
```typescript
export interface IFormatterConfiguration {
    orderedList: IFromatterRule;
    unorderedList: IFormatterRule;
    bold: IFormatterRule;
    italic: IFormatterRule;
    hyperlink: IFormatterRule;
    [ruleName: string]: IFormatterStyleRule
}
```

which would allow developers to add/remove custom rules as follows, with the guarantee they can't conflict with a built-in rule:
```typescript
// Add a custom rule
MarkdownFormatter.configuration.myCustomRule = { ... };
// Remove a custom rule
delete MarkdownFormatter.configuration.myCustomRule;

```

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -0,0 +1,182 @@
...
+                html:"<a href='{1}'>{0}</a>"
+            },
+			pattern: ['[]()'],
+			patternType: 'asymmetric',
+			groups: 2,
+		},
+    ];
+
+    //combines default config and user config.
+    static setCustomMarkdownRegex(regexArray: any[]){
```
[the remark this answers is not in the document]
dclaux: I still believe this isn't optimal/necessary.

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
+
+    private initailize(configuration: IFormatterConfiguration): void {
+
+        let patternArray = configuration.patterns,
+            patternType = configuration.patternType,
+            pattern = patternArray[0],
+            groups = configuration.groups;
+
+        if (patternType === "list") {
+            pattern = this.encodeSpecialCharacters(patternArray[0]) + '[\\s]+((.*?)[' + this.encodeSpecialCharacters(patternArray[1]) + "](?=" + this.encodeSpecialCharacters(patternArray[1]) + "[\\s]+)|(.*?)$)";
```
dclaux: That's better, but now `this.encodeSpecialCharacter(patternArray[0])` and `this.encodeSpecialCharacter(patternArray[1])` should really be put in variables given there are called 4 and 2 times respectively.

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
+                this.initailize(configurations[j]);
+                continue;
+            } else {
+                this.configuration.push(configurations[j]);
+            }
+        }
+    }
+
+    /**
+     * To add '\' before all the special characters in the regex.
```
dclaux: Can you add an example input and output? I'm still unclear what this does.

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
+     * 
+     * @param pattern 
+     */
+    private encodeSpecialCharacters(pattern: string) {
```
dclaux: The return type should be explicitly specified. Does this not raise an error with `noImplicitAny`?

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
-            
-            if(patternType === "custom"){                
-                pattern = patternArray[0];
+
+    /**
+     * To add custom markdown rules. This will update the default markdown rule is the same type is passed.
+     * 
+     * @param configurations 
+     */
+    public setCustomRules(configurations: IFormatterConfiguration[]) {
```
dclaux: I still believe this isn't necessary or ideal. Please refer to my main comment.

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
-                let p = MarkdownFormatter.MD_FORMATTER_CONFIG[i].pattern[0];
-                for (var j = 0; j < p.length; j++) {
-                    regexForm = regexForm + "\\" + p[j];
+            pattern = regex;
+        }
+        configuration.regexp = new RegExp(pattern, 'gim');
+    }
+
+    formatText(text: string): string {
+
```
dclaux: Unnecessary empty line

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
-                pattern = patternArray[0].replace(MarkdownFormatter.SPECIAL_CHAR_REGEX, "\\$&") + '(.*?)' + patternArray[0].replace(MarkdownFormatter.SPECIAL_CHAR_REGEX, "\\$&");    
-            }else if(patternType == "asymmetric"){                
-                pattern = patternArray[0].replace(MarkdownFormatter.SPECIAL_CHAR_REGEX, "\\$&");    
-                let regexForm = '';
-                let p = MarkdownFormatter.MD_FORMATTER_CONFIG[i].pattern[0];
-                for (var j = 0; j < p.length; j++) {
-                    regexForm = regexForm + "\\" + p[j];
+            pattern = regex;
+        }
+        configuration.regexp = new RegExp(pattern, 'gim');
```
dclaux: Blank line after closing }

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
-                pattern = patternArray[0].replace(MarkdownFormatter.SPECIAL_CHAR_REGEX, "\\$&") + '[\\s]+((.*?)[' + patternArray[1].replace(MarkdownFormatter.SPECIAL_CHAR_REGEX, "\\$&")+ "](?="+patternArray[1].replace(MarkdownFormatter.SPECIAL_CHAR_REGEX, '\\$&')+"[\\s]+)|(.*?)$)";
-            }else if(patternType == "symmetric"){
-                pattern = patternArray[0].replace(MarkdownFormatter.SPECIAL_CHAR_REGEX, "\\$&") + '(.*?)' + patternArray[0].replace(MarkdownFormatter.SPECIAL_CHAR_REGEX, "\\$&");    
-            }else if(patternType == "asymmetric"){                
-                pattern = patternArray[0].replace(MarkdownFormatter.SPECIAL_CHAR_REGEX, "\\$&");    
-                let regexForm = '';
-                let p = MarkdownFormatter.MD_FORMATTER_CONFIG[i].pattern[0];
-                for (var j = 0; j < p.length; j++) {
-                    regexForm = regexForm + "\\" + p[j];
+            pattern = regex;
```
dclaux: Blank line after closing }

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
+        if (patternType === "list") {
+            pattern = this.encodeSpecialCharacters(patternArray[0]) + '[\\s]+((.*?)[' + this.encodeSpecialCharacters(patternArray[1]) + "](?=" + this.encodeSpecialCharacters(patternArray[1]) + "[\\s]+)|(.*?)$)";
+        } else if (patternType == "symmetric") {
+            pattern = this.encodeSpecialCharacters(patternArray[0]) + '(.*?)' + this.encodeSpecialCharacters(patternArray[0]);
+        } else if (patternType == "asymmetric") {
+            let regexForm = this.encodeSpecialCharacters(patternArray[0]);
+
+            let part = regexForm.length / groups;
+            let regex = "";
+            for (var j = 0; j < groups; j++) {
```
dclaux: Please use `let` instead of `var` everywhere. I have [a pending PR](https://github.com/microsoft/AdaptiveCards/pull/3508) which replaces all `var`s into `let`s.

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
+    private initailize(configuration: IFormatterConfiguration): void {
+
+        let patternArray = configuration.patterns,
+            patternType = configuration.patternType,
+            pattern = patternArray[0],
+            groups = configuration.groups;
+
+        if (patternType === "list") {
+            pattern = this.encodeSpecialCharacters(patternArray[0]) + '[\\s]+((.*?)[' + this.encodeSpecialCharacters(patternArray[1]) + "](?=" + this.encodeSpecialCharacters(patternArray[1]) + "[\\s]+)|(.*?)$)";
+        } else if (patternType == "symmetric") {
```
dclaux: Please move `else` to a new line

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
+     * To add '\' before all the special characters in the regex.
+     * 
+     * @param pattern 
+     */
+    private encodeSpecialCharacters(pattern: string) {
+        return pattern.replace(MarkdownFormatter.SPECIAL_CHAR_REGEX, "\\$&")
+    }
+
+    private initailize(configuration: IFormatterConfiguration): void {
+
```
dclaux: Unnecessary empty line

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
+        let patternArray = configuration.patterns,
+            patternType = configuration.patternType,
+            pattern = patternArray[0],
+            groups = configuration.groups;
+
+        if (patternType === "list") {
+            pattern = this.encodeSpecialCharacters(patternArray[0]) + '[\\s]+((.*?)[' + this.encodeSpecialCharacters(patternArray[1]) + "](?=" + this.encodeSpecialCharacters(patternArray[1]) + "[\\s]+)|(.*?)$)";
+        } else if (patternType == "symmetric") {
+            pattern = this.encodeSpecialCharacters(patternArray[0]) + '(.*?)' + this.encodeSpecialCharacters(patternArray[0]);
+        } else if (patternType == "asymmetric") {
```
dclaux: Please move `else` to a new line

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
+            let regexForm = this.encodeSpecialCharacters(patternArray[0]);
+
```
dclaux: Unnecessary empty line

source/nodejs/adaptivecards/src/markdown-formatter.ts
dclaux: Add empty line before `for`

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
+    formatText(text: string): string {
+
+        for (let i = 0; i < this.configuration.length; i++) {
+            let first = true;
+            let parsed,
+                pattern = this.configuration[i].regexp,
+                //pattern = this.patterns[i],
+                patternType = this.configuration[i].patternType,
+                styles = this.configuration[i].styles;
+            while (pattern != null && (parsed = pattern.exec(text)) !== null) {
```
dclaux: Consistency: use `!==` everywhere

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
-                    text = text.replace(text.substring(indexOL, indexOL + styles.end.length), "");
-                    text = text.replace(parsed[0], "<li>" + parsed[1] + "</li>" + styles.end);
+
+                //For lists.
+                if (patternType === "list" && styles.start && styles.end) {
+                    //If it is the first element of the list then add the list start tag(<ol> / <ul>) 
+                    if (first) {
+                        text = text.replace(parsed[0], styles.start + "<li>" + parsed[1] + "</li>" + styles.end);
+                        first = false;
+                    } else {
```
dclaux: Please move `else` to its own line

source/nodejs/adaptivecards/src/markdown-formatter.ts
```diff
@@ -1,182 +1,186 @@
...
-            else if(patternType == "asymmetric"){ 
-                styles.html = styles.html.replace("{0}",parsed[1]);
-                styles.html = styles.html.replace("{1}",parsed[2]);
-                text = text.replace(parsed[0], styles.html);
-            }
-
-		}
-		return text;
+        }
+        return text;
```
dclaux: Empty line after closing }

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5,7 +5,7 @@ import * as Shared from "./shared";
 import * as Utils from "./utils";
 import * as HostConfig from "./host-config";
 import * as TextFormatters from "./text-formatters";
-import * as MarkdownFormatter from "./markdown-formatter";
+import {MarkdownFormatter, TextFormatter} from "./markdown-formatter";
```
dclaux: Space after { and before }

source/nodejs/adaptivecards/src/markdown-formatter.ts
bhuvana-guna: It didn't show an error.

source/nodejs/adaptivecards/src/markdown-formatter.ts
bhuvana-guna: When a user adds his own custom regex we need to call the initialize method to update the patterns list. I tried to make it a part of constructor but the constructor in the card-elements gets called even before the user sets custom rules. How else to do?

source/nodejs/adaptivecards/src/markdown-formatter.ts
bhuvana-guna: When we try to match '**' in the text we need the tell regex to match the exact '**' instead of the actual usage of '*' which is zero or more of the previous expression. For that we need to add '\'before '*'. Similarly for all the special characters.

source/nodejs/adaptivecards/src/markdown-formatter.ts
dclaux: How about this: design configuration as I suggest, keep the `MarkdownFormatter._configuration` field private, allow the developer to pass a custom configuration to the constructor and remove `setCustomRules`. This way you can initialize everything just once at construction time, and if the developer wants to use multiple different configurations they can simply instantiate multiple instances of `MarkdownFormatter`.

source/nodejs/adaptivecards/src/markdown-formatter.ts
dclaux: So you're escaping special characters in order for the rules to not catch them?

bhuvana-guna:
@dclaux Can you please check now.

dclaux:
> @dclaux Can you please check now.

@bhuvana-guna what did you change? I'm not seeing anything new.

source/nodejs/adaptivecards/src/markdown-formatter.ts
bhuvana-guna: Then I would have to expose MarkdownFormatter class as a static variable right? I did the following changes to make it work. 

I added this is card-elements.ts
```
static MarkdownFormatter = MarkdownFormatter;
```
This is how my a user can pass the custom rules.
```
AdaptiveCards.AdaptiveCard.textFormatter = new AdaptiveCards.AdaptiveCard.MarkdownFormatter(customRules);
```

Am I doing it right?

bhuvana-guna:
> > @dclaux Can you please check now.
> 
> @bhuvana-guna what did you change? I'm not seeing anything new.

Sorry I missed your reply to the comments. I have replied to that. Let me know if that will work. I will push those changes.

source/nodejs/adaptivecards/src/markdown-formatter.ts
dclaux: Here is what I am suggesting:
```typescript
export type FormatterPatternType = "start-end" | "symmetric" | "asymmetric";

export interface IFormatterRule {
    styles: {
        start?: string;
        end?: string;
        html?: string;
    };
    patterns: string[];
    patternType: FormatterPatternType;
    groups: number;
}

export interface IFormatterConfiguration {
    orderedList: IFromatterRule;
    unorderedList: IFormatterRule;
    bold: IFormatterRule;
    italic: IFormatterRule;
    hyperlink: IFormatterRule;
    [name: string]: IFormatterRule;
}

export class MarkdownFormatter {
    static getDefaultConfiguration(): IFormatterConfiguration {
        return {
            orderedList: {
                styles:{
                    start: "<ol>",
                    end: "</ol>"
                },
                patterns: ["\\d.", '\\r|\\n'],
                patternType: "start-end",
                groups: 1
            },
            unorderedList: {
                ...
            },
            ...
        }
    }

    private _configuration: IFormatterConfiguration;

    constructor(configuration?: IFormatterConfiguration) {
        super();

        this._configuration = configuration ? configuration : MarkdownFormatter.getDefaultConfiguration();

        // Initialize configuration
    }
}
```

No need to have a `static MarkdownFormatter`. Developers can write this:
```typescript
// This instantiates a MarkdownFormatter with the default configuration
AdaptiveCards.AdaptiveCard.textFormatter = new AdaptiveCards.AdaptiveCard.MarkdownFormatter();

// The below instantiates a MarkdownFormatter with a custom configuration

// Get a copy of the default configuration
let myConfiguration = MarkdownFormatter.getDefaultCOnfiguration();

// Customize the orderedList rule
myConfiguration.orderedList = {
    ...
}

// Add a custom rule
myConfiguration["myCustomRule"] = {
    patternType: "start-end",
    ...
};

// Alternatively, the custom configuration can be created entirely from scratch
myConfiguration = {
    orderedList: {
        ...
    },
    unorderedList: {
        ...
    },
    ...,
    myCustomRule: {
        ...
    }
};

// Pass the custom configuration to the MarkdownFormatter constructor
AdaptiveCards.AdaptiveCard.textFormatter = new AdaptiveCards.AdaptiveCard.MarkdownFormatter(myConfiguration);

```
The point for all this really is that everything is strongly typed, and thus the developer gets "IntelliSense" hints as they code.

Does that make sense?

bhuvana-guna: @dclaux  Yes. This is brilliant. Thanks so much for the detailed explanation. I have made these changes. Can you please review it.

dclaux:
@matthidinger it would be great if you could take a look at this. It brings a simple built-in Markdown implementation to the JS renderer, which is something we had discussed doing anyway.

bhuvana-guna:
@matthidinger Can you please review.

bhuvana-guna:
@matthidinger Can you please check this pull request when you get time. It's been a while since @dclaux reviewed it.

matthidinger:
Hey folks, apologies for the delay here! @bhuvana-guna, thanks so much for this PR. I've got two questions (also for your @dclaux):

1. What is our guidance on when to use `onProcessMarkdown` vs the `textFormatter`. At first glance they seem to do similar things?
2. This PR removed the check for `window["markdownIt"]`, but if people were relying on the old behavior will it continue to work as expected? Presumably they will automatically get the new built-in markdown formatter? And if for some reason they do want to use markdown-it, then they would read the README and follow the guidance there. I just want to make sure there are no breaking changes in this for existing users based on our previous behavior, and if so, we should call that our explicitly in the upcoming 2.0 release.

Thanks again!

bhuvana-guna:
@matthidinger Thanks for your reply. Please find my answers below.

1. When to use,
**onProcessMarkdown** - This should be used when you are integrating with a third-party library. Just call that library's render method in onProcessMarkdown. eg: mardownIt. How to do it has been added to the readme file.

**textFormatter** - This can be used when you are going to build your own markdown formatter, specifically for your adaptive card, inside your project.

2. People who want to continue to use markdownIt should specify it explicitly in onProcessMarkdown. But once added nothing will change.

ACMarkdown

@dclaux Please correct me if I'm wrong.

matthidinger:
@dclaux I know this is a late ask but do you think we should add this to the 2.0.0-rc before we ship it in the coming weeks? Since it's a major rev it seems like a good time to add such a "big" feature change.

paulcam206:
@dclaux / @matthidinger - are we still pursuing this?

jonmill:
Closing due to 2 years of merge conflicts...if this is still a relevant PR, please fix up and re-open. Thanks!

Issue #2293, opened 2019-01-11
minHeight property on containers


andrewleader:
<table>
<tr>
<th>Release</th>
<th>Renderer status</th>
<th>Tasks status</th>
</tr>
<tr>
<td>1.2
<td>🔄 .NET (#2618)<br/>✔️ Android (#2619)<br/>🔄 iOS (#2620)<br/>✔️ TS (#2621)<br/>✔️ UWP (#2622)</td>
<td>✔️ Shared (#2616)<br/>✔️ Designer (#2617)</td>
</tr>
</table>

# Solves requests

* Borderless header/footer images #2290
* Full-width images with fixed height and correct fill/crop behavior #2291

# Summary
This proposal is an alternative way of solving the above two requests, written up so we can consider both options. Containers can already almost achieve the two above requests except for one thing... you can't set a height on them.

# Schema
**New** property on the following elements...

* `Container`
* `ColumnSet`
* `Column`
* `ActionSet`

Property|Type|Required|Description
---|---|---|---
minHeight | `null` (default) or `px` | false | Specifies the minimum height of the container.

# Example

To achieve the following, where the image is full width but has a fixed height...

![image](https://user-images.githubusercontent.com/13246069/51007006-9a9ef900-14fa-11e9-977c-f20586a47184.png)

I would add a container with a background image, and set the minimum height of the container to 100px (so it's my exact height).

```json
        {
            "type": "Container",
            "backgroundImage": {
              "url": "https://picsum.photos/300/400?image=882"
            },
            "minHeight": "100px"
        }
```

Now, what if I want the image to go to the edge of the card? I don't want that padding around the edges?

![image](https://user-images.githubusercontent.com/13246069/51008026-cc19c380-14fe-11e9-9ec4-52579db4c34d.png)

I would simply set bleed=true, and that'll remove the padding.
```json
        {
            "type": "Container",
            "backgroundImage": {
              "url": "https://picsum.photos/300/400?image=882"
            },
            "minHeight": "100px",
            "bleed": true
        }
```

# Host Config
None.

# Down-level impact
**High**. Image won't appear at all (since height on container is ignored down-level, and therefore the container won't have any height).

# Host burden
None.

# Auto-generated task status

- [x] Shared
- [x] Designer
- [ ] .NET
- [x] Android
- [ ] iOS
- [x] TS
- [x] UWP

andrewleader:
Proposal approved!

mdtauk:
[...]

PR #3641, opened 2019-11-25
Implemented the minheight property for the container

IbrahimSulai:
Implemented and handled the minHeight property for the container

Added the Container.minHeight.json in the payload for validation.

This feature is related to:
https://github.com/microsoft/adaptivecards/issues/2293

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3641)

### DIFF
--- a/source/community/reactnative/src/components/containers/container.js
+++ b/source/community/reactnative/src/components/containers/container.js
@@ -52,18 +52,8 @@ export class Container extends React.Component {
 
 	internalRenderer = () => {
 		const payload = this.payload;
-		let containerStyle;
-		let minHeight = payload.minHeight;
-		if (minHeight) {
-			if (typeof (minHeight) == 'string') {
-				const lastIndex = minHeight.lastIndexOf('px');
-				minHeight = lastIndex != -1 ? parseInt(minHeight.substring(0, lastIndex)) : minHeight
-			}
-			//If it is number, then directly we can assign minHeight
-			containerStyle = [styles.container, { minHeight }];
-		} else {
-			containerStyle = [styles.container];
-		}
+		const minHeight = Utils.convertStringToNumber(payload.minHeight);
+		const containerStyle = [styles.container, minHeight && { minHeight }];
 
 		var containerContent = (
 			<ContainerWrapper json={this.payload} style={containerStyle} containerStyle={this.props.containerStyle}>
--- a/source/community/reactnative/src/components/containers/container.js
+++ b/source/community/reactnative/src/components/containers/container.js
@@ -53,7 +53,7 @@ export class Container extends React.Component {
 	internalRenderer = () => {
 		const payload = this.payload;
 		const minHeight = Utils.convertStringToNumber(payload.minHeight);
-		const containerStyle = [styles.container, minHeight && { minHeight }];
+		const containerStyle = typeof minHeight === "number" ? [styles.container, { minHeight }] : [styles.container];
 
 		var containerContent = (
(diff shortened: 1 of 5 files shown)


source/community/reactnative/src/components/containers/container.js
```diff
@@ -51,9 +52,21 @@ export class Container extends React.Component {
 
 	internalRenderer = () => {
 		const payload = this.payload;
+		let containerStyle;
+		let minHeight = payload.minHeight;
+		if (minHeight) {
+			if (typeof (minHeight) == 'string') {
+				const lastIndex = minHeight.lastIndexOf('px');
+				minHeight = lastIndex != -1 ? parseInt(minHeight.substring(0, lastIndex)) : minHeight
```
dclaux: The value "ABCpx" will evaluate to minHeight == NaN, which I think should be explicitly handled as an invalid minHeight value.

IbrahimSulai: Thanks, @dclaux. Will evaluate that pattern also.

IbrahimSulai: @dclaux - Created a method to handle the above pattern. Please provide your comment on this.

source/community/reactnative/src/components/containers/container.js
```diff
@@ -52,18 +52,8 @@ export class Container extends React.Component {
...
-				const lastIndex = minHeight.lastIndexOf('px');
-				minHeight = lastIndex != -1 ? parseInt(minHeight.substring(0, lastIndex)) : minHeight
-			}
-			//If it is number, then directly we can assign minHeight
-			containerStyle = [styles.container, { minHeight }];
-		} else {
-			containerStyle = [styles.container];
-		}
+		const minHeight = Utils.convertStringToNumber(payload.minHeight);
+		const containerStyle = [styles.container, minHeight && { minHeight }];
```
dclaux: Won't this generate an array with 2 items, with the second one being `false` if `minHeight == 0` or is `null`? And if so is that ok?

IbrahimSulai: Since it is a style container. I hope, It's ok to have the condition like this.

IbrahimSulai: I have tested and verified for all the below patterns.
minHeight: "100"
minHeight: "100px"
minHeight: "100a5"
minHeight: 100  
minHeight: "ABC"
minHeight: null   
minHeight: ''ABC100"

Please let me know if I am missing any other patterns. Thanks!

IbrahimSulai: @dclaux - We need to consider only the string with "px" or also we can have a pattern like below.

eg: "100px" / "100" / "100dp"

the value for the above strings are the same - "minHeight: 100"

dclaux: Only "100px" is valid.

dclaux: BTW when you say you "hope it's ok", I'd rather you be sure :) Maybe rewriting the code like this would be better?

```typescript
const containerStyle = typeof minHeight === "number" ? [styles.container, { minHeight }] : [styles.container];
```

IbrahimSulai: Got it. Have updated the container style assignment as above.

Issue #3081, opened 2019-06-18
Input validation & Labels Support


productboard-ac:
# Summary

Authors need input validation, both just simple "isRequired" and more advanced than just saying that an input is required.

![ValidateOnSubmit](https://user-images.githubusercontent.com/13246069/60541698-0c484a80-9cc7-11e9-8471-32cb0d49a855.gif)

# Dev Task List Breakdown
## Input Labels Support

- [x] JS implementation | 10 days
- [x] C++ Object Model | 4 days
- [x] UWP Implementation | 4 days
- [x] Android Implementation | 4 days
- [x] iOS Implementation | 4 days
- [x] C# Object Model | 4 days
- [x] .NET implementation | 4 days
- [ ] HTML implementation | 4 days
- [ ] Bug Bash | 2 days
- [ ] Update Documentation | 2 days
- [ ] Merge from topic branch into master] | 2 days

## Input Validation Support 
- [x] JS implementation | 5 days
- [x] C++ Object Model | 6 days
- [x] UWP Implementation | 5  days
- [x] Android Implementation | 5  days
- [x] iOS Implementation | 5 days
- [x] C# Object Model | 5 days
- [x] .NET implementation | 5 days
- [ ] HTML implementation | 6 days
- [ ] Bug Bash | 7 days
- [ ] Update Documentation | 6 days

# Feature Spec including prioritized feature list & proposed schema changes
In code review at 
<https://github.com/microsoft/AdaptiveCards/pull/4190>

productboard-ac:
Link to feature: https://adaptivecards.productboard.com/feature-board/planning/features/1734121

dclaux:
My suggestions:
- `inputNecessityIndicators` (why is this an enum?) --> `showRequiredInputIndicators: true|false`
- `showInputValidationErrors`: this needs to be introduced as well as a way for a host to custom render these error messages. We've had a very long debate around this.
- `displayInputSuccessIndicators`: don't implement this feature. There is at best anecdotal evidence that we need it
- `errorMessage` --> `validationErrorMessage`

andrewleader:
# 7/12/2019

* More spec details needed on behavior of ShowCard and which inputs it validates on
* `showInputValidationErrors` for renderers (via code) makes sense

Proposal approved

shalinijoshi19:
Assigning to you @RebeccaAnne since you are driving this feature. thanks

KoradaRajesh:
{
[...]

PR #3663, opened 2019-12-05
Input Validation Spec Documents

RebeccaAnne:
## Related Issue
Spec updates for #3081

## Description
Spec and documentation for input validation. Includes:
- specs\DesignDiscussions\InputValidation.md file discussing the design of the input validation feature. (Start here)
- specs\DesignDiscussions\InputLabels.md file updated based on discussions coming out of review of the InputValidation spec.
- Updates to the schemas\src files to reflect changes discussed in InputValidation.md
- Generated schema and spec files in schemas\1.3.0\adaptive-card.json and specs\elements respectively.
- Manual updates to the Rendering sections of the specs\elements files to reflect spec'd behavior for rendering input validation.
- Scenario files using input validation in samples\v1.3\Scenarios. (I will make sure to run the UWP test app on these files before checkin).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3663)


specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Schema Additions to Adaptive Card for Validation
+
+In order to allow the card author to visually indicate whether inputs in a card are required or optional, we will add a new property, `showRequiredInputHints`. When this is true, visual indicators will be added to required inputs inputs to signal to the user that they are required. This property will default to true.
+
+```json
+{
+	"type": "AdaptiveCard",
+	"version": "1.3",
+	"showRequiredInputHints":true,
```
RebeccaAnne: Is this needed? 

Discussion: 
The host could do this manually by adding their own *.
Semantic information is good.
Do we want host styles for how to style required and options inputs?

We don't need this. Card author can do it today, or we may want to add host features for this in the future.

RebeccaAnne: On labels: We need to balance between host control and card author control. Current spec calls for card author control.

---
In reply to: [355022762](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r355022762) [](ancestors = 355022762)

RebeccaAnne: If there are not labels (the card author is using placeholders), we may still need hints.

---
In reply to: [355023402](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r355023402) [](ancestors = 355023402,355022762)

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
...
+
+The error text and invalid indication will be rendered in the host's `attention` color.
+
+![img](assets/InputValidation/InputValidationExample.png)
+
+## Input Validation Behavior
+
+### When to Validate
+
+Issue [#3081](https://github.com/microsoft/AdaptiveCards/issues/3081) covers research and best practices behind when to validate. The conclusion reached there is as follows:
```
RebeccaAnne: Lots of discussion on this. Outlook.com has behavior that allows clicking the button and *then* validates on navigate.

Choice sets and other inputs cause problems with the model spec'd here.

Matt feels strongly we should scroll the first bad one into view.

I want to follow up with WinUI to see if they've thought about this.

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
...
+
+![img](assets/InputValidation/Required.PNG)
+
+The error text and invalid indication will be rendered in the host's `attention` color.
+
+![img](assets/InputValidation/InputValidationExample.png)
+
+## Input Validation Behavior
+
+### When to Validate
```
RebeccaAnne: Discussion today:
- David not a fan of focus lost behavior
- Less aggressive if we only validate if they typed something
- WinUI choosing focus lost behavior is a compelling argument for us to do the same (do they validate if no typing? I think so, but should confirm)
- Our hosts are varied, likely want the behavior to conform with the rest of their app
- Proposal: Have this be configurable via host config, enum of choices would be: focus lost, focus lost if typing, only on submit.

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
...
+	"inputs": "All",
+}
+```
+
+The second option would be "None". This would be useful for, for example, a "Cancel" button where validation is not required, and would be the default for non-Submit (or Invoke action being discussed for Universal Actions) action types (see further discussion of validation for different action types below).
+```json
+{
+	"type": "Action.Submit",
+	"title": "Cancel",
+	"inputs": "None",
```
RebeccaAnne: >"inputs": "None", [](start = 1, length = 17)

Discussion: What is the scenario for "None"? Should we have an explicit Action.Cancel instead?
- Fire company has an Action.Cancel in their stuff, but what are the semantics?
- Non-submit action types default to None.

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
...
+
+Because by default Toggles and ShowCards don't necessarily need to validate inputs, these actions will default to `"inputs": "None"`. We will, however, support the ability to set a different value to enforce validation of some or all of the inputs before allowing the user to take one of these actions.
+
+> Question: Is it confusing to have an `inputs` property on a type that doesn't actually return those inputs (only validates them)? Would we want to add support for returning inputs on other action types? Is this an indication we need a different name?
+
+##### OpenUrl
```
RebeccaAnne: Discussion:
- This could be useful if we someday have a scenario where the URL is configured by the values of the inputs.
- If we had that, the inputs would be explicit there, so this would be redundant.
- We could have the behavior be that if there are explicit inputs they are validated in addition to anything in input.

Ultimately we're okay with having inputs on the base class.

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
...
+Actions that depend on validation of the inputs should be greyed out until all related inputs have been validated. Because we are validating on focus lost, it would be possible for a user to navigate through a form using tab, resulting in all error messages and validation being shown. Clicking on a non-greyed out submit button at this time would not give any additional feedback, just silently fail to submit. To avoid this, we should grey out the submit button until such time as all inputs pass validation.
+
+## Custom Rendering Error Messages
+Comments in [#3081](https://github.com/microsoft/AdaptiveCards/issues/3081) from dclaux:
+> showInputValidationErrors: this needs to be introduced as well as a way for a host to custom render these error messages. We've had a very long debate around this.
+
+And andrewleader:
+>showInputValidationErrors for renderers (via code) makes sense
+
+TODO: Follow up on specifying this behavior.
```
RebeccaAnne: Discussion: This is about custom rendering the errors, for example in a pop up bubble rather then underneath. David feels this is important, but may be P2 for other renderers. Because this is a renderer feature, not a card schema thing, we'd be okay with an inconsistency between renderers if this is implemented for Javascript but not others.

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
...
+}
+```
+
+> Questions:
+>
+> - Do we want to support a "CurrentContainer" or "CurrentContainerAndParent" type option? I think no. Allowing authors to specify containers allows for this behavior, and introducing an option with possibly non-obvious logic in the edge cases may be more confusing than helpful to card authors.
+>
+> - Do we want a typed "Form" container? The design described above allows card authors to use containers as "forms" by specifying them as the inputs for a submit action. Is it helpful to have an explicit "FormContainer" type to use for this? There doesn't seem to be an explicit use for it today. One scenario may be if we have "Form" type properties in the future that would be useful to put on a typed element.
+> - I called the property `"inputs"` rather than something validation specific (`validationInputs`?) because in the submit case it indicates not only the inputs that will be validated, but also those that will be submitted. Are there scenarios in which a submit button would need to validate a different set of inputs than those that it wants to send? I would think **no**, as sending un-validated inputs seems like it could cause problems.
```
RebeccaAnne: >I called the property `"inputs"` [](start = 4, length = 32)

Discussion: On naming, Matt was concerned calling it `inputs` might imply to card authors that input types should go in there. Maybe include "Id" in the name? "AssociatedInputIds"?

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
...
+	"title": "Cancel",
+	"inputs": ["input1Id", "input2Id"] 
+}
+```
+In order to allow authors to group their inputs, we could also accept container ids. This allows the card author to functionally define a "form" and indicate that the submit button relates to that form:
+```json
+{
+	"type": "Action.Submit",
+	"title": "Cancel",
+	"inputs": ["input1Id", "input2Id", "formContainerId"]
```
RebeccaAnne: Discussion: Can this be a card Id? We don't have card ids in shared model today, but we do in JavaScript. We probably should have them in all platforms, at which point yes, this could be a card id.

specs/DesignDiscussions/InputValidation.md
RebeccaAnne: Further discussion of labels today: 
- All agreed that P0 base behavior is a plain text label on the input, required inputs marked by a *.
- Next most important thing is allowing host control. Two dimensions: styling, and required indicator string and position
- David feels strongly that all styling should go through CSS (or equivalent) and should not be in host config.
- Matt and i feel some base host config control may be useful. Discussion of underlying use and usefulness of host config in general
- All agreed that native styling should be possible for required vs optional.
- On string (*) configuration, we could use the common expression language. David is concerned about taking a dependency on it for size reasons.
- If host can configure the string, they need to be able to localize it.
- David thinks that allowing hosts to change the string (without loc) should be high priority.

---
In reply to: [355029970](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r355029970) [](ancestors = 355029970,355023402,355022762)

RebeccaAnne:
Apologies for the noise in the commit history on this, I flew too close to the sun with my merges and rebases. Now the changes from master show up in code flow, but don't show up in a rebase -i, so I can't figure out how to squash them.

specs/DesignDiscussions/InputValidation.md
RebeccaAnne: Spec is updated with plan to have three possible values in host config. This is POR for now, pending further data from the input survey.

---
In reply to: [367152372](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367152372) [](ancestors = 367152372)

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
...
+``` json
+{
+    "type": "Input.Text",
+    "id": "username",
+    "isRequired": true,
+    "regex": "^\\w{5,50}$",
+    "errorMessage": "Username must be between 5 and 50 characters long."
+}
+```
+The main additions to the schema for input validation are the `"isRequired"`, `"regex"`, and `"errorMessage"` properties on inputs. These properties allow a card author to specify whether or not an input is required, what regular expression defines valid input, and what error message should be shown in case of invalid user input.
```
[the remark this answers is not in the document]
RebeccaAnne: There are further additions to the schema that are described later in the document (specifically: associated inputs on actions and id on adaptive card). I'll update the wording here to indicate that this is the full list of schema changes for the inputs themselves.

---
In reply to: [367012948](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367012948) [](ancestors = 367012948)

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
...
+
+Although `regex` would only be supported on `Input.Text`, the `isRequired` and `errorMessage` properties would be supported on all `Input.*` types.
```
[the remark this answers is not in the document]
RebeccaAnne: Reworded this section. Let me know if it makes more sense after the update.

---
In reply to: [367015376](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367015376) [](ancestors = 367015376)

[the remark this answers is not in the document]
RebeccaAnne: We didn't have any plans to add any particular restriction here beyond valid regex. Was there something particular you were thinking of?

I don't think we have other scenarios where we take a regex as card author input, but we do use regexs internally to enforce, for example, date/time formatting.

---
In reply to: [367214151](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367214151) [](ancestors = 367214151)

specs/DesignDiscussions/InputValidation.md
RebeccaAnne: All of this discussion is out of the scope of this document, and needs to be reflected in the labels spec. I'll make an update to that soon.

---
In reply to: [367159737](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367159737) [](ancestors = 367159737,355029970,355023402,355022762)

specs/DesignDiscussions/InputValidation.md
RebeccaAnne: Updated the spec to indicate we should scroll the first invalid input into view. Label discussion will continue in the label spec.

---
In reply to: [355034372](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r355034372) [](ancestors = 355034372)

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
+# Input Validation
```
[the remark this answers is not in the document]
RebeccaAnne: Hmm... I'm not sure what would go into a higher level spec. There's basically two pieces, labels and validation. They did end up more connected than i originally imagined they would, but I'm hesitant to combine them as both specs are quite large.  

I did add a link from here to the label spec to help make the connection, and I'll make sure i have the reverse as well when i send out an update to the label spec.

---
In reply to: [367012288](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367012288) [](ancestors = 367012288)

specs/DesignDiscussions/InputValidation.md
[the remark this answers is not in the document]
RebeccaAnne: We don't have an existing isRequired property that officially shipped. 

As we thought we were doing this earlier, there were some places where isRequired may have slipped in. I think maybe it existed in shared model (but wasn't hooked up to anything) and there was at one point a working version of validation for JavaScript that must have included it, but we haven't officially shipped/documented/supported it. 

Let me know once i update if with the new wording you still feel a table up here would help. There is one a few sections down that calls out the properties that are being validated.

---
In reply to: [367015051](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367015051) [](ancestors = 367015051)

specs/DesignDiscussions/InputValidation.md
RebeccaAnne: Updated the spec to call out the schema change for card ids.

---
In reply to: [367156982](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367156982) [](ancestors = 367156982)

specs/DesignDiscussions/InputValidation.md
RebeccaAnne: Put this context in the spec.

---
In reply to: [367155556](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367155556) [](ancestors = 367155556)

specs/DesignDiscussions/InputValidation.md
RebeccaAnne: renamed to associatedInputIds

---
In reply to: [367156518](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367156518) [](ancestors = 367156518)

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,260 @@
...
+    "errorMessage": "Username must be between 5 and 50 characters long."
+}
+```
+The main additions to the schema for input validation are the `"isRequired"`, `"regex"`, and `"errorMessage"` properties on inputs. These properties allow a card author to specify whether or not an input is required, what regular expression defines valid input, and what error message should be shown in case of invalid user input.
+
+Although `regex` would only be supported on `Input.Text`, the `isRequired` and `errorMessage` properties would be supported on all `Input.*` types.
+
+## Schema Additions to Adaptive Card for Validation
+
+In order to allow the card author to visually indicate whether inputs in a card are required or optional, we will add a new property, `showRequiredInputHints`. When this is true, visual indicators will be added to required inputs inputs to signal to the user that they are required. This property will default to true.
```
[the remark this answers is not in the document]
RebeccaAnne: We've removed showRequireInputHints from this version, but added a section at the end to address backwards compatibility impact.

---
In reply to: [367215587](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367215587) [](ancestors = 367215587)

specs/DesignDiscussions/InputValidation.md
RebeccaAnne: We've got a discussion scheduled to talk about host config and defaults in general. Do you have a specific concern related to this spec?

---
In reply to: [367215908](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367215908) [](ancestors = 367215908)

RebeccaAnne: Added updates to the label spec to this PR to cover the current state of this discussion.

---
In reply to: [373689342](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r373689342) [](ancestors = 373689342,367159737,355029970,355023402,355022762)

specs/DesignDiscussions/InputLabels.md
```diff
@@ -123,7 +62,67 @@ One option would be to make the label text a property of the input itself. This
...
+
+This conflict speaks to a general tension inherent in the design of Adaptive Cards. There's always going to be a trade off between giving the author more power, and letting the host control the experience for their users. We need to determine what balance we want to strike in this case.
+
+In order to gather more data to help us balance the needs of the card authors and hosts, we will send a survey to customers who have expressed an interest in improving out input story. The following options are under consideration, and will be evaluated against survey data to determine which if any will be included in our v1 implementation.
+
+### Host Formatting
```
almedina-ms: something that has not been taken into consideration is the spacing between the label and the input and the spacing between the input and the error message, should we use the default or none?

specs/DesignDiscussions/InputLabels.md
```diff
@@ -204,59 +169,36 @@ An alternative option would be to have labels as separate text elements, but pro
 	]
 }
 ``` 
-#### Non-adjacent labels
-Like HTML “label for”, this approach doesn’t require the label to be near the input it labels. In most cases, it seems likely that this is not something we want or need to support, and may in fact be confusing to users. However, we should consider if there are valid scenarios for this use case. Possible scenarios that I could come up with include:
 
-##### Labels to the left of the input
+The downside of this is that it provides a less clean story to our card authors. They would now have two ways to specify a label, and without reading the documentation it would not be clear that one option results in host formatting being applied and the other doesn't. If, for example, a card author wants to emphasize a particular word, and uses a RichTextBlock to highlight that word, they might unknowingly opt themselves out of host formatting. 
+
+##### Non-adjacent labels
```
almedina-ms: This has not been talked about but what happens with Input.Text with inline actions, where does the label go? Only on top of the input or above the input and the action?? the same happens for the error message

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,238 @@
...
+
+In addition to these new properties, there are several existing properties that indicate valid values for an input. Currently, these existing properties are enforced on some platforms whose UI controls allow them to be enforced natively in the UI. For example `Input.Date` has `min` and `max` properties. Ideally, this is implemented as a calendar picker in which dates outside the supported range are greyed out. Currently, however, only some platforms support this behavior due to the presence or absence of a native control with that functionality.
+
+With the advent of this input validation feature, those platforms which do not support these properties should use the input validation behavior described in this spec to add support. Enforcing them natively in the UI by preventing the user from picking an invalid value is still preferred, but when that is not feasible, input validation should be used in these cases. Considering this, the full list of properties that should be either enforced or validated are:
+
+|Input|Properties||
+|---|---|---
+|Input.Text|regex <br> maxLength|New for this feature <br> ExistingProperty
+|Input.Number|min, max|Existing Properties
```
almedina-ms: Should we also validate for integer or floating point numbers?

specs/DesignDiscussions/InputValidation.md
almedina-ms: this are incremental values, right? Focus lost if typing should also validate on submit and the focus lost should also validate as focus lost if typing

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,238 @@
...
+|Input.Date|min, max|Existing Properties
+|Input.Time|min, max|Existing Properties
+|Input.Date|min, max|Existing Properties
+|**All inputs** |isRequired |New for this feature
```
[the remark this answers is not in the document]
almedina-ms: I've made some images to try to help convey my comments about how errors should be rendered on selectable elements

Render all text with attention color
image

 Draw a rectangle around the input
image

Just color the boxes (would this make the input confusing as people may think they have to select all inputs? )
image

specs/DesignDiscussions/InputValidation.md
almedina-ms: TODO: identify what host config will allow and make proposals on how would this be done

specs/DesignDiscussions/InputLabels.md
```diff
@@ -32,76 +32,15 @@ Consider the following card:
 In that card, the label for the input is “Please enter your name:”, however the card renderer has no way to determine that this is the case. We’ve run into a few cases where this is an issue:
  - Accessibility: The input will be read by screen readers as “Last, First”, rather than “Please enter your name” making it more difficult for screen reader users to use the card. 
 
- - Validation indicators: We have considered allowing card authors to mark inputs as optional or required. One way to indicate that visually would be to put and indicator next to the input’s label. This requires knowing which text is the label.
+ - Validation indicators: As part of input validation, we are allowing card authors to mark inputs as optional or required. One way to indicate that visually would be to put and indicator next to the input’s label. This requires knowing which text is the label.
```
almedina-ms: March 27th: Matt, Paul, David and Alberto agreed on labels will get the benefit of isRequired markers, stand alone text blocks will not have markers

specs/DesignDiscussions/InputValidation.md
almedina-ms: We went back to this issue, David thinks Inputs should only be a property on Action.Submit instead of all actions. This would make it easier to extend Inputs to all Action elements; if we consider Inputs to be necessary in all actions we can introduce it later

shalinijoshi19:
Closing this out now that we are tracking spec updates for Inputs via #3897

PR #3669, opened 2019-12-06
[UWP] Fix null dereference in Media code path

### LINKED ISSUE #3668: [UWP] Null dereference in Media.json file
# Platform

* UWP

# Details

WrapInTouchTarget in ActionHelpers.cpp introduced a new dereference of the action parameter with change #3593. The media code path passes in nullptr for the action, leading to a null dereference.

```
        ABI::AdaptiveNamespace::ActionType actionType;
        action->get_ActionType(&actionType);
```

RebeccaAnne:
## Related Issue
Fixes #3668

## Description
WrapInTouchTarget in ActionHelpers.cpp introduced a new dereference of the action parameter with change #3593. The media code path passes in nullptr for the action, leading to a null dereference.

Updated the CreateAppropriateButton method to take the action directly rather than the ActionType, and to check the action for null before dereferencing.

## How Verified
Verified that Media.json file now opens without crashing. This case is covered by the UWP test app.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3669)


shalinijoshi19:
Thanks for catching + fixing. Any chance this can be included as part of our automated unit test suite - if not end-end via the media.json scenario file, at least via a unit test for this specific function in question here (ActionHelpers::CreateAppropriateButton)?

RebeccaAnne:
Added a basic media rendering test to the unit tests that would have caught this case.

---
In reply to: [569335441](https://github.com/microsoft/AdaptiveCards/pull/3669#issuecomment-569335441) [](ancestors = 569335441)

PR #3678, opened 2019-12-12
Background image is hide by column background

venniladhanasekaran:
Background image is hide by column background

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3678)


venniladhanasekaran:
> Can't we set the background color in the container wrapper component itself instead of column and column-set ?

Yes. This works well by setting background color to the container-wrapper. Reverted the changes in column and column-set. Thank you.

Issue #3671, opened 2019-12-10
[Android][Inconsistency with desktop] [Date input component does not honour the default value]


shbhadra:
# Platform
* Android

# Author or host
Host

# Version of SDK
1.2.5

# Details
The date input component does not honour the default value and always shows the current date by default. For the card as in the attached json, following are the views on desktop and android.
(https://github.com/microsoft/AdaptiveCards/files/3944744/default_date.json.zip)

 View on desktop
![image](https://user-images.githubusercontent.com/51158485/70527594-4a993e00-1b72-11ea-8077-2fb71230326c.png)
View on android
<img width="320" alt="Screenshot 2019-12-10 at 5 33 21 PM" src="https://user-images.githubusercontent.com/51158485/70528004-3570df00-1b73-11ea-919f-ccaf0970a29b.png">
[default_date.json.zip](https://github.com/microsoft/AdaptiveCards/files/3944750/default_date.json.zip)

```json
{
   "type":"AdaptiveCard",
   "body":[
      {
         "type":"TextBlock",
         "size":"medium",
         "weight":"bolder",
         "text":"Test title"
      },
      {
         "type":"TextBlock",
         "text":"Testing Date input.",
         "wrap":true
      },
      {
         "type":"Input.Date",
         "value":"03/12/2019",
         "id":"date"
      }
   ],
   "$schema":"http://adaptivecards.io/schemas/adaptive-card.json",
   "version":"1.0"
}
```

nesalang:
Time and Date value have to be expressed in specific format.
Date: YYYY-MM-DD
Time:HH:MM
Resolution: Created [PR](https://github.com/microsoft/AdaptiveCards/pull/3706) for updated doc

shalinijoshi19:
@nesalang,  thanks! Are we now consistent across all renderers/platforms wrt time and date expressions?

shbhadra:
@nesalang @shalinijoshi19 I believe this has been erroneously closed, the issue still persists in the 1.2.6 drop. Note that the bug is not regarding the format of the date, but that the date input element picks up todays date by default, instead of the specified date value.

nesalang:
@shbhadra ,

Thanks for reaching out to us.

'''json
{
         "type":"Input.Date",
         "value":"**03/12/2019**",
         "id":"date"
      }
'''

The date string, "03/12/2019",  is in wrong format. 
The correct format is 2019-03-12 . i.e YYYY-MM-DD.

PR #3706, opened 2020-01-07
updated the doc

nesalang:
## Related Issue
#3671
## Description
Updated doc to better represent the current state of Input.date and Input.time.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3706)

### DIFF
--- a/schemas/src/elements/inputs/Input.Time.json
+++ b/schemas/src/elements/inputs/Input.Time.json
@@ -5,11 +5,11 @@
   "properties": {
     "max": {
       "type": "string",
-      "description": "Hint of maximum value (may be ignored by some clients)."
+      "description": "Hint of maximum value expressed in HH:MM (may be ignored by some clients)."
     },
     "min": {
       "type": "string",
-      "description": "Hint of minimum value (may be ignored by some clients)."
+      "description": "Hint of minimum value expressed in HH:MM (may be ignored by some clients)."
     },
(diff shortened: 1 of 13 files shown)


schemas/src/elements/inputs/Input.Time.json
```diff
@@ -5,19 +5,19 @@
...
+      "description": "Hint of minimum value expressed in HH:MM (may be ignored by some clients)."
     },
     "placeholder": {
       "type": "string",
       "description": "Description of the input desired. Displayed when no time has been selected."
     },
     "value": {
       "type": "string",
-      "description": "The initial value for this field expressed in ISO-8601 format."
+      "description": "The initial value for this field expressed in ISO-8601 format or in HH:MM."
```
RebeccaAnne: >in ISO-8601 [](start = 65, length = 11)

Is there a reason this one is still here as ISO-8601? I though we only supported the simplified times for all of these (UWP only supports HH:MM)

nesalang: thanks for reviewing it; i will update it.

---
In reply to: [364408373](https://github.com/microsoft/AdaptiveCards/pull/3706#discussion_r364408373) [](ancestors = 364408373)

PR #3713, opened 2020-01-09
[Android] Fix to have HH:mm in Input.Time elements

### LINKED ISSUE #3672: [Android][Inconsistency between android and web] [Time rendering format on android is different as that of web]
# Platform
* Android
* iOS
# Author or host
HOST

# Version of SDK
1.2.5

# Details
The time rendered in the Input.Time component, on android also shows the seconds whereas it doesn't on web. With reference to the card json attached
View of the card in web
Screenshot 2019-12-10 at 5 59 09 PM
View of the card on android
Screenshot 2019-12-10 at 6 00 16 PM
[time_render.json.zip](https://github.com/microsoft/AdaptiveCards/files/3944830/time_render.json.zip)
```json
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "TextBlock",
      "text": "What time do you want to meet?"
    },
    {
      "type": "Input.Time",
      "id": "time",
      "min": "09:00",
      "max": "17:00",
      "value": "15:30"
    }
  ],
  "actions": [
    {
      "type": "Action.Submit",
      "title": "OK"
    }
  ]
}
```

iOS would not show "PM" in this use case.

almedina-ms:
## Related Issue
Fixes #3672 

## Description
This PR addresses the inconsistency between Android and Web as Android's renderer shows the time in HH:mm:ss format while Web shows it as HH:mm . With this fix, Input.Time in android now uses the HH:mm format

## How Verified
The fix was tested manually, the main scenarios tested were checking default values, checking that selected values were kept and that values introduced were persistent, aka, basic functionality

### Default value check
Screenshot_20200109-135152_AdaptiveCards Visualizer

### Value changed
Screenshot_20200109-135159_AdaptiveCards Visualizer

### Checking that value changed is persistent in the pop up
Screenshot_20200109-135215_AdaptiveCards Visualizer

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3713)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/TimePickerFragment.java
```diff
@@ -38,7 +36,7 @@ public Dialog onCreateDialog(Bundle savedInstanceState)
         try
         {
             // Get current value shown to user in the TextEdit
-            Date value = DateFormat.getTimeInstance().parse(m_editText.getText().toString());
+            Date value = DateFormat.getTimeInstance(DateFormat.SHORT).parse(m_editText.getText().toString());
```
RebeccaAnne: >Date value = DateFormat.getTimeInstance(DateFormat.SHORT).parse(m_editText.getText().toString()); [](start = 12, length = 97)

Should this line use the helper function as well?

shalinijoshi19:
@almedina-ms are there unit tests we can add for this case? Also if there's a scenario json, let's get into the habit of checking those in as part of our PRs as well. Thanks. @nesalang FYI.

RebeccaAnne:
Do we need a new json file for this? Are there existing files that use these properties? Or existing files that could be updated to use them? I'd rather not add unnecessary json files if we can avoid it as the tests that run through all json files will take longer the more files we have.

---
In reply to: [575289561](https://github.com/microsoft/AdaptiveCards/pull/3713#issuecomment-575289561) [](ancestors = 575289561)

shalinijoshi19:
@almedina-ms we should be able to go ahead and check this in at this point yes?

PR #3714, opened 2020-01-09
[Android] Fix scrolling behaviour in multiline Input.Text

### LINKED ISSUE #3693: Input.Text element with multi line isn't vertically scrollable
# Platform

* Android

# Author or host

Host

# Version of SDK

1.2.5

# Issue
 
An Input.Text element with `"isMultiline": true` isn't vertically scrollable. If the input has multiple lines, then there is no way to scroll up and view the previously typed text.
However, the same element is scrollable on Desktop/Web and iOS.

Below is the sample Input.Text element JSON.

> {
>             "id": "MultiLineVal",
>             "placeholder": "Comments",
>             "type": "Input.Text",
>             "isMultiline": true
>  }

Also, attaching the relevant JSON.
[multi-line-input-text.json.zip](https://github.com/microsoft/AdaptiveCards/files/4004709/multi-line-input-text.json.zip)

cc: @VikrantSingh01 .

almedina-ms:
## Related Issue
Fixes #3693 

## Description
This changes fix the current behaviour of multiline Input.Text elements as it doesn't allow scrolling when the element is inside a tall card (tall enough so that the scrollbar indicators show up)

## How Verified
This fix was tested manually and this is the result showing the new behavior

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3714)


shalinijoshi19:
@almedina-ms can we add this sample card to our list of samples ? @nesalang FYI. Do our existing Android Visualizer tools allow for testing of this card or are there app changes on Android that need to be made as well? Also can we test this out in this month's bugbash for all platforms (ios,Android, web, uwp etc) ?

Issue #3719, opened 2020-01-10
[UWP][Unit Tests] Remove image comparison from unit tests


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* UWP

# Details

Remove image comparison from the UWP unit tests. The image comparison is not stable enough to run consistently as a unit test, and requires manual updates to consider and accept new cards. Updating the unit tests to verify json round tripping and errors only allows us to decouple efforts to stabilize the test app from the ability to get the other UWP unit tests running in CI.

PR #3724, opened 2020-01-13
[UWP] Remove image comparison from unit tests

RebeccaAnne:
## Related Issue
Fixes: #3719

## Description
Updated the unit tests that runs all files to not require image comparisons to match. Image comparison is not stable enough to run as an automated blocking test at this time, and this change decouples stabilizing the test app from getting a basic rendering pass of all files into CI.

A file will pass for the unit test if it 
1. Is expected to fail, and does
2. It is expected to succeed and does, and the expected round tripped json is unchanged
3. It is a new file, and renders successfully.

## How Verified
Unit test suite passes consistently

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3724)

### DIFF
--- a/source/uwp/UWPTestLibrary/TestResultViewModel.cs
+++ b/source/uwp/UWPTestLibrary/TestResultViewModel.cs
@@ -490,7 +490,7 @@ namespace UWPTestLibrary
             {
                 return "Passed";
             }
-            else if (NewCard == true)
+            else if (NewCard)
             {
                 if (Error == null)
                 {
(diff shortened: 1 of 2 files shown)


source/uwp/UWPTestLibrary/TestResultViewModel.cs
```diff
@@ -488,7 +492,14 @@ public override string ToString()
             }
             else if (NewCard == true)
```
paulcam206:
```suggestion
            else if (NewCard)
```

not your bug, but... :)

RebeccaAnne: Probably is my bug, to be honest :)

Fixed.

---
In reply to: [366065171](https://github.com/microsoft/AdaptiveCards/pull/3724#discussion_r366065171) [](ancestors = 366065171)

Issue #3723, opened 2020-01-13
[Javascript][Safari/Media] [Media Attachments not loading in Safari]


tdurnford:
# Platform
* JavaScript

# Author or host
Host [Web Chat](https://github.com/microsoft/BotFramework-WebChat/)

If you're an author, who are you sending cards to?

# Version of SDK
Javascript v1.2.5

# Details
Videos embedded in an Adaptive Card do not load on iOS devices when the user presses the play button that is overlaid on the video unless the user presses pause and restarts the video or navigates away from Safari and then returns to the page.

## Adaptive Card
```json
{
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "type": "AdaptiveCard",
  "version": "1.1",
  "fallbackText": "This card requires Media to be viewed. Ask your platform to update to Adaptive Cards v1.1 for this and more!",
    "body": [
        {
            "type": "Media",
            "poster": "https://adaptivecards.io/content/poster-video.png",
            "sources": [
                {
                    "mimeType": "video/mp4",
                    "url": "https://adaptivecardsblob.blob.core.windows.net/assets/AdaptiveCardsOverviewVideo.mp4"
                }
            ]
        }
    ],
    "actions": [
        {
            "type": "Action.OpenUrl",
            "title": "Learn more",
            "url": "https://adaptivecards.io"
        }
    ]
}
```

## Screen Recording
![RPReplay_Final1578941642](https://user-images.githubusercontent.com/17039790/72284205-3f17b780-35f5-11ea-9968-d370b2cd3b3c.gif)

## Additional Context
https://github.com/microsoft/BotFramework-WebChat/issues/2697

shalinijoshi19:
Hi @tdurnford thanks for reporting this issue as well as the additional context. Could you help us understand specific repro steps/debugging details that lead you to believe this to be an AdaptiveCards issue? @nesalang FYI.

nesalang:
@tdurnford 
does your bot-framework use ios adaptive cards or java script adaptive cards sdk?
if iOS, can you make sure this delegate is implemented?
- (void)didFetchMediaViewController:(AVPlayerViewController *)controller
                               card:(ACOAdaptiveCard *)card
you can check the sample app's implementation from ViewController.m

arman-g:
Is this going to be included in the next release? And when is the next release? Thanks!

arman-g:
Any updates on this one?

PR #3732, opened 2020-01-17
[JS] Fix video not autoplaying on iOS Safari

dclaux:
## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/3723

## Description
The `<video>` tag on mobile OSes (iOS and Android) behaves in a non-standard way. In the context of this issue:
- Non-muted videos will not auto play, even if the `autoplay` attribute is specified
- The JS API `videoTag.play()` will not work if not triggered via explicit user interaction
- iOS only: The `playsinline` attribute must be specified on the `<video>` tag, otherwise the video will only play full screen

Refer this this page for more details: https://webkit.org/blog/6784/new-video-policies-for-ios/

This fix detects if the renderer in running on a mobile OS and adds the appropriate attributes accordingly.

## How Verified
- Verified that the implementation still works as before in "old" Edge, "new" Edge, Chrome, Firefox, Opera using the visualizer
- Pending verification on iOS. Test at https://adaptivecardsci.z5.web.core.windows.net/pr/3732/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3732)


source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -4,6 +4,12 @@ import * as Enums from "./enums";
 import * as Shared from "./shared";
 import { HostConfig } from "./host-config";
 
+export function isMobileOS(): boolean {
+    let userAgent = window.navigator.userAgent;
+
+    return !!userAgent.match(/Android/i) || !!userAgent.match(/iPad/i) || !!userAgent.match(/iPhone/i);
```
shalinijoshi19: Are these the only 2 iOS device types we care about? (iPad/iPhone) ?

shalinijoshi19:
@nesalang is there an existing scenario file that we could leverage for testing this (outside of the useragent requirement) today? If not @dclaux we should add the scenario file to a folder someplace in our samples to ensure we can subsequently test this for now on-demand for bug bashes for eg. Also is this something that we want to include in the upcoming v1.2.6 patch this month? If so do tag the issue with the label pls. Thanks!

source/nodejs/adaptivecards/src/utils.ts
dclaux: There is only one more device, the iPod, and I don't think we really care.

matthidinger: This feels a bit crude of a check. Do we have a desire to behave differently if a user checked on "Show Desktop Version" or any other mobile-browser toggles?

dclaux: Why is it crude? It detects if the renderer is running on a mobile OS, in which case things must be done differently for autoplay to work.

nesalang:
yes, we do have a existing test file. 

---
In reply to: [575829211](https://github.com/microsoft/AdaptiveCards/pull/3732#issuecomment-575829211) [](ancestors = 575829211)

PR #3737, opened 2020-01-22
[JS] Fix Angular build break

### LINKED ISSUE #3736: [JavaScript] Error encountered in metadata generated for exported symbol 'Media'
# Platform

* JavaScript

# Version of SDK

1.2.5

# Issue

I'm making an Angular library that uses AdaptiveCards. When I bundle the library, I get two Errors in the source code from Adaptive Cards:

`
ERROR: /node_modules/adaptivecards/src/card-elements.ts:2487:1: Error encountered in metadata generated for exported symbol 'Media':
/node_modules/adaptivecards/src/card-elements.ts:2659:12: Metadata collected contains an error that will be reported at runtime: Only initialized variables and constants can be referenced because the value of this variable is needed by the template compiler.
  {"__symbolic":"error","message":"Variable not initialized","line":2658,"character":11}
`

and

`
ERROR: /node_modules/adaptivecards/src/card-elements.ts:6612:1: Error encountered in metadata generated for exported symbol 'AdaptiveCard':
/node_modules/adaptivecards/src/card-elements.ts:6642:16: Metadata collected contains an error that will be reported at runtime: Only initialized variables and constants can be referenced because the value of this variable is needed by the template compiler.
{"__symbolic":"error","message":"Variable not initialized","line":6641,"character":15}
`

the first error can be solved by initializing the static variable onPlay with null:
`static onPlay: (sender: Media) => void = null;`

the second error needs another fix: because the static variable exists only as setter and getter. Here the solution is to add the dynamic annotation as comment in front of the class.

`
// @dynamic

export class AdaptiveCard extends ContainerWithActions {
`

with these two simple fixes I can build my library.

ivanhofer:
## Related Issue

Fixes #3736

## Description
 
Attempting to build AdaptiveCards JS in an Angular project fails. These changes fix that.

## How Verified

Library builds correctly

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3737)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2668,7 +2668,7 @@ export class Media extends CardElement {
         return element;
     }
 
-    static onPlay: (sender: Media) => void;
+    static onPlay: (sender: Media) => void = null;
```
dclaux: Can you elaborate why this is necessary here only? There are other event declarations in the library that also do not include `= null`. Also why `null` and not `undefined`?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -6607,6 +6607,7 @@ export interface IMarkdownProcessingResult {
     outputHtml?: any;
 }
 
+// @dynamic
```
dclaux: Can you elaborate what this is, why it is necessary and why only here?

source/nodejs/adaptivecards/src/card-elements.ts
ivanhofer: I got the error `Only initialized variables and constants can be referenced because the value of this variable is needed by the template compiler`
This is the only static field that is not initialized in this class.
You're right: `undefined ` would be better.

source/nodejs/adaptivecards/src/card-elements.ts
ivanhofer: in this class there is not a real static field that is not initialized. The static field has only a getter and a setter, so it can't be initialized (at least of my knowledge). 
I found this [issue](https://github.com/ng-packagr/ng-packagr/issues/641) and applied the solution from there

source/nodejs/adaptivecards/src/card-elements.ts
ivanhofer: i was wrong about the `undefined`. I took again a look in at the code and the reason I used `null` is because of the click() method on this class. There is a if-statement that checks for `this.onClick != null` . So undefined would fail the test and it would the invocation of the method `this.onClick(this);` in the next line would fail.

shalinijoshi19:
@paulcam206 could you help debug the .NET CI failures? thanks

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: So there are a bunch of static events in the AdaptiveCard class, for instance:
```typescript
static onExecuteAction?: (action: Action) => void;
```

Your solution for AdaptiveCard seems to be to decorate it with `// @dynamic` whereas on the `Media` class you opted to initialize the event declaration. Is that correct?

It seems to me that the issue should be solved the same way everywhere. Given that `// @dynamic` seems to be an Angular-specific concept, I think the right solution is to initialize the various static fields in the AdaptiveCard class, and remove the `// @dynamic` directive.

ivanhofer: I agree that it would be better to use the same approach everywhere. But like I sayed in my comment above, I don't know a solution to rewrite the getter and setter of `processMarkdown `without changing the behavior of the code. The only solution I see is to remove `processMarkdown` (it is deprecated) completely from the class.

dclaux: Oh my mistake, I did read your comment about that, then forgot. Sorry about that. Then I too don't see an alternative for now.

PR #3739, opened 2020-01-22
[UWP] Remove weak reference to ellipse object in image renderer

### LINKED ISSUE #3720: [UWP] Person image with background color not rendering
# Platform

* UWP

# Details
 [Image.BackgroundColor.json](https://github.com/microsoft/AdaptiveCards/blob/master/samples/v1.1/Tests/Image.BackgroundColor.json) test doesn't render the person image with a background color. Only a circle with the background color is shown. Removing either the background color or the person style causes the image to render properly. Adding a size also causes it to render properly.

image

paulcam206:
Fixes #3720

## Description
Commit 8058022ea7d9dc440e4aa693149538d14192f92d fixed a big leak in our rendering stack, but unfortunately regressed the behavior of `Image` elements that have both a `backgroundColor` and `"style": "person"`. In particular:

https://github.com/microsoft/AdaptiveCards/blob/fe5fcefb4275b6bfd0fbf395e0fc80f71a01277d/source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp#L707-L715

By the time the callback in this snippet is invoked, `ellipseAsUIElement` has passed out of scope and been deleted, so `lambdaEllipseAsFrameworkElement` ands up being `nullptr`. We actually need a stronger reference here, but moving back to a strong reference reintroduces the large leak we had. However, the source for the leak actually ends up being `imageSourceAsBitmap`. It's not intuitively obvious from a casual inspection of this function, but `imageSourceAsBitmap` is actually `brushAsImageBrush`'s `ImageSource` QI'd to an `IBitmap`. In other words, the unused `IInspectable*` parameter passed to the callback is actually the same source as the lambda-captured `imageSourceAsBitmap`. The fix is to not capture `imageSourceAsBitmap`, but rather to reacquire the same thing from within the callback (breaking the circular reference).

## How Verified
* Added `"style": "person"` to `Image.BackgroundColor.json` and updated reference images
* Verified no large leak using PerfApp
* Visually verified in the visualizer:
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3739)


source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -704,24 +704,26 @@ namespace AdaptiveNamespace
                 WeakRef weakParent;
```
paulcam206: need to update comment above

shalinijoshi19:
Thanks for debugging & fixing @paulcam206! Can we add a unit test to cover this regression (even if the actual CI inclusion of these unit tests is a work-in-progress?  I am assuming also that since hte refactor is at the moment only in master, this doesnt warrant CP'ing over to release/1.2?

shalinijoshi19:
>                         ComPtr<IFrameworkElement> lambdaImageAsFrameworkElement;

So is this block of code subject to a similar bug where the image object goes out of scope by the time the callback gets invoked? 

---
Refers to: source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp:777 in 1c3f806. [](commit_id = 1c3f806b46e4bb7e27e06881aa0ac49be7d95575, deletion_comment = False)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -704,24 +704,26 @@ namespace AdaptiveNamespace
...
                 THROW_IF_FAILED(localParentElement.AsWeak(&weakParent));
 
-                WeakRef weakEllipse;
-                THROW_IF_FAILED(ellipseAsUIElement.AsWeak(&weakEllipse));
                 THROW_IF_FAILED(brushAsImageBrush->add_ImageOpened(
```
shalinijoshi19: >add_ImageOpened [](start = 51, length = 15)

Would switching to wpp/winrt help simplify this eventregistration pattern? Do we actually want the callback to be invoked if the parent object teh weakref in the lambda is pointing to, gets destructed?

paulcam206:
this work will need to be ported to `release/1.2`, but was easier to debug/fix in master (I could partially revert the original change out of `master`, but not the release branch since release's commit was a large set of changes)

paulcam206:
> Can we add a unit test to cover this regression (even if the actual CI inclusion of these unit tests is a work-in-progress?

I don't think we currently do any visual verification in our unit tests. However, when we get sample cards verified in UT, this will be covered due to the addition of the person style to `Image.BackgroundColor.json`. :)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
paulcam206: The registration pattern *would* look nicer in cpp/winrt. However, we aren't currently using that pattern anywhere in our codebase, so this would be the first instance. Since this is a regression fix, I didn't want to hold up shipping it to figure out the lifetime patterns or change a bunch of code...

---
In reply to: [369931944](https://github.com/microsoft/AdaptiveCards/pull/3739#discussion_r369931944) [](ancestors = 369931944)

paulcam206: Also, I didn't really dig in to why `weakEllipse` was resolving as `nullptr`.

---
In reply to: [370266579](https://github.com/microsoft/AdaptiveCards/pull/3739#discussion_r370266579) [](ancestors = 370266579,369931944)

shalinijoshi19:
> > Can we add a unit test to cover this regression (even if the actual CI inclusion of these unit tests is a work-in-progress?
> 
> I don't think we currently do any visual verification in our unit tests. However, when we get sample cards verified in UT, this will be covered due to the addition of the person style to `Image.BackgroundColor.json`. :)

I was actually wondering if testing the async registration code as a unit, vs a functional unit could help add coverage without needing visual verification.. We shouldnt need full on functional verification to add code coverage for this path right?

PR #3742, opened 2020-01-24
Jwoo/gradle update

nesalang:
## Related Issue
android can't be built without commenting an azure related token.

## Description
provided a null value, so the token will be recognized.

## How Verified
built locally using the updated gradle.properties file.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3742)


shalinijoshi19:
@nesalang can you document the purpose for this and other variables used in teh pipeline.

PR #3773, opened 2020-02-11
[UWP][Visualizer Accessibility] Tab Order Fixes

### LINKED ISSUE #3771: [UWP][Visualizer Accessibility] Keyboard Focus goes to non-interactive error section
# Platform

* UWP

# Details

When navigating the visualizer using the tab key, after the json edit section the focus goes to the error section. This section is non-interactive, and should not have a tab stop.

### LINKED ISSUE #3772: [UWP][Visualizer Accessibility] Keyboard Focus goes to card before separator
# Platform

* UWP

# Details

When tab navigating the visualizer, the focus moves to the rendered card before the separator slider which is unexpected because the slider is to the left of the rendered card.

RebeccaAnne:
## Related Issue
Fixes #3771 
Fixes #3772

## Description
Two fixes related to tab order in the UWP visualizer:
1. Tab order went json panel->rendered card->grid splitter, when json panel->grid splitter->rendered card would be more appropriate. Changed the order of those two elements in xaml to update this ordering.
2. The error panel was included in the tab order even though it is not interactive. Set IsEnabled to false to indicate to xaml that the error panel should not have a tab stop.

## How Verified
Manually confirmed tab order now behaves as expected in these cases.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3773)

### DIFF
--- a/source/uwp/Visualizer/DocumentView.xaml
+++ b/source/uwp/Visualizer/DocumentView.xaml
@@ -19,7 +19,19 @@
         </Grid.ColumnDefinitions>
 
         <!--Editor and errors-->
-        <local:GenericDocumentView />
+        <local:GenericDocumentView x:Name="GenericDocumentViewName"/>
+
+        <!--Vertical draggable column splitter-->
+        <toolkitControls:GridSplitter
+            Width="11"
+            Background="{ThemeResource SystemControlBackgroundAccentBrush}"
+            GripperCursor="Default"
+            HorizontalAlignment="Left"
+            Grid.Column="1"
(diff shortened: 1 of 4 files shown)


shalinijoshi19:
thanks! Not high priority, but we should probably consider pushing the visualizer from our nightlies from master into appcenter as well much like we are doing for our iOS/Android visualizers now @nesalang for fyi who's coordinated this for the tools for the other platforms.

PR #3778, opened 2020-02-12
[UWP] Set background image tile control to IsEnabled false to avoid generating extraneous tab stops

### LINKED ISSUE #3777: [UWP][Accessibility] Cards with background images contain an additional tab stop
# Platform

* UWP

# Details

When a card with a background image is rendered in UWP (i.e. [AdaptiveCard.BackgroundImage.FillMode.Cover.json](https://github.com/microsoft/AdaptiveCards/blob/master/samples/v1.2/Elements/AdaptiveCard.BackgroundImage.FillMode.Cover.json)), there is an extra tab stop. This can cause confusion for keyboard users.

RebeccaAnne:
## Related Issue
Fixes #3777 

## Description
The additional tab stop is caused by the custom TileControl which hosts the background image. Because this is not intended as an interactive control, set IsEnabled to false to remove the tab stop.

## How Verified
Manually verified additional tab stop is no longer present using [AdaptiveCard.BackgroundImage.FillMode.Cover.json](https://github.com/microsoft/AdaptiveCards/blob/master/samples/v1.2/Elements/AdaptiveCard.BackgroundImage.FillMode.Cover.json)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3778)

### DIFF
--- a/source/uwp/Renderer/lib/XamlHelpers.cpp
+++ b/source/uwp/Renderer/lib/XamlHelpers.cpp
@@ -334,6 +334,12 @@ namespace AdaptiveNamespace::XamlHelpers
         // Creates the background image for all fill modes
         ComPtr<TileControl> tileControl;
         THROW_IF_FAILED(MakeAndInitialize<TileControl>(&tileControl));
+
+        // Set IsEnabled to false to avoid generating a tab stop for the background image tile control
+        ComPtr<IControl> tileControlAsControl;
+        THROW_IF_FAILED(tileControl.As(&tileControlAsControl));
+        THROW_IF_FAILED(tileControlAsControl->put_IsEnabled(false));
+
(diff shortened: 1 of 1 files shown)


shalinijoshi19:
Nice catch. How did you find it?

RebeccaAnne:
Saw a comment somewhere that this repro'd whenever there was a background image, which pointed me at the thing hosting the background images (TileControl). Since that inherits from Control, it needs to be marked as not enabled if it's not interactive otherwise it will generate a tab stop.

---
In reply to: [585551792](https://github.com/microsoft/AdaptiveCards/pull/3778#issuecomment-585551792) [](ancestors = 585551792)

PR #3798, opened 2020-02-21
Jwoo/dotnet templating vs

jwoo-msft:
## Description
This encompasses .Net Template SDK's design spec, documentation, and source code.
It also has some prior commits before ANTLR was used.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3798)


specs/DesignDiscussions/DotNetTemplate.md
```diff
@@ -0,0 +1,338 @@
...
+
+Once binding is defined, adaptive cards will be updated by changing only the data.
+```json
+{
+  "latestPrice": 128.9,
+}
+```
+### Syntax
+- Binding syntax starts with { and ends with }. E.g., {myProperty}
+- Dot-notation to access sub-objects
```
RebeccaAnne: >ot-notation to access sub-objects [](start = 3, length = 33)

Matt: Add here that we need to gracefully handle null objects (if manager is null, it should just return empty)

specs/DesignDiscussions/DotNetTemplate.md
```diff
@@ -0,0 +1,338 @@
...
+- $when conditional expression that determines if current element to be included.
+- functions are provided to help SDK users
+
+# Parsing
+We use ANTLR to generate parser and lexer
+
+## ANTLR
+From a grammar, ANTLR generates a parser that can build parse trees and also generates a listener interface (or visitor) that makes it easy to respond to the recognition of phrases of interest.
+ANTLR is really two things: a tool that translates your grammar to a parser/lexer in Java (or other target language) and the runtime needed by the generated parsers/lexers.
+Once parsers and lexers are generated, the tool is no longer needed, so it won't be part of template SDK. CEL uses ANTLR, so CEL SDK already includes ANTRL runtime. There will be no additional penaly of using ANTLR with careful planning
```
RebeccaAnne: >careful planning [](start = 220, length = 16)

Discussion in the room: What is careful planning? This refers to ensuring that we don't use a different version of ANTLR which would cause us to bring in two versions.

specs/DesignDiscussions/DotNetTemplate.md
```diff
@@ -0,0 +1,338 @@
...
+- $data assigns data context to current element
+- $when conditional expression that determines if current element to be included.
+- functions are provided to help SDK users
+
+# Parsing
+We use ANTLR to generate parser and lexer
+
+## ANTLR
+From a grammar, ANTLR generates a parser that can build parse trees and also generates a listener interface (or visitor) that makes it easy to respond to the recognition of phrases of interest.
+ANTLR is really two things: a tool that translates your grammar to a parser/lexer in Java (or other target language) and the runtime needed by the generated parsers/lexers.
```
RebeccaAnne: >or other target language [](start = 91, length = 24)

Question in the room: What languages do we need? Do we need objective C? This will be happening in a C++ layer and we won't need objective C.

Current plan is for .NET and JavaScript to handle server scenarios. Long term we'll want to support client side for mobile/UWP as well. It won't be a problem to handle this in C++ and consume it on all shared model platforms.

specs/DesignDiscussions/DotNetTemplate.md
```diff
@@ -0,0 +1,338 @@
...
+**Parse Tree with Templated String**
+```json
+{
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"type": "AdaptiveCard",
+	"version": "1.0",
+	"body": [
+		{
+			"type": "TextBlock",
+			"text": "**wrap: true** {content}",
```
RebeccaAnne: >**wrap: true** [](start = 12, length = 14)

Document nit: It would probably be clearer if this string didn't look like a json property.

RebeccaAnne: >{ [](start = 27, length = 1)

This is going to change to ${

specs/DesignDiscussions/DotNetTemplate.md
RebeccaAnne: >CEL [](start = 106, length = 3)

Update the doc to use Adaptive Expression Language everywhere.

specs/DesignDiscussions/DotNetTemplate.md
```diff
@@ -0,0 +1,338 @@
...
+
+**Signature of TryEvaluate**
+```dotnet
+/// <summary>
+/// Evaluate the expression.
+/// </summary>
+/// <param name="state">
+/// Global state to evaluate accessor expressions against.  Can be <see cref="IDictionary{String}{Object}"/>, <see cref="IDictionary"/> otherwise reflection is used to access property and then indexer.
+/// </param>
+/// <returns>Computed value and an error string.  If the string is non-null, then there was an evaluation error.</returns>
```
RebeccaAnne: >f the string is non-nul [](start = 51, length = 23)

Discussion in room: These comments seem not entirely up to date.

RebeccaAnne: Add a link to wherever this came from.

---
In reply to: [382839131](https://github.com/microsoft/AdaptiveCards/pull/3798#discussion_r382839131) [](ancestors = 382839131)

RebeccaAnne: Nevermind, the comment is fine, but adding a link would still help.

---
In reply to: [382839335](https://github.com/microsoft/AdaptiveCards/pull/3798#discussion_r382839335) [](ancestors = 382839335,382839131)

specs/DesignDiscussions/DotNetTemplate.md
```diff
@@ -0,0 +1,338 @@
...
+                    "$data": {
+                        "key": "3"
+                    },
+                    "text": "key should be 3, actual:{key}",
+                    "size": "Medium",
+                    "wrap": true
+                },
+                {
+                    "type": "TextBlock",
+                    "text": "key should be 2, actual:{key}",
```
RebeccaAnne: >key [](start = 54, length = 3)

Discussion: Should this have to be parent.key?

RebeccaAnne: Lots of discussion of the fact that the child nodes data context masks the entire data context of the parent. Some concern that we can't get to the parent's properties. This could be solved with support of an implicit parent.foo. This is not currently in plan but could theoretically be added to address this issue.

---
In reply to: [382840789](https://github.com/microsoft/AdaptiveCards/pull/3798#discussion_r382840789) [](ancestors = 382840789)

specs/DesignDiscussions/DotNetTemplate.md
```diff
@@ -0,0 +1,348 @@
...
+],
+```
+### Parsing Strategy
+Since the introduction of a new data context happens only in a pair, and only an obj can include the pair, the Obj is the natural place to handle the data context. 
+In VisitObj method, it filters through the method's current context's children which are pairs, and tries to catch a data context node. 
+ Current design tentatively uses the first one it sees for the performance reason.  
+When a data context node is encountered, it call a base visit method with the node, then the method resolved into VisitTemplateData method.
+The method, creates a data context object with a field that indicates Array if set. The data context object parse the value in the pair of the data context node into JArray object if it's array, otherwise into JObject
+The newly created data context object is set to current context. In current design, it uses a stack for this purpose. When the method returns to the VisitObj method, data context is set, and translation can continue. Please refer to **Figure 4**
+* Do we use first data context we see, or do we have to see all of them?
```
RebeccaAnne: Discussion in the room: This is invalid json. We should do whatever we usually do with invalid json. (In shared model, for example, we don't render cards that are invalid json) 

Relatedly, as part of the inconsistency work we should confirm that we handle invalid json the same way on all platforms.

specs/DesignDiscussions/DotNetTemplate.md
```diff
@@ -0,0 +1,348 @@
...
+    "body": [
+        {
+            "type": "Container",
+            "items": [
+                {
+                    "type": "TextBlock",
+                    "text": "Header"
+                },
+                {
+                    "$data": [
```
RebeccaAnne: >$data [](start = 21, length = 5)

Discussion: This will repeat the fact set. Might make more sense to update the example to update the facts instead.

specs/DesignDiscussions/DotNetTemplate.md
chrimc62: And also so you know, string interpolation has been checked in and uses backquotes so if you call the expression engine with `my name is ${name}`, you will get the concatenation of the constant and evaluated parts.  This works inside of a ${} expression as well so you could do `my name is ${`awesome ${name}`}`.

---
In reply to: [382836292](https://github.com/microsoft/AdaptiveCards/pull/3798#discussion_r382836292) [](ancestors = 382836292)

specs/DesignDiscussions/DotNetTemplate.md
```diff
@@ -0,0 +1,348 @@
...
+ - if they have templated string that needs to be expanded, then they are linked as linked lists without merging the nodes
+- In the VisitObj method, once visit to children finishes, we have one merged result object
+- if data context object was discovered in the current node, and it's array type, expansion happens with the merged result object
+ - it repeats the following steps n times where n is number of elements in data context
+ - it expands the result with data context. 
+ - the expanded result is merged
+- the result is returned from the method.
+- if data context object wasn't discovered in this node, simply return the current result without expanding, so it can be handled by a parent node which will have the data context
+
+## $when
```
RebeccaAnne: >$when [](start = 3, length = 5)

Discussion: Should this be $where like SQL? It's not really a filtering thing. This is really $showWhen or $showIf for a single object. Should we change to one of those? $when is consistent with Amazon's templating language (APL), is it valuable to match?

RebeccaAnne: Seems like we're agreed to stay with $when.

---
In reply to: [385959190](https://github.com/microsoft/AdaptiveCards/pull/3798#discussion_r385959190) [](ancestors = 385959190)

specs/DesignDiscussions/DotNetTemplate.md
nesalang: it's noted. I'll also add unit tests to ensure it's handled properly

specs/DesignDiscussions/DotNetTemplate.md
nesalang: when I updated to the latest and greatest version of AEL, I will change it '${', 

---
In reply to: [385954354](https://github.com/microsoft/AdaptiveCards/pull/3798#discussion_r385954354) [](ancestors = 385954354,382836292)

nesalang: Thanks Chris, I will try it out soon.

---
In reply to: [385954354](https://github.com/microsoft/AdaptiveCards/pull/3798#discussion_r385954354) [](ancestors = 385954354,382836292)

specs/DesignDiscussions/DotNetTemplate.md
nesalang: I will consider it as p1 item, and re-evaluate it once all p0 items are complete.

---
In reply to: [382845909](https://github.com/microsoft/AdaptiveCards/pull/3798#discussion_r382845909) [](ancestors = 382845909,382840789)

shalinijoshi19:
@jwoo-msft is this PR still valid?

jwoo-msft:
closed and replaced with https://github.com/microsoft/AdaptiveCards/pull/3958 and https://github.com/microsoft/AdaptiveCards/pull/3946

PR #3816, opened 2020-03-03
[Designer/templating] Fixes + add toString

dclaux:
## Related Issue
Fixes https://github.com/MicrosoftDocs/AdaptiveCards/issues/251

## Description
- Actually create a copy of the expression nodes which were previously overwritten causing the same data object to be reused for all items in an array
- Introduce toString to support basic explicit conversion of numbers into string

## How Verified
Verified in:
- adaptivecards-designer-app
- Web site running locally

## How to test
Card payload:
```json
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "Container",
            "items": [
                {
                    "type": "ColumnSet",
                    "columns": [
                        {
                            "type": "Column",
                            "width": "auto",
                            "items": [
                                {
                                    "type": "TextBlock",
                                    "text": "{if(changePriceRatio >= 0, '▲', '▼')}",
                                    "color": "{if(changePriceRatio >= 0, 'good', 'attention')}"
                                }
                            ]
                        },
                        {
                            "type": "Column",
                            "width": "stretch",
                            "items": [
                                {
                                    "type": "TextBlock",
                                    "text": "{name}"
                                }
                            ]
                        },
                        {
                            "type": "Column",
                            "width": "stretch",
                            "items": [
                                {
                                    "type": "TextBlock",
                                    "text": "{toString(lastPrice)}",
                                    "horizontalAlignment": "Center"
                                }
                            ],
                            "horizontalAlignment": "Center"
                        },
                        {
                            "type": "Column",
                            "width": "auto",
                            "items": [
                                {
                                    "type": "TextBlock",
                                    "text": "{changePriceRatio}%",
                                    "color": "{if(changePriceRatio >= 0, 'good', 'attention')}"
                                }
                            ],
                            "horizontalAlignment": "Right"
                        }
                    ]
                }
            ],
            "$data": "{$root}"
        }
    ]
}
```

Data payload:
```json
[
    {
        "name": "Object-1",
        "lastPrice": 1.10762,
        "changePriceRatio": -0.17
    },
    {
        "name": "Object-2",
        "lastPrice": 1578.205,
        "changePriceRatio": -0.68
    },
    {
        "name": "Object-3",
        "lastPrice": 51.475,
        "changePriceRatio": -0.23
    },
    {
        "name": "Object-4",
        "lastPrice": 28324,
        "changePriceRatio": 0.35
    },
    {
        "name": "Object-5",
        "lastPrice": 9338.87,
        "changePriceRatio": -1.04
    }
]
```

Card should render as follows in the designer in "Preview mode":
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3816)

### DIFF
--- a/source/nodejs/adaptivecards-designer-app/src/app.ts
+++ b/source/nodejs/adaptivecards-designer-app/src/app.ts
@@ -12,13 +12,11 @@ import "./app.css";
 
 window.onload = function() {
     // Uncomment to enabled preview features such as data binding
-    /*
     ACDesigner.GlobalSettings.showVersionPicker = true;
     ACDesigner.GlobalSettings.enableDataBindingSupport = true;
     // Note the below two flags are ignored if enableDataBindingSupport is set to false
     ACDesigner.GlobalSettings.showDataStructureToolbox = true;
     ACDesigner.GlobalSettings.showSampleDataEditorToolbox = true;
-    */
 
     // Uncomment to configure default toolbox titles
     /*
--- a/source/nodejs/adaptivecards-templating/src/expression-parser.ts
+++ b/source/nodejs/adaptivecards-templating/src/expression-parser.ts
@@ -145,7 +145,49 @@ function ensureValueType(value: any): LiteralValue {
 }
 
 type FunctionCallback = (...params: any[]) => any;
-type FunctionDictionary = { [key: string]: FunctionCallback };
+
+interface FunctionDefinition {
+    name: string;
+    callback: FunctionCallback;
+}
+
+class FunctionCollection {
+    private _functions: FunctionDefinition[] = [];
+
+    indexOf(name: string): number {
+        return this._functions.findIndex((f: FunctionDefinition) => { return f.name === name; });
+    }
+
+    register(name: string, callback: FunctionCallback) {
+        let index = this.indexOf(name);
+
+        if (index >= 0) {
+            this._functions[index].callback = callback;
+        }
+        else {
+            this._functions.push(
+                {
+                    name: name,
+                    callback: callback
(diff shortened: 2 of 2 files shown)


source/nodejs/adaptivecards-designer-app/src/app.ts
```diff
@@ -12,13 +12,11 @@ import "./app.css";
 
 window.onload = function() {
     // Uncomment to enabled preview features such as data binding
-    /*
```
shalinijoshi19: Are we uncommenting this to expose the databinding feature by default now?

shalinijoshi19:
@dclaux, @matthidinger, guys, question: how/when are we expecting to spin up a package update for this and also should we update the version up on vnext.adaptivecards.io for bug bash purposes ?  Also David, can you actually include the sample update that we can use both for the bash as well as to update our documentation site if needed?

source/nodejs/adaptivecards-designer-app/src/app.ts
dclaux: Yes. Also keep in mind that adaptivecards-designer-app is basically a test app.

matthidinger:
@shalinijoshi19 I'd love to try and get an update out tomorrow if the fix is confirmed working in the site. If not we might need a few more days into next week, but we should definitely sync on the overall status of the latest site deployments and close those ends. The Project board has the 4 items I'm hoping to close for the next publish

PR #1377, opened 2018-04-14
[UWP] Add height property to element properties

almedina-ms:
Adds the height property so elements may stretch over an empty space in the card height. As we don't have a property to define. 

Issue: https://github.com/Microsoft/AdaptiveCards/issues/1295

PR #3820, opened 2020-03-05
[UWP][Accessability] Use element tag instead of access key for internal tracking

### LINKED ISSUE #3819: [UWP][Accessibility] Unexpected access keys appearing for select actions
# Platform

* UWP

# Details

When alt is pressed on a card containing a select action on a container with stretch height, an access key of '0' appears. This is unexpected, can lead to many select actions with the same access key, and can interfere with access keys set by the host app.

Example:
The below card renders as follows:
image

When the user presses 'Alt' a '0' access key appears
image

```json
{
	"type": "AdaptiveCard",
	"version": "1.2",
	"body": [
		{
			"type": "Container",
			"items": [
				{
					"type": "TextBlock",
					"text": "This is a select action in a container with stretch"
				}
			],
			"selectAction": {
				"type": "Action.Submit",
				"title": "Submit action"
			},
			"height": "stretch"
		}
	]
}
```

RebeccaAnne:
## Related Issue
Fixes #3819

## Description
The WholeItemsPanel keeps track of which items have been marked as `"height":"stretch"` in order to render them properly. In order to do this, the code was stashing an identifier in the `AccessKey` property of the xaml elements (#1377). Unfortunately, that was causing access keys to show up in the UI when the user presses 'Alt'.

image

Updated the code to use the `Tag` property to handle this internal tracking, as that is a place for custom data that won't have a UI impact. Because we're already using the Tag property for a variety of purposes, added the `IsStretchable` property to the existing `ElementTagContent` type.

## How Verified
Confirmed that the card from bug #3819 no longer generates an access key. Verified existing VerticalStretch cards still render correctly.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3820)

### DIFF
--- a/source/uwp/Renderer/lib/WholeItemsPanel.cpp
+++ b/source/uwp/Renderer/lib/WholeItemsPanel.cpp
@@ -408,16 +408,18 @@ namespace AdaptiveNamespace
         ComPtr<IElementTagContent> elementTagContent;
         if (tagAsInspectable != nullptr)
         {
-            tagAsInspectable.As(&elementTagContent);
-            elementTagContent->put_IsStretchable(true);
+            THROW_IF_FAILED(tagAsInspectable.As(&elementTagContent));
+            THROW_IF_FAILED(elementTagContent->put_IsStretchable(true));
         }
         else
         {
             ComPtr<IElementTagContent> tagContent;
-            THROW_IF_FAILED(MakeAndInitialize<ElementTagContent>(&tagContent, nullptr, nullptr, nullptr, nullptr, false, true));
+            THROW_IF_FAILED(MakeAndInitialize<ElementTagContent>(&tagContent));
 
-            tagContent.As(&tagAsInspectable);
-            elementAsFrameworkElement->put_Tag(tagAsInspectable.Get());
+            THROW_IF_FAILED(tagContent->put_IsStretchable(true));
+
+            THROW_IF_FAILED(tagContent.As(&tagAsInspectable));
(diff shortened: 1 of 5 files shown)


source/uwp/Renderer/lib/WholeItemsPanel.cpp
```diff
@@ -399,35 +399,48 @@ namespace AdaptiveNamespace
...
         {
-            std::string elementAccessKey = std::to_string(m_accessKeyCount);
-            ++m_accessKeyCount;
+            tagAsInspectable.As(&elementTagContent);
+            elementTagContent->put_IsStretchable(true);
+        }
+        else
+        {
+            ComPtr<IElementTagContent> tagContent;
+            THROW_IF_FAILED(MakeAndInitialize<ElementTagContent>(&tagContent, nullptr, nullptr, nullptr, nullptr, false, true));
```
almedina-ms: what are all these nullptr and booleans? Could we get a comment to make it more descriptive?

RebeccaAnne: They're non-relevant constructor parameters. But, this is silly. Added a constructor that doesn't take all this to allow callers to construct a default tag content object and make the code a little clearer.

---
In reply to: [389054235](https://github.com/microsoft/AdaptiveCards/pull/3820#discussion_r389054235) [](ancestors = 389054235)

Issue #3688, opened 2019-12-19
[Javascript][Not able to render a TextBlock with no data & $when condition]


atishayv:
# Platform

* JavaScript

# Author or host

Host Name : Microsoft Search

# Version of SDK

adaptivecards-designer@0.7.1

# Details

Not able to render a TextBlock element if I give a $when condition with no data : 
```
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "TextBlock",
            "text": "{title}",
            "$when": "{title != \"\"}"
        }
    ],
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
   "$data": {}
}
```

Getting this error : 
`Template expansion error: Invalid value type: undefined`

Not sure if its related with this bug : 
https://github.com/microsoft/AdaptiveCards/issues/3523

shalinijoshi19:
@dclaux / @matthidinger ideas on this one?

dclaux:
@atishayv please try `"$when": "{title != ''}"`

atishayv:
@dclaux Tried with 
```
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "TextBlock",
            "text": "{title}",
            "$when": "{title != ''}"
        }
    ],
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
     "$data": {}
}
```

Still getting the same error

dclaux:
I think we have a limitation there is currently no solution for here. I think `title != ''` evaluates to true because `title` is actually `undefined`, but there is no way currently to check for undefined.

This is not something I'm going to be able to fix quickly unfortunately. @matthidinger this is something we might need to discuss, at least look into whether CEL has a solution for it.

PR #3823, opened 2020-03-05
[Template] Don't drop card if $when fails to eval

dclaux:
## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/3688

### $when clause handling change
Currently, two things happen:
- When the `$when` clause fails to evaluate, an exception is thrown which causes the whole card to not render
- When the `$when` clause evaluates to a non-boolean value, it is considered to be `true`

With this PR, the behavior becomes:
- When the `$when` clause fails to evaluate, it is considered to be `false`
- When the `$when` clause evaluates to a non-boolean value, it is considered to be `false`

In turn this means that an invalid $when clause will now cause the element on which it is defined to be dropped, but the card will still render.

### Handling of templatized strings containing expressions that evaluate to undefined
Currently, when an expression refers to a field that isn't defined (doesn't have a value) in the data context, it often evaluates to `undefined` and the string "undefined" gets emitted in the output card, and eventually displayed on screen That behavior is less than optimal.

With this PR:
- Such expressions will still evaluate to `undefined` but the display result (e.g. what ends up in the card) will be the original expression string. For instance "{title}"
- Applications can override that behavior by providing their own default string. For instance:

```typescript
ACTemplating.GlobalSettings.undefinedExpressionValueSubstitutionString = "<undefined value>";
```
 
## How Verified

### For $when clause handling
```json
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "TextBlock",
            "text": "{title}",
            "$when": "{title!=\"\"}"
        }
    ],
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json"
}

```

### For expressions evaluating to undefined
```json
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "type": "TextBlock",
            "text": "[{title}]({titleUrl})",
            "color": "Accent",
            "size": "Medium",
            "weight": "Bolder"
        }
    ],
    "$data": {
        "titleUrl": "https://modernacdesigner.azurewebsites.net"
    }
}
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3823)

### DIFF
--- a/source/nodejs/adaptivecards-templating/src/template-engine.ts
+++ b/source/nodejs/adaptivecards-templating/src/template-engine.ts
@@ -205,11 +205,22 @@ export class Template {
             let when = node["$when"];
 
             if (when instanceof TemplatizedString) {
-                let whenValue = when.evaluate(this._context);
+                let whenValue: any;
+                
+                try {
+                    whenValue = when.evaluate(this._context);
 
-                if (typeof whenValue === "boolean") {
-                    dropObject = !whenValue;
+                    // If $when doesn't evaluate to a boolean, consider it is false
+                    if (typeof whenValue !== "boolean") {
+                        whenValue = false;
+                    }
                 }
+                catch {
+                    // If we hit an exception, consider $when to be false
+                    whenValue = false;
+                }
+
+                dropObject = !whenValue;
             }
 
             if (!dropObject) {
--- a/source/nodejs/adaptivecards-templating/src/template-engine.ts
+++ b/source/nodejs/adaptivecards-templating/src/template-engine.ts
@@ -111,7 +111,7 @@ class TemplatizedString {
                     let evaluatedPart = this.evalExpression(<Binding>part, context);
 
                     if (evaluatedPart === undefined) {
-                        evaluatedPart = Shared.GlobalSettings.undefinedExpressionValueSubstitutionString ?? (<Binding>part).expressionString;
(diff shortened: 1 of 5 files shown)


source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -94,14 +97,24 @@ class TemplatizedString {
...
             for (let part of this._parts) {
                 if (typeof part === "string") {
                     s += part;
                 }
                 else {
-                    s += this.evalExpression(<Binding>part, context);
+                    let evaluatedPart = this.evalExpression(<Binding>part, context);
+
+                    if (evaluatedPart === undefined) {
+                        evaluatedPart = Shared.GlobalSettings.undefinedExpressionValueSubstitutionString ? Shared.GlobalSettings.undefinedExpressionValueSubstitutionString : (<Binding>part).expressionString;
```
matthidinger: Minor change, but it seems this could be shortened to the following?
```
evaluatedPart =  Shared.GlobalSettings.undefinedExpressionValueSubstitutionString ?? (<Binding>part).expressionString;

dclaux: Indeed, changed.

Issue #2844, opened 2019-05-03
[UWP Visualizer] Trying to save an Untitled card gives a NullReferenceException


shalinijoshi19:
# Platform

What platform is your issue or question related to? (Delete other platforms).
* UWP

# Version of SDK
master; Latest store app (published nightly at 7pm per Andrew)

# Issue

1. Launch UWP Visualizer app
2. Create an untitled JSON
3. Hit Save 
Notice the nullereference error :

![image](https://user-images.githubusercontent.com/20822397/57166408-f265b680-6dae-11e9-803d-c2d6a69f2b2b.png)

shalinijoshi19:
Dont have VS intalled on my laptop for debugging details; I'll get that going in a bit and get back

PR #3825, opened 2020-03-07
[UWP] Add ESC support to MessageDialog

almedina-ms:
## Related Issue
Fixes: #3826
Fixes: #3881

## Description
Adds support for ESC character to get out of the exception MessageDialog

## How Verified
The Visualizer app was executed where I created a new file and then tried to save it so a messageDialog would pop up with an exception. This message dialog was where the issue was located. From this point on, OnScreen Keyboard and narrator were turned on so the correct behaviour could be tested.

As can be seen in the gif below, the ESC key is pressed and the Dialog is dismissed.

bugFix

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3825)


shalinijoshi19:
Nit: Can you be slightly more explicit about "what you executed" in the How Verified section?

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
```diff
@@ -117,15 +120,58 @@ protected async Task LoadFromFileAsync(IStorageFile file, string token, bool ass
             }
         }
 
+        private async Task CreateMessageDialog(string text)
+        {
+            var dontWait = new MessageDialog(text);
```
RebeccaAnne: >dontWait [](start = 16, length = 8)

Why is this called dontWait?

shalinijoshi19:
Is the bug relevant only in the case where (like today) hitting "Save" results in an exception dialog? Does it work also once we fix the exceptions (which we should bundle in with this fix if it's containable and not big of a churn) in the Save workflow?

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
```diff
@@ -117,15 +120,58 @@ protected async Task LoadFromFileAsync(IStorageFile file, string token, bool ass
...
+                    await CreateMessageDialog("File " + File.Name + " was saved");
+                }
+                else
+                {
+                    await CreateMessageDialog("File " + File.Name + " couldn't be saved");
+                }
+            }
+            else
+            {
+                await CreateMessageDialog("Save new file was cancelled");
```
RebeccaAnne: >CreateMessageDialog [](start = 22, length = 19)

Do we need another dialog to pop up in this case? Generally when i click cancel i expect to go back to the app, not see another dialog. Similarly for the success case, i'm not sure we need a dialog there either. Maybe just for the error case? What do you think?

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
almedina-ms: that was the original name, I thought there was a reason behind so I just kept the name. I'll change it to messageDialog


---
In reply to: [389910458](https://github.com/microsoft/AdaptiveCards/pull/3825#discussion_r389910458) [](ancestors = 389910458)

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
```diff
@@ -117,15 +120,58 @@ protected async Task LoadFromFileAsync(IStorageFile file, string token, bool ass
             }
         }
 
+        private async Task CreateMessageDialog(string text)
+        {
+            var dontWait = new MessageDialog(text);
+            dontWait.CancelCommandIndex = 0;
```
RebeccaAnne: >dontWait.CancelCommandIndex = 0; [](start = 12, length = 32)

Could you add a comment here explaining what this line does?

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
RebeccaAnne: Ah, i see. Looks like before we were not waiting on the async call, and the dontwait variable would have been the async action we were explicitly not waiting on. Now it's a message dialog, so changing the name makes sense. 

But now I'm wondering: why the change from not waiting to waiting?

---
In reply to: [389913301](https://github.com/microsoft/AdaptiveCards/pull/3825#discussion_r389913301) [](ancestors = 389913301,389910458)

shalinijoshi19: >dontWait [](start = 16, length = 8)

Variable name: this could be better now that we are moving hte old code around :) - how about escMessageDialog or just even messageDialog?

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
```diff
@@ -117,15 +120,58 @@ protected async Task LoadFromFileAsync(IStorageFile file, string token, bool ass
...
             }
-            catch (Exception ex)
+            else
             {
-                var dontWait = new MessageDialog(ex.ToString()).ShowAsync();
+                try
+                {
+                    await FileIO.WriteTextAsync(File, Payload);
+                }
+                catch (Exception ex)
```
shalinijoshi19: We should figure out why we are getting exceptions in the "Save" workflow to begin with here (one of them is issue #2844)

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
almedina-ms: There's no regular Show method, there's only the ShowAsync method

---
In reply to: [389916462](https://github.com/microsoft/AdaptiveCards/pull/3825#discussion_r389916462) [](ancestors = 389916462,389913301,389910458)

RebeccaAnne: Right, but previously we didn't "await" the result from ShowAsync. In the old code, we put the IAsyncOperation in a variable called dontWait, and didn't "await" it. In the new code we await the IAsyncOperation.

---
In reply to: [389920998](https://github.com/microsoft/AdaptiveCards/pull/3825#discussion_r389920998) [](ancestors = 389920998,389916462,389913301,389910458)

almedina-ms:
Any dialog exception in the "Save" workflow should be fixed with this. The actuals exceptions are not. Unauthorized access exception could not be reproduced.

---
In reply to: [596736022](https://github.com/microsoft/AdaptiveCards/pull/3825#issuecomment-596736022) [](ancestors = 596736022)

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
almedina-ms: I didn't know that you could capture an IAsyncOperation with a variable. changed it to not wait again

---
In reply to: [389924506](https://github.com/microsoft/AdaptiveCards/pull/3825#discussion_r389924506) [](ancestors = 389924506,389920998,389916462,389913301,389910458)

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
```diff
@@ -117,15 +120,58 @@ protected async Task LoadFromFileAsync(IStorageFile file, string token, bool ass
...
         }
 
+        private async Task CreateMessageDialog(string text)
+        {
+            var dontWait = new MessageDialog(text);
+            dontWait.CancelCommandIndex = 0;
+            await dontWait.ShowAsync();
+        }
+
+        private async Task SaveNewFile()
```
shalinijoshi19: Nit: This should be named to reflect its non-blocking implementation - ie SaveNewFileAsync

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
almedina-ms: I couldn't repro the System.UnauthorizedAccessException. Have you tried building the app in your local machine and check if the same exception pop ups. Otherwise we should check if it's an issue with the app in the store

---
In reply to: [389920350](https://github.com/microsoft/AdaptiveCards/pull/3825#discussion_r389920350) [](ancestors = 389920350)

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
almedina-ms: I can agree with the cancel dialog not showing. On the case of saving success and fail I wonder if not getting feedback on what's happening may be the best scenario. I could remove everything and if we receive a new bug I can re add them

---
In reply to: [389912842](https://github.com/microsoft/AdaptiveCards/pull/3825#discussion_r389912842) [](ancestors = 389912842)

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
```diff
@@ -117,15 +120,58 @@ protected async Task LoadFromFileAsync(IStorageFile file, string token, bool ass
...
+            }
+        }
+
         public async Task SaveAsync()
         {
-            try
+            if (File == null)
             {
-                await FileIO.WriteTextAsync(File, Payload);
+                await SaveNewFile();
```
shalinijoshi19: If this fails (the case when you click on Save without hitting the new file workflow) you'll continue to see the default messagebox with only the Close button and no Esc/cancel handler i think since it's not inside of a catch block right?

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
```diff
@@ -117,15 +120,55 @@ protected async Task LoadFromFileAsync(IStorageFile file, string token, bool ass
...
+            }
+        }
+
         public async Task SaveAsync()
         {
-            try
+            if (File == null)
             {
-                await FileIO.WriteTextAsync(File, Payload);
+                await SaveNewFileAsync();
```
shalinijoshi19: What happens if this fails?

source/uwp/Visualizer/ViewModel/GenericDocumentViewModel.cs
almedina-ms: I could repro the issue now and I think it was a big oversight by us
https://social.msdn.microsoft.com/Forums/en-US/a32d16fd-7855-4fce-9091-8067f712a1b6/access-denied-on-windowsstoragefileiowritetextasync-to-write-a-text-file-in-a-local-application?forum=winappswithhtml5

It looks like we can't save on the app folder, during the build process we copy some files to the app folder (i.e. AppX\Samples\Agenda.json), this means that all cards inside the app folder can we overwritten as we don't have permission by UWP's design. I'm not sure what scenarios we use the Save function for as any cards loaded from another location or newly created won't have a token and they are not loaded on app startup

almedina-ms:
I ran the android job manually and it built correctly

PR #3837, opened 2020-03-11
[UWP] Set runtime class name on ElementTagContent

### LINKED ISSUE #3836: [UWP] Assert when using LivePropertyViewer
# Platform

* UWP

# Details

When viewing elements in Visual Studio's Live Visual Tree, an Assert fires, and properties are not shown.

Repro steps:
1. Open [TextBlock.json](https://github.com/microsoft/AdaptiveCards/blob/master/samples/v1.0/Elements/TextBlock.json) in the UWP Visualiser under Visual Studio.
2. Select "Select Element" from the toolbar:
image
3. Select one of the TextBlocks in the card
4. Go to the Live Visual Tree in Visual Studio and right click on the selected TextBlock and choose "Show Properties"

Result:
image

Continuing past the assert the property window is empty.

RebeccaAnne:
## Related Issue
Fixes #3836 

## Description
ElementTagContent didn't have it's runtime class name set, causing Visual Studio's Live Property Tree viewer to Assert and fail to show element properties. Set the runtime class name using the InspectableClass macro via our AdaptiveRuntimeStringClass macro.

## How Verified
Verified Live Property Tree now working as expected.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3837)


shalinijoshi19:
Great catch! Isn't this class used anywhere for this to have been caught earlier? Also is this one a recent addition to our interface?

RebeccaAnne:
This class is used to associate various content (such as the card element) with the xaml objects. It's used on most of our xaml objects, and functionally for what we've been using it for it works fine without this declaration. For whatever reason, the only place we're hitting this assert is via the usage of the live visual tree. We've had this class around for a while (since May of last year), but as this seems to be the only error caused by this bug, and because the cause and effect wasn't immediately obvious, we didn't find this until i really needed the Live Visual Tree to work to debug something else.

---
In reply to: [597902793](https://github.com/microsoft/AdaptiveCards/pull/3837#issuecomment-597902793) [](ancestors = 597902793)

PR #3839, opened 2020-03-11
[UWP] Change visibility of Close button to always visible

### LINKED ISSUE #3838: [UWP][Accessibility] 'Close' control is not accessible through keyboard at left panel.
# Platform

* UWP

# Version of SDK

master

# Details

Repro Steps:
1. Launch the application.
2. New screen starts appearing. 
3. Navigate to various json file like Activity Update.JSON, Agenda.JSON etc. at left panel.
4. Navigate to 'Close' control for each json file.
5. Verify 'Close' control by using keyboard.

Actual Result: 
'Close' control is not accessible through keyboard appearing over each json file like Agenda.JSON, Agenda.JSON at left panel which will impact the functionality of relevant control. Not only this, user won't be able to perform functionality over such control.

Expected Result: 
'Close' control should be accessible through keyboard appearing over each json file like Agenda.JSON, Agenda.JSON at left panel so that end user could easily perform the relevant functionality.

almedina-ms:
## Related Issue
Fixes #3838

## Description
Changes the visibility of the Close button to be always visible so it can be accessed through keyboard navigation.

## Why the change?
It was deemed not possible (at this time) to access the children views for a ListView in UWP. This access was necessary as it would have allowed changing the visibility of the button with the "OnSelectionChanged" event. More changes would have been necessary to use this approach so this change be sufficient in the meantime.

## How Verified
The visualizer app was executed and the behaviour can be checked as the Close button is now selectable

bugFix4

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3839)


source/uwp/Visualizer/TabView.xaml.cs
```diff
@@ -42,12 +42,10 @@ private void ButtonClose_Click(object sender, RoutedEventArgs e)
 
         private void OnPointerEntered()
```
paulcam206, line 43: can we remove `OnPointerEntered()` and `OnPointerExited()` altogether?

Issue #3842, opened 2020-03-13
[UWP][Accessibility] Luminosity contrast ratio is less than 3:1 for outlines under 'FlightItinerary.JSON' list item


almedina-ms:
# Platform

* UWP

# Version of SDK

master

# Issue

Repro Steps:​

1. Launch the application.​

2. New screen starts appearing. Navigate to Left pane JSON files.
3. Now select "FlightItinerary.JSON" button. It will update the Middle and Right Pane. Navigate all the elements of both the panes.

4. Open Color contrast Analyzer tool.
5. Select foreground and background color of outlines using color picker control of tool.
6. Verify the result of Color contrast Analyzer tool.

Actual Result:

Luminosity contrast ratio is less than 3:1 for outlines under FlightItinerary.JSON list item.

Expected Result:

Luminosity ratio should be greater than or equal to 3:1 for outlines.

PR #3843, opened 2020-03-13
[UWP] Fix contrast issues

almedina-ms:
## Related Issue
Fixes: #3842 

## Description
Values were changed for some of the styles and host config so gray lines would have 3:1 contrast ratio.

## How Verified
The visualizer app was run along with the Colour Contrast application and all inputs now have at least 3:1 contrast ratio

image

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3843)


source/uwp/Visualizer/ViewModel/DocumentViewModel.cs
```diff
@@ -238,6 +238,22 @@ public static void InitializeRenderer(AdaptiveHostConfig hostConfig)
             // Custom resource resolvers
             _renderer.ResourceResolvers.Set("symbol", new MySymbolResourceResolver());
 
+            _renderer.OverrideStyles = new ResourceDictionary();
+
+            Style textInputBorderStyle = new Style(typeof(TextBox));
```
shalinijoshi19, line 243: what is this change about again?

almedina-ms: That is a change in the visualizer to darken the border of input elements

Issue #3866, opened 2020-03-23
[JS][Templating] Move template engine to AdaptiveExpression


dclaux:
# Platform

* JavaScript

# Details

Update the JS template engine to use [adaptive-expressions](https://www.npmjs.com/package/adaptive-expressions) instead of the built-in expression parser/evaluator.

shalinijoshi19:
Closing as dup of #3901

PR #3867, opened 2020-03-23
[JS][Templating] Move to AdaptiveExpressions

dclaux:
## NOTE
This PR is obsolete. The changes have been moved to this other PR: https://github.com/microsoft/AdaptiveCards/pull/3508

## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/3866

## Description
- Drops the built-in expression parser/evaluator
- Makes use of the [adaptive-expressions](https://www.npmjs.com/package/adaptive-expressions) library instead
- As a result, the start marker for an expression is now `${` as opposed to just `{` previously

## Notes
- The `${?#expression}` (note the ?# characters) syntax originally planned is not supported anymore

## How to build
This PR is currently dependent on daily builds of the adaptive-expressions library. These builds are sourced from https://botbuilder.myget.org/F/botbuilder-v4-js-daily/npm/

To build, you first have to update your npm registry:
`npm config set registry https://botbuilder.myget.org/F/botbuilder-v4-js-daily/npm/`

Then run `npm install`

**NOTE**: The JS build checks **will not pass** while the code uses the daily builds.
**NOTE**: To revert your npm registry to the default run:
`npm config set registry https://registry.npmjs.org/`

## How Verified
Verified manually in adaptivecards-designer-app using our existing set of templates

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3867)


paulcam206:
Code changes themselves look good :)

dclaux:
@paulcam206 this will not be merged until the adaptive-expressions 4.9 is pushed to the public npm registry.

dclaux:
Plus I have other changes to make.

shalinijoshi19:
Could you add a note under the "How Verified" section of hte description on how you are verifying some of hte changes? Perhaps running existing sample cards we have and ensuring they continue to render? Add new cards where needed to cover the integration scenarios with AE(AdaptiveExpressions) , existing partner scenarios etc

source/nodejs/adaptivecards-templating/package-lock.json
```diff
@@ -1,9 +1,14 @@
...
 	"name": "adaptivecards-templating",
-	"version": "0.1.0-alpha.1",
+	"version": "0.1.1-alpha.1",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
+		"@microsoft/recognizers-text-data-types-timex-expression": {
+			"version": "1.3.0",
+			"resolved": "https://botbuilder.myget.org/F/botbuilder-v4-js-daily/npm/@microsoft/recognizers-text-data-types-timex-expression/-/@microsoft/recognizers-text-data-types-timex-expression-1.3.0.tgz",
+			"integrity": "sha1-/F1YbIJuR46Ed7f8sh6eKDDoHGc="
```
shalinijoshi19: How will the code change, depending on our discussion outcome on how we want to ship this (whether update to the templating package or update to the rendering package)? Trying to understand whether this should go in as is or if we should reissue the PR once that decision is made (while continuing to iterate on this branch/PR with actual integration changes) ..

source/nodejs/adaptivecards-templating/src/shared.ts
```diff
@@ -2,4 +2,6 @@
 // Licensed under the MIT License.
 export class GlobalSettings {
     static undefinedExpressionValueSubstitutionString?: string = undefined;
+    static expressionStartMarker = "${";
```
shalinijoshi19: >${"; [](start = 36, length = 4)

This would need to be updated in the documentation/samples ? If so can you do me a favor, update your issue with a tasklist style todo in there (- [ ] is the syntax) and capture that in there for starters (unsure yet if we need a completely separate issue for that for now)?

source/nodejs/adaptivecards-templating/src/shared.ts
```diff
@@ -2,4 +2,6 @@
...
+    static expressionEndMarker = "}";
```
shalinijoshi19: @nesalang is your design spec updated with this new marker syntax ? Also i thought we could use the antlr based parser for JS also - no? is that a discussion item ?

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -1,142 +1,105 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { Binding, ExpressionParser, EvaluationContext } from "./expression-parser";
-import * as Shared from "./shared";
+import * as CEL from "adaptive-expressions";
```
shalinijoshi19: >CEL [](start = 12, length = 3)

Nit: Should probably rename this alias/variable at this point since CEL is deprecated naming for adaptiveExpressions

source/nodejs/adaptivecards-templating/package-lock.json
dclaux: Either the code will remain in the adaptivecards-templating package, or it will be moved to the adaptivecards package. If the latter I will have to do that work.

source/nodejs/adaptivecards-templating/src/shared.ts
dclaux: None of the flags in GlobalSettings are used at this time. If these end up staying then I'll make sure the capabilities are documented.

source/nodejs/adaptivecards-templating/src/shared.ts
dclaux: I don't have a design spec. I am literally just porting the existing template engine, which is a very small piece of code which isn't done using Antlr. The Expression parser/evaluator, e.g. Adaptive Expressions, which the templating code utilizes, is based on Antlr.

nesalang: the configurable start and end tokens for templated string is captured in the project board. I think we discussed it, and decided during the last wednesday (3/25) work group meeting that it will be JS only feature to help current JS libary user migrate to new version of JS SDK. We don't have plan for JS to use antlr for parsing templated adaptive cards. .Net Antlr parsers/lexer can be used for ios/android/uwp, and I believe that is the current plan.

---
In reply to: [398591949](https://github.com/microsoft/AdaptiveCards/pull/3867#discussion_r398591949) [](ancestors = 398591949)

nesalang:
since code is not complete. I will wait until code is done :-)

---
In reply to: [380622246](https://github.com/microsoft/AdaptiveCards/pull/3867#issuecomment-380622246) [](ancestors = 380622246)

dclaux:
This PR is obsolete. The changes have been moved to this PR: https://github.com/microsoft/AdaptiveCards/pull/3508

PR #3871, opened 2020-03-25
[JS] Initial commit of aclint tool

paulcam206:
## Description
This is the initial version of a tool that can be extended to perform linting operations on Adaptive Cards. Right now it doesn't have any custom rules -- it only verifies that cards parse correctly and prints warnings.

Here's what it looks like when run (for demonstration purposes, I created some test files):
image
(alt: screenshot of commandline tool being used)

Next steps include adding a custom set of linting rules to run (beyond things surfaced as warnings). 

## Try it out!
1. Build `nodejs` directory like you normally would (see [README](https://github.com/microsoft/AdaptiveCards/blob/master/source/nodejs/README.md)).
2. In `source/nodejs/aclint` directory, run `npm i -g .`
3. `aclint` command should be available for use

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3871)


source/nodejs/aclint/tslint.json
```diff
@@ -0,0 +1,75 @@
+{
```
paulcam206: This should probably not be part of this project -- need to finalize move to eslint

source/nodejs/aclint/src/index.ts
```diff
@@ -0,0 +1,236 @@
...
+    });
+}
+
+async function main()
+{
+    return testItemsInArgv();
+}
+
+main().then((results) => {
+    if (results.length === 0) {
```
paulcam206: need to sprinkle in some comments

matthidinger:
That's really neat! So presumably this could be used to verify the Sample scenario/element payloads as part of our normal builds? What other processes could we streamline with this?! :)

paulcam206:
> That's really neat! So presumably this could be used to verify the Sample scenario/element payloads as part of our normal builds? What other processes could we streamline with this?! :)

Yep -- we can validate our samples with this tool, as well as add "best practices" rules. This started with an email thread where @dclaux pointed out that using `"width": "auto"` on two columns in a columnset is usually not what a user wants. We'll be able to add more rules as we come up with them. Ideally at some point these rules will be made available in the designer...

paulcam206:
I should also note that this is my first from-scratch node project, so please be as nitpicky as possible around best practices, etc.

source/nodejs/aclint/.vscode/launch.json
```diff
@@ -0,0 +1,36 @@
...
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "program": "${workspaceFolder}\\lib\\src\\index.js",
+            "args": ["c:\\ac\\samples\\v1.0\\Scenarios"],
+            "preLaunchTask": "tsc: build - tsconfig.json",
+            "outFiles": [
+                "${workspaceFolder}/lib/**/*.js"
+            ]
+		},
```
nesalang: nit: tab here sorry!

nesalang:
i think you should have a session showing how to add a custom rule. this would really help close the gap between JS renderer and Mobile.

---
In reply to: [603638038](https://github.com/microsoft/AdaptiveCards/pull/3871#issuecomment-603638038) [](ancestors = 603638038)

paulcam206:
> @paul campbell can you create a quick learning session showing how to add custom rules? this would really help close the gap between JS renderer and Mobile.
> 
> In reply to: [603638038](https://github.com/microsoft/AdaptiveCards/pull/3871#issuecomment-603638038) [](ancestors = 603638038)

Codeflow doesn't work for me, so I have to comment here -- this tool doesn't yet support adding custom rules. That was going to be in a follow-on PR :)

paulcam206:
fixed up whitespace in launch.json :)

paulcam206:
ping! :)

paulcam206:
> Code itself looks good as far as I can tell.
> 
> The one thing to me is that I think it'd be better to author this as an API that can be used indifferently by a command line tool and a web site. As it is now, and as far as I can tell, it is dependent on being invoke from the command line and on NodeJS only APIs to browse the file system.
> 
> If instead things were decoupled and there was a simple API that, given a payload, returns hints, you could use it in a NodeJS command line tool and I could use it in the designer.

Definitely agree on separating out a linting library. For now, since we don't have custom rules support yet, I figured it was less important. :)

PR #3893, opened 2020-03-30
[JS] Add doc generation support

paulcam206:
## Description
Recently I did a little side project standing up the scaffolding for an Adaptive Cards linting tool. In writing the tool, I found that our docs are sorely lacking at the API level (this isn't unique to our Typescript codebase -- we should visit each platform in turn). Thankfully, for Typescript projects, there's a solution: [TypeDoc](https://typedoc.org/). TypeDoc uses Typescript's type information in combination with Typescript's flavor of [JSDoc](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#supported-jsdoc) comments to generate API docs. By default, the output is static HTML, but through a plugin we're able to get markdown output.

The [initial commit](https://github.com/microsoft/AdaptiveCards/pull/3893/commits/a61c5fd56bd6a9d95f319d9a9d73738aca111c6e) here has all of the relevant configuration and setup (this is the part that really needs review), and the [second commit](https://github.com/microsoft/AdaptiveCards/pull/3893/commits/6a70a4ae37f13dd8e241ec9a5e88b1747e5df91d) contains the output. A couple of notes:

* Because of the way our source is structured, the shipping version of TypeDoc doesn't generate great structure (it's assumed that you have one module per file *or* that each file is an independent collection of code). `typedoc@next` has support for `library` output, which actually merges everything across files. This is a much better structure, even if we have to depend on prerelease software for now.
* There are a lot of [options](https://typedoc.org/guides/options/) that can be tweaked. I chose what I believe to the be the right set of defaults
* For now we don't have any CI provisions for doc generation. We need to decide how we want to distribute the docs. We can either continue to have them checked in to our repo and point our official docs to them (easy but perhaps a bit odd, though it may endear us to some devs), *or* we can consume them into our website (harder, but more official-looking).
* The docs are generated based on type information, but are greatly aided by additional comments. Bonus points: markdown is supported in comments. For a decent example, check out `getEnumValue`: https://github.com/microsoft/AdaptiveCards/blob/6a70a4ae37f13dd8e241ec9a5e88b1747e5df91d/source/nodejs/adaptivecards/src/utils.ts#L66-L82 and the resulting [output](https://github.com/microsoft/AdaptiveCards/blob/paulcam/ac-docs-infra/source/nodejs/adaptivecards/docs/README.md#getenumvalue).
* The easiest way to see the output is probably to just navigate to the first run of docs [here](https://github.com/microsoft/AdaptiveCards/tree/paulcam/ac-docs-infra/source/nodejs/adaptivecards/docs).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3893)

### DIFF
--- a/source/nodejs/adaptivecards/docs/README.md
+++ b/source/nodejs/adaptivecards/docs/README.md
@@ -1,0 +1,556 @@
+[Adaptive Cards Javascript SDK](README.md)
+
+# Adaptive Cards Javascript SDK
+
+## Index
+
+### Enumerations
+
+* [ActionAlignment](enums/actionalignment.md)
+* [ActionIconPlacement](enums/actioniconplacement.md)
+* [ContainerFitStatus](enums/containerfitstatus.md)
+* [FillMode](enums/fillmode.md)
+* [FontType](enums/fonttype.md)
+* [HorizontalAlignment](enums/horizontalalignment.md)
+* [ImageStyle](enums/imagestyle.md)
+* [InputTextStyle](enums/inputtextstyle.md)
+* [InputValidationNecessity](enums/inputvalidationnecessity.md)
+* [Orientation](enums/orientation.md)
+* [ShowCardActionMode](enums/showcardactionmode.md)
+* [Size](enums/size.md)
+* [SizeUnit](enums/sizeunit.md)
+* [Spacing](enums/spacing.md)
+* [TextColor](enums/textcolor.md)
+* [TextSize](enums/textsize.md)
+* [TextWeight](enums/textweight.md)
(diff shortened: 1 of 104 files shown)


shalinijoshi19:
This is great stuff @paulcam206 ! Thanks for doing this. I quite like putting these up and checking them in to our docs repo, to supplement our official microsoftdocs content..

shalinijoshi19:
Also do we need to put up documentation best practices/guides as well checked into our source repo as next steps or does this already include it?

PR #3897, opened 2020-04-01
Update some open discussion and comments on Input Validation Spec

almedina-ms:
## How to approach this document

The [InputLabels](https://github.com/microsoft/AdaptiveCards/pull/3897/files#diff-114ac7f9164cd2eef456300b74c45aea) spec and the [InputValidation ](https://github.com/microsoft/AdaptiveCards/pull/3897/files#diff-cae2d7f5c79c829c005f09319a97bde2) spec included in this PR details our current agreed upon plan of action to land a basic Labels + Basic Input validation feature set for v1.3 of the renderers. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3897)


almedina-ms:
Here's the list of open discussions from the original PR: [https://github.com/microsoft/AdaptiveCards/pull/3663](https://github.com/microsoft/AdaptiveCards/pull/3663)

almedina-ms:
Further discussion of labels today:

All agreed that P0 base behavior is a plain text label on the input, required inputs marked by a *.
Next most important thing is allowing host control. Two dimensions: styling, and required indicator string and position
David feels strongly that all styling should go through CSS (or equivalent) and should not be in host config.
Matt and i feel some base host config control may be useful. Discussion of underlying use and usefulness of host config in general
All agreed that native styling should be possible for required vs optional.
On string (*) configuration, we could use the common expression language. David is concerned about taking a dependency on it for size reasons.
If host can configure the string, they need to be able to localize it.
David thinks that allowing hosts to change the string (without loc) should be high priority.

almedina-ms:
[https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367153579](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367153579)

> @RebeccaAnne

Discussion: What is the scenario for "None"? Should we have an explicit Action.Cancel instead?

Fire company has an Action.Cancel in their stuff, but what are the semantics?
Non-submit action types default to None.

almedina-ms:
[https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367154651](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367154651)

> @RebeccaAnne

Discussion:

This could be useful if we someday have a scenario where the URL is configured by the values of the inputs.
If we had that, the inputs would be explicit there, so this would be redundant.
We could have the behavior be that if there are explicit inputs they are validated in addition to anything in input.
Ultimately we're okay with having inputs on the base class. #Resolved

> @almedina-ms

April 3rd: We went back to this issue, David thinks Inputs should only be a property on Action.Submit instead of all actions. This would make it easier to extend Inputs to all Action elements; if we consider Inputs to be necessary in all actions we can introduce it later

almedina-ms:
[https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367155556](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367155556)

> @RebeccaAnne

Discussion: This is about custom rendering the errors, for example in a pop up bubble rather then underneath. David feels this is important, but may be P2 for other renderers. Because this is a renderer feature, not a card schema thing, we'd be okay with an inconsistency between renderers if this is implemented for Javascript but not others. #Resolved

> @RebeccaAnne

Put this context in the spec.

almedina-ms:
[https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367156518](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367156518)

> @RebeccaAnne

Discussion: On naming, Matt was concerned calling it inputs might imply to card authors that input types should go in there. Maybe include "Id" in the name? "AssociatedInputIds"? #Resolved

> @RebeccaAnne

renamed to associatedInputIds

almedina-ms:
[https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367156982](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367156982)

> @RebeccaAnne

Discussion: Can this be a card Id? We don't have card ids in shared model today, but we do in JavaScript. We probably should have them in all platforms, at which point yes, this could be a card id. #Resolved

> @RebeccaAnne

Updated the spec to call out the schema change for card ids.

almedina-ms:
[https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367012948](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367012948)

> @shalinijoshi19

Do we expect other additions to the schema around this space? (trying to understand why 'main' vs definitive list?) #Resolved

> @shalinijoshi19

Wasn't there an 'isRequired' in use in our schema already someplace? (i can't find it now as i sift through the schema explorer :() Are we repurposing that property to apply also in the context of Input.*? It may be worth clarifying and outlining upfront (perhaps a table or list or something) where you highlight the Input type and the new properties it would have #Resolved

>  @RebeccaAnne

There are further additions to the schema that are described later in the document (specifically: associated inputs on actions and id on adaptive card). I'll update the wording here to indicate that this is the full list of schema changes for the inputs themselves.

> @RebeccaAnne

We don't have an existing isRequired property that officially shipped.

As we thought we were doing this earlier, there were some places where isRequired may have slipped in. I think maybe it existed in shared model (but wasn't hooked up to anything) and there was at one point a working version of validation for JavaScript that must have included it, but we haven't officially shipped/documented/supported it.

Let me know once i update if with the new wording you still feel a table up here would help. There is one a few sections down that calls out the properties that are being validated.

almedina-ms:
[https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367015376](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r367015376)

> @shalinijoshi19

Thanks for adding this. Would be nice to see this called out explicitly upfront and then delve into an explanation on the properties' description #Resolved

> @shalinijoshi19

What's the regex syntax that we'd be permitting? Do we call that out anyplace? How permissive are we planning to be? Do we use regex's anywhere else today in our stack already or is this introducing regex's for the first time (outside of Markdown which i presume is a specialized syntax yes?) #Resolved

> @RebeccaAnne

Reworded this section. Let me know if it makes more sense after the update.

> @RebeccaAnne

We didn't have any plans to add any particular restriction here beyond valid regex. Was there something particular you were thinking of?

I don't think we have other scenarios where we take a regex as card author input, but we do use regexs internally to enforce, for example, date/time formatting.

almedina-ms:
[https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r397555263](https://github.com/microsoft/AdaptiveCards/pull/3663#discussion_r397555263)

> @almedina-ms

This has not been talked about but what happens with Input.Text with inline actions, where does the label go? Only on top of the input or above the input and the action?? the same happens for the error message

shalinijoshi19:
>  - Validation indicators: As part of input validation, we are allowing card authors to mark inputs as optional or required. One way to indicate that visually would be to put and indicator next to the input’s label. This requires knowing which text is the label.

nitnit: typo: "an"

---
Refers to: specs/DesignDiscussions/InputLabels.md:35 in 00a3e36. [](commit_id = 00a3e364c493265f392d7f673a93e94157b85186, deletion_comment = False)

shalinijoshi19:
>  - Interaction model. Many forms move focus to the input when the label is clicked/tapped on. We can’t support that behavior without knowing which label goes with which input.

super extreme nit: typo ":" instead of '.'

---
Refers to: specs/DesignDiscussions/InputLabels.md:37 in 00a3e36. [](commit_id = 00a3e364c493265f392d7f673a93e94157b85186, deletion_comment = False)

specs/DesignDiscussions/InputLabels.md
```diff
@@ -72,17 +72,55 @@ This conflict speaks to a general tension inherent in the design of Adaptive Car
 
 In order to gather more data to help us balance the needs of the card authors and hosts, we will send a survey to customers who have expressed an interest in improving out input story. The following options are under consideration, and will be evaluated against survey data to determine which if any will be included in our v1 implementation.
```
shalinijoshi19, line 73: > we will send a survey to customers [](start = 88, length = 35)

@ almedina-ms can you update the spec with a summary of what the results of the survey were around this - Summary: 
At this point looks like app developers want that control in hostconfig to match label formatting that they have for their respective apps; while the majority of the card authors are also used to having flexibility around being able to format the default label text.

specs/DesignDiscussions/InputLabels.md
```diff
@@ -72,17 +72,55 @@ This conflict speaks to a general tension inherent in the design of Adaptive Car
...
 
+> ## Rendering for string labels
```
shalinijoshi19, line 75: Are these the options described above for Formatting? If so these need to be subheadings under Formatting (###?)

specs/DesignDiscussions/InputLabels.md
```diff
@@ -72,17 +72,55 @@ This conflict speaks to a general tension inherent in the design of Adaptive Car
...
+>
+> The default rendering for labels should be performed with the default values for all of the following properties:
```
shalinijoshi19, line 77: >ith the default values [](start = 56, length = 22)

What are the defaults again and how are they specified?

specs/DesignDiscussions/InputLabels.md
```diff
@@ -72,17 +72,55 @@ This conflict speaks to a general tension inherent in the design of Adaptive Car
...
+> * spacing
+> * fontFamily
+> * fontSizes
+> * fontWeights
+> * foregroundColors
+>
+> This would allow authors to emulate writing a simple TextBlock element which only non-default property would be the text, making the string version look like this
```
shalinijoshi19, line 84: >which only non-default property would be the text, [](start = 73, length = 50)

I am not sure i can parse this correctly? Do you mean the only non-default property would be the text that is specified as value for the newly-introduced Label property?

specs/DesignDiscussions/InputLabels.md
```diff
@@ -72,17 +72,55 @@ This conflict speaks to a general tension inherent in the design of Adaptive Car
...
+>
+> ```json
+> {
+>	"type": "Input.Text",
+>	"id": "textInput",
+>	"label": "Label for Input"
+> }
+> ```
+>
+> While the TextBlock version would look like this
```
shalinijoshi19, line 94: >hile the TextBlock version would look like this [](start = 3, length = 47)

For continuity, discuss string and richtext support options as separate options here for formatting support in labels.

specs/DesignDiscussions/InputLabels.md
shalinijoshi19, line 77: ### Card Author Formatting Support
1. Simple Text String in Labels
2. Formatting support in Labels
- TextBlock version

.......................
<Final image card>

specs/DesignDiscussions/InputLabels.md
```diff
@@ -72,17 +72,55 @@ This conflict speaks to a general tension inherent in the design of Adaptive Car
...
+> ```
+>
+> Both of the afore mentioned cards would produce the same rendered result that would look like  this:
+> 
+> ![img](assets/InputLabels/LabelDefault.PNG)
+>
+
 ### Host Formatting
 
 If we want to allow the host to control formatting of the labels, either in general or for required inputs, we need a way for them to specify the desired format. There are two main things we may want to include:
```
shalinijoshi19, line 114: >f we want to allow the h [](start = 1, length = 24)

Should probably now update the spec to reflect that we indeed are adding host-formatting support (no longer a hypothetical)

specs/DesignDiscussions/InputLabels.md
```diff
@@ -91,6 +129,42 @@ If we want to allow the host to control formatting of the labels, either in gene
 ```
 Another option is to not put this in host config, and instead control this styling via support of native styling on required and optional labels (i.e. CSS, Xaml Styles, etc). This decision speaks to a larger discussion of the direction of host config that is beyond the scope of this document.
 
+> ### Host side styling
```
shalinijoshi19, line 132: >### [](start = 2, length = 3)

This is the 2nd of the 2 options mentioned in the Host Formatting section yes? If so, this should be a subhead at the same level as TextFormatting (####) - though ideally use numbered values to indicate numbered options for clarity/ease of discussion. so 1. and 2. in addition to or inlieu of sub-heads.

specs/DesignDiscussions/InputLabels.md
```diff
@@ -91,6 +129,42 @@ If we want to allow the host to control formatting of the labels, either in gene
 ```
 Another option is to not put this in host config, and instead control this styling via support of native styling on required and optional labels (i.e. CSS, Xaml Styles, etc). This decision speaks to a larger discussion of the direction of host config that is beyond the scope of this document.
```
shalinijoshi19, line 130: Is this ok to be noted as option we considered but discarded owing to the fact that it's out of scope here?

specs/DesignDiscussions/InputLabels.md
```diff
@@ -72,17 +72,55 @@ This conflict speaks to a general tension inherent in the design of Adaptive Car
...
+> 
+> ![img](assets/InputLabels/LabelDefault.PNG)
+>
+
 ### Host Formatting
 
 If we want to allow the host to control formatting of the labels, either in general or for required inputs, we need a way for them to specify the desired format. There are two main things we may want to include:
 
 #### Text formatting
  We can introduce the option to specify formatting of input labels for required and optional inputs. This could include things such as bold, italics, size, etc. We could either allow full text formatting (as in the formatting properties on a text block) or a subset specified as an enum. Since this would be host control, full formatting is probably fine here. For example:
```
shalinijoshi19, line 117: Let's just say "We are introducing the option .etc" since we seem to be leaning towards it from teh looks of the discussion?

shalinijoshi19, line 117: >llow full text formatting [](start = 178, length = 25)

Are there any blockers to supporting fulltext formatting at this point? If not, lets just then go with that approach in the proposal here to close unless there are open comments/issues/implementation costs here that are making this complicated.

specs/DesignDiscussions/InputLabels.md
shalinijoshi19, line 132: >st side styling [](start = 8, length = 15)

Question: How is host-side formatting different from styling?

specs/DesignDiscussions/InputLabels.md
```diff
@@ -91,6 +129,42 @@ If we want to allow the host to control formatting of the labels, either in gene
 ```
 Another option is to not put this in host config, and instead control this styling via support of native styling on required and optional labels (i.e. CSS, Xaml Styles, etc). This decision speaks to a larger discussion of the direction of host config that is beyond the scope of this document.
 
+> ### Host side styling
+>
```
shalinijoshi19, line 133: Any reason we need to get into host side styling if it's out of scope ?

specs/DesignDiscussions/InputLabels.md
```diff
@@ -91,6 +129,42 @@ If we want to allow the host to control formatting of the labels, either in gene
...
+>		"size": "medium",
+>		"color": "good"
+>	},
+>	"optionalInputs": {
+>		"isSubtle": true
+>	}	
+> }
+> ```
+>
+> While it's mentioned that we should allow almost all TextBlock properties to be modified by the label property, some of them must stay out of bounds for hosts as they would provide a bad user experience:
```
shalinijoshi19, line 162: >hem must stay out of bounds for hosts as they would provide a bad user experien [](start = 123, length = 79)

Great -- should this metapoint around ensuring we arent enabling poor end-user card experience, be included in the HostConfig discussion also?

specs/DesignDiscussions/InputLabels.md
```diff
@@ -91,6 +129,42 @@ If we want to allow the host to control formatting of the labels, either in gene
...
+> * maxLines and wrap: Labels should be descriptive enough so users can provide the information needed, if the user can't read the whole question the information they will not be able to provide a correct answer
+> * isVisible: labels should always be visible as long as the associated input is
+> * height: input elements should control the height for the whole set of label, input and error message 
+>
+
 #### Required and Optional indicators
 
  By default, we will mark required inputs with a `*`. We may want, however, to provide the option for the host to configure a suffix to the label for required or optional inputs. This allows the host to add a `*` to required labels, or to add the words "required" or "optional". Consider a host config similar to the below:
```
shalinijoshi19, line 170: >with a `*`. [](start = 42, length = 11)

Great! When we say by default, that means the default host config(urable) value will be "*" for suffix property for requiredinputs right?  And are we proposing we have any defaults for the optionalInputs defined "suffix" ?

specs/DesignDiscussions/InputLabels.md
```diff
@@ -91,6 +129,42 @@ If we want to allow the host to control formatting of the labels, either in gene
 ```
 Another option is to not put this in host config, and instead control this styling via support of native styling on required and optional labels (i.e. CSS, Xaml Styles, etc). This decision speaks to a larger discussion of the direction of host config that is beyond the scope of this document.
 
+> ### Host side styling
+> 
+> Following the previously mentioned text formatting the proposed set of properties we could allow the host application to override would be almost the same set of properties we provide for configuring a TextBlock: 
+>
```
almedina-ms: We need to specify that the string will allow markdown support

specs/DesignDiscussions/InputLabels.md
```diff
@@ -91,6 +129,42 @@ If we want to allow the host to control formatting of the labels, either in gene
...
+> 
+> | Property | Default Value |HostConfig | Native |
+> | --- | --- | --- | --- |
+> | spacing | "default" | X | |
+> | fontType | "default" | X | X |
+> | size | "default" | X | X |
+> | weight | "default" | X | X |
+> | color | "default" | X | X |
+> | isSubtle | false | X | |
+> | separator | false | X | |
```
almedina-ms: we may want to remove:
- fontType
- weight (we're supporting markdown)
- isSubtle
- separator
- horizontalAlignment

specs/DesignDiscussions/InputLabels.md
```diff
@@ -224,6 +308,9 @@ Either of these approaches would allow card authors who know their cards will be
 
 In the case of a host using the `labelFor` property there would be no negative impact on down-level platforms other than not providing the association. Because the TextBlock will still exist, only the association will be dropped when rendered on a platform that doesn't support this feature.
 
+> ### Conclusion 
+> It was decided that the approach to follow was the one presented in the Option 1: Providing the "label" property in all Input elements. isRequired hints will also only be rendered in label elements.
+
 ## Placeholders
```
shalinijoshi19, line 314: >Placeholder [](start = 3, length = 11)

This doesnt cover adding in PLaceholders for Input.ChoiceSet yes?

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+
+ - Validation indicators: As part of input validation, we are allowing card authors to mark inputs as optional or required. One way to indicate that visually would be to put an indicator next to the input’s label. This requires knowing which text is the label.
+
+ - Interaction model: Many forms move focus to the input when the label is clicked/tapped on. We can’t support that behavior without knowing which label goes with which input.
+
+ - Responsiveness: One direction for the future of cards involves re-flowing the layout of a card based on the size of the display. Knowing that a particular label goes with a particular input may help us ensure that semantically related elements stay together on reflow.
+
+## Basic Label Support in Adaptive Cards
+
+The basic addition that we are going to make to the schema is to add a label property to all input types. This approach provides a simple and concise way to label inputs, that hopefully encourages card authors to use it and receive the benefits of improvements to the issues called out above.
```
shalinijoshi19, line 43: If we do this how does it impact Teams consumption of the labels feature @siduppal?

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+	"actions": [
+		{
+			"type": "Action.Submit",
+			"title": "OK"
+		}
+	]
+}
+```
+
+## Formatting
```
shalinijoshi19, line 66: - How important of a feature is card author formatting support and can this be included in InputValidation part 2? @paulcam206 / @siduppal ; Can we have simple strings here as values for starters? 
- What's the min-bar here wrt formatting as far as Accessibility is concerned?

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+
+There is another concern, however, which must be balanced against the desire to provide card authors with the ability to format their labels. Hosts may have label formatting in their app which they want to apply consistently to cards they are displaying. This applies particularly to how they mark fields as required or optional (per spec at [InputValidation.md](https://github.com/microsoft/AdaptiveCards/blob/master/specs/DesignDiscussions/InputValidation.md)). While by default we will mark required inputs with a `*`, we may want to allow hosts more control. Perhaps some apps wish to show labels for required inputs in bold or italics. If we allows hosts to specify formatting based on what is required, however, and we allow authors to specify formatting via a TextBlock or RichTextBlock, we can easily run into cases where the formatting comes into conflict.
```
shalinijoshi19, line 69: @siduppal thoughts here ?

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+
+In order to gather more data to help us balance the needs of the card authors and hosts, we sent a survey to customers who have expressed an interest in improving out input story. The following options are under consideration, and will be evaluated against survey data to determine which if any will be included in our v1 implementation.
+
+The survey sent included 10 questions including the topics of input validation and input labels, for this spec we will only review the 4 questions that focus on the former topic. The results for this questions were used to leverage on some of the decisions taken on this spec document and can be found in the [Appendix](Appendix).
+
+> ### Default formatting for labels
+>
+> Reviewing the results from Q8 of the survey, we can see that most voters usually have a good experience with the default rendering of a TextBlock for rendering a label. From this result we can decide that string label rendering should be performed as if we were rendering a default TextBlock.
```
shalinijoshi19, line 79: Great.. so let's just specify that upfront then? That a label for now will just be a simple string value @almedina-ms ? (@paulcam206 what was your suggested approach again for when we choose to enhance label support to be richer than plain strings?)

specs/DesignDiscussions/InputLabelsOptionLabel.md
shalinijoshi19, line 69: How will a card author specify a given Input as "Required" ? What does it take to light up a simple label + specifying "isRequired" and AC rendering it by default with an asterisk? (@paulcam206 / @almedina-ms )

specs/DesignDiscussions/InputLabelsOptionLabel.md
paulcam206: I personally think that card author formatting is less important than host formatting

specs/DesignDiscussions/InputLabelsOptionLabel.md
paulcam206: One possibility is that we can ship v1 where the `label` is specified as a simple string. In subsequent releases we can support `label` being an object (i.e. an `AdaptiveElement`, or the equivalent):

v1:
``` json
{
    "type": "Input.Text",
    "label": "First name"
}
```

v2 (could use container or columnset... whatever):
``` json
{
   "type": "Input.Text",
    "label": {
        "type": "RichTextBlock",
        "inlines": [
            "some sample text",
            "maybe something with a glyph here!"
        ]
    }
}
```

specs/DesignDiscussions/InputLabelsOptionLabel.md
paulcam206: inputs are marked required by setting the `isRequired` property on an input element to `true`.

rendering by marking with an asterisk should be pretty straightforward in all renderers... should take less than a day of clock time per renderers (+- depending on test collateral and hostconfig entanglements... and as always there's lead time for shared model renderers)

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+> }
+> ```
+>
+> Both of the afore mentioned cards would produce the same rendered result that would look like  this:
+> 
+> ![img](assets/InputLabels/LabelDefault.PNG)
+>
+> It is important to note that wrapping should always be performed for this type of labels as not doing so would provide a bad user experience. If users can't read the whole label for an input they may not be able to provide a correct answer.
+
+### Host Formatting
```
shalinijoshi19, line 110: What are the implications if we dont supply host formatting options for labels for v1? @siduppal for Teams perspective..@almedina-ms / @paulcam206 ? Also can we start to include a "Dev Cost" line item associated with the features here? Thanks!

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+* maxLines and wrap: Labels should be descriptive enough so users can provide the information needed, if the user can't read the whole question the information they will not be able to provide a correct answer
+* isVisible: labels should always be visible as long as the associated input is
+* height: input elements should control the height for the whole set of label, input and error message 
+
+Allowing the control of this styling via support of native styling on required and optional labels (i.e. CSS, Xaml Styles, etc) may also be required but will be set out of scope for v1 of this feature. This decision speaks to a larger discussion of the direction of host config that is beyond the scope of this document.
+
+#### Required and Optional indicators
+
+ By default, we will mark required inputs with a `*`. We may want, however, to provide the option for the host to configure a suffix to the label for required or optional inputs. This allows the host to add a `*` to required labels, or to add the words "required" or "optional". Consider a host config similar to the below:
```
shalinijoshi19, line 146: @siduppal is the default behavior we are proposing sufficient for Teams for v1? And can greater host formatting options for labels follow with v2 of inputs/label support?

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+
+> When using these host properties, we would need to only apply them if all inputs on the card have labels so that required fields are marked consistently. We don't want to end up in a situation where some inputs in the card are marked as required in the label and others on the field itself, as this may lead to a confusing user experience.
+
+### Card Author Formatting
+
+We may wish to prioritize Host formatting over card author formatting in order to allow our hosts to impose a specified look and feel to forms displayed in their app. That said, if we don't allow card authors to specify formatting of labels, those authors who wish to do so will simply opt-out of our label approach and revert to using the existing unassociated TextBlocks and RichTextBlocks. This has accessibility and usability impact on our end customers, as they won't have access to the improved experiences made possible when our renderers have semantic knowledge of which text blocks are the labels for a given input.
```
shalinijoshi19, line 169: I think this is a fair point. But having labels without support for card author formatting in v1 is acceptable as long as we document and provide guidance samples etc on building an Accessible Card with the current AdaptiveCards stack, and then iterate on our feature set after shipping the basic feature.

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+
+Given that, we should consider options for card author label formatting. Consider the two following option.
```
shalinijoshi19, line 171: We should categorize this in the heading/title as P2 and have a section in the Appendix below around what each priority maps to, potential landing vehicle for it. Here's at least P0 definition as I see it, followed by the rest (@paulcam206 / @almedina-ms / @dclaux / @matthidinger )
P0: Minimum set of label + input validation features needed to ship in July to satisfy Accessibility and Teams requirements
P1: V2 with Label custom host-formatting support - Q4 possibly
P2: V3 with Label custom card-formatting support - Q1 2021

etc

specs/DesignDiscussions/InputLabels.md
almedina-ms: Based on the information gathered from asking the Accessibility team we decided not to use labels as placeholders as that affects user experience and accessibility

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+This card is another possible example of a use case for non-adjacent labels, where there is additional information between the label and the input:
+
+![img](assets/InputLabels/AdditionalLabelContent.png)
+
+This scenario could be covered by a RichTextBlock that contains both pieces of text being used as the label (arguably the entire text should be read as the accessibility label for the input). 
+
+This approach will limit what elements can be rendered as the label to only text.
+
+## Backwards Compatibility
+On earlier versions of renderers the label would be completely dropped, leading to a bad user experience. The card may be completely unusable if without the labels the user cannot deduce what information is meant to be provided in the unlabeled inputs.
```
shalinijoshi19, line 213: Hmm this one indeed sounds precarious and from what @paulcam206 demo'd to me  quite confusing indeed. On the other hand this option indeed seems to be the most intuitive :( -- Isnt there some novel use of fallback on hte new cards that could be leveraged? @paulcam206 ?

shalinijoshi19: It's going to impact new cards that use the label property on inputs rendering on say older versions of Teams hosts etc -- but i wonder if the newer cards can include feature fallback that shipped in 1.2 and we provide that guidance?

shalinijoshi19: Aah never mind - i see your note around fallback down below! thanks

specs/DesignDiscussions/InputLabelsOptionLabel.md
almedina-ms: I think accessibility wise, most of the work will be done through rendering, as long as the user has enough information on the input they should be able to work through it. I agree that we should provide guidance on how to make accessible cards.

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+				"id": "NameId"
+			}
+		]
+    }
+}
+```
+We could also possibly add a specific named feature to the requires dictionary to allow card authors to address it directly using something like:
+```json
+"requires": {
+	"inputLabels": "1.0"
```
shalinijoshi19, line 243: I quite like this latter very specific approach. it can make things simpler to use for the card author/improves readability of the card json and we dont have to remember the AC schema version to feature mapping necessarily. @paulcam206 / @matthidinger / @dclaux  thoughts?

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+}
+```
+Either of these approaches would allow card authors who know their cards will be hitting platforms targeting both 1.2 and 1.3 to take advantage of the labels for accessibility or interactivity purposes in 1.3 without breaking their experience on 1.2 clients. That this will not help card authors who are targeting 1.0 or 1.1 hosts where fallback is not supported. Card authors who know their hosts support version 1.3 (or any future version) will not need to worry about this concern.
+
+## Placeholders
```
shalinijoshi19, line 248: Great. Also arent we getting guidance from the Accessibility team that Placeholders in general arent that useful? 
<snip>
While placeholders have historically been very common in apps, there’s a growing view these days that from a usability perspective, placeholders should be avoided. At the end of Sarah’s [What's in a name?](https://sarahmhigley.com/writing/whats-in-a-name/#tooltips%2C-placeholders%2C-names%2C-and-duplicate-announcements) article, she references a number of articles which discuss why placeholders shouldn’t be used. Some of the content relates to web UI, but the principles around design apply to any type of UI.

So you may want to review some of the articles, such as [Placeholders in Form Fields Are Harmful](https://www.nngroup.com/articles/form-design-placeholders/), and consider if perhaps your customers would be better helped by having (as the article says) “clear, visible labels that are placed outside empty form fields.”
</snip>

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+
+Based on the answers from Q7, no users answered to using placeholders only for labeling their inputs. For v1 of this feature we decided to not modify placeholders.
+
+## Label Position
+
+As alluded to above, we should consider how the label position is determined. One of the goals of this feature is to allow us to someday intelligently reflow inputs, so it may be the case that we want to maintain control of label positioning. Today we would always draw the label above the input, but in the future we may put it to the side if our rendering space is wide enough, or push it into placeholders if our space is constrained. For the time being, we can consider label position to be work for the future so it can line with intelligent reflowing so on v1 it's not going to be considered.
```
shalinijoshi19, line 254: Great - let's annotate this P2 and also mention in teh proposal upfront that labels will be positioned on top of the input. @siduppal will this work for you guys?

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
...
+Based on the answers from Q7, no users answered to using placeholders only for labeling their inputs. For v1 of this feature we decided to not modify placeholders.
+
+## Label Position
+
+As alluded to above, we should consider how the label position is determined. One of the goals of this feature is to allow us to someday intelligently reflow inputs, so it may be the case that we want to maintain control of label positioning. Today we would always draw the label above the input, but in the future we may put it to the side if our rendering space is wide enough, or push it into placeholders if our space is constrained. For the time being, we can consider label position to be work for the future so it can line with intelligent reflowing so on v1 it's not going to be considered. 
+
+## Other Input Types
+
+This document has primarily focused on `Input.Text` in it's examples, but the label property should be present on all input types. See examples below:
+
```
shalinijoshi19, line 259: What's our thinking here wrt implementation ? Would we need to add it to each individual input type or hopefully add it to the base class directly (is it BaseInputElement?)

specs/DesignDiscussions/InputLabelsOptionLabel.md
```diff
@@ -0,0 +1,440 @@
+# Input/Label Associations
+
+This document covers design considerations related to issue [#203](https://github.com/microsoft/AdaptiveCards/issues/203).
+
```
shalinijoshi19, line 4: @almedina-ms / @dclaux / @matthidinger / @paulcam206 : Why am i already seeing labels /requires both referenced as part of Inputs already here in our [public documentation](https://adaptivecards.io/explorer/Input.Number.html)?

specs/DesignDiscussions/InputLabels.md
```diff
@@ -1,3 +1,52 @@
+# Table of Contents
+1. [Input/Label Associations](#InputLabel-Associations)
+	1. [Basic Label Support in Adaptive Cards](#Basic-Label-Support-in-Adaptive-Cards) 
+	2. [Formatting](#Formatting)
+		1. [Default rendering for labels](#Default-rendering-for-labels)
+		2. [Required indicators](#Required-indicators) 
+		3. [Card Author Formatting](#Card-Author-Formatting)
+			1. [Option 1 – Allow TextBlocks and RichTextBlocks as Labels](#Option-1-–-Allow-TextBlocks-and-RichTextBlocks-as-Labels)
+			2. [Option 2 – LabelFor Property on TextBlocks/RichTextBlocks]([#Option-2-–-LabelFor-Property-on-TextBlocks/RichTextBlocks])
+			3. [Final Conclussion](#Final-conclussion)
```
shalinijoshi19, line 10: Nit: Typo; "Conclusion"

specs/DesignDiscussions/InputLabels.md
```diff
@@ -32,9 +81,9 @@ Consider the following card:
...
 
- - Validation indicators: As part of input validation, we are allowing card authors to mark inputs as optional or required. One way to indicate that visually would be to put and indicator next to the input’s label. This requires knowing which text is the label.
+ - Validation indicators: As part of input validation, we are allowing card authors to mark inputs as optional or required. One way to indicate that visually would be to put an indicator next to the input’s label. This requires knowing which text is the label.
 
- - Interaction model. Many forms move focus to the input when the label is clicked/tapped on. We can’t support that behavior without knowing which label goes with which input.
+ - Interaction model: Many forms move focus to the input when the label is clicked/tapped on. We can’t support that behavior without knowing which label goes with which input.
 
  - Responsiveness: One direction for the future of cards involves re-flowing the layout of a card based on the size of the display. Knowing that a particular label goes with a particular input may help us ensure that semantically related elements stay together on reflow.
```
shalinijoshi19, line 88: >ds involves re-flowing the layout of a ca [](start = 54, length = 41)

@paulcam206 can i have a visual here to understand this scenario a bit better? Thanks

specs/DesignDiscussions/InputLabels.md
```diff
@@ -1,3 +1,52 @@
...
+			1. [Expanded](#Expanded)
+			2. [Compact](#Compact)
+		5. [Toggle Inputs](#Toggle-Inputs)
+	7. [Feature Cost Estimation](#Feature-Cost-Estimation)
+		1. [Follow ups](#Follow-ups) 
+	8. [Accessibility](#Accessibility)
+		1. [1.3.1 Info and Relationships](#131-Info-and-Relationships)
+		2. [2.4.6 Headings and Labels](#246-Headings-and-Labels)
+		3. [3.3.2 Labels and Instructions](#332-Labels-and-Instructions)
+2. [Appendix](#Appendix)
```
shalinijoshi19, line 28: We should also annotate requirements wiht Priorities and explain the mapping in the Appendix: (Here's a stab at this  - @matthidinger / @dclaux let me know your thoughts)
P0 - Required feature included in v1 feature release; Counts towards Accessibility and/or Teams MVP scenario
P1 - Good to have feature that should be considered due to feature completeness/Survey demand
P2 - Nice to have feature that has been demanded via customer surveys, but can be included in a follow up release.

specs/DesignDiscussions/InputLabels.md
```diff
@@ -413,8 +376,84 @@ This document has primarily focused on `Input.Text` in it's examples, but the la
...
+| --- | --- | --- | --- |
+| C++ Object Model | 4 | Spec completion | Tests |
+| UWP Implementation | 4 | C++ Object Model | Tests and override sample |
+| Android Implementation | 4 | C++ Object Model | Tests and override sample | 
+| iOS Implementation | 4 | C++ Object Model | Override sample |
+| C# Object Model | 4 | Spec completion | Tests |
+| .NET Implemenation | 4 | C# Object Model | Override sample |
+| HTML Implementation | 4 | C# Object Model | Tests | 
+| JavaScript Implementation | 14 | Spec completion | Changes to designer |
+| Documentation | 6 | Spec completion, UWP, Android, iOS and .NET implementations | Changes to schema explorer and guidance on how to override inputs |
```
shalinijoshi19, line 395: Bug Bash ? Probably add in a couple of days worth here to account for prep + bash

specs/DesignDiscussions/InputLabelsOptionLabel.md
shalinijoshi19: Thanks @Paulcam206; i think that's precisely the direction we ended up going last week where subsequently, per @matthidinger's follow up proposal, v2's type would potentially be a separate "Label" type..

PR #3910, opened 2020-04-08
[ReactNative] Added requires/fallback support

pragadeeshk:
## Solves requests
* Fallback and requires properties on elements #2019

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3910)

### DIFF
--- a/source/community/reactnative/src/models/model-factory.js
+++ b/source/community/reactnative/src/models/model-factory.js
@@ -1,9 +1,28 @@
 import * as Models from './index'
 import { ElementType } from '../utils/enums'
 import * as Utils from '../utils/util'
+import { HostConfigManager, HostCapabilities } from '../utils/host-config';
 
 export class ModelFactory {
     static createElement(payload, parent) {
+        if (!payload) {
+            return undefined;
+        }
+        if(payload.requires) {
+            let requirements = new HostCapabilities(payload.requires)
+            let hostCapabilities = HostConfigManager.getHostConfig().getHostCapabilities()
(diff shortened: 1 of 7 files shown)


source/community/reactnative/src/utils/host-config.js
```diff
@@ -516,6 +516,7 @@ export class HostConfig {
 	factSet = new FactSetConfig();
 	fontStyles = new FontStyleConfig();
 
+	hostCapabilities = new HostCapabilities()
```
vivekvijayakrishnan, line 519: Just a question - So by `default`, we are not giving any `capabilities` right?

source/community/reactnative/src/models/model-factory.js
```diff
@@ -1,9 +1,28 @@
...
+                return ModelFactory.checkForFallBack(payload, parent);
+            }
+        } else {
+            return this.getElement(payload, parent)
+        }
+        
+    }
+
+    static getElement(payload, parent) {
         if (!payload) {
```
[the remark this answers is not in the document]
pragadeeshk: Just in case if getElement is called from some other place. good to have a validation

source/community/reactnative/src/utils/host-config.js
pragadeeshk: Thats correct. capabilities need to be set in the hostConfig and if the items in the 'requires' object matches the host capabilities, it will be rendered. Else it will fallback.

PR #3925, opened 2020-04-13
[JS] Add docs and generation support to templating library

paulcam206:
## Description
As per the last docs PR, this one adds doc generation support and generated docs (as two commits) to the `adaptivecards-templating` project. Note that I have not opted to make this project-wide via lerna. This is because it looks like we need to reconcile our projects before hoisting our dependencies up.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3925)

### DIFF
--- a/source/nodejs/adaptivecards-templating/docs/README.md
+++ b/source/nodejs/adaptivecards-templating/docs/README.md
@@ -1,0 +1,13 @@
+[Adaptive Cards Templating Javascript SDK](README.md)
+
+# Adaptive Cards Templating Javascript SDK
+
+## Index
+
+### Classes
+
+* [Binding](classes/binding.md)
+* [EvaluationContext](classes/evaluationcontext.md)
+* [ExpressionParser](classes/expressionparser.md)
+* [GlobalSettings](classes/globalsettings.md)
+* [Template](classes/template.md)
--- a/source/nodejs/adaptivecards-templating/typedoc.json
+++ b/source/nodejs/adaptivecards-templating/typedoc.json
@@ -1,0 +1,17 @@
+{
+	"name": "Adaptive Cards Templating Javascript SDK",
+	"inputFiles": "./src",
+	"mode": "library",
+	"out": "docs",
+	"exclude": [
+		"__tests__"
+	],
+	"excludeNotExported": true,
+	"excludePrivate": true,
+	"readme": "none",
+	"plugin": [
+		"typedoc-plugin-markdown"
+	],
+	"theme": "markdown",
+	"stripInternal": true
+}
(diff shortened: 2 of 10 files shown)


source/nodejs/adaptivecards-templating/docs/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Index
+
+### Classes
+
+* [Binding](classes/binding.md)
+* [EvaluationContext](classes/evaluationcontext.md)
+* [ExpressionParser](classes/expressionparser.md)
+* [GlobalSettings](classes/globalsettings.md)
+* [Template](classes/template.md)
```
shalinijoshi19, line 13: What does it take to consolidate this into a single API reference .md for easy reference to exported interfaces something like [https://docs.microsoft.com/en-us/dotnet/api/overview/azure/?view=azure-dotnet](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/?view=azure-dotnet)?

source/nodejs/adaptivecards-templating/typedoc.json
```diff
@@ -0,0 +1,17 @@
...
+	"name": "Adaptive Cards Templating Javascript SDK",
+	"inputFiles": "./src",
+	"mode": "library",
+	"out": "docs",
+	"exclude": [
+		"__tests__"
+	],
+	"excludeNotExported": true,
+	"excludePrivate": true,
+	"readme": "none",
```
shalinijoshi19, line 11: Why is this none and yet a ReadMe.md did get generated? What's the guidance here (i am hoping we'll eventually flip this switch to actually write up a readme that gets automatically included?

source/nodejs/adaptivecards-templating/docs/README.md
paulcam206: I'll have to dig in to see what it'll take. It looked like it would be a bit more complicated than just generating from one level higher, though, so for now it seemed better to get docs working independently.

source/nodejs/adaptivecards-templating/typedoc.json
paulcam206: [This option](http://typedoc.org/guides/options/#readme) allows you to either supply `none`, in which case it generates its own toplevel README.md linking directly into the docs, or you can supply `path/to/README.md`, which is the README it will use on the index page. IIRC, when I was configuring all this stuff for the `adaptivecards` package, not supplying any `readme` directive caused it to just use `source/nodejs/adaptivecards/README.md`. This meant that there was no aggregate page with links into all the various docs.

PR #3967, opened 2020-04-29
Changes to accommodate Schema changes

muthurathinam:
* Removed appId property
* Renamed autoRefresh property to refresh
* Removed userIds and displayCurrentCardWhileRefreshingProperty properties inside refresh property

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3967)

### DIFF
--- a/source/nodejs/adaptivecards-aaf/src/core.ts
+++ b/source/nodejs/adaptivecards-aaf/src/core.ts
@@ -504,7 +504,7 @@ export class AdaptiveApplet {
                         }
 
                         if (this._card.refresh) {
-                            this.internalExecuteAction(this._card.refresh.action, ActivityInvocationContext.Refresh);
+                            this.internalExecuteAction(this._card.refresh.action, ActivityInvocationContext.Automatic);
                         }
                     }
                 }
--- a/source/nodejs/adaptivecards-aaf/src/core.ts
+++ b/source/nodejs/adaptivecards-aaf/src/core.ts
@@ -18,8 +18,8 @@ export class ExecuteAction extends Adaptive.SubmitAction {
     }
 }
 
-export class AutoRefreshActionProperty extends Adaptive.PropertyDefinition {
-    parse(sender: AutoRefreshDefinition, source: Adaptive.PropertyBag, context: Adaptive.SerializationContext): ExecuteAction | undefined {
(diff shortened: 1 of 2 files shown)


source/nodejs/adaptivecards-aaf/src/invoke-activity.ts
```diff
@@ -28,7 +27,7 @@ export interface InvokeActivity {
 }
 
 export enum ActivityInvocationContext {
-    AutoRefresh,
+    Refresh,
```
dclaux, line 30: I think we should name this enum value "Automatic" (vs the other one that is "UserInteraction")

muthurathinam: Yeah sounds good, will change.

dclaux: Actually, if that's not much to ask - could you also rename the enum itself into **ActivityInvocationTrigger** and rename variables/function parameters of that type from **context** to **trigger**?

If you'd rather not I can always do it later.

muthurathinam: Updated with the changes

source/nodejs/adaptivecards-aaf/src/core.ts
```diff
@@ -134,10 +134,10 @@ export class AdaptiveApplet {
         }
     }
 
-    private createActivityRequest(action: ExecuteAction, context: ActivityInvocationContext): ActivityRequest | undefined {
+    private createActivityRequest(action: ExecuteAction, trigger: ActivityInvocationTrigger): ActivityRequest | undefined {
         if (this.card) {
             let request: ActivityRequest = {
-                context: context,
+                trigger: trigger,
```
dclaux, line 140: Sorry I missed that before - per spec, `trigger` should be passed as a property of the `activity.value` object rather than a top level. And its value should be "automatic" or "manual":
image

It might make sense to change the ActivityInvocationTrigger enum values to "Automatic" and "Manual". Would you mind making those changes too?

muthurathinam: Update the changes

PR #3968, opened 2020-04-29
Adding config.yml to enable Question template

shalinijoshi19:
## Description
Following along the lines of the [typescript repo](https://github.com/microsoft/TypeScript/issues/new/choose) adding a redirect-only question template via config.yml that redirects the community to StackOverflow for questions.
For now, allowing for blank issues as a catch-all but eventually goal is to only enable issues that fit our allowed templates for better project management/tracking.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3968)

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+--- 
+blank_issues_enabled: true
+contact_links: 
+  - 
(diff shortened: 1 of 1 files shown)


shalinijoshi19:
@matthidinger do we want to add anything about website feedback channels for providing feedback on adaptivecards.io?

PR #3979, opened 2020-05-04
Allow comma to be taken as a decimal point in the Number only input

AparnaHP:
Locales like Swedish (SV) use comma "," instead of "." as the decimal point

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/3979)


paulcam206:
@BalajiR - care to take a look?

PR #4006, opened 2020-05-12
Updated Nuget Package Version for AEL

jwoo-msft:
## Description
Replaces nuget source for AEL to public nuget feed

## How Verified
Ran the unit tests & verified with samples

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4006)


paulcam206:
can we revert back to not using `nuget.config` with this change?

jwoo-msft:
> can we revert back to not using `nuget.config` with this change?

yes.

PR #4010, opened 2020-05-13
[JS] Only assign img alt text if it's supplied

### LINKED ISSUE #2235: Name/ Alt-text property is not proprly defined for Images.
See dependent bug for repro steps and videos:
https://github.com/Microsoft/BotFramework-WebChat/issues/1339

paulcam206:
## Related Issue
Fixes #2235

## Description
When a card author fails to supply alt text, we actually create an img element with the alt text set to `"undefined"` rather than not set the property in the first place.

## How Verified
1. New unit test
2. Local repro and inspection with F12 tools

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4010)

### DIFF
--- a/source/nodejs/adaptivecards/src/__tests__/components/image.spec.ts
+++ b/source/nodejs/adaptivecards/src/__tests__/components/image.spec.ts
@@ -5,4 +5,5 @@ import {Image} from "../../card-elements";
 test('Image should be instantiated', ()=>{
     const image = new Image();
     expect(image).toEqual(expect.anything());
+    expect(image.altText).toBeUndefined();
 })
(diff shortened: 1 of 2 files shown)


source/nodejs/adaptivecards/src/__tests__/components/image.spec.ts
```diff
@@ -5,4 +5,5 @@ import {Image} from "../../card-elements";
 test('Image should be instantiated', ()=>{
     const image = new Image();
     expect(image).toEqual(expect.anything());
+    expect(image.altText).toBeUndefined();
```
shalinijoshi19, line 8: I know you were working on more comprehensive UTs for the platform.. so potentially some coming up which actually test the altText flowing through if specified/not undefined?

paulcam206:
> Something we should consider backporting to release/1.2 and potentially prep for an AdaptiveCards v1.2.9 patch.. Thoughts?

Should be easy enough (though it'll likely be a manual port rather than a cherry-pick).

PR #4013, opened 2020-05-15
Jwoo/template json value handling update

### LINKED ISSUE #4014: [.NET][Template] [Ordinary JSON Value As $data's Context's Value]
# Platform

* .NET HTML

# Version of SDK

Latest master

# Details
In parsing data context of form, "$data" : value
value can be any valid json value.
but current .Net doesn't handle bool, string, and number.

valid ordinary json value examples in setting data context.
"$data" : 1
"$data" : TRUE
"$data" : FALSE
"$data" : "Hello"

jwoo-msft:
## Related Issue
Fixes #4014 

## Description
In parsing data context of form, "$data" : value
value can be any valid json value.
but current .Net doesn't handle bool, string, and number. 

## How Verified
1. ran all unit tests
2. updated test card and verified it.
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4013)

### DIFF
--- a/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
+++ b/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
@@ -197,21 +197,8 @@ namespace AdaptiveCards.Templating
             // get value node from pair node
             // i.e. $data : "value"
             IParseTree templateDataValueNode = context.value();
-            // value was json object or json array, we take this json value and create a new data context
-            if (templateDataValueNode is AdaptiveCardsTemplateParser.ValueObjectContext || templateDataValueNode is AdaptiveCardsTemplateParser.ValueArrayContext)
-            {
-                string childJson = templateDataValueNode.GetText();
-                try
-                {
-                    PushDataContext(childJson, root);
-                }
-                catch (Exception e)
-                {
-                    throw new Exception($"parsing data failed at line, '{context.Start.Line}', '{childJson}' was given", e);
-                }
-            }
             // refer to label, valueTemplateStringWithRoot in AdaptiveCardsTemplateParser.g4 for the grammar this branch is checking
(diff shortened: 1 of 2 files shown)


source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
```diff
@@ -241,8 +228,17 @@ public override AdaptiveCardsTemplateResult VisitTemplateData([NotNull] Adaptive
...
             {
-                throw new Exception($"'{context.value().GetText()}' at line, '{context.Start.Line}' is invalid value type for '$data : ' pair");
+                string childJson = templateDataValueNode.GetText();
+                try
+                {
+                    PushDataContext(childJson, root);
+                }
+                catch (Exception e)
+                {
+                    throw new Exception($"parsing data failed at line, '{context.Start.Line}', '{childJson}' was given", e);
```
shalinijoshi19, line 240: @jwoo-msft / @paulcam206,  should we be handling and re-throwing all System.Exception's or try to be specific about templating exceptions? What are the exceptions types that can result in the call to PushDataContext() ? Ideally we'd only be handling those.. 

Something like: 
try
{
    PushDataContext(childJson, root);
}
catch(Exception innerException)
{
 if (e is JsonSerializationException || e is InvalidOperationException) ==> or something similar
{
  throw new AdaptiveSerializationException("your string", innerException );
}
 throw; // rethrow the original "exceptional"/unexpected exception as-is to avoid overwriting original context/stacktrace.
}

shalinijoshi19: Btw we should add unit tests that invoke the exception code path also to test the non-golden-path behavior the customer will see

PR #4026, opened 2020-05-18
Adaptive Template SDK Exception Class

jwoo-msft:
## Description
Added AdaptiveTemplateException Class.

## How Verified

1. All unit tests are ran and passed.
2. Added a new unit test for testing invalid test case.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4026)

### DIFF
--- a/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
+++ b/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
@@ -97,13 +97,9 @@ namespace AdaptiveCards.Templating
                     root = JToken.Parse(data);
                     PushDataContext(data, root);
                 }
-                catch (JsonException e)
+                catch (JsonException innerException)
                 {
-                    throw new Exception("Setting root data failed with given data context", e);
-                }
-                catch
-                {
-                    throw;
+                    throw new AdaptiveTemplateException("Setting root data failed with given data context", innerException);
                 }
                 
             }
@@ -225,15 +221,10 @@ namespace AdaptiveCards.Templating
                     {
                         throw new AdaptiveTemplateException($"'{templateLiteral.Symbol.Text}' at line, '{templateLiteral.Symbol.Line}' is invalid because '{e.Message}'");
                     }
-                    catch (JsonException e)
+                    catch (JsonException innerException)
                     {
-                        throw new AdaptiveTemplateException($"'{templateLiteral.Symbol.Text}' at line, '{templateLiteral.Symbol.Line}' is malformed for '$data : ' pair", e);
+                        throw new AdaptiveTemplateException($"'{templateLiteral.Symbol.Text}' at line, '{templateLiteral.Symbol.Line}' is malformed for '$data : ' pair", innerException);
                     }
-                    catch
-                    {
-                        throw;
-                    }
-
                 }
             }
             else
@@ -244,13 +235,9 @@ namespace AdaptiveCards.Templating
                 {
                     PushDataContext(childJson, root);
                 }
-                catch (JsonException e)
+                catch (JsonException innerException)
                 {
-                    throw new AdaptiveTemplateException($"parsing data failed at line, '{context.Start.Line}', '{childJson}' was given", e);
-                }
-                catch 
-                {
-                    throw;
+                    throw new AdaptiveTemplateException($"parsing data failed at line, '{context.Start.Line}', '{childJson}' was given", innerException);
                 }
             }
 
@@ -301,13 +288,9 @@ namespace AdaptiveCards.Templating
             {
                 throw new AdaptiveTemplateException($"value of '$data : ', json pair, '{context.GetText()}'  at line, '{context.Start.Line}'is malformed");
             }
-            catch (JsonException e)
+            catch (JsonException innerException)
             {
-                throw new AdaptiveTemplateException($"value of '$data : ', json pair, '{child.TEMPLATEROOT().Symbol.Text}' at line, '{child.TEMPLATEROOT().Symbol.Line}' is malformed", e);
-            }
-            catch
-            {
-                throw;
+                throw new AdaptiveTemplateException($"value of '$data : ', json pair, '{child.TEMPLATEROOT().Symbol.Text}' at line, '{child.TEMPLATEROOT().Symbol.Line}' is malformed", innerException);
             }
 
             return new AdaptiveCardsTemplateResult();
--- a/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
+++ b/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
@@ -219,8 +219,7 @@ namespace AdaptiveCards.Templating
                     }
                     catch (ArgumentNullException)
                     {
-                        string errorMessage = $"Please enter a non-null value for '{templateLiteral.Symbol.Text}' at line, '{templateLiteral.Symbol.Line}'";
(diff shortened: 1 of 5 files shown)


source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
```diff
@@ -97,10 +97,15 @@ public AdaptiveCardsTemplateVisitor(Func<string, object> nullSubstitutionOption,
                     root = JToken.Parse(data);
                     PushDataContext(data, root);
                 }
-                catch (Exception e)
+                catch (JsonException e)
                 {
                     throw new Exception("Setting root data failed with given data context", e);
```
shalinijoshi19, line 102: >Exception [](start = 30, length = 9)

This should also be an AdaptiveException type instead of system.exception.

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
```diff
@@ -97,10 +97,15 @@ public AdaptiveCardsTemplateVisitor(Func<string, object> nullSubstitutionOption,
                     root = JToken.Parse(data);
                     PushDataContext(data, root);
                 }
-                catch (Exception e)
+                catch (JsonException e)
```
shalinijoshi19, line 100: Nit nit: naming; instead of 'e' probably innerException?

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
```diff
@@ -97,10 +97,15 @@ public AdaptiveCardsTemplateVisitor(Func<string, object> nullSubstitutionOption,
...
                 {
                     throw new Exception("Setting root data failed with given data context", e);
                 }
+                catch
+                {
+                    throw;
+                }
+
```
shalinijoshi19, line 108: Probably just remove this catch block since we are only rethrowing here and nothing else.

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
```diff
@@ -218,11 +223,15 @@ public override AdaptiveCardsTemplateResult VisitTemplateData([NotNull] Adaptive
                     }
                     catch (ArgumentNullException e)
                     {
-                        throw new Exception($"'{templateLiteral.Symbol.Text}' at line, '{templateLiteral.Symbol.Line}' is invalid because '{e.Message}'");
+                        throw new AdaptiveTemplateException($"'{templateLiteral.Symbol.Text}' at line, '{templateLiteral.Symbol.Line}' is invalid because '{e.Message}'");
                     }
-                    catch (Exception e)
+                    catch (JsonException e)
+                    {
+                        throw new AdaptiveTemplateException($"'{templateLiteral.Symbol.Text}' at line, '{templateLiteral.Symbol.Line}' is malformed for '$data : ' pair", e);
```
shalinijoshi19: How are you distinguishing "invalid" vs ""malformed" as far as the customer is concerned? Is this an error in the template or the data? Is JSONexception for the for the data portion? If so we should probably correlate it explicitly in the error message we are forming out for the exception..

jwoo-msft: if parsing can be initiated, and failed, then it's malformed. if parsing could even be started such as null selection path, then, it's invalid.

---
In reply to: [426855989](https://github.com/microsoft/AdaptiveCards/pull/4026#discussion_r426855989) [](ancestors = 426855989)

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
```diff
@@ -282,15 +295,19 @@ public override AdaptiveCardsTemplateResult VisitTemplateRootData([NotNull] Adap
...
             {
-                throw new Exception($"value of '$data : ', json pair, '{context.GetText()}'  at line, '{context.Start.Line}'is malformed");
+                throw new AdaptiveTemplateException($"value of '$data : ', json pair, '{context.GetText()}'  at line, '{context.Start.Line}'is malformed");
+            }
+            catch (JsonException e)
+            {
+                throw new AdaptiveTemplateException($"value of '$data : ', json pair, '{child.TEMPLATEROOT().Symbol.Text}' at line, '{child.TEMPLATEROOT().Symbol.Line}' is malformed", e);
             }
-            catch (Exception e)
+            catch
```
shalinijoshi19: Same: you can remove this block that is doing nothing but throwing the inner exception

jwoo-msft: please see iteration 2.

---
In reply to: [426861321](https://github.com/microsoft/AdaptiveCards/pull/4026#discussion_r426861321) [](ancestors = 426861321)

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
shalinijoshi19: Ok thanks. So it almost feels like we can simply alter the string then to be more actionable for the user by including this bit of context int the error message for the ArgumentNullException handling and maybe bubble up the ArgumentNullException as-is to distinguish between teh 2 scenarios explicitly - ([https://docs.microsoft.com/en-us/dotnet/standard/exceptions/](https://docs.microsoft.com/en-us/dotnet/standard/exceptions/))

How about this? 

try
{
 ...
}
catch(ArgumentNullException)
{
   var errorMessage = $"Please enter a non-null value for '{templateLiteral.Symbol.Text}' at line, '{templateLiteral.Symbol.Line}''"; // Do we need to relay the innerException.Message? What more context is it providing to the user? 
   throw new ArgumentNullException(errorMessage);
}
catch(JsonException innerEx)
{
  throw new AdaptiveTemplateException($"'{templateLiteral.Symbol.Text}' at line, '{templateLiteral.Symbol.Line}' is malformed for '$data : ' pair", innerException);
}


---
In reply to: [426858661](https://github.com/microsoft/AdaptiveCards/pull/4026#discussion_r426858661) [](ancestors = 426858661,426855989)

shalinijoshi19: 1. Nit: i used a string variable in my suggestion since that's my preference -- if you want to do it, then use the same format for the handler below too for consistency; 
2. We should extend your unit tests to cover both these cases so we can use that later for sample output/clarifying the behavior in our documentation also.. Wll we need to regenerate the API docs here? 


---
In reply to: [426946169](https://github.com/microsoft/AdaptiveCards/pull/4026#discussion_r426946169) [](ancestors = 426946169,426858661,426855989)

source/dotnet/Test/AdaptiveCards.Templating.Test/TestTransform.cs
```diff
@@ -258,6 +258,60 @@ public void TestExternalDataContextInternalReference()
...
+            {
+                Root = jsonData
+            };
+
+            try
+            {
+                string cardJson = transformer.Expand(context);
+                Assert.Fail("There should be an exception");
+            }
+            catch (AdaptiveTemplateException e)
```
shalinijoshi19, line 304: Probably add another one to cover the ArgumentNullException case

PR #4040, opened 2020-05-20
[UWP] Fix UWP builds in master

paulcam206:
## Description
An update to VS2019's SDKs breaks builds in CI. This change fixes that.

## How Verified
* local build
* CI build

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4040)

### DIFF
--- a/source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
+++ b/source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
@@ -96,7 +96,7 @@
       <AdditionalOptions>/profile %(AdditionalOptions)</AdditionalOptions>
     </Link>
     <CustomBuildStep>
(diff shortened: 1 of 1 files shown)


shalinijoshi19:
Thanks @paulcam206 -- but are we broken yet again? I thought there was some break earlier last month that we'd needed to work around yes?

paulcam206:
> Thanks @paulcam206 -- but are we broken yet again? I thought there was some break earlier last month that we'd needed to work around yes?

This is a different issue -- related to changes in the SDK rather than changes in the build toolchain. When it rains it pours! :)

Issue #4034, opened 2020-05-19
[Designer] "type" property at the top element JSON


matthidinger:
With the latest Designer update, the `type` property is no longer at the top of the generated JSON. This has caused some friction with my muscle memory and makes the payloads harder to scan.

1. Load vnext Designer
2. Drag in a TextBlock from the Card Elements toolbox

Expected: 

```json
{
   "type": "TextBlock",
   "text": "New TextBlock"
}
```

PR #4044, opened 2020-05-20
[TS] Serialize properties in declaration order

dclaux:
## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/4034

## Description
This change serializes properties in the order they are declared in serializable object classes. The most important improvement this change introduces is that the `type` property is now serialized first, as used to be the case before the move to the PropertyBag model, making JSON payloads easier to read.

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4044)

### DIFF
--- a/source/nodejs/adaptivecards/src/serialization.ts
+++ b/source/nodejs/adaptivecards/src/serialization.ts
@@ -744,7 +744,6 @@ export abstract class SerializableObject {
                 let propertyValue = ctor[propertyName];
 
                 if (propertyValue instanceof PropertyDefinition) {
-                    // schema.add(propertyValue);
                     properties.push(propertyValue);
                 }
             }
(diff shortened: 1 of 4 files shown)


source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -729,13 +737,15 @@ export abstract class SerializableObject {
...
         let ctor = <any>this.constructor;
+        let properties: PropertyDefinition[] = [];
 
         for (let propertyName in ctor) {
             try {
                 let propertyValue = ctor[propertyName];
 
                 if (propertyValue instanceof PropertyDefinition) {
-                    schema.add(propertyValue);
+                    // schema.add(propertyValue);
```
paulcam206, line 747: nit: probably don't need to keep commented code

dclaux: Indeed - removed.

PR #4045, opened 2020-05-20
[JS] Paint aria-current attribute on selected ChoiceSet option

### LINKED ISSUE #4027: [JS] Selected state not reflected for comboboxes in WebChat
Per email thread:

> Combo box for state selection does not announce “Selected” state. If screen reader user selects “Washington”, it only announces “Washington” instead of “Washington, selected.”

paulcam206:
## Related Issue
Fixes #4027

## Description
For accessibility purposes, we need to apply the `aria-current` attribute to the `<option/>` that's currently selected.

Also fixed a couple of minor typos while I was in there :)

## How Verified
* Manually verified properties in F12 tools
* Verified speech output for Narrator
* Verified speech output for NVDA

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4045)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -2806,18 +2806,18 @@ export class ChoiceSetInput extends Input {
     private static uniqueCategoryCounter = 0;
 
     private static getUniqueCategoryName(): string {
-        let uniqueCwtegoryName = "__ac-category" + ChoiceSetInput.uniqueCategoryCounter;
+        let uniqueCategoryName = "__ac-category" + ChoiceSetInput.uniqueCategoryCounter;
 
         ChoiceSetInput.uniqueCategoryCounter++;
 
-        return uniqueCwtegoryName;
+        return uniqueCategoryName;
     }
 
     private _uniqueCategoryName: string;
     private _selectElement: HTMLSelectElement;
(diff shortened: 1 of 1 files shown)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2887,20 +2887,34 @@ export class ChoiceSetInput extends Input {
         return element;
     }
 
+    // Make sure `aria-current` is applied to the currently-selected item
+    protected internalApplyAriaCurrent(): void {
+        const options = this._selectElement.options;
+        if (options) {
```
dclaux, line 2893: Nit: empty line before if statement

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2887,20 +2887,34 @@ export class ChoiceSetInput extends Input {
...
+            for (const i in options) {
+                if (options[i].selected) {
+                    options[i].setAttribute("aria-current", "true");
+                } else {
```
dclaux, line 2897: For consistency with rest of the code, please move else to its own line

paulcam206:
@dclaux - thanks for the review! :)

PR #4059, opened 2020-05-27
[JS] Generate correct aria role depending on Action type

### LINKED ISSUE #2240: Incorrect role is define for 'View details' control.
Please review the following bug for repro steps and video:
https://github.com/Microsoft/BotFramework-WebChat/issues/1350

paulcam206:
## Related Issue
Fixes #2240 

## Description
Any time a card element has an `Action.OpenUrl` `selectAction`, we should assign the `link` role.

## How Verified
* local build and inspection

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4059)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -869,7 +869,7 @@ export class TextBlock extends BaseTextBlock {
 
                 if (hostConfig.supportsInteractivity) {
                     element.tabIndex = 0
-                    element.setAttribute("role", "button");
+                    element.setAttribute("role", this.selectAction.getAriaRole());
 
                     if (this.selectAction.title) {
                         element.setAttribute("aria-label", this.selectAction.title);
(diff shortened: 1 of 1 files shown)


paulcam206:
verified behavior on generated test site as well

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -869,7 +869,7 @@ export class TextBlock extends BaseTextBlock {
 
                 if (hostConfig.supportsInteractivity) {
                     element.tabIndex = 0
-                    element.setAttribute("role", "button");
+                    element.setAttribute("role", Utils.roleFromAction(this.selectAction));
```
dclaux, line 872: This should be a `getAriaRole()` method off the `Action` class, that returns "button" by default. It would be overridden in `OpenUrlAction` to return "link". With the way you have it now, it's not possible for a custom action type to decide what aria role it's associated with.

PR #4066, opened 2020-05-28
[JS] Render Action.OpenUrl with role=link (#3914)

### LINKED ISSUE #3914: [JS][Renderer][Accessibility] Action.OpenUrl rendered buttons should have the "link" aria role
# Platform
* JavaScript

# Details
Currently all actions get renderered as <button> elements with role "button". When a button opens a link, as is the case for Action.OpenUrl, it should have the "link" role instead.

paulcam206:
## Related Issue
Fixes #3914 in `release/1.2` (direct cherry-pick)

## Description
Render `Action.OpenUrl` buttons with `role="link"`

## How Verified
* local build and inspection

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4066)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -4216,6 +4216,13 @@ export class OpenUrlAction extends Action {
         this.url = Utils.getStringValue(json["url"]);
     }
 
+    render(baseCssClass: string = "ac-pushButton") {
+        super.render(baseCssClass);
+
+        // OpenUrl actions behave like a hyperlink. Make sure screenreaders treat them that way.
+        this.renderedElement.setAttribute("role", "link");
+    }
+
     getHref(): string {
         return this.url;
     }


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4216,6 +4216,13 @@ export class OpenUrlAction extends Action {
         this.url = Utils.getStringValue(json["url"]);
     }
 
+    render(baseCssClass: string = "ac-pushButton") {
+        super.render(baseCssClass);
+
+        // OpenUrl actions behave like a hyperlink. Make sure screenreaders treat them that way.
+        this.renderedElement.setAttribute("role", "link");
```
dclaux, line 4223: Shouldn't this code be in the `Action` class' render method, and look like this:
```typescript
this.renderedElement.setAttribute("role", this.getAriaRole());
```

paulcam206: I believe you're thinking about the PR from yesterday (#4059) rather than this original fix (#3914)...

dclaux: Yes, I am. Given the fix from yesterday, I think this fix should be implemented using the newly introduced getAriaRole() method. Or is this fix going into another branch that doesn't have getArialRole()?

paulcam206: yeah, this is targeting `release/1.2`... that said, I think you're right. I might need to reconcile these two changes...

paulcam206:
going to re-do this from the unified approach (see #4080)

PR #4108, opened 2020-06-04
Fixed action set cut off issue #4058

### LINKED ISSUE #4058: [IOS][Layout] [When put submit button on the right column of the card, it could be cut off]
# Platform

* iOS

# Author or host

Host - Teams IOS

# Version of SDK

Both SDK 1.2.7 and the latest master 

# Details

Our demand is to put the "submit" button on the right of the card and cover a fixed-width space. Then the text would cover other spaces on the left. So we update our payload like the attachment.
[feedback_payload.txt](https://github.com/microsoft/AdaptiveCards/files/4687985/feedback_payload.txt)

It could work fine on the Android side, but for IOS, we can find that the second join button is truncated as the text has more content. We have marked the text with "maxLines": 2, not sure why it still shrink the button space.
Screen Shot 2020-05-27 at 4 52 39 PM

Our expected result is as below. I can trigger this by hardcoded the column width as 59px. But we don't want server-side know the client render logic and pass through the hardcoded value. Is there any way we can implement the demands without hard coded value?
Screen Shot 2020-05-27 at 4 55 24 PM

Besides, I also notice that when there is enough space, the auto layout for the right column's width would respect the button's instrinsicContentSize. Add width constraint for the button would only increase button width, column width won't increase accordingly and cut off the button.

For example, we initialize to ACRButton with instrinsicContentSize = 59.0. And set the constraint as below. The result is that button's width is increased to 72.0, but cut off by column width. Is there any way we can increase them both?
`[button.widthAnchor constraintEqualToConstant:72.0].active = YES;`
Screen Shot 2020-05-27 at 5 50 16 PM

jwoo-msft:
## Related Issue
Fixed #4058

## Description
ActionsSet's intrinsic content size was not working. When container view that held button were added to UIScrollView, the container view's intrinsic content size didn't work because by the time intrinsic content size was updated to actual value after buttons are rendered, it was too late. Following modifications have been made:
1. delayed adding the container view until the intrinsic content size is computed.
2. added insert arranged subview at index method. by making #1 change, there was a negative side effect to layout. originally container view was added first, then buttons, followed by subview that is created by ShowCard Action. To preserve the original layout, the insert arranged subview at index method is added. An index is recorded, then after all views including subview is added, container view is inserted at the index.
3. updated intrinsic content size to dynamically update its size when icons are loaded to buttons. UIKit wasn't updating the size dynamically, so code is added to update it.
4. if buttons don't fill the space initially, and the buttons grow the container view after icons are loaded, buttons shrank. This is now fixed by dynamically updating content size of UIScroll view as button's icons are loaded.
5.  Host config's action alignment is now honored. 
6. Some minor clean-ups.
- removed an unused variable
- updated a custom renderer interface as interface has been updated.

## How Verified
1. cut off issue is fixed as it can be observed below.
Simulator Screen Shot - iPhone 8 - 2020-06-03 at 20 02 03
previously, buttons were shrunken, now buttons behave correctly even after images are loaded later 
Simulator Screen Shot - iPhone 8 - 2020-06-03 at 20 03 28

2. existing units tests were run, and verified with all of the cards with different combinations of action alignment host config values.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4108)


source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/CustomInputNumberRenderer.mm
```diff
@@ -18,7 +18,7 @@ + (CustomInputNumberRenderer *)getInstance
 
 - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
            rootView:(ACRView *)rootView
-             inputs:(NSArray *)inputs
+             inputs:(NSMutableArray *)inputs
```
shalinijoshi19, line 21: Didnt we make a similar change for 1.2.9? Was it someplace else?

jwoo-msft: No, this is applied to sample app's custom renderer.

PR #4132, opened 2020-06-05
[JS] Fix adaptivecards-fabric tests on Linux

paulcam206:
## Description
I noticed that the `adaptivecards-fabric` tests were failing in Linux. The failures are due to path case sensitivity -- fix is easy.

## How Verified
* Ran tests :)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4132)


paulcam206:
thanks @dfisenko

PR #4157, opened 2020-06-09
[JS] Add aria-expanded state to showcard action button

paulcam206:
## Related Issue
Fixes VSO #23940376

## Description
Expandable showcard actions lack the `aria-expanded` attribute, which is helpful for screenreaders.

image

## How Verified
* local build and devtools testing

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4157)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -3290,6 +3290,20 @@ class ActionButton {
                     this.action.renderedElement.classList.add(...hostConfig.makeCssClassNames("style-" + this.action.style.toLowerCase()));
                 }
             }
+
+            this.updateAriaExpandedState();
+        }
+    }
+
+    private updateAriaExpandedState() {
+        // update aria-expanded property
+        if (this.action.renderedElement && this.action.renderedElement.classList.contains("expandable")) {
+            if (this.state === ActionButtonState.Expanded) {
+                this.action.renderedElement.setAttribute("aria-expanded", "true");
+            }
+            else {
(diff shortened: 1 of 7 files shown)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3290,6 +3290,20 @@ class ActionButton {
                     this.action.renderedElement.classList.add(...hostConfig.makeCssClassNames("style-" + this.action.style.toLowerCase()));
                 }
             }
+
+            this.updateAriaExpandedState();
+        }
+    }
+
+    private updateAriaExpandedState() {
```
dclaux, line 3298: Sorry for being picky, but there are two things I think should be done differently:
- As a general principle, the logic should not rely on inspecting the DOM or checking if a CSS style is applied.
- Instead of introducing yet another method that updates the rendered element's CSS class list, I think we should reuse the existing `updateActionButtonCssStyle`:

  - Change its declaration to add a `buttonState` paremeter of type `ActionButtonState`. To make the method backwards compatible, have that parameter default to `ActionButtonState.Normal`
  - In the `ShowCardAction.updateActionButtonCssStyle` implementation, set `aria-expanded` as appropriate

paulcam206: picky is good :)

Working on fix

paulcam206: fixed 👍

paulcam206:
thanks @dclaux!

PR #4161, opened 2020-06-10
[Designer] Place keyboard focus back on "New card" button after dialog close

paulcam206:
## Related Issue
Fixes VSO #24068381

## Description
Per accessibility rules, when a button shows a dialog, focus should be returned to the button after the dialog is closed. This does that for the designer's "New card" button.

## How Verified
* local build, keyboard

### DIFF
--- a/source/nodejs/adaptivecards-designer/src/card-designer.ts
+++ b/source/nodejs/adaptivecards-designer/src/card-designer.ts
@@ -638,6 +638,12 @@ export class CardDesigner extends Designer.DesignContext {
                         };
                         dialog.selectedSample.download();
                     }
+
+                    const newCardButton = document.getElementById(CardDesigner.ToolbarCommands.NewCard);
+
+                    if (newCardButton) {
+                        newCardButton.focus();
+                    }
                 };
                 dialog.open();
             });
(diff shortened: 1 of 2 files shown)


shalinijoshi19:
> Hi @paulcam206. Thanks for helping make the AdaptiveCards JS renderer + tooling better. As additional verification, once the JS build succeeds, please go to the [test site](https://adaptivecardsci.z5.web.core.windows.net/pr/4161) to test out your website/designer changes.

I checked the changes against the test website - while the New Card flow int eh designer looks OK the visualizer seems broken; Can you actually do me a favor and as you are making this change, remove references to the Classic Visualizer altogether since we're deprecating it and don't want it exposed via the designer or the website at this point for starters.. Also takes away the overhead of having to support it or maintain it etc Thanks!

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -638,6 +638,12 @@ export class CardDesigner extends Designer.DesignContext {
                         };
                         dialog.selectedSample.download();
                     }
+
+                    const newCardButton = document.getElementById(CardDesigner.ToolbarCommands.NewCard);
```
dclaux, line 642: There is already a private `_newCardButton` field that you can use for this:
```typescript
this._newCardButton.renderedElement.focus();
```

source/nodejs/adaptivecards-designer/src/toolbar.ts
```diff
@@ -94,6 +94,7 @@ export class ToolbarButton extends ToolbarElement {
     protected internalRender(): HTMLElement {
         let element = document.createElement("button");
 
+        element.id = this.id;
```
dclaux, line 97: Shouldn't be necessary, although why not.

PR #4177, opened 2020-06-15
[Site] Remove visualizer link in Designer

paulcam206:
## Related Issue
Fixes VSO #24081771

## Description
Per advice in VSO item, removing the visualizer link from the Designer. The visualizer itself will remain up, but should have some work done per conversation (see #4180).

## How Verified
* local build and testing

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4177)

### DIFF
--- a/source/nodejs/adaptivecards-site/themes/adaptivecards/layout/designer.ejs
+++ b/source/nodejs/adaptivecards-site/themes/adaptivecards/layout/designer.ejs
@@ -96,19 +96,6 @@
 
 			designer.toolbar.getElementById(ACDesigner.CardDesigner.ToolbarCommands.HostAppPicker).width = "auto";
 
-
-			let visualizerButton = new ACDesigner.ToolbarButton(
-				"visualizerButton",
-				"Classic Visualizer",
-				null,
-				function () {
-					window.open("<%- config.root %>visualizer")
-				}
-			);
-			visualizerButton.separator = true;
-			visualizerButton.alignment = ACDesigner.ToolbarElementAlignment.Right;
-			designer.toolbar.addElement(visualizerButton);
-
 			let feedbackButton = new ACDesigner.ToolbarButton(
 				"feedbackButton",
 				"Feedback",


matthidinger:
Removing the visualizer means no most Host Config editor, which is a pretty big loss for an arguably small group of folks. Any workarounds suggested?

dclaux:
I don't think we have any workaround at this point.

shalinijoshi19:
> Removing the visualizer means no most Host Config editor, which is a pretty big loss for an arguably small group of folks. Any workarounds suggested?

Hmm to @matthidinger's point, there's still ~4% of our site users that are hitting the Visualizer page. Given that there is no workaround for them, then @paulcam206 should we then just remove references to it from the designer page while still hosting the visualizer on the site (so we aren't advertising it anymore but people who actually want to use it can manually hit it over adaptivecards.io/visualizer should they choose to? Is HostConfig editing the only thing needed to get feature parity in the designer vis-a-vis classic visualizer?

paulcam206:
> > Removing the visualizer means no most Host Config editor, which is a pretty big loss for an arguably small group of folks. Any workarounds suggested?
> 
> Hmm to @matthidinger's point, there's still ~4% of our site users that are hitting the Visualizer page. Given that there is no workaround for them, then @paulcam206 should we then just remove references to it from the designer page while still hosting the visualizer on the site (so we aren't advertising it anymore but people who actually want to use it can manually hit it over adaptivecards.io/visualizer should they choose to? Is HostConfig editing the only thing needed to get feature parity in the designer vis-a-vis classic visualizer?

I'm willing to bet that the majority of users of the classic visualizer are _not_ using it for hostconfig editing. I bet it's a combination of folks with old bookmarks and folks that don't like the designer.

Perhaps the right thing to do here is:
* Remove visualizer link from the designer
* Leave the visualizer live
* Give the user a banner prompting them to move to the designer
* Offer up a survey seeking feedback if they don't like the designer

paulcam206:
Filed #4180 to cover the banner and survey.

PR #4188, opened 2020-06-16
[pic2card] Evaluate the model performance using mAP score.

haridas:
* Measure the quality of the object detection.

    - Using mAP metric
    - Evaluating image label vs custom image pipeline, metrics shows model
    does good job.
    - Tried pytorch implementations, have to do more iterations to improve.
    - Some cleanups

    * Fixed the test failure issue.

    - Separated the tf and pytorch dependencies, we don't need to install
    one if not required.
    - Only the stable pipeline will be enabled as default, other one will be
    used at dev stage only.
    - Migrated the requirements and docker files under a folder.

    * Reverted the separate folder for requirements.

    As the actions in the uptream need a change to make it work.

    * pandas version fix.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4188)


IbrahimSulai:
@paulcam206 - Could you please help us to find out the details, why the checks are getting failed in this PR?

Thanks in advance!

paulcam206:
Looks like this is failing in the `git checkout` stage:

```
git checkout --progress --force refs/remotes/pull/4188/merge
error: invalid path 'source/pic2card/app/assets/samples/Icon?'
##[error]Git checkout failed with exit code: 128
```

haridas:
Thanks @paulcam206. I have fixed it.

@Vasanth-S @IbrahimSulai Please review.

PR #4190, opened 2020-06-16
[DevSpec] Updates to input validation spec

almedina-ms:
This is the PR for the spec document to go into main; the 2 documents to review are [InputValidation.md](https://github.com/microsoft/AdaptiveCards/pull/4190/files#diff-cae2d7f5c79c829c005f09319a97bde2) and [InputLabels.md](https://github.com/microsoft/AdaptiveCards/pull/4190/files#diff-114ac7f9164cd2eef456300b74c45aea)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4190)


shalinijoshi19:
@almedina-ms can this now be checked into main ?

specs/DesignDiscussions/InputLabels.md
```diff
@@ -123,140 +115,130 @@ One option would be to make the label text a property of the input itself. This
...
+		"text": "Label for Input"
+	}
 }
 ```
-Either of these approaches would allow card authors who know their cards will be hitting platforms targeting both 1.2 and 1.3 to take advantage of the labels for accessibility or interactivity purposes in 1.3 without breaking their experience on 1.2 clients. That said, needing to add fallback to the card defeats the conciseness benefits of this approach, and will not help card authors who are targeting 1.0 or 1.1 hosts. Card authors who know their hosts support version 1.3 (or any future version) will not need to worry about this concern.
 
-### Option 2 – LabelFor Property on TextBlocks/RichTextBlocks
-An alternative option would be to have labels as separate text elements, but provide an id that links the inputs to the labels:
-``` json
+Both of the afore mentioned cards would produce the same rendered result that would look like  this:
```
RebeccaAnne, line 153: >afore mentioned [](start = 12, length = 15)

[nit] aforementioned

specs/DesignDiscussions/InputLabels.md
```diff
@@ -269,107 +251,26 @@ Similarly to option 1, a card author could use fallback to address this if they
...
 ```json
-{
-	"type": "Input.Text",
-	"id": "nameInput",
-	"label": "Please enter your name:",
-	"labelPosition": "Above",
-	"placeholder": "Last, First"
+"requires": {
+	"inputLabels": "1.0"
 }
```
RebeccaAnne, line 262: Has there been any further discussion/decision around whether we want to do this? If we're not planning to, we should probably say that here or remove this idea from the spec.

schemas/src/elements/inputs/Input.json
```diff
@@ -10,6 +10,16 @@
...
+			"description": "Error message to display when entered input is invalid",
+			"version": "1.3"
+		},
+		"isRequired": {
+			"type": "boolean",
+			"description": "Whether or not this input is required",
+			"version": "1.3"
+		},
 		"label": {
 			"type": "string|TextBlock|RichTextBlock",
```
RebeccaAnne, line 24: >|TextBlock|RichTextBlock [](start = 18, length = 24)

If we're not supporting this for v1 we should remove these from here. In general we should make sure the schema files are up-to-date with the latest spec changes.

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,503 @@
...
+![img](assets/InputValidation/Required.PNG)
+
+The error text and invalid indication will be rendered in the host's `attention` color.
+
+![img](assets/InputValidation/InputValidationExample.png)
+
+
+### Host side  styling
+
+As has been proposed, the error message is a string property which is rendered in the `attention` color by default, card hosts may need more control over how this messages and the visual representation of an error appear. As can be seen in the [Annex] there's a wide array on how error messages are represented in different websites. The table below details the proposed set of properties that host apps will be able to modify based on the examples located in the annex as well as if they are going to be modified through the host config or the native styling in each platform. It's important to note that some of this properties will act on the input element rather than the error message.
```
RebeccaAnne, line 70: >[Annex] [](start = 244, length = 7)

Link?

RebeccaAnne, line 70: >It's important to note that some of this properties will act on the input element rather than the error message. [](start = 578, length = 112)

Can you clarify this a bit? I assume "size" and "weight" do refer to the error message and that "spacing" refers to the space between the input and the error?

specs/DesignDiscussions/InputLabels.md
```diff
@@ -123,140 +115,130 @@ One option would be to make the label text a property of the input itself. This
...
+
+This conflict speaks to a general tension inherent in the design of Adaptive Cards. There's always going to be a trade off between giving the author more power, and letting the host control the experience for their users. We need to determine what balance we want to strike in this case.
 
-We should also consider which item types are valid as labels. Probably just TextBlock and RichTextBlock make sense, but as Xaml supports other types we should consider whether we want to as well.
+In order to gather more data to help us balance the needs of the card authors and hosts, we sent a survey to customers who have expressed an interest in improving out input story. The survey sent included 10 questions including the topics of input validation and input labels, for this spec we will only review the 4 questions that focus on the former topic. The results for this questions were used to leverage on some of the decisions taken on this spec document and can be found in the [Appendix](Appendix).
```
RebeccaAnne, line 126: >for this spec we will only review the 4 questions that focus on the former topic [](start = 277, length = 80)

Thanks for this detailed discussion of how the survey results were used! Great to see this feedback in action.

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,503 @@
...
+
+### Native Styling
+
+When an input fails validation it is visually signalized with a change of the input border color or with other visual cues. As this feature has to do with error message rendering no changes will be made for input rendering. On the other hand, we currently provide native styling for inputs in almost all platforms, this allows host apps to use the platform's capabilities to render the inputs as they want and the platform allows; platforms usually provide ways for signalizing an error in input fields through element states; for clarification we will provide examples on two platforms:
+
+#### UWP (XAML)
+
+For the UWP renderer we can still support the current styles that host apps have defined for their inputs, we would only have to provide the naming convention for the visual states for valid (or neutral) and invalid inputs. The following snippet should provide a template on how hosts could specify this custom rendering by overloading the `InvalidState` on the VisualStateGroup:
```
RebeccaAnne, line 98: >For the UWP renderer we can still support the current styles that host apps have defined for their inputs [](start = 0, length = 105)

This is cool. Is this implemented already?

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,503 @@
...
+To explain the input retrieving behavior we landed on consider the following diagram where the main card A has two `ShowCard` actions.
+
+![img](assets/InputValidation/CardCascadeSample.png)
+
+The behaviour is as follows:
+* If the `Submit` action on card A is clicked, then only the inputs in card A are retrieved
+* If the `Submit` action on card B is clicked, then inputs in card A and card B are retrieved, but not the inputs in card C
+* If the `Submit` action on card C is clicked, then inputs in card A and card C are retrieved, but not the inputs in card B
+
+To summarize, all the inputs in the card where the `Submit` action was clicked and all the inputs in all the "parent" cards are retrieved, no inputs in "sibling" cards are retrieved. In the diagram above, card A would be a "parent" of B and C, and card B and C are siblings.
```
RebeccaAnne, line 366: >To summarize [](start = 0, length = 12)

What happens in a scenario where an input is hidden behind a toggle action? Are those retrieved regardless of whether or not they are hidden?

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,503 @@
...
+
+- By default we will validate on focus lost, so in a host with that behavior a user who has navigated to and away from an input will have seen the validation error, even if they subsequently hid it using a toggle or show card action. This means that in scenarios where the input is not required, but has a regex or other validation (such as the personalization example above) the user will already be aware that their field has not validated correctly. The scenario where a **required** input is not shown may be confusing to a user, but one could make the case that it's questionable card design to hide required fields behind toggle or show card actions.
+
+- If "All" doesn't work well for a particular  card, allowing the card author to explicitly specify inputs allows them to override this default behavior to configure the validation as appropriate for their scenario.
+
+##### Conclusion
```
RebeccaAnne, line 353: This conclusion section doesn't have a link in the table of contents

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,503 @@
...
+
+If there was a validation error on the first "page", bringing the hidden element back into focus on submit would result in one of the following, depending on whether we brought back the entire hidden element or just the invalid input. In either case, this is not the expected behavior of this card author, and the subsequent behavior if the toggle buttons are pressed again would not be well defined:
+
+Bring back non-visible card element            |  Bring back single input
+:-------------------------:|:-------------------------:
+![img](assets/InputValidation/BrokenToggleWizard.PNG)  |  ![img](assets/InputValidation/BrokenToggleWizard2.PNG)
+
+> Note on this card: This "Wizard" scenario is implemented here via toggle visibility. We should consider whether we want to support a first class wizard experience where a new element type could list a series of subcards to display in order. Whether or not we ever introduce such a concept, the point remains that toggle logic may be complex, and we shouldn't be arbitrarily toggling things visible outside the scope of what the card author has defined in their toggle buttons.
+
+#### Current Card/Container
```
RebeccaAnne, line 268: This is effectively the approach you land on at the end, so maybe combine, reorganize, or mention that here. Also, this section brings up concerns with this approach that aren't addressed in the conclusion, so we should probably add something around how we're addressing them and/or why they're less pressing than other considerations.

specs/DesignDiscussions/InputValidation.md
```diff
@@ -0,0 +1,503 @@
...
+| `Input.Date` | `min` | Input value is greater or equal to min. |
+| `Input.Date` | `max` | Input value is less or equal to max. |
+| `Input.Date` | `isRequired` | A date has been selected and populates the input. |
+| `Input.Time` | `min` | Input value is greater or equal to min. |
+| `Input.Time` | `max` | Input value is less or equal to max. |
+| `Input.Time` | `isRequired` | A time has been selected and populates the input. |
+| `Input.ChoiceSet` | `isRequired` | Any value has been selected. In the case of multiselect, 1 or more values have been selected. | 
+| `Input.Toggle` | `isRequired` | ValueOn has been selected. |
+
+### Which Inputs to Validate
```
RebeccaAnne, line 201: >Which Inputs to Validate [](start = 4, length = 24)

I think this section is a little unwieldy now, and honestly it probably was already when i handed it off.  Consider reorganizing this to put what we're actually shipping up front, and maybe move some of the detailed discussion of other options to the appendix and reference them from the text. Ideally someone would be able to come to this document in the future and easily see what the expected behavior is. We don't want to lose the discussion of tradeoffs, but now that we've decided on something we should probably lead with that. 

Also, because we both added to this at different times and with different opinions, i feel like it's maybe not super consistent about it's message on the tradeoffs of the various approaches (a section i wrote says one way is the best, and a section you wrote says something else). Consider re-reading through the whole section and seeing if there are places where we should update the language to present a consistent point of view. Again, we don't want to lose the substance of the discussions/disagreements, but someone reading should be able to come away from it with a clear feeling that "The team decided X approach was best for Y reasons."

shalinijoshi19:
@RebeccaAnne could you work @almedina-ms  to get this merged into the topic branch? Thanks!

PR #4203, opened 2020-06-22
Improve C++ JSON parsing performance

### LINKED ISSUE #4202: [UWP] JSON parsing performance
# Platform

* Android
* iOS
* UWP

# Author or host

# Version of SDK

Nuget 1.2.8

# Details

CPU profiling suggests performance of the C++ JSON parsing code can be improved as time is spent unnecessarily converting strings to stringstream objects and back again.

chriche-ms:
## Related Issue
Resolves #4202 

## Description
This PR modifies the ParseUtil::GetJsonValueFromString function to improve JSON parsing performance. The replaced code used a stream extraction operator (i.e. >>) to convert the passed string into a Json::Value instance. While this code was straightforward and easy to understand it required the string data to be copied into a stringstream (twice) and then copied back into a string before parsing could begin. The updated code creates a Json::CharReader directly to avoid these copies.

## How Verified
The changes mean that the GetJsonValueFromString function goes from being around 10% of CPU samples to 4% of samples when profiling adaptive card parsing and rendering.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4203)


shalinijoshi19:
Very neat. Thanks for finding and fixing!

PR #4212, opened 2020-06-23
Avoid unnecessary std::shared_ptr copies

### LINKED ISSUE #4211: [UWP] Avoid unnecessary std::shared_ptr ownership changes
# Platform

* Android
* iOS
* UWP

# Author or host

# Version of SDK

NuGet 1.2.8

# Details

Copying and assigning std::shared_ptr(s) to local variables should be avoided if possible as the incrementing and decrementing of the underlying reference cout can end up having a non-negligible effect on performance. Use of std::move to transfer ownership and passing by const-ref can help avoid some of this overhead.

chriche-ms:
## Related Issue
Resolves #4211 

## Description
This PR updates various frequently called functions in order to minimize the frequency with which ownership of various std::shared_ptr variables is shared. Instead the variables are either passed by reference (instead of by value) or ownership is transferred using std::move.

## How Verified
Existing unit tests were successfully run.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4212)


shalinijoshi19:
@almedina-ms is added to the review. #Closed

shalinijoshi19:

shalinijoshi19:
@RebeccaAnne is added to the review. #Closed

PR #4218, opened 2020-06-24
Remove try-catch from ParseUtil::GetBackgroundImage and other small fixes

### LINKED ISSUE #4217: [UWP] Don't use exceptions for control flow
# Platform

* Android
* iOS
* UWP

# Author or host

# Version of SDK

NuGet 1.2.8

# Details

The ParseUtil::GetBackgroundImage function supports the backgroundImage property being either a string or JSON object value. However, it does this by utilizing exception handling even though there is no error condition that needs signaling.

chriche-ms:
## Related Issue
Resolves #4217 

## Description
The ParseUtil::GetBackgroundImage throws and catches exception during 'good path' execution. This is confusing for new users of the library and goes against good programming practices. This PR replaces the try-catch with an if-else statement. There are also a number of other small fixes included in this PR:

- Ensure all PopulateKnownPropertiesSet functions are private, not protected.
- Update BaseElement::GetAdditionalProperties and Inline::GetAdditionalProperties to return a const reference instead of a value type.
- Update GetEnumValue function to use a template type parameter rather than std::function.
- Avoid use of unnecessary local variable in JsonObjectToJsonCpp for performance reasons.

## How Verified
Ran the existing unit tests and manually tested various cards with background images.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4218)

### DIFF
--- a/source/shared/cpp/ObjectModel/ParseUtil.cpp
+++ b/source/shared/cpp/ObjectModel/ParseUtil.cpp
@@ -150,29 +150,34 @@ namespace AdaptiveSharedNamespace
 
     std::shared_ptr<BackgroundImage> ParseUtil::GetBackgroundImage(const Json::Value& json)
     {
-        try
-        {
-            // handle "backgroundImage": <string>
-            std::string backgroundImageUrl = ParseUtil::GetString(json, AdaptiveCardSchemaKey::BackgroundImage, false);
-            if (backgroundImageUrl != "")
-            {
-                return std::make_shared<BackgroundImage>(backgroundImageUrl);
-            }
+        // handle "backgroundImage": <string>
+        const std::string& backgroundImagePropertyName = AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::BackgroundImage);
+        Json::Value propertyValue = json.get(backgroundImagePropertyName, Json::Value());
 
+        if (propertyValue.empty())
+        {
             // handle "backgroundImageUrl": <string>
-            backgroundImageUrl = ParseUtil::GetString(json, AdaptiveCardSchemaKey::BackgroundImageUrl, false);
+            const std::string& backgroundImageUrlPropertyName = AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::BackgroundImageUrl);
+            propertyValue = json.get(backgroundImageUrlPropertyName, Json::Value());
+        }
+
+        if (propertyValue.isString())
+        {
+            const std::string backgroundImageUrl = propertyValue.asString();
+
             if (backgroundImageUrl != "")
             {
                 return std::make_shared<BackgroundImage>(backgroundImageUrl);
             }
-            return nullptr;
         }
-        catch (AdaptiveCardParseException)
+        else if (!propertyValue.empty())
         {
             // handle "backgroundImage": { <content> }
             auto jsonValue = ParseUtil::ExtractJsonValue(json, AdaptiveCardSchemaKey::BackgroundImage, false);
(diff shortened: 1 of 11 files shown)


shalinijoshi19:
@RebeccaAnne is added to the review. #Closed

shalinijoshi19:

source/shared/cpp/ObjectModel/ParseUtil.h
```diff
@@ -103,11 +103,11 @@ namespace AdaptiveSharedNamespace
         std::string ToLowercase(const std::string& value);
     };
 
-    template<typename T>
+    template<typename T, typename Fn>
```
shalinijoshi19, line 106: >typename [](start = 25, length = 8)

Hey @chriche-ms  are there known perf issues with std::function that are prompting this change? Quick search seems to indicate so ? Also I'm wondering if you'd be up for a quick "Perf gotchas" type of learnings for us for education as you've been wrking on profiling the library here -- are there specific best practices resources you've been looking at as guiding your analysis here? It might be a great topic to visit all up say in our next sync maybe - the worst offenders as far as perf is concerned based on your profiling and case study?

source/shared/cpp/ObjectModel/ParseUtil.cpp
```diff
@@ -150,29 +150,34 @@ namespace AdaptiveSharedNamespace
...
-        try
-        {
-            // handle "backgroundImage": <string>
-            std::string backgroundImageUrl = ParseUtil::GetString(json, AdaptiveCardSchemaKey::BackgroundImage, false);
-            if (backgroundImageUrl != "")
-            {
-                return std::make_shared<BackgroundImage>(backgroundImageUrl);
-            }
+        // handle "backgroundImage": <string>
+        const std::string& backgroundImagePropertyName = AdaptiveCardSchemaKeyToString(AdaptiveCardSchemaKey::BackgroundImage);
```
shalinijoshi19, line 154: Curious about the guideline/best practice here -- is it that we should avoid exception code paths int eh "golden" path scenarios but that it's ok to report back errors directly from here via exceptions in the error scenarios (eg GetValueAsString above that throws)

chriche-ms: Best practice is that exceptions shouldn't be used for implementing behaviour that is normal and expected (in this case a backgroundImage property that has an object value). In isolation it's fine for ParseUtil::GetString to throw but using such a function in GetBackgroundImage, when we expect to be be passing the function an 'invalid' argument, isn't a good idea. It's the wrong tool for the job, so to speak. The [CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Re-throw) (E2 & E3) do a good job of explaining this.

On a practical level I want to set the debugger to break when all exceptions are thrown because it allows me to discover potential bugs - I know that if any exception is thrown during development or testing that something unexpected has happened and I should immediately investigate. It's not really possible to do this if the software is constantly generating exceptions.

source/shared/cpp/ObjectModel/ParseUtil.h
chriche-ms: Yep, there definitely can be performance concerns with using std::function although in this case the change was more to do with it being used incorrectly. Unlike C# delegates, std::function is mostly for storing (and calling obviously!) callable types (e.g. used as a member variable) and not meant to be used as a function parameter. Where possible templated parameters should be used for passing callable types.

PR #4253, opened 2020-06-29
Updating docs for Image to clarify acceptable formats

matthidinger:
Updating the Schema Explorer docs for `Image` to clarify that only PNG, JPEG, and GIF are supported formats. This is to reduce the confusion around SVG support

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4253)


schemas/src/elements/Image.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/microsoft/AdaptiveCards/6f39aedce45864ae1067ed44a5551dc973790bb5/source/nodejs/typed-schema/schema/lib/Type.json",
   "extends": "Element",
-  "description": "Displays an image.",
+  "description": "Displays an image. Acceptable formats are PNG, JPEG, and GIF",
```
RebeccaAnne, line 4: Are we also supposed to check in the generated schema file under, for example, c:\AdaptiveCards\schemas\1.3.0\?

matthidinger: Oh that's a good question, I suppose I need to run the tool locally if it doesn't run in CI

RebeccaAnne: Also, i just remembered, the docs under \specs\elements generate from here as well, if we want to keep those up to date too.

---
In reply to: [447096959](https://github.com/microsoft/AdaptiveCards/pull/4253#discussion_r447096959) [](ancestors = 447096959)

matthidinger: Sweet, @paulcam206  fixed the Generate Specs build and it looks good now. The CI system will auto gen the files, and I've confirmed in this PR it looks good. Does anything look out of place to you?

Issue #4186, opened 2020-06-16
issue in Input.Date control in adaptive cards


Vishnuvk33:
we have created an adaptive card for outlook with Input.Date field and one http.action.

On click of  http.action , we are always getting the date from input.date filed as one day before.

For ex, if we set 13 may 2020 in date field, we are always getting 12 may 2020 as api input.

Channel :  O365

Any thing need to be given in Input.Date fiels?

shalinijoshi19:
@dclaux for FYI on this one. @Vishnuvk33 can you please share out your card payload here? Also it would be great if in the future you could use our [Bug Template](https://github.com/microsoft/AdaptiveCards/issues/new?assignees=&labels=Bug&template=bug_report.md&title=%5BPlatform%5D%5BBroad+Issue+Category+if+applicable+%28eg+Accessibility%29%5D+%5BBug+Title%5D) to file bugs for us with more details! Thanks

dclaux:
@Vishnuvk33 could you indee share a card payload as well as:
- Which Outlook version you're referring to (Windows, OWA, iOS, Android?)
- The time zone configured on the system running Outlook

"13 may 2020" basically means "13 may 2020 at 0:00 in the system's time zone". If the system's time zone is, for instance, set to Central European Summer Time, or GMT+2, "13 may 2020 at 0:00 GMT+2" converts into "12 may 2020 at 22:00 GMT", e.g. the previous day.

Vishnuvk33:
@dclaux  have uploaded the json flle in this.

1. we are using outloook windows 0365
2.  The timezone configured in the outlook is IST

[customer_card.txt](https://github.com/microsoft/AdaptiveCards/files/4802438/customer_card.txt)

Vishnuvk33:
Yeah sure @shalinijoshi19 , will follow the mentioned steps to report bugs.

dclaux:
@Vishnuvk33 the team was able to repro and is investigating.

Vishnuvk33:
Hi @dclaux , Any update on this issue.

dclaux:
The problem has been identified and a fix will be implemented. It's a server-side fix so it won't require an Outlook update. I don't have an ETA.

Vishnuvk33:
Hi @dclaux  Thanks for the update.

Vishnuvk33:
Hi @dclaux what is the fix which is implemeted.

Can you please just describe it.

The Date filed will always consider the Current date time.

[...]

Issue #2496, opened 2019-03-01
 [Android] Input.Date and Input.Time value/min/max consistency


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* Android

# Issue

The Input.Date value/min/max properties should only be recognized if they are of the form YYYY-MM-DD. The Input.Time value property should only be recognized if it's of the form HH:MM. The Android code should use the shared model's DateTimePreparser::TryParseSimpleDate and DateTimePreparser::TryParseSimpleTime methods added in #2450 to ensure consistency.

shalinijoshi19:
@almedina-ms assigning to you to take a look

RebeccaAnne:
Fixed by #2732

PR #4255, opened 2020-06-30
[Spec][InputEvolution]Update schema and documentation files to match the latest spec

RebeccaAnne:
This change updates the schema and documentation files for input evolution to match the latest spec. Specifically:

 - json files under schemas\src updated to match latest spec.
 - Re-generated adaptive-card.json from those files under schemas\1.3.0
 - Re-generated element specs under specs\elements from schema changes
 - Updated non-generated Rendering sections on all Action and Input files under specs\elements to document the current expected rendering behavior.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4255)

### DIFF
--- a/schemas/1.3.0/adaptive-card.json
+++ b/schemas/1.3.0/adaptive-card.json
@@ -217,7 +217,7 @@
               "version": "1.0"
             }
           ],
-          "description": "Specifies the background image of the card. Acceptable formats are PNG, JPEG, and GIF",
+          "description": "Specifies the background image of the card.",
           "version": "1.2"
         },
         "minHeight": {
@@ -243,12 +243,6 @@
             "fr",
             "es"
           ]
-        },
-        "showRequiredInputHints": {
-          "type": "boolean",
-          "description": "True if required fields should be visually indicated",
-          "version": "1.3",
-          "default": true
         },
         "verticalContentAlignment": {
           "$ref": "#/definitions/VerticalContentAlignment",
--- a/schemas/1.3.0/adaptive-card.json
+++ b/schemas/1.3.0/adaptive-card.json
@@ -20,7 +20,6 @@
         "iconUrl": {},
         "style": {},
         "fallback": {},
-        "inputs": {},
         "requires": {}
       },
       "type": "object",
@@ -51,7 +50,6 @@
         "iconUrl": {},
         "style": {},
         "fallback": {},
-        "inputs": {},
         "requires": {}
       },
       "type": "object",
@@ -82,60 +80,19 @@
           ],
           "description": "Initial data that input fields will be combined with. These are essentially ‘hidden’ properties."
         },
-        "inputs": {
-          "anyOf": [
-            {
-              "type": "array",
-              "items": {
-                "type": "string"
-              }
-            },
-            {
-              "$ref": "#/definitions/ActionInputs"
-            }
-          ],
-          "description": "None, All, or an array of ids for inputs (or containers of inputs) that should be submitted with this action and validated before allowing it.",
-          "version": "1.3",
-          "default": "All"
-        },
-        "title": {
-          "type": "string",
-          "description": "Label for button or link that represents this action."
-        },
-        "iconUrl": {
-          "type": "string",
-          "format": "uri-reference",
-          "description": "Optional icon to be shown on the action in conjunction with the title. Supports data URI in version 1.2+",
-          "version": "1.1"
-        },
-        "style": {
-          "$ref": "#/definitions/ActionStyle",
-          "description": "Controls the style of an Action, which influences how the action is displayed, spoken, etc.",
-          "version": "1.2"
-        },
-        "fallback": {
-          "anyOf": [
-            {
-              "$ref": "#/definitions/ImplementationsOf.Action"
-            },
-            {
-              "$ref": "#/definitions/FallbackOption"
-            }
-          ],
-          "description": "Describes what to do when an unknown element is encountered or the requires of this or any children can't be met.",
-          "version": "1.2"
-        },
-        "requires": {
-          "type": "object",
-          "additionalProperties": {
-            "type": "string"
-          },
-          "description": "A series of key/value pairs indicating features that the item requires with corresponding minimum version. When a feature is missing or of insufficient version, fallback is triggered.",
-          "version": "1.2"
-        }
+        "title": {},
+        "iconUrl": {},
+        "style": {},
+        "fallback": {},
+        "requires": {}
       },
       "type": "object",
-      "additionalProperties": false
+      "additionalProperties": false,
+      "allOf": [
+        {
+          "$ref": "#/definitions/Extendable.Action"
+        }
+      ]
     },
     "Action.ToggleVisibility": {
       "description": "An action that toggles the visibility of associated card elements.",
@@ -158,7 +115,6 @@
         "iconUrl": {},
         "style": {},
         "fallback": {},
-        "inputs": {},
         "requires": {}
       },
       "type": "object",
@@ -2419,22 +2375,6 @@
           "description": "Describes what to do when an unknown element is encountered or the requires of this or any children can't be met.",
           "version": "1.2"
         },
-        "inputs": {
-          "anyOf": [
-            {
-              "type": "array",
-              "items": {
-                "type": "string"
-              }
-            },
-            {
-              "$ref": "#/definitions/ActionInputs"
-            }
-          ],
-          "description": "None, All, or an array of ids for inputs (or containers of inputs) that should be validated before allowing this action.",
-          "version": "1.3",
-          "default": "None"
-        },
         "requires": {}
       },
       "type": "object",
@@ -2500,17 +2440,7 @@
           "version": "1.3"
         },
         "label": {
-          "anyOf": [
-            {
-              "type": "string"
-            },
-            {
-              "$ref": "#/definitions/TextBlock"
-            },
-            {
-              "$ref": "#/definitions/RichTextBlock"
-            }
-          ],
+          "type": "string",
           "description": "Label for this input",
           "version": "1.3"
         },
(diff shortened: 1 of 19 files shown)


schemas/1.3.0/adaptive-card.json
```diff
@@ -982,19 +938,19 @@
         },
         "max": {
           "type": "string",
-          "description": "Hint of maximum value expressed in ISO-8601 format (may be ignored by some clients)."
+          "description": "Hint of maximum value expressed in YYYY-MM-DD(may be ignored by some clients)."
```
dclaux, line 941: This should indeed be ISO8601. It has to support a time offset, otherwise it will not work properly across time zones.

schemas/1.3.0/adaptive-card.json
```diff
@@ -982,19 +938,19 @@
...
         },
         "min": {
           "type": "string",
-          "description": "Hint of minimum value expressed in ISO-8601 format (may be ignored by some clients)."
+          "description": "Hint of minimum value expressed in YYYY-MM-DD(may be ignored by some clients)."
```
dclaux, line 945: This should indeed be ISO8601. It has to support a time offset, otherwise it will not work properly across time zones.

schemas/1.3.0/adaptive-card.json
```diff
@@ -982,19 +938,19 @@
...
         },
         "placeholder": {
           "type": "string",
           "description": "Description of the input desired. Displayed when no selection has been made."
         },
         "value": {
           "type": "string",
-          "description": "The initial value for this field expressed in ISO-8601 format."
+          "description": "The initial value for this field expressed in YYYY-MM-DD."
```
dclaux, line 953: This should indeed be ISO8601. It has to support a time offset, otherwise it will not work properly across time zones.

schemas/1.3.0/adaptive-card.json
dclaux: Should we remove "may be ignored by some clients"?

schemas/1.3.0/adaptive-card.json
dclaux: Should we remove "may be ignored by some clients"?

schemas/1.3.0/adaptive-card.json
dclaux: Should we remove "may be ignored by some clients"?

schemas/1.3.0/adaptive-card.json
```diff
@@ -1127,19 +1083,19 @@
         },
         "max": {
           "type": "string",
-          "description": "Hint of maximum value (may be ignored by some clients)."
+          "description": "Hint of maximum value expressed in HH:MM (may be ignored by some clients)."
```
dclaux, line 1086: Should we remove "may be ignored by some clients"?

schemas/1.3.0/adaptive-card.json
```diff
@@ -1127,19 +1083,19 @@
...
         },
         "min": {
           "type": "string",
-          "description": "Hint of minimum value (may be ignored by some clients)."
+          "description": "Hint of minimum value expressed in HH:MM (may be ignored by some clients)."
```
dclaux, line 1090: Should we remove "may be ignored by some clients"?

schemas/1.3.0/adaptive-card.json
RebeccaAnne: This wasn't something i intentionally changed. It changed as a result of re-generating the docs.

That said, I'm pretty sure at least UWP is only supporting is only supporting YYYY-MM-DD for all fields in Input.Date. We had a whole discussion about this at some point, let me see if I can dig it up. 

---
In reply to: [447778889](https://github.com/microsoft/AdaptiveCards/pull/4255#discussion_r447778889) [](ancestors = 447778889)

schemas/1.3.0/adaptive-card.json
RebeccaAnne: Good point.

---
In reply to: [447780023](https://github.com/microsoft/AdaptiveCards/pull/4255#discussion_r447780023) [](ancestors = 447780023)

RebeccaAnne: Thinking about it, we may still want to document that this is ignored on earlier versions. How about "(may be ignored by some clients prior to version 1.3)"?

---
In reply to: [448481109](https://github.com/microsoft/AdaptiveCards/pull/4255#discussion_r448481109) [](ancestors = 448481109,447780023)

schemas/1.3.0/adaptive-card.json
RebeccaAnne: So, this YYYY-MM-DD format is what's currently showing up on the website schema explorer. (Presumably because the source files were already showing that, not sure why this 1.3 version isn't)

As for this history of this decision, I couldn't quite find the smoking gun, but here's what i've tracked down. In February 2019 we had a long mail thread about what to do with this scenario, which ended in us deciding to have a meeting and talk about it in person. I didn't find meeting notes, but following that meeting I implemented the YYYY-MM-DD (and HH:MM for time) behavior in UWP (#2450) and we logged several bugs to standardize on YYYY-MM-DD and HH:MM for Input.Date and Input.Time (#2495, #2496, #2497). 

All of which is to say i believe the current spec is for YYYY-MM-DD for Input.Date (inline dates in TextBlocks are different) and this apparent change showing up in my PR is an artifact of the 1.3 version being out of date for some reason (The source file was changed in January as part of #3706). If we want to talk about the merits of that decision, that's probably outside the scope of this PR.

---
In reply to: [448480840](https://github.com/microsoft/AdaptiveCards/pull/4255#discussion_r448480840) [](ancestors = 448480840,447778889)

dclaux: Well, that simply doesn't work... Just for context, we just uncovered a bug in Outlook a couple weeks ago, check out the details here: https://github.com/microsoft/AdaptiveCards/issues/4186

Any time we are in the business of manipulating dates, we must support a time offset.

Looks like we should have yet another conversation about this.

specs/elements/Input.Toggle.md
```diff
@@ -97,17 +84,17 @@ Controls the amount of spacing between this element and the preceding element.
 ## Rendering
 
 ### Labels
-The `label` property should be rendered to the left of the toggle box. Clicking/tapping on the rendered label should toggle the input.
+The `label` property should be rendered above the input box if present. Clicking/tapping on the rendered label should put focus in the input box.
```
shalinijoshi19, line 87: @RebeccaAnne should we include a quick note someplace encouraging authors to replace any hand-rolled label support in their cards with these to guarantee Accessible card elements ?

specs/elements/Input.Toggle.md
```diff
@@ -97,17 +84,17 @@ Controls the amount of spacing between this element and the preceding element.
 ## Rendering
 
 ### Labels
-The `label` property should be rendered to the left of the toggle box. Clicking/tapping on the rendered label should toggle the input.
+The `label` property should be rendered above the input box if present. Clicking/tapping on the rendered label should put focus in the input box. 
+
+The `color`, `isSubtle`, `size`, and `weight` properties provided in the host config under `"inputs":"label":"requiredInputs"` for inputs with `"isRequired":true`, and under `"inputs":"label":"optionalInputs"` for inputs with `"isRequired":false` should be respected. The spacing between the label and the input should be set to the value of `inputSpacing` specified in those same locations in host config.
 
 ### Accessibility
 The `label` property should be set as the accessibility text when present. If the `label` property is not present, the `placeholder` property should be used instead if present.
```
shalinijoshi19, line 92: What precisely do we imply by "set as the Accessibility text" ? What if the element has both a placeholder as well as a title/label defined since ideally placeholder wouldn't be use as an alternative to label ?

specs/elements/Input.Toggle.md
RebeccaAnne: Hmmm... Maybe in the release notes for the release this goes out in? The target audience for this particular document is intended to be renderer developers so here's probably not the place for that.

---
In reply to: [449975851](https://github.com/microsoft/AdaptiveCards/pull/4255#discussion_r449975851) [](ancestors = 449975851)

specs/elements/Input.Toggle.md
RebeccaAnne: This is a good question that I don't know the answer to.  @almedina-ms do you know what the expected behavior is here, or how it behaves today on the implemented platforms?

---
In reply to: [449977721](https://github.com/microsoft/AdaptiveCards/pull/4255#discussion_r449977721) [](ancestors = 449977721)

almedina-ms: When I wrote this I meant "accessibility text" as the text that is read by narrator but since I have changed my view in the matter, I don't think placeholder should be used as something the can be a replacement for labels as they don't behave the same and it's not recommended by the accessibility teams. Only label should be read by narrator, while we should make card authors aware of using placeholders as labels is bad practice

specs/elements/Input.Toggle.md
almedina-ms: I was thinking that we could take on this type of issues in blog posts and in the "Authoring cards" section of the website, authors should know about the accessibility support we're giving with this feature

specs/elements/Input.Toggle.md
RebeccaAnne: What is the best practice for exposing placeholder text to narrator in the case where it's not being used as a label? Presumably whatever is in there (a hint, or whatever) would be useful to narrator users as well, right? Does the accessibility team have a recommendation there? 

Also, if using placeholders as labels is bad practice, we probably shouldn't have our our main input scenario card sample card doing that :) (See [Inputs.json](https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.0/Scenarios/Inputs.json)). I can log a separate issue to update it if that makes sense to people. (cc @matthidinger)

image

PR #4262, opened 2020-06-30
[Android] Don't render empty text blocks

### LINKED ISSUE #3702: [Adaptive Cards /  Platform Discrepancy] Excess spacing in empty text blocks
# Web / Desktop

- [x] Android
- [x] iOS
- [x] JavaScript
- [x] UWP

#4305 tracks this issue for .NET

# Author or host

This is issue seen in the empty text playlod in Adaptive Card.

# Version of SDK

1.2.5

# Details

For empty text blocks – there is too much spacing rendered on Mobile as compared to Desktop (example in images below).
Empty text block elements should not be rendered at all.

JSON for these cards attached

Desktop: 
image

Mobile:
image

This issues in see in empty text block in Adaptive Card.

**Payload :**
```json
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "Container",
      "items": [
          {
          "type": "TextBlock",
          "text": "before"
        },
        {
          "type": "TextBlock",
          "text": ""
        },
        {
          "type": "TextBlock",
          "text": ""
        },
        {
          "type": "TextBlock",
          "text": "after"
        },
        {
          "type": "Image",
        }
      ]
    }
  ]
}
```

Input from @Susan Shen : 

@Vikrant Singh Checked the SDK code and I'm seeing that if the text is an empty string, the SDK renderer returns null for the object, and therefore, the TextBlock is not rendered. The behavior in Web/Desktop holds in the Adaptive Card Designer as well - you can use the same payload and then select Preview Mode to see that there's no spacing between the before and after. 

I think the overall behavior should be that if the developer wants to display an empty text block, they can create one by setting the text to " " (one space) instead of an empty string. 

This look like a discrepancy where for Desktop client A.C SDK Returns the null instance instead of a new line for empty strings?

golddove:
## Related Issue
Fixes #3702 

## Description
Ignore empty text blocks in Android renderer

## How Verified
Manual testing w/ sample payload in linked issue:

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4262)


jwoo-msft:
> @golddove
> Could you make sure that separator doesn't get rendered when TextBlock is not rendered?
> Renderers put separator between ui elements unless the separator is set to none.

NVM, i saw the separator codes few lines below.

almedina-ms:
Same for toggleVisibility, check that there's nothing broken from it, you could use this payload

```json

{
"type": "AdaptiveCard",
"version": "1.0",
"body": [
{
"type": "TextBlock",
"text":"",
"id": "id1"
}
],
"actions": [
{
"type":"Action.ToggleVisibility",
"id": ["id1"],
"title": "Toggle!"
}
]
}


```

almedina-ms:
Also, if the string is a space, should it be rendered?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
```diff
@@ -158,6 +158,11 @@ else if ((textBlock = TextBlock.dynamic_cast(baseCardElement)) == null)
             throw new InternalError("Unable to convert BaseCardElement to TextBlock object model.");
         }
 
+        if (textBlock.GetText().isEmpty())
```
almedina-ms, line 161: is a string of spaces valid?

jwoo-msft:
> Also, if the string is a space, should it be rendered?

We haven't decided it yet. fwiw, JS doesn't render the space or textblock but renders its separator in such case.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
jwoo-msft: We haven't decided it yet. fwiw, JS doesn't render the space or textblock but renders its separator in such case.

golddove:
> Same for toggleVisibility, check that there's nothing broken from it, you could use this payload
> 
> ```json
> {
> "type": "AdaptiveCard",
> "version": "1.0",
> "body": [
> {
> "type": "TextBlock",
> "text":"",
> "id": "id1"
> }
> ],
> "actions": [
> {
> "type":"Action.ToggleVisibility",
> "id": ["id1"],
> "title": "Toggle!"
> }
> ]
> }
> ```

@almedina-ms what should be the expected behavior here? Currently, the action does nothing, which is what I would expect.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
golddove: Right, we said spaces are the way to force-show an empty TextBlock, pending further discussions..

almedina-ms:
the expected result was for the renderer to not crash, if it's working, aka not crashing, then I'll approve

---
In reply to: [652061972](https://github.com/microsoft/AdaptiveCards/pull/4262#issuecomment-652061972) [](ancestors = 652061972)

golddove:
> the expected result was for the renderer to not crash, if it's working, aka not crashing, then I'll approve

Yup, working!

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
shalinijoshi19: Were you able to test this out with the samples that @jwoo-msft checked in with his change? 

---
In reply to: [447998584](https://github.com/microsoft/AdaptiveCards/pull/4262#discussion_r447998584) [](ancestors = 447998584)

shalinijoshi19:
we should add these test card cases to the tests folder for future regression testing.

---
In reply to: [652061972](https://github.com/microsoft/AdaptiveCards/pull/4262#issuecomment-652061972) [](ancestors = 652061972)

shalinijoshi19:
@golddove we'll also need to make the same change to UWP, .NET renderers as well before we call it done - are you comfortable enough to do that also?

PR #4265, opened 2020-06-30
[Android] Don't render empty text blocks

### LINKED ISSUE #3702: [Adaptive Cards /  Platform Discrepancy] Excess spacing in empty text blocks

golddove:
## Related Issue
Patches #4262 to v1.2 (fixes #3702)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4265)

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
@@ -158,6 +158,11 @@ public class TextBlockRenderer extends BaseCardElementRenderer
             throw new InternalError("Unable to convert BaseCardElement to TextBlock object model.");
         }
 
+        if (textBlock.GetText().isEmpty())
+        {
+            return null;
+        }
+
         TextView textView = new TextView(context);
 
         View separator = setSpacingAndSeparator(context, viewGroup, textBlock.GetSpacing(), textBlock.GetSeparator(), hostConfig, true);


shalinijoshi19:
Ideally you'd do this after you merged the changes to main to avoid duplicate code review comments and potentially needing to redo your changes twice!

PR #4282, opened 2020-07-02
[UWP] Sample custom input

### LINKED ISSUE #4281: [UWP][Extensibility] Adding an input value to a custom element crashes
# Platform

* UWP

# Details

If a new element is defined and that element is an input then it has to call the AddInputValue element, that element should have an underlying InternalId which is not a fact as new types defined by the user don't derive from AdaptiveCardElementBase, the implementation should change to use a string (which every input has) to use as index for getting the inputs

almedina-ms:
## Related Issue
Fixes #4281

## Description
Adds a sample custom Input element to test that labels and error messages are properly rendered for custom elements.

Input retrieval change
- The previous model we used for retrieving the inputs located in a card was perfect for our own data types but not for custom elements. So what was the difference? Inheriting from BaseCardElement. The aforementioned class provides us with an access to the InternalId but CustomElements must inherit from IAdaptiveCardElement and/or IAdaptiveInputElement so casting them to BaseCardElement in impossible and thus InternalId is not accessible at all. The change leaves us using the Id property found in all inputs as that is something we can access from all elements (our own or custom) and we enforce they must be non empty and unique so we don't have to do any more work.

## How Verified
1. A custom element was introduced so this card can be tested in future releases
2. A sample card was introduced so other renderers can use for testing purposes

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4282)


source/uwp/Visualizer/CustomElements/CustomInput.cs
```diff
@@ -0,0 +1,124 @@
+using System;
```
paulcam206, line 1: nit: missing copyright banner

source/uwp/Visualizer/CustomElements/CustomInput.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
```
paulcam206, line 4: are these using statements up to date?

source/uwp/Visualizer/CustomElements/CustomInput.cs
```diff
@@ -0,0 +1,124 @@
...
+        IList<AdaptiveRequirement> IAdaptiveCardElement.Requirements => throw new NotImplementedException();
+    }
+
+
+    public class CustomInputParser : IAdaptiveElementParser
+    {
+        public IAdaptiveCardElement FromJson(JsonObject inputJson, AdaptiveElementParserRegistration elementParsers, AdaptiveActionParserRegistration actionParsers, IList<AdaptiveWarning> warnings)
+        {
+            CustomInput customInput = new CustomInput();
+            customInput.Id = "id1";
```
paulcam206, line 56: I know we're avoiding having to write a full parser here, but consider using a random `Id` here (maybe a GUID?) just so we don't have to worry about future test cards using more than one `CustomInput` :)

source/uwp/Renderer/lib/AdaptiveInputs.h
```diff
@@ -38,7 +38,7 @@ namespace AdaptiveNamespace
         // Map with key: input internal id, value: input value class
         // This one has the collection of all input element values, this was introduced to be able to set the error message to the input value
```
shalinijoshi19, line 39: Do we need to update hte comment here?

source/uwp/Renderer/lib/AdaptiveInputs.cpp
```diff
@@ -85,12 +88,16 @@ namespace AdaptiveNamespace
...
-        RETURN_IF_FAILED(localInputElement.As(&inputElementImpl));
+        ComPtr<IAdaptiveCardElement> cardElement;
+        RETURN_IF_FAILED(localInputElement.As(&cardElement));
+
+        HString elementId;
+        RETURN_IF_FAILED(cardElement->get_Id(elementId.GetAddressOf()));
 
-        InternalId inputId = inputElementImpl->GetInternalId();
+        std::string utf8Id;
+        RETURN_IF_FAILED(HStringToUTF8(elementId.Get(), utf8Id));
```
shalinijoshi19, line 98: Can you explain this change a bit? So we are not working with the internalId but card id now?

shalinijoshi19:
> ## Related Issue
> Fixes #4281
> 
> ## Description
> Adds a sample custom Input element to test that labels and error messages are properly rendered for custom elements.
> 
> ## How Verified
> 1. A custom element was introduced so this card can be tested in future releases
> 2. A sample card was introduced so other renderers can use for testing purposes
> 
> ###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4282)

Can you also explain the code change in the description (this is more than a sample change) tx

source/uwp/Renderer/lib/AdaptiveInputs.cpp
almedina-ms: custom elements don't have an internal id (or they may but it's not accessible), that breaks the first solution I came up with, now we'll use the input Id for registering the input, this will increase the search complexity but at least we know that all inputs have an unique Id is an enforced requirement

PR #4293, opened 2020-07-03
Updated background image

### LINKED ISSUE #3700: [iOS][UI] Container background image does not cover card
# Platform
* iOS

# Author or host

*Host - Teams

# Version of SDK

*1.2.5

# Details

Container background image does not cover card 

Sample JSON

{"$schema":"http://adaptivecards.io/schemas/adaptive-card.json","type":"AdaptiveCard","version":"1.0","body":[{"type":"Container","items":[{"type":"TextBlock","text":"This is some text"},{"type":"Container","backgroundImage":{"url":"http://104.211.228.145:3000/images/assets/icons/Background.png"},"items":[{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"TextBlock","text":"New TextBlock"},{"type":"FactSet","facts":[{"title":"Fact 1","value":"Value 1"},{"title":"Fact 2","value":"Value 2"}]}]}],"height":"stretch"}]}

image

jwoo-msft:
## Related Issue
Fixed #3700 

## Description
The logic for cover is accurate.
 m * a >= x
 m * b >= y
 we want factor m that produces width and height when multiplied to a and b, that are equal or greater than x and y where a, b is the background image size, and x, y are size of super view we are trying to fill
 then m is max of (a/x, b/y)
however, the assumption that UIKit will increase the corresponding axis when one axis is set was not correct. the change is made such that constraint for the corresponding axis is explicitly set.

Additionally, refactored code, and fixed vertical alignment for cover mode as it was missing.

## How Verified

1. All relevant test cards are run to check for regression.
2. ran the repro steps.
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4293)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/UtiliOS.mm
```diff
@@ -713,3 +640,60 @@ void TextRunToRichTextElementProperties(const std::shared_ptr<TextRun> &textRun,
...
+        case HorizontalAlignment::Left:
+        default:
+            [imageView.leadingAnchor constraintEqualToAnchor:superView.leadingAnchor].active = YES;
+            break;
+    }
+}
+
+void configWidthAndHeightAnchors(UIImageView *imageView, UIView *superView, bool isComplimentaryAxisHorizontal)
+{
+    if (imageView && imageView.image && superView) {
```
shalinijoshi19, line 686: >mageView && imageView.image && superView [](start = 9, length = 40)

Super Nit: Might be easier to be consistent with parameter type placements for these config* routines..( for the others you have UIImageView as the last parameter while here it's the first etc

shalinijoshi19, line 686: >mageView && imageView.image && superView [](start = 9, length = 40)

What's the behavior for !imageView || !imageView.image || !superView? Should you explictly call that case out much like the pattern for the configHorizontal/Vertical routines above for consistent readability?

shalinijoshi19:
We should add teh specific test card as regression test for this scenario.

PR #4320, opened 2020-07-07
[Android] Custom input behaviour check

almedina-ms:
## Related Issue
Untracked issue

## Description
Adds samples for custom inputs to verify that labels are automatically rendered when a custom input element is rendered

## How Verified
1. A "CustomInput" class was added to the mobile app to test the correct behaviour, as can be seen from the image below, the label renders corrrectly

Screenshot_20200707-194158_AdaptiveCards Visualizer

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4320)


source/android/mobile/src/main/java/io/adaptivecards/adaptivecardssample/CustomObjects/CardElements/CustomInput.java
```diff
@@ -0,0 +1,127 @@
+package io.adaptivecards.adaptivecardssample.CustomObjects.CardElements;
+
```
shalinijoshi19, line 2: missing the Msft copyright header

source/android/mobile/src/main/java/io/adaptivecards/adaptivecardssample/CustomObjects/CardElements/CustomInput.java
```diff
@@ -0,0 +1,127 @@
...
+            renderedInput.setTag(new TagContent(baseCardElement));
+            viewGroup.addView(renderedInput);
+
+            BaseInputElement inputElement = Util.castTo(baseCardElement, BaseInputElement.class);
+            renderedCard.registerInputHandler(new CustomInputHandler(inputElement, renderedInput), renderArgs.getContainerCardId());
+            return renderedInput;
+        }
+    }
+
+
```
shalinijoshi19, line 85: Nit: Too many stray newlines

source/android/mobile/src/main/java/io/adaptivecards/adaptivecardssample/CustomObjects/CardElements/CustomInput.java
```diff
@@ -0,0 +1,127 @@
...
+            try
+            {
+                JSONObject obj = new JSONObject(jsonString);
+                element.SetId(obj.getString("id"));
+                element.SetLabel(obj.getString("label"));
+            }
+            catch (JSONException e)
+            {
+                e.printStackTrace();
+            }
```
shalinijoshi19, line 121: Is that right? If we can't parse the input json for the custom input element, do we just return defaults? If so this is yet another place where we should both annotate the sample and in case this is what we are doing for rendering native elements, update our documentation.

almedina-ms: the native elements parse in C++ and as so they throw a different type of exception, for samples we try to keep the behaviour simple, that's why we print the exception trace

shalinijoshi19:
@almedina-ms / @jwoo-msft / @RebeccaAnne what is the behavior here for other platforms outside of Android?

shalinijoshi19 (approved):

PR #4321, opened 2020-07-08
[Android] fix focus on elements

### LINKED ISSUE #4247: [Android] Focus not on Inputs (Input.ChoiceSet/Input.Number/input.Toggle) for validation failures
# Platform

* Android

# Details
Similar to #4230 looks like; Here's the screenshots for the various inputs:
## Input.ChoiceSet

image

## Input.Toggle

image

almedina-ms:
## Related Issue
Fixes #4247 

## Description
Only EditTexts in android are focusable by default so checkbox and radiobuttons had to be set focus in a different way, as focus is not as simple as first thought, input handlers will also have to implement the SetFocusToView method to catter for different needs.

## How Verified
The fix was verified manually using the InputChoiceSet.ErrorMessage.json card, now the behaviour makes it easier for users to know what input failed to validate
AndroidFocusFix


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/RadioGroupInputHandler.java
```diff
@@ -70,4 +71,18 @@ public void setInput(String choice)
...
 
+    @Override
+    public void setFocusToView()
+    {
+        RadioGroup radioGroup = (RadioGroup)m_view;
+
+        if (radioGroup.getChildCount() > 0)
+        {
+            RadioButton radioButton = (RadioButton)(radioGroup.getChildAt(0));
+
```
shalinijoshi19, line 82: Nit: additional newline creeped in

Issue #4318, opened 2020-07-07
[.NET WPF][RenderCardToImageAsync fails on certain cards]


Hammad-H:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* .NET WPF

# Details
The .NET WPF method RenderCardToImageAsync is unable to process certain payloads. I have tested with all our samples and the following are not working: 
* v1.1/Tests/Image.Stretch
* v1.1/Tests/ImageSet.Stretch
* v1.2/Tests/ColumnSet.AllElements.Stretch.ToggleVisibility
* v1.2/Tests/Container.HideAndShowElement
* v1.2/Tests/ToggleVisibility.AllElements
* v1.2/Elements/Action.ToggleVisibility

I have tested these cards with the WPF Visualizer and they are rendered correctly. However when I select the View as Image option, the visualizer crashes due to an unhandled stack overflow exception. 

![CardRender](https://user-images.githubusercontent.com/33607998/86846266-a67c6480-c078-11ea-9c5f-447ae0600747.PNG)
_AdaptiveVisitor.cs_
![exception](https://user-images.githubusercontent.com/33607998/86846322-b5fbad80-c078-11ea-805e-497078bc110f.PNG)

shalinijoshi19:
Hey @Vasanth-S would you be able to take a look at this one? Thanks!

PR #4326, opened 2020-07-08
Hammad h/test card consistency

Hammad-H:
## Link to Spec
[Consistent Cards Spec](https://github.com/microsoft/AdaptiveCards/blob/main/specs/DesignDiscussions/ConsistentCards.md)

# Description
The consistent cards project that currently supports .NET WPF, Node Js, Android and UWP. I have implemented the driver programs, REST API and the node Js web interface. The Readme in the ClientApp directory contains instructions to build and run the project locally on a windows machine.

## How Verified
1. Added integration tests for all the driver progams. These test the driver programs against all the cards present in the samples directory. Some cards mentioned in #4318 do not work with the .NET WPF driver program.
2. Added integration tests for the Flask Apps. These test the Flask Apps by sending various types of payloads and ensure that each time the Flask App is able to receive the data, send it to the driver and return the results back.
3. Added integration tests for the node js web interface. These test the web inteface by sending various types of payloads and ensure that each time the web interface is able to receive the data, send it to the Flask App and get a result back with a response code of 200. 

indexWithData
indexWithResults

## Deployment
- Deployment details can be found in the spec.
- The docker files and specific details to host each piece can be found in the branch hammad-h/deployment

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4326)


source/tools/testConsistency/ClientApp/app.js
```diff
@@ -0,0 +1,196 @@
+const express = require("express");
```
paulcam206, line 1: nit: we should have the standard copyright banner here (and in all source files):

https://github.com/microsoft/AdaptiveCards/blob/9b27ee8d79f0010ec246ec736e5722bb54f436f8/source/shared/cpp/ObjectModel/BaseElement.h#L1-L2

source/tools/testConsistency/ClientApp/app.js
```diff
@@ -0,0 +1,196 @@
...
+app.post('/', function(req, res) {
+  var cardData = req.body.cardData;
+  var cardTemplate = req.body.cardTemplate;
+
+  const formData = querystring.stringify({
+    "data" : cardData,
+    "template" : cardTemplate
+  });
+
+  const optionsForNodeJs = {
```
paulcam206, line 26: here and below -- we should separate out the per-platform bits of the code from the main parts of the code. where possible, configuration tasks should be separated out as data that is iterated over. this makes debugging, maintaining, and extending functionality easier (e.g. if I wanted to add a new platform, what all would I need to edit to get it working?)

source/tools/testConsistency/ClientApp/app.js
```diff
@@ -0,0 +1,196 @@
...
+    headers: {
+      'Content-Type' : 'application/x-www-form-urlencoded',
+      'Content-Length': Buffer.byteLength(formData)
+    }
+  };
+
+  const optionsForWPF = Object.assign({}, optionsForNodeJs);
+  optionsForWPF["port"] = 4000;
+
+
```
paulcam206, line 41: nit: unnecessary extra newline here

source/tools/testConsistency/ClientApp/app.js
```diff
@@ -0,0 +1,196 @@
...
+      nodeWarnings: nodeWarnings,
+      nodeErrors: nodeErrors,
+    });
+  });
+  
+});
+  
+//This test route is is simply to illustrate that we can handle the situation where we get incomplete JSON from the POST request
+app.post('/testIncompleteResult', function(req, res) {
+  var cardData = req.body.cardData;
```
paulcam206, line 108: a lot of the functionality here is duplicated from the `post` handler for `/` above -- common functionality should be factored out into helper functions to reduce duplication (as a bonus, this also improves readability)

source/tools/testConsistency/ClientApp/app.test.js
```diff
@@ -0,0 +1,113 @@
+// To run these tests, first ensure that the flask app for nodeJs is running at 127.0.0.1 port 5000
```
paulcam206, line 1: if possible, it would be nice if this was configurable on the command line

source/tools/testConsistency/ClientApp/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "clientapp",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "jest"
+  },
+  "author": "",
+  "license": "ISC",
```
paulcam206, line 10: these fields should align with the values we use in our public nodejs packages:

https://github.com/microsoft/AdaptiveCards/blob/9b27ee8d79f0010ec246ec736e5722bb54f436f8/source/nodejs/adaptivecards/package.json#L5-L6

source/tools/testConsistency/ClientApp/views/index.ejs
```diff
@@ -0,0 +1,15 @@
...
+<html>
+  <head>
+
+  </head>
+  <body>
+
+    <h1>Consistent Cards</h1>
+    <form action = "/" method=POST>
+      <textarea placeholder="card data" name ="cardData"></textarea><br>
+      <textarea placeholder="card template" name ="cardTemplate"></textarea><br>
```
paulcam206, line 11: totally fine for now, but if you want to snazz this up later, let me know -- we can get this hooked up with a proper editor (like the one used in the designer)

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,114 @@
+using System;
+using System.IO;
+using System.Collections.Generic;
+using System.Linq;
```
paulcam206, line 4: I'm guessing that you probably don't use all of this `using` statements -- I think VS has a cleanup you can run to remove the ones you don't need :)

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using AdaptiveCards;
+using AdaptiveCards.Rendering.Wpf;
+using AdaptiveCards.Templating;
+using Newtonsoft.Json.Linq;
+
+
```
paulcam206, line 13: nit: unnecessary extra newline

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+            JObject resultJson = new JObject();
+            List<string> errorList = new List<string>();
+            List<string> warningList = new List<string>();
+            resultJson.Add("imageData", "");
+            resultJson.Add("warnings", "");
+            resultJson.Add("errors", "");
+            Console.OutputEncoding = Encoding.UTF8;
+            try
+            {  
+                string inputString = "";
```
paulcam206, line 29: consider using a `StringBuilder` here for perf

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+                string currLine;
+            
+                //read a multi line input from stdin
+                do
+                {
+                    currLine = Console.ReadLine();
+                    inputString = string.Concat(inputString, currLine);
+                }
+                while ((currLine != null) && (currLine != ""));
```
paulcam206, line 38: are you sure that `currLine != ""` is correct here?

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+                string cardData = jsonInput.GetValue("data").ToString();
+                string cardTemplate = jsonInput.GetValue("template").ToString();
+          
+                AdaptiveCardTemplate adaptiveCardTemplate = new AdaptiveCardTemplate(cardTemplate);
+                string adaptiveCardJson = adaptiveCardTemplate.Expand(cardData);
+                AdaptiveCardParseResult parseResult = AdaptiveCard.FromJson(adaptiveCardJson);
+                AdaptiveCardRenderer renderer = new AdaptiveCardRenderer();
+
+                GetResultJsonString(parseResult, renderer, resultJson, warningList).GetAwaiter().GetResult();
+
```
paulcam206, line 50: nit: unnecessary newline

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+          
+                AdaptiveCardTemplate adaptiveCardTemplate = new AdaptiveCardTemplate(cardTemplate);
+                string adaptiveCardJson = adaptiveCardTemplate.Expand(cardData);
+                AdaptiveCardParseResult parseResult = AdaptiveCard.FromJson(adaptiveCardJson);
+                AdaptiveCardRenderer renderer = new AdaptiveCardRenderer();
+
+                GetResultJsonString(parseResult, renderer, resultJson, warningList).GetAwaiter().GetResult();
+           
+            }
+
```
paulcam206, line 52: nit: there shouldn't be additional newlines between each catch statement

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRendererTests/UnitTest1.cs
```diff
@@ -0,0 +1,137 @@
+using System;
```
paulcam206, line 1: this file should probably have a more descriptive name (as should the test class)

source/tools/testConsistency/ClientApp/app.test.js
Hammad-H: added a bat file in the client app folder that does all this work for us

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
shalinijoshi19: @paulcam206 what's our story/guidance for .NET so far? I was trying to look at our existing .editorconfigs and looks like .NET doesnt have any defined yet..Do we have stylecop running in teh meanwhile (stylecop.analyzers nuget package).. Or adding an .editorconfig for .NET following [these docs](https://docs.microsoft.com/en-us/visualstudio/ide/code-styles-and-code-cleanup?view=vs-2019)

shalinijoshi19:
Go ahead check in the spec for this (.docx is fine for now until we can convert it) and then add a link to it here in this PR for cross-reference.

source/tools/testConsistency/ClientApp/app.js
```diff
@@ -0,0 +1,120 @@
+ // Copyright (c) Microsoft Corporation. All rights reserved. 
+ // Licensed under the MIT License. 
+const express = require("express");
```
shalinijoshi19, line 3: Are we running this through Semmle automatically at this point? @paulcam206?

shalinijoshi19: If we not we should run linters manually through the code base to ensure we are clean/kosher.

---
In reply to: [453430695](https://github.com/microsoft/AdaptiveCards/pull/4326#discussion_r453430695) [](ancestors = 453430695)

source/tools/testConsistency/ClientApp/app.js
```diff
@@ -0,0 +1,120 @@
...
+const path = require('path');
+const bodyParser = require('body-parser');
+const querystring = require('querystring');
+const optionsArray = require("./options");
+const postRequest = require("./postRequest");
+
+const getPostRequestPromise = postRequest.getPostRequestPromise;
+
+const app = express();
```
shalinijoshi19, line 12: somewhere there should be a class comment explaining what exactly it is that this "app" does; Ideally we'd also include a readme.md

source/tools/testConsistency/ClientApp/app.js
```diff
@@ -0,0 +1,120 @@
...
+app.set('view engine', 'ejs');
+app.set('views', path.join(__dirname, 'views'));
+app.use(bodyParser.urlencoded({extended: false}));
```
shalinijoshi19, line 15: >{extended: false [](start = 30, length = 16)

Apparently using extended: false has been deprecated - See [https://expressjs.com/en/resources/middleware/body-parser.html](https://expressjs.com/en/resources/middleware/body-parser.html); I would also recommend explaining wiht quick comments your non-default choices in general for parameters. Also good to link to the corresponding doc links for reference for the code reader in your broader class/file comments.

source/tools/testConsistency/ClientApp/postRequest.js
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. 
+// Licensed under the MIT License. 
+const http = require('http');
+
+function getPostRequestPromise(options, platformName, formData) {
+    const postReqPromise = new Promise(function(resolve, reject) {
```
shalinijoshi19, line 6: A comment here would be nice explainng what is going on - esp with this long of an anonymous function :))

source/tools/testConsistency/ClientApp/server.js
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. 
+// Licensed under the MIT License. 
+const app = require("./app");
+
+app.listen(8000, function(){
```
shalinijoshi19, line 5: >8000 [](start = 11, length = 4)

Can we make the listen port configurable instead of the hardcoded 8000?

source/tools/testConsistency/ClientApp/startProject.bat
```diff
@@ -0,0 +1,7 @@
+@echo off
+cd ../nodejs-renderer
+START flask run 
+cd ../dotnetWPF-renderer
+START flask run -p 4000
```
shalinijoshi19, line 5: >-p 4000 [](start = 16, length = 7)

What does this port (?) signify? Also we should have some comments here explaining what this script does and what it's used for. Comments here but also a readme.md to explain how to use/run this

source/tools/testConsistency/ClientApp/app.test.js
```diff
@@ -0,0 +1,115 @@
...
+      .send(cardTemplate=template, cardData="")
+      .then(function(response){
+          expect(response.statusCode).toBe(200);
+          done();
+      });
+});
+
+test("Sending no card template should still return a response code 200", function(done) {
+    const data = {
+                    "name": "Matt Hidinger"
```
shalinijoshi19, line 32: >Matt Hidinger [](start = 29, length = 13)

:) I think we can cook up a different name here instead of using Matt's yet again (also privacy in general though not that big of a deal since he's everywhere :))

source/tools/testConsistency/ClientApp/app.test.js
```diff
@@ -0,0 +1,115 @@
...
+test("The GET request to the index page should return a response code 200", function(done) {
+     request(app)
+      .get("/")
+      .then(function(response) {
+        expect(response.statusCode).toBe(200);
+        done();
+      });
+});
+
+test("Sending no card data should still return a response code 200", function(done) {
```
shalinijoshi19, line 18: >Sending no card data should still return a response code 200", [](start = 6, length = 62)

Nit nit: I think you can simply describe a very high level expecatation of success here (rather than going into 200 (which is implied here when talking about REST and successfult HTTP Status codes)-- So i am more interested here in the actual success scenario you are testing for - eg NonTemplatedCardsAreSupported or something like that as teh test name and the description which can explain "static cards without any template parameter specified should be accepted by the API" or something like that.

source/tools/testConsistency/ClientApp/app.test.js
```diff
@@ -0,0 +1,115 @@
...
+    request(app)
+      .post("/")
+      .send(cardTemplate="", cardData="")
+      .then(function(response){
+          expect(response.statusCode).toBe(200);
+          done();
+      });
+});
+
+test("Sending invalid JSON data should still return a response code 200", function(done) {
```
shalinijoshi19, line 56: >d JSON data should still  [](start = 20, length = 25)

Interesting - so can i send in any old string here as well instead of only well-formed json? If we are not validating our parameters for being well-formed json then why are we defining our parameters as accepting json data when it really should be any string right? (I am guessing the reason is because we want to let the actual per-platform serialization and renderer stacks catch and report the invalid json errors rather than intervening ourselves and causing potential confusion?)

source/tools/testConsistency/ClientApp/app.test.js
```diff
@@ -0,0 +1,115 @@
...
+    const data = '{"title" : "invalid card data" ';
```
shalinijoshi19, line 57: We should actually test for invalid parameters as far as the actual POST is concerned - eg both null strings (which i am guessing is invalid, or an otherwise empty POST request body altogether (unsure if you're actually able to simulate that here using the test library?)

source/tools/testConsistency/ClientApp/views/index.ejs
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <head>
```
shalinijoshi19, line 3: Probably need the msft copy right header using the comment syntax here as well *<%# Comment tag, no execution, no output)

source/tools/testConsistency/dotnetWPF-renderer/test_app.py
```diff
@@ -0,0 +1,152 @@
+# Copyright (c) Microsoft Corporation. All rights reserved. 
+# Licensed under the MIT License. 
+import unittest
+import json
+from app import app
+from copy import deepcopy
+
+
+POST_URL = 'http://127.0.0.1:5000/'
```
shalinijoshi19, line 9: Ideally this would be configurable to allow for different listening ports for the server

source/tools/testConsistency/dotnetWPF-renderer/test_app.py
```diff
@@ -0,0 +1,152 @@
...
+TEST_PARTIAL_RESULT_URL = 'http://127.0.0.1:5000/test'
+
+class TestFlaskApi(unittest.TestCase):
+    def setUp(self):
+        #self.backup_items = deepcopy(app.items)  # no references!
```
shalinijoshi19, line 14: > #self.backup_items = deepcopy(app.items)  # no references! [](start = 7, length = 59)

Why is this commented out again? Ideally avoid commented out source code; If you haev to then add a comment explaining the comment :)

source/tools/testConsistency/dotnetWPF-renderer/test_app.py
```diff
@@ -0,0 +1,152 @@
...
+        self.app = app.test_client()
+        self.app.testing = True
+
+    def testEmptyData(self):
+        jsonFile = open('../../../../samples/v1.0/Scenarios/ActivityUpdate.json')
```
shalinijoshi19, line 19: >('../../../../samples/v1.0/Scenarios/ActivityUpdate.json') [](start = 23, length = 58)

This would also come in from a test execution parameter or something to allow the test to run from a location other than the current location of the source file here

source/tools/testConsistency/dotnetWPF-renderer/test_app.py
```diff
@@ -0,0 +1,152 @@
...
+import unittest
+import json
+from app import app
+from copy import deepcopy
+
+
+POST_URL = 'http://127.0.0.1:5000/'
+TEST_PARTIAL_RESULT_URL = 'http://127.0.0.1:5000/test'
+
+class TestFlaskApi(unittest.TestCase):
```
shalinijoshi19, line 12: >TestFlaskApi [](start = 6, length = 12)

Do add comments in the code explaining what the general purpose of this class is and also add names/descriptions to the actual tests themselves to make it very self-evident say in a run log as to what the test is and thus what it is that it's testing.

source/tools/testConsistency/dotnetWPF-renderer/test_app.py
```diff
@@ -0,0 +1,152 @@
...
+                        "version": "1.0",
+                        "body": [
+                                    {
+                                        "type": "TextBlock",
+                                        "text": "Hello ${name}!"
+                                    }
+                                ]
+                    }
+        data = {
+                   "name": "Matt Hidinger"
```
shalinijoshi19, line 126: > "Matt Hidinger" [](start = 26, length = 16)

 Remove; Pls use fake data/names for privacy concerns.. tx

source/tools/testConsistency/dotnetWPF-renderer/app/routes.py
```diff
@@ -0,0 +1,67 @@
...
+def test():
+        """This POST request is primarily for testing purposes. 
+
+            It accepts JSON data via HTTP POST, creates a new JSON string with the new 
+            fields and sends it a mock WPF app. It receives an incomplete JSON string that contains only 
+            part of a base64encoded string. It serves to demonstrate the scenario where the driver program 
+            crashes midway or is not able to send back all the data.
+
+            Returns:
+                A string that contains only part of the base64 encodeded image.
```
shalinijoshi19, line 51: nitnit: typo.

source/tools/testConsistency/dotnetWPF-renderer/app/routes.py
```diff
@@ -0,0 +1,67 @@
...
+                                    input=card_byte_buffer,
+                                    capture_output=True 
+                                    )
+        img_json_str = (res_json.stdout).decode('utf-8')
+        return img_json_str
+
+@app.route('/test', methods=['POST'])
+def test():
+        """This POST request is primarily for testing purposes. 
+
```
shalinijoshi19, line 44: Ideally this would be simulated on the test client instead of adding mocks in the actual app code. Either ways though this should be behind a config switch that only your test client sets and is allowed to call and not the whole world

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/sendPartialResult/SendPartialResult.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. 
+// Licensed under the MIT License.
+using System;
+using System.Text;
+
+namespace sendPartialResult
```
shalinijoshi19, line 6: >sendPartialResult [](start = 10, length = 17)

I am not sure I understand what is going on here with this separate project and why you need to include this as part of the renderer project?

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace WPFRenderer
+{
+    public class Program
```
shalinijoshi19, line 17: again class comments here would be nice and so would a project level readme; Also we already have a WPF renderer in our SDK - can we name this to be slightly less ambiguous? Is the WPFDriver?

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+            catch (AdaptiveSerializationException e)
+            {
+                errorList.Add("Exception: " + e.Message);
+                errorList.Add("The supplied json input is not a valid Adaptive Card");
+
+            }
+            catch (ArgumentNullException e)
+            {
+                errorList.Add("Exception: " + e.Message);
+                errorList.Add("Unable to capture the image of the card. Ensure that the card template and card data are valid and try again.");
```
shalinijoshi19, line 72: Hmm who's throwing this? Also this program-specific null checks for the input parameters should be performed upfront if they aren't already - we should know at this point whether the data or template they'd specified was null or not and thus detect this upfront and fail the request; If this is somethign that the renderer is throwing then that's different (though I'd still love to udnerstand why it's an ArgumentNullException)

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+            catch (Exception e)
+            {
+                errorList.Add("Exception: " + e.Message);
+                errorList.Add("An error was encountered: " + e.Message);
```
shalinijoshi19, line 77: How are you disambiguating WPF driver errors from say JS driver errors etc? Do we want to annotate the error/warning list or does it get disambiguated on the "client"?

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
```
shalinijoshi19, line 78: These 2 cases actually look like these should result in failing the actual POST request eventually - is that the case? Do errors end up completing the originating client request with a non-200 (OK) response or does the client always have to dig into the POST response body to figure that out?

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+                resultJson["warnings"] = new JArray(warningList);
+                resultJson["errors"] = new JArray(errorList);
+                Console.WriteLine(resultJson.ToString());
+            }
+        }
+
+        private static async Task GetResultJsonString(AdaptiveCardParseResult parseResult, AdaptiveCardRenderer renderer, JObject resultJson, List<string> warningList)
+        {
+            AdaptiveCard adaptiveCard = parseResult.Card;
+            CancellationTokenSource cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
```
shalinijoshi19, line 91: >ds(30)) [](start = 88, length = 7)

Nit: hardcoded magic numbers! This should be a configurable value encapsulated in a variable - you can then actually configure this variable to be say 1s to simulate failures in your testing.

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+            RenderedAdaptiveCardImage adaptiveCardImage = await renderer.RenderCardToImageAsync(adaptiveCard, true, 400, cts.Token);
+
+            foreach (var warning in parseResult.Warnings.Union(adaptiveCardImage.Warnings))
+            {
+                warningList.Add(warning.Message);
+            }
+
+            Stream cardImgStream = adaptiveCardImage.ImageStream;
```
shalinijoshi19, line 99: >Stream cardImgStream = adaptiveCardImage.ImageStream; [](start = 12, length = 53)

Are you potentially leaking any native resources here? Does ImageStream require you to dispose off the stream after using? (in other words does it implement a Dispose() pattern?) If so this would be under a using() statement to automatically invoke stream.Dispose() and release any native handles it might have been holding on to after going out of scope.

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+                JObject jsonInput = JObject.Parse(inputSb.ToString());
+                string cardData = jsonInput.GetValue("data").ToString();
+                string cardTemplate = jsonInput.GetValue("template").ToString();
+
+                AdaptiveCardTemplate adaptiveCardTemplate = new AdaptiveCardTemplate(cardTemplate);
+                string adaptiveCardJson = adaptiveCardTemplate.Expand(cardData);
+                AdaptiveCardParseResult parseResult = AdaptiveCard.FromJson(adaptiveCardJson);
+                AdaptiveCardRenderer renderer = new AdaptiveCardRenderer();
+
+                GetResultJsonString(parseResult, renderer, resultJson, warningList).GetAwaiter().GetResult();
```
shalinijoshi19, line 50: >GetResultJsonString(parseResult, renderer, resultJson, warningList).GetAwai [](start = 16, length = 75)

Hmm.. so you are using Tasks but then ultimately blocking on it anyways - so what's the point of spinning this up in the background again?

source/tools/testConsistency/ClientApp/app.test.js
Hammad-H: That is correct. I want to ensure that in case of an invalid json string, all errors are handled by the renderers so that we can ensure consistent error handling as well

source/tools/testConsistency/dotnetWPF-renderer/test_app.py
Hammad-H: So each test has a description. There is a general way of running python unit tests which is 
`python -m unittest -v` in the directory that contains the testing script. This will run all the tests along with descriptions as to what is being tested

source/tools/testConsistency/dotnetWPF-renderer/app/routes.py
Hammad-H: I wasn't sure how to simulate the situation where we only get an incomplete result. So my solution was to create a test route that would be call a different driver program. The driver program would function in a similar manner as the actual driver except it would not actually render the card. It would simply send a partial/incomplete base64 encoded string. 
Does that sound reasonable? Let me know if there is a better way to simulate such a situation

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/sendPartialResult/SendPartialResult.cs
Hammad-H: This is primarily for testing purposes. It simulates the situation where we are only able to receive partial results. This is called in the test route of the flask app

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
Hammad-H: It gets disambiguated on the client. The drivers are not aware of other drivers and each sends its own set of results. These are all disambiguated at the client level. The client collects all the results and displays them appropriately

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
Hammad-H: The response code will be 200. The client will look into the body and display the error it received.

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
Hammad-H: The method renderCardToImageAsync in the .NET WPF is an asynchronous method. I wanted to ensure synchronous execution of my code. Therefore, this was my workaround. I would like to know if there is a better way since I do not have much experience with asynchronous code in C#

source/tools/testConsistency/ClientApp/startProject.bat
Hammad-H: Added comments in the bat file and in the readme

source/tools/testConsistency/ClientApp/app.test.js
Hammad-H: I tested for null strings. There is a test there.  Regarding long strings, I tested some cards that were extremely long (the imageStretch under samples) and they do work

source/tools/testConsistency/ClientApp/views/index.ejs
Hammad-H: Added

shalinijoshi19:
> ## Link to Spec
> #4475
> 
> # Description
> The consistent cards project that currently supports .NET WPF and Node Js. I have implemented the driver programs, REST API and the node Js web interface. To run the project navigate to the nodejs-renderer directory and enter the command `flask run`. Then navigate to the dotnetWPF-renderer directory and enter the command `flask run -p 4000`. Finally navigate to the Client App directory and enter the command `node server.js`. Then, from a browser, go to http://localhost:8000/
> 
> ## How Verified
> How you verified the fix, including one or all of the following:
> 
> 1. Added integration tests for both of the driver progams. These test the driver programs against all the cards present in the samples directory. Some cards mentioned in #4318 do not work with the .NET WPF driver program.
> 2. Added integration tests for the Flask Apps. These test the Flask Apps by sending various types of payloads and ensure that each time the Flask App is able to receive the data, send it to the driver and return the results back.
> 3. Added integration tests for the node js web interface. These test the web inteface by sending various types of payloads and ensure that each time the web interface is able to receive the data, send it to the Flask App and get a result back with a response code of 200.
> 
> indexWithData
> indexWithResults
> 
> ## Deployment
> * Deployment details can be found in the spec.
> * The docker files and specific details to host each piece can be found in the branch hammad-h/deployment
> 
> ###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4326)

Probably want to update the description now at this point to reflect the current state of the project? Thanks!

shalinijoshi19:
There's also conflicts in your branch that you probably want to resolve.

shalinijoshi19:
Hey @Hammad-H can you actually resolve and/or respond to the comments from earlier so it makes it easier to track where we are at for subsequent code reviews? Thanks!

samples/v1.3/Scenarios/FoodOrderv2.json
```diff
@@ -1,151 +0,0 @@
-{
-	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
-	"type": "AdaptiveCard",
-	"version": "1.0",
-	"body": [
-		{
-			"type": "TextBlock",
-			"text": "Your registration is almost complete",
```
shalinijoshi19, line 8: hmm - do you have the rebasing done right ? I dont think we want to delete these samples from main.

source/nodejs/adaptivecards/src/strings.ts
```diff
@@ -1,37 +0,0 @@
-export class Strings {
-    static readonly errors = {
-        unknownElementType: (typeName: string) => `Unknown element type "${typeName}". Fallback will be used if present.`,
-        unknownActionType: (typeName: string) => `Unknown action type "${typeName}". Fallback will be used if present.`,
-        elementTypeNotAllowed: (typeName: string) => `Element type "${typeName}" is not allowed in this context.`,
-        actionTypeNotAllowed: (typeName: string) => `Action type "${typeName}" is not allowed in this context.`,
-        invalidPropertyValue: (value: any, propertyName: string) => `Invalid value "${value}" for property "${propertyName}".`,
-        showCardMustHaveCard: () => `"An Action.ShowCard must have its "card" property set to a valid AdaptiveCard object.`,
-        invalidColumnWidth: (invalidWidth: string) => `Invalid column width "${invalidWidth}" - defaulting to "auto".`,
```
shalinijoshi19, line 9: Definitely dont check this in :) There's a bunch of deletes here which i am not sure is what we want!

source/tools/testConsistency/android-renderer/AndroidDriver/build.gradle
```diff
@@ -0,0 +1,24 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
+buildscript {
+    repositories {
+        google()
+        jcenter()
```
shalinijoshi19, line 5: @golddove / @almedina-ms could you double check the gradle files here to make sure these match the new format you guys are trying to move to and/or remove any redundancies? thanks

source/tools/testConsistency/android-renderer/AndroidDriver/app/src/androidTest/assets/README.md
```diff
@@ -0,0 +1 @@
+Add the samples directory here to run the tests
```
shalinijoshi19, line 1: Do we actually need this readme or can we have a top-level readme instead?

shalinijoshi19: Also why assets + resources both? I see samples appear in the resources folder and not much in the assets folder? 

---
In reply to: [471243745](https://github.com/microsoft/AdaptiveCards/pull/4326#discussion_r471243745) [](ancestors = 471243745)

source/tools/testConsistency/android-renderer/AndroidDriver/app/src/androidTest/resources/ActivityUpdate.json
```diff
@@ -0,0 +1,822 @@
+{
+    "type": "AdaptiveCard",
+    "body": [
+      {
+        "type": "Container",
```
shalinijoshi19, line 5: Why do we need to actually check the same file in different places?

source/tools/testConsistency/android-renderer/AndroidDriver/app/src/test/java/com/example/androiddriver/ExampleUnitTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+public class ExampleUnitTest {
+    @Test
```
shalinijoshi19, line 13: What's the point of this file? Do we need it? Where are the actual tests again?

source/tools/testConsistency/dotnetWPF-renderer/WPFRenderer/WPFRenderer/Program.cs
Hammad-H: This sometimes occurs when an image is not captured. Pretty sure it is thrown by the RenderCardToImageaAync

source/tools/testConsistency/dotnetWPF-renderer/test_app.py
Hammad-H: I modified it so that the POST_URL is simply the route instead of the URL

source/tools/testConsistency/android-renderer/AndroidDriver/app/src/test/java/com/example/androiddriver/ExampleUnitTest.java
Hammad-H: They are under the directory android test. This is an unnecessary test file that is created by android studio by default. i can get rid of it

source/nodejs/adaptivecards/src/strings.ts
Hammad-H: I resolved these conflicts :)

paulcam206:
I'm going to close this PR for now -- we can look at resurrecting it when we have resources to stand it up :)

PR #4346, opened 2020-07-10
[Android] Concurrent image load

almedina-ms:
## Related Issue
Untracked issue

## Description
The issue was that we were using the method "execute" when running the download image async tasks, quoting the [official documentation](https://developer.android.com/reference/android/os/AsyncTask#execute(Params...)), this api has suffered some changes across different android api versions

> Note: this function schedules the task on a queue for a single background thread or pool of threads depending on the platform version. When first introduced, AsyncTasks were executed serially on a single background thread. Starting with Build.VERSION_CODES.DONUT, this was changed to a pool of threads allowing multiple tasks to operate in parallel. Starting Build.VERSION_CODES.HONEYCOMB, tasks are back to being executed on a single thread to avoid common application errors caused by parallel execution. If you truly want parallel execution, you can use the executeOnExecutor(Executor, Params...) version of this method with THREAD_POOL_EXECUTOR; however, see commentary there for warnings on its use.

Because of this, all calls to the method ```execute``` were changed to ```executeOnExecutor```

## How Verified
The fix was verified using a modified version of the ImageGallery card, by bumping the number of images loaded we can see how images are not being downloaded sequentially

AndroidMultiDownload
 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4346)

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
@@ -4,6 +4,7 @@ package io.adaptivecards.renderer;
 
 import android.content.Context;
 import android.graphics.Color;
+import android.os.AsyncTask;
 import android.support.v4.app.FragmentManager;
 import android.view.Gravity;
 import android.view.View;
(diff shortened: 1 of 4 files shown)


shalinijoshi19:
@almedina-ms this should first get into main; Then cherry-picked into release/1.2.. Btw is this supposed to go into July 20.07 patch for 1.2?

PR #4363, opened 2020-07-10
[Spec] Update Input Label Spec to improve readability and consistency

RebeccaAnne:
## Description
Updated the input label spec to contain the latest plan, and to move extraneous information to the appendix. It was not my intent with this change to make any changes to what is in plan for the release, so let me know if something seems inaccurate or not consistent with your understanding of what is in plan for this version

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4363)

### DIFF
--- a/specs/DesignDiscussions/InputLabels.md
+++ b/specs/DesignDiscussions/InputLabels.md
@@ -116,7 +116,7 @@ The basic addition that we are going to make to the schema is to add a label str
 
 ### Rendering
 
-The label will be rendered above the input as if the card author had used a TextBlock to label their input. It will ld have the same behavior as a TextBlock regarding handling empty/white space only text, and be drawn only if there is a non-empty non-whitespace value. 
+The label will be rendered above the input as if the card author had used a TextBlock to label their input. It will have the same behavior as a TextBlock regarding handling empty/white space only text, and be drawn only if there is a non-empty non-whitespace value. 
 
 It is important to note that wrapping will always be performed for this type of labels as not doing so would provide a bad user experience. If users can't read the whole label for an input they may not be able to provide a correct answer.
 
@@ -179,6 +179,8 @@ By default, we will mark required inputs with a `*`. We will also, however, prov
 	}
 }
 ```
+
+> Note: If an input is marked as required but does not have the label property set, no visual indicator will be shown indicating that the input is required. Renderers will return a warning in this case that can be surfaced to card authors in the designer and visualizers to warn authors of this behavior.
 
 ### Card Author Formatting
 This version of input labels does not allow for card authors to provide their own formatting for labels. Survey results show that this is not commonly done, and introducing it would create conflict with the styling provided by the host. For further discussion of ways to possibly add card author formatting in the future, see the Appendix. 
--- a/specs/DesignDiscussions/InputLabels.md
+++ b/specs/DesignDiscussions/InputLabels.md
@@ -1,7 +1,8 @@
 # Table of Contents
 1. [Input/Label Associations](#InputLabel-Associations)
 	1. [Basic Label Support in Adaptive Cards](#Basic-Label-Support-in-Adaptive-Cards) 
-		1. [Host Formatting](#Host-Formatting)
+		1. [Rendering](#Rendering)
+		2. [Host Formatting](#Host-Formatting)
 			1. [Text Formatting](#Text-formatting)
 			2. [Required indicators](#Required-indicators) 
 		3. [Card Author Formatting](#Card-Author-Formatting)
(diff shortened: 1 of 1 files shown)


specs/DesignDiscussions/InputLabels.md
```diff
@@ -91,7 +88,7 @@ In that card, the label for the input is “Please enter your name:”, however
 
  - Responsiveness: One direction for the future of cards involves re-flowing the layout of a card based on the size of the display. Knowing that a particular label goes with a particular input may help us ensure that semantically related elements stay together on reflow.
 
-## Basic Label Support in Adaptive Cards
+ ## Basic Label Support in Adaptive Cards
 
 The basic addition that we are going to make to the schema is to add a label property to all input types. This approach provides a simple and concise way to label inputs, that hopefully encourages card authors to use it and receive the benefits of improvements to the issues called out above.
```
shalinijoshi19, line 93: >dd a label property t [](start = 66, length = 21)

We should specify upfront that this will accept values of type "string"

specs/DesignDiscussions/InputLabels.md
```diff
@@ -115,55 +112,16 @@ The basic addition that we are going to make to the schema is to add a label pro
...
-		"type": "TextBlock",
-		"text": "Label for Input"
-	}
-}
-```
-
-Both of the afore mentioned cards would produce the same rendered result that would look like  this:
- 
-![img](assets/InputLabels/LabelDefault.PNG)
+The label should be rendered above the input as if the card author had used a TextBlock to label their input.
```
shalinijoshi19, line 115: >label should be rendered a [](start = 4, length = 26)

Does this need to be in its own "Renderer" sub-section? Is this the default behavior or the only behavior for hosts at this point? Can hosts customize this layout?

shalinijoshi19:
> ## Accessibility

This section is missing discussion of per-platform native Accessibility styles/support that we will be using; For ex there should ideally be a discussion around for JS, how the new label element is going to be annotated so as to correctly guide accessibility tools like Narrator etc for each platform. -That;s still happening for v1 yes? 

---
Refers to: specs/DesignDiscussions/InputLabels.md:349 in f38a439. [](commit_id = f38a439cf1ad170d3de190086a5eef7b8a54a7ca, deletion_comment = False)

shalinijoshi19:
> ### XAML

This is great - this is what i was looking for - a quick TL;DR summary of what per-platform properties we actually ended up using to implement labels

---
Refers to: specs/DesignDiscussions/InputLabels.md:405 in f38a439. [](commit_id = f38a439cf1ad170d3de190086a5eef7b8a54a7ca, deletion_comment = False)

RebeccaAnne:
> ### XAML

Unfortunately, this section isn't actually doing that. I wrote this before we had decided on an approach to look at other platforms for comparison to inspire our own schema choices. It's true that it could serve a double purpose as implementation guidance, but it's not explicitly serving that purpose now (for example, we didn't end up using XAML headers to implement our Adaptive Card labels at least in UWP).

That's not to say that platform specific details wouldn't be useful. We could certainly add a section for that. But it's beyond the scope of what this document is currently doing and would require additional investment here.

---
In reply to: [658460570](https://github.com/microsoft/AdaptiveCards/pull/4363#issuecomment-658460570) [](ancestors = 658460570)

---
Refers to: specs/DesignDiscussions/InputLabels.md:405 in f38a439. [](commit_id = f38a439cf1ad170d3de190086a5eef7b8a54a7ca, deletion_comment = False)

RebeccaAnne:
> ## Accessibility

We're not really covering platform specific design and implementation details in this doc at all. I can log a separate issue to track updating the docs with platform specific design choices.

---
In reply to: [658455149](https://github.com/microsoft/AdaptiveCards/pull/4363#issuecomment-658455149) [](ancestors = 658455149)

---
Refers to: specs/DesignDiscussions/InputLabels.md:349 in f38a439. [](commit_id = f38a439cf1ad170d3de190086a5eef7b8a54a7ca, deletion_comment = False)

specs/DesignDiscussions/InputLabels.md
RebeccaAnne: Sure, i put a section header here. There's no position formatting at this time (all the available host formatting is described in the host formatting section) but the discussion of possible positioning changes still exists in the Appendix.

---
In reply to: [454691954](https://github.com/microsoft/AdaptiveCards/pull/4363#discussion_r454691954) [](ancestors = 454691954)

RebeccaAnne:
> ## Accessibility

Logged #4363 to track adding platform specific implementation details.

---
In reply to: [658916743](https://github.com/microsoft/AdaptiveCards/pull/4363#issuecomment-658916743) [](ancestors = 658916743,658455149)

---
Refers to: specs/DesignDiscussions/InputLabels.md:349 in f38a439. [](commit_id = f38a439cf1ad170d3de190086a5eef7b8a54a7ca, deletion_comment = False)

RebeccaAnne:
> ### XAML

Logged #4363 to track adding platform specific implementation details.

---
In reply to: [658915174](https://github.com/microsoft/AdaptiveCards/pull/4363#issuecomment-658915174) [](ancestors = 658915174,658460570)

---
Refers to: specs/DesignDiscussions/InputLabels.md:405 in f38a439. [](commit_id = f38a439cf1ad170d3de190086a5eef7b8a54a7ca, deletion_comment = False)

Issue #4416, opened 2020-07-17
[JavaScript][Validation] Customize parse warning messages for unsupported element/action types


mokandre:
**What platform is your feature request related to? (Delete other platforms).**

* JavaScript

**Is your feature request related to a problem? Please describe.**

We would like to modify the warning message which displays in designer while trying to parse/render the unsupported element/action types of adaptive cards

**Describe the solution you'd like**

If users try to add unsupported element/action to their JSON then the warning message is showing up like `[Parse] Unknown element type: Media. Fallback will be used if present.` (Example for unsupported element) 

We would like to have a way to customize parse warning messages for unsupported element/action types

**Example for unsupported element** `[Parse] Unknown element type: Media`

Issue #4414, opened 2020-07-17
[JavaScript][Deregistering a property from the parser/render]


mokandre:
**What platform is your feature request related to?**

* JavaScript

**Is your feature request related to a problem? Please describe.**

We would like to disable the functionality of unsupported properties on parser/render

**Describe the solution you'd like**
It would be great to have a way to `Deregister a property from the parser/render`

**Deregister has to work for the following cases:**

* Deregister the property for a particular element( For Example: Deregister the `Height` property of `columnSet`)
* Deregister the property for all elements(For Example:  Deregister the `verticalContentAlignment` property  for all the elements where it supports)

### For Example

Unsupported Property  : `maxLines` for `TextBlock` element.

After deregistering a  property, though users try to add `maxLines` property to their JSON then that property shouldn't render/parse.

**Current behavior:**

<img width="1680" alt="Screenshot 2020-07-17 at 11 37 18 AM" src="https://user-images.githubusercontent.com/55378170/87754154-e9c99680-c821-11ea-93f0-d5e393a428e0.png">

**If possible, validator have to point out that the unsupported properties are not supported as warning message**

For Example:(Warning Message)

**Unknown/Unsupported property :`maxLines` of `TextBlock` Element**

dclaux:
This is actually already supported, via the SrializableObject.onRegisterCustomProperties event. Example:

```typescript
Adaptive.SerializableObject.onRegisterCustomProperties = (sender: Adaptive.SerializableObject, schema: Adaptive.SerializableObjectSchema) => {
    if (sender instanceof Adaptive.TextBlock) {
        schema.remove(Adaptive.TextBlock.heightProperty);
    }
}

```

Can you please try this and close this issue if that solves your problem?

mokandre:
> @mokandre This is actually already supported, via the SrializableObject.onRegisterCustomProperties event. Example:
> 
> ```ts
> Adaptive.SerializableObject.onRegisterCustomProperties = (sender: Adaptive.SerializableObject, schema: Adaptive.SerializableObjectSchema) => {
>     if (sender instanceof Adaptive.TextBlock) {
>         schema.remove(Adaptive.TextBlock.heightProperty);
>     }
> }
> ```
> 
> Can you please try this and close this issue if that solves your problem?

@dclaux  Able to deregister some of the properties but
 **Following are some issues facing while deregistering  (`fallback` and `requires`) properties**

* Deregistering `requires` property:
[...]

Issue #4413, opened 2020-07-17
[Javascript][Validation][Warning should be returned  for unsupported action types]


mokandre:
# Platform

* JavaScript

# Author or host
 Cisco Adaptive Cards Designer

# Version of SDK
Adaptive Cards Designer: 1.0
Adaptive Cards: 2.0

# Details

After deregistering the actions, warning should be returned for unsupported action types.

**Example of warning message for the unregistered element**

<img width="799" alt="Screenshot 2020-07-17 at 10 44 12 AM" src="https://user-images.githubusercontent.com/55378170/87750964-838d4580-c81a-11ea-9a17-670abe2063ed.png">

PR #4419, opened 2020-07-17
Introduce resource strings in renderer

dclaux:
## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/4416
Fixes https://github.com/microsoft/AdaptiveCards/issues/4414
Fixes https://github.com/microsoft/AdaptiveCards/issues/4413

## Description
- Hard-coded strings are now replaced with localizable/customizable resource strings
- Warnings correctly emitted for unknown action types
- Added an `isSerializationEnabled` property on schema property definitions to allow control of (de)serialization at the property level
- Added a `GlobalSettings.enableFallback` boolean flag to allow hosts to completely disable fallback

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4419)


source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -4,6 +4,25 @@ import * as Enums from "./enums";
 import * as Shared from "./shared";
 import { HostConfig } from "./host-config";
 
+export function formatString(s: string, ...args: any[]): string {
```
paulcam206, line 7: would this perhaps be a little more discoverable if it were an extension method on string?

source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -4,6 +4,25 @@ import * as Enums from "./enums";
...
+    let result = s;
+
+    if (result) {
+        let regEx = /{(\d+)}/g;
+        let matches: RegExpExecArray | null;
+
+        while ((matches = regEx.exec(result)) != null) {
```
paulcam206, line 14: I know that this is mainly just for our own internal use, but it would be nice if this method were a *bit* more robust -- in particular:

```javascript
const templateString = "This is my {0} string.";
foo.formatString(templateString, "format", "extra"); // should probably throw an exception
```

```javascript
const templateString = "My {0} string.";
foo.formatString(templateString, "{1}", 8); // do we want this sort of chaining to work?
```

```javascript
const templateString = "{0} is my {1} {2}.";
foo.formatString(templateString, "This", "template"); // should probably throw an exception
```

```javascript
const templateString = "My {0} string.";
foo.formatString(templateString, { foo: "bar" }); // maybe an exception here? I'm not sure...

File: source/nodejs/adaptivecards/src/strings.ts [hunk 3]
```diff
@@ -0,0 +1,37 @@
+export class Strings {
+    static readonly errors = {
+        unknownElementType: "Unknown element type \"{0}\". Fallback will be used if present.",
```
paulcam206, line 3: I think it would improve legibility to use `'` instead of `"` when the string has embedded `"`s :)

```suggestion
        unknownElementType: 'Unknown element type "{0}". Fallback will be used if present.',
```

source/nodejs/adaptivecards/src/utils.ts
dclaux: The recommendation in the wild is to not extend string to avoid conflicts.

source/nodejs/adaptivecards/src/utils.ts
dclaux: This was designed to be resilient and explicitly not throw in those cases. But as per Risheek's suggestion, I'm moving to resource functions, so this is moot.

golddove:
Just posting my offline suggestion here.

What if the `Strings.ts` definitions were like this:

``unknownElementType: (typeName: string) => `Unknown element type ${typeName}. Fallback will be used if present.`,``

and used like this:

`Strings.errors.unknownElementType(typeName)`

so that we have named identifiers and better intellisense w/ typescript

source/nodejs/adaptivecards/src/strings.ts
dclaux: I tend to never use `'`. But anyway, I am making these resource strings functions, and this is addressed in the process.

dclaux:
@golddove and I have discussed his proposal, and I am convinced. I am making the change to use resource functions instead of strings.

source/nodejs/adaptivecards/src/utils.ts
golddove, line 7: Is this function still used somewhere?

dclaux: Oops - no, removed.

source/nodejs/adaptivecards/src/card-object.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import * as Enums from "./enums";
+import * as Utils from "./utils";
+import { Strings} from "./strings";
```
paulcam206, line 5: nit:

```suggestion
import { Strings } from "./strings";
```

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -212,21 +214,22 @@ export abstract class BaseSerializationContext {
     }
 
     logEvent(
+        source: SerializableObject | undefined,
         phase: Enums.ValidationPhase,
         event: Enums.ValidationEvent,
-        message: string,
-        source?: SerializableObject) {
+        message: string) {
```
paulcam206, line 220: something to consider as a future enhancement -- if there's a way we could strongly type the string helpers (e.g. via a type alias or something), it would be an easy way to enforce the "don't use a hardcoded string" rule :)

PR #4432, opened 2020-07-17
[UWP] Fix Input.Number validation logic to ensure value is a number

### LINKED ISSUE #4421: [UWP][Inputs] Input.Number accepts non-number values
# Platform

* UWP

# Details

This is not new with the input work, but given input validation we should consider enforcing the fact that Input.Number should only accept numbers.

image

RebeccaAnne:
## Related Issue
Fixes #4421

## Description
Updated number validation logic to ensure value is a number even if min and max are not set

## How Verified
Validated manually in visualizer that error message is now shown for non number values.

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4432)

### DIFF
--- a/source/uwp/Renderer/lib/InputValue.cpp
+++ b/source/uwp/Renderer/lib/InputValue.cpp
@@ -253,40 +253,34 @@ HRESULT NumberInputValue::RuntimeClassInitialize(_In_ IAdaptiveNumberInput* adap
 HRESULT NumberInputValue::IsValueValid(_Out_ boolean* isInputValid)
 {
     // Call the base class to validate isRequired
-    boolean isBaseValid;
-    RETURN_IF_FAILED(InputValue::IsValueValid(&isBaseValid));
+    boolean isValid;
+    RETURN_IF_FAILED(InputValue::IsValueValid(&isValid));
 
-    // Check that min and max are satisfied
     int max, min;
     RETURN_IF_FAILED(m_adaptiveNumberInput->get_Max(&max));
     RETURN_IF_FAILED(m_adaptiveNumberInput->get_Min(&min));
 
-    // For now we're only validating if min or max was set. Theoretically we should probably validate that the input is
-    // a number either way, but since we haven't enforced that in the past and the card author likely hasn't set an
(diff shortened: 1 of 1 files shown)


shalinijoshi19:
thanks.. how easy would it be to add a unit test for this? Also am assuming a sample already exists? If not, can we get one added to the tests\ folder?

source/uwp/Renderer/lib/InputValue.cpp
```diff
@@ -253,40 +253,34 @@ HRESULT NumberInputValue::RuntimeClassInitialize(_In_ IAdaptiveNumberInput* adap
 HRESULT NumberInputValue::IsValueValid(_Out_ boolean* isInputValid)
```
shalinijoshi19, line 253: Is this something that can eventually land as part of shared code in the shared model at all?

RebeccaAnne: We could theoretically put a helper method for this and/or other validation methods in the shared model. I'm not sure how much we'd get out of that, because some portion of what's going on here is getting things out of xaml controls and whatnot, and what we have in hand is a UWP number input object not a shared model one. We could convert back, or just get the min and max values to pass to a helper, but the actual shared code is probably just a few not very interesting lines. (Although, admittedly, I'm fixing a consistency bug in those few lines so i guess they're at least a little interesting :)). 

Probably not worth doing now, but it's something we can keep in mind if we start seeing more consistency issues in this logic or if our local validation becomes more complex.

---
In reply to: [457623190](https://github.com/microsoft/AdaptiveCards/pull/4432#discussion_r457623190) [](ancestors = 457623190)

RebeccaAnne:
We don't currently have any interactive tests that do things like fill in inputs and press buttons, and this class isn't exposed publicly to easily call the method directly from tests. There are existing samples of number inputs.

---
In reply to: [661271503](https://github.com/microsoft/AdaptiveCards/pull/4432#issuecomment-661271503) [](ancestors = 661271503)

PR #4446, opened 2020-07-20
[Shared][UWP][Android][iOS] Rework Input.Number property handling to be nullable

### LINKED ISSUE #4382: [Shared Model][Input Evolution] [Input.Number can't have nullable value]
# Platform

* Android
* iOS
* UWP

# Details
Input.Number has Value field.
If Value field is left empty, the shared model should let the renderer knows that value is null.
Currently, at renderer, if GetValue is called, it will return 0. This will leads to "required" validation not to work and causes inconsistency issue. WPF and JS leaves field empty, and "required" validation will work.

paulcam206:
## Related Issue
Fixes #4382

## Description
Several things in this PR:
* **Shared Model**
  * Move `NumberInput`'s `Get[Min|Max|Value]` methods to working with `std::optional<int>` (basically, a `Nullable<T>` for C++17)
  * Update unit test build to use C++17 for all flavors
  * Update unit tests to consume new interface
* **UWP**
  * `AdaptiveNumberInput::[Min|Max|Value]` now projected as `Windows.Foundation.IReference<Int32>` instead of `INT32`. (`IReference<T>` is how `Nullable` is [implemented in UWP](https://docs.microsoft.com/en-us/uwp/api/windows.foundation.ireference-1?view=winrt-19041))
  * Updates to implementation to map between shared model's `std::optional<int>` and API boundary's `IReference<int32_t>`
  * New unit test to verify nullability
* **Android**
  * Updated SWIG
  * Update `CMakeLists.txt` to account for C++17 requirement
  * Updates to shared model to squelch clang errors related to C++17
* [x] **iOS**
  * [x] @jwoo-msft to push updates here :)

### Open issue?
One open issue is around how we should handle mutation of min/max/value properties: suppose I had a `NumberInput` with `min == -1`, `max == 1`, and `value == 0`. What should happen if I `min = 2`? `value = 1000`?

## How Verified
* local build/tests/debugging

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4446)


source/uwp/Renderer/lib/AdaptiveNumberInput.cpp
```diff
@@ -30,9 +27,24 @@ namespace AdaptiveNamespace
...
         }
 
-        m_min = sharedNumberInput->GetMin();
-        m_max = sharedNumberInput->GetMax();
-        m_value = sharedNumberInput->GetValue();
+        const auto sharedMin = sharedNumberInput->GetMin();
+        if (sharedMin)
+        {
+            m_min = winrt::box_value(sharedMin.value()).as<ABI::Windows::Foundation::IReference<int32_t>>();
+        }
```
paulcam206, line 34: Should probably be a helper...

source/uwp/Renderer/lib/AdaptiveNumberInput.cpp
```diff
@@ -97,9 +109,33 @@ namespace AdaptiveNamespace
...
-        numberInput->SetMax(m_max);
-        numberInput->SetValue(m_value);
+        std::optional<int> min;
+        if (m_min)
+        {
+            int minValue;
+            RETURN_IF_FAILED(m_min->get_Value(&minValue));
+            min = minValue;
+        }
+        numberInput->SetMin(min);
```
paulcam206, line 119: helper?

paulcam206:
I need to update Android tests too :)

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -778,19 +778,19 @@ AdaptiveNamespaceStart
...
         internal
 #endif
     ]
     runtimeclass AdaptiveNumberInput : IAdaptiveInputElement, IAdaptiveCardElement
     {
         AdaptiveNumberInput();
 
-        Int32 Max;
-        Int32 Min;
+        Windows.Foundation.IReference<Int32> Max;
```
RebeccaAnne, line 790: >Windows.Foundation.IReference<Int32> Max; [](start = 8, length = 41)

There's going to need to be some more fixes once this change comes into contact with the validation branch. Might be worth considering going into there even though this isn't technically a validation feature.

shalinijoshi19: @paulcam206 might be worth considering basing this against feature/inputEvolution if only to prevent merge issues?

paulcam206:
I need some `typemap` magic in our `.i` file to get Android mapping working correctly (need to map `std::optional<int>` to `Integer`)

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
paulcam206: > @paulcam206 might be worth considering basing this against feature/inputEvolution if only to prevent merge issues - is there a reason why you decided to go against main directly here?

This functionality isn't dependent on the Input Evolution work, so I just automatically decided to go against `main`. It's no problem to bring the changes over to the IE feature branch, though, so I will do that :)

shalinijoshi19: @paulcam206  heads up that we've since merged into main -- so you should be good to continue to go against main at this point!

paulcam206:
@jwoo-msft - this should be ready for iOS fixes now :)

shalinijoshi19:
@jwoo-msft i believe this is awaiting your input?

jwoo-msft:
@shalinijoshi19  @paulcam206 committed my changes here.

paulcam206:
now just need a review! :)

source/uwp/Renderer/lib/InputValue.cpp
```diff
@@ -266,12 +268,24 @@ HRESULT NumberInputValue::IsValueValid(_Out_ boolean* isInputValid)
...
+                int maxInt;
+                RETURN_IF_FAILED(max->get_Value(&maxInt));
+                isValid &= (currentInt < maxInt);
+            }
+
+            if (min.Get())
+            {
+                int minInt;
+                RETURN_IF_FAILED(min->get_Value(&minInt));
+                isValid &= (currentInt > minInt);
```
jwoo-msft, line 287: is this correct? currentInt can be equal to minInt. The same applies to maxInt.

RebeccaAnne: Good catch, and not introduced with this change. @paulcam206, is this going to make it into 1.3? it might be good to try to get the fix for this in there. I can make a separate change for that if it would be easier.

---
In reply to: [463375533](https://github.com/microsoft/AdaptiveCards/pull/4446#discussion_r463375533) [](ancestors = 463375533)

paulcam206: sent mail on logistics for fixing this 👍

paulcam206: this is 1.3-only behavior (duh). fix inbound shortly.

PR #4458, opened 2020-07-21
[WPF] Retrieve all choiceset inputs, even if empty

### LINKED ISSUE #4429: [WPF][Input Evolution] [Choice Set Compact's Input is not collected]
# Platform

* .NET WPF

# Details
image

If no choices are made for ChoiceSet compact, its input is not gathered.

almedina-ms:
## Related Issue
Fixes #4429 

## Description
Change input retrieval for ChoiceSet to return empty strings rather than null, this makes the AsJson method to populate the fields.

## How Verified
The fix was verified manually and as can be seen below, it now behaves as expected

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4458)


shalinijoshi19:
@almedina-ms is there a test card that exercises this scenario? If not can you get one added please? tx

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveInputValue.cs
```diff
@@ -363,7 +363,7 @@ public override string GetValue()
                         AdaptiveChoice adaptiveChoice = item.DataContext as AdaptiveChoice;
                         return adaptiveChoice.Value;
                     }
-                    return null;
+                    return "";
                 }
```
shalinijoshi19, line 367: Can you add a quick code comment here explaining what this will do- are we stripping out original empty strings before teh cde gets here that needs us to manually return an empty string?

PR #4461, opened 2020-07-21
[Android] Use ConstraintLayout for Image sizing

### LINKED ISSUE #4439: [Android][Rendering] Conflict between image size and column weight
# Platform
* Android

# Details

With the introduction of image resizing to a bigger scale the ```v1.2/Elements/Column.Weighted.Image.json``` card became broken as there's a conflict between the image size and the column weight.

image

### LINKED ISSUE #4469: [Android][Regression] Image is not centered
# Platform
* Android

# Version of SDK
main

# Details

The image in the middle must be horizontally centered, as can be seen, it's set all the way to the left

image

image

golddove:
## Related Issue
Fixes #4439 , Fixes #4469 

## Description
As we see in the issue above, weird edge cases occur with interactions between image scaling, height stretching, column widths, etc. Android's ConstraintLayout provides a more robust way to define these constraints.

Updated the ImageSizing.json sample to cover more cases.

## How Verified
Manual verification with the following payloads:

`samples/v1.2/Tests/ImageSizing.json`

image
image
image
image
image
image
image
image
image

`samples/v1.1/Tests/ColumnSet.Image.VerticalStretch.json`

image
image

`samples/v1.1/Tests/ColumnSet.ImageSet.VerticalStretch.json`

image
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4461)


shalinijoshi19:
@almedina-ms / @golddove what's the status of this PR esp wrt #4469 ?

shalinijoshi19:
@golddove FYI, when using a single PR to target multiple issues, you'd need to use the github keywords (fixes/resolves etc) mltiple times per issue that is being fixed for the PR to be autolinked to these issues correctly. (_went ahead and fixed your description this time around)_

samples/v1.2/Tests/ImageSizing.json
```diff
@@ -1,153 +1,187 @@
...
-                {
-                    "type": "Image",
-                    "url": "https://picsum.photos/id/133/300/200"
-                }
-            ]
-        }
-    ]
-}
\ No newline at end of file
+  "type": "AdaptiveCard",
```
almedina-ms, line 2: I can't access codeflow, but if these lines are changing because it converting from single spaces to tabs then ignore this comment, otherwise revert to tabs

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
```diff
@@ -143,56 +145,125 @@ else if (imageSize == ImageSize.Large)
...
             if (explicitWidth != 0 && explicitHeight != 0)
             {
+                // If both are set, ignore aspect ratio
                 imageView.setScaleType(ImageView.ScaleType.FIT_XY);
             }
             if (explicitWidth != 0)
             {
-                viewWidth = Util.dpToPixels(context, explicitWidth);
+                // Stretch width + max width = exact width
+                imageConstraints.constrainMaxWidth(imageView.getId(), Util.dpToPixels(context, explicitWidth));
```
almedina-ms, line 239: why is the constraint for width set with MaxWidth and height with just Height?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
```diff
@@ -143,56 +145,125 @@ else if (imageSize == ImageSize.Large)
...
         {
-            viewWidth = getImageSizePixels(context, imageSize, hostConfig.GetImageSizes());
+            // Disable stretch width
+            imageConstraints.constrainWidth(imageView.getId(), ConstraintSet.WRAP_CONTENT);
         }
-        else if (imageSize != ImageSize.Auto && imageSize != ImageSize.None) {
-            // TODO: Instead of failing, proceed to render w/ default size "auto"
+        else if (imageSize != ImageSize.Stretch)
+        {
+            // TODO: Don't fail. Provide warning and then make a best-effort render w/ size "auto"
```
almedina-ms, line 261: Now that you're here, you could add the warning and remove the illegalargumentexception

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
```diff
@@ -143,56 +145,125 @@ else if (imageSize == ImageSize.Large)
...
+        if(imageView.getId() == View.NO_ID) {
+            imageView.setId(View.generateViewId());
+        }
+
         long explicitWidth = image.GetPixelWidth();
         long explicitHeight = image.GetPixelHeight();
         ImageSize imageSize = image.GetImageSize();
+        ConstraintSet imageConstraints = new ConstraintSet();
+
+        // Set horizontal alignment
```
almedina-ms, line 203: you can add this section in a different method, this method is almost 90 lines long

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
```diff
@@ -143,56 +145,125 @@ else if (imageSize == ImageSize.Large)
...
-    private static void setImageSize(Context context, ImageView imageView, Image image, HostConfig hostConfig, boolean isInImageSet)
+    private static View setImageConstraints(Context context, ImageView imageView, Image image, HostConfig hostConfig, TagContent tagContent)
     {
+        // Wrapper layout around the ImageView
+        ConstraintLayout wrapper = new ConstraintLayout(context);
+        wrapper.addView(imageView);
+        tagContent.SetStretchContainer(wrapper);
+
+        // Grow wrapper layout if height is stretch (assumes the parent is a vertical LinearLayout)
+        if (image.GetHeight() == HeightType.Stretch)
```
almedina-ms, line 184: could you test this cards to verify that stretch works as expected?

samples\v1.1\Tests\ColumnSet.Image.VerticalStretch.json
samples\v1.1\Tests\ColumnSet.ImageSet.VerticalStretch.json

almedina-ms:
Could you add some pictures to your pull request, please? It's usually a good way for us to have some documentation that the features worked as expected at some point

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
golddove: Setting explicit width can cause the image to grow larger than the width of its parent (which is what happened in the issue you filed w/ large image in a small column). Setting it to stretch and then placing a maximum allows stretch as much as it can, up to the explicit width, without exceeding its parent.

For height, we never want to stretch, we either want to wrap or set fixed height.

almedina-ms: by "stretch" you mean the image size stretch, right? so we want to make the image as large as the container width allows but keeping the ratio so we don't  mind the height at all?

golddove: This is when an explicit width is specified. This will effectively be like `width = Math.min(container width, explicit width)`. I will try to clarify this better in the code comment.

Aspect ratio is always maintained, unless both width and height are specified.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
golddove: Works. Screenshots added to PR details at top.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
golddove: Done. It's now ~55 lines :)

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
golddove: Resolved. Error case is unreachable.

samples/v1.2/Tests/ImageSizing.json
golddove: Done. Converted all whitespace to tabs.

PR #4462, opened 2020-07-22
Fixing .NET CI tests for feature/inputEvolution

shalinijoshi19:
## Related Issue
Not being tracked (CI failure in .NET)

## Description
1. Updating the serialization behavior for the newly introduced LabelConfig.InputSpacing property to ignore default values from being written out while serializing to allow the current HTML renderer that doesn't yet support Labels to continue to work.
2. In AdaptiveContainerRenderer, fallback to working with the earlier elementForAccessibility element for inputs with null values _(appears at least the v1.0 samples fell into this category)._
3. Fixing a typo in the XML comments for the API doc
4. Including the regenerated API doc 

## How Verified
1. Ran RunAllTests.ps1 with defaults (samples\v1.0\scenarios) 
2. Manually ran the ImageRenderer sample for WPF, the WPF visualizer as well as the HTML Test suite, with the problem scenario files and verified they now successfully render as expected.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4462)

### DIFF
--- a/source/dotnet/Library/AdaptiveCards/docs/AdaptiveCards.md
+++ b/source/dotnet/Library/AdaptiveCards/docs/AdaptiveCards.md
@@ -2673,7 +2673,7 @@ Properties which control rendering of input labels
 
 ##### Summary
 
-Specifies the spacing bewteen the label and the input
+Specifies the spacing between the label and the input
 
 <a name='T-AdaptiveCards-Rendering-MediaConfig'></a>
 ## MediaConfig `type`
--- a/source/dotnet/Samples/AdaptiveCards.Sample.ImageRender/Properties/launchSettings.json
+++ b/source/dotnet/Samples/AdaptiveCards.Sample.ImageRender/Properties/launchSettings.json
@@ -2,7 +2,7 @@
   "profiles": {
     "AdaptiveCards.Sample.ImageRender": {
       "commandName": "Project",
-      "commandLineArgs": "-r  -o ./out"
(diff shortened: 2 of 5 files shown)


source/dotnet/Samples/AdaptiveCards.Sample.ImageRender/Properties/launchSettings.json
```diff
@@ -2,7 +2,7 @@
   "profiles": {
     "AdaptiveCards.Sample.ImageRender": {
       "commandName": "Project",
-      "commandLineArgs": "-r  -o ./out"
+      "commandLineArgs": "\" C:\\AdaptiveCards\\samples\\v1.0\\Scenarios\" -r  -o ./out"
```
shalinijoshi19, line 5: Oops reverting.

source/dotnet/Library/AdaptiveCards/docs/AdaptiveCards.md
```diff
@@ -2673,7 +2673,7 @@ Properties which control rendering of input labels
 
 ##### Summary
 
-Specifies the spacing bewteen the label and the input
+Specifies the spacing between the label and the input
```
shalinijoshi19, line 2676: @paulcam206 i was wanting to add in specifics about the default values also for these- is that just manual text at this point or some magic metadata being read someplace?

PR #4476, opened 2020-07-23
updated xcode to c++17 compiler and updated to use optional

jwoo-msft:
## Related Issue
#4382 

## Description
* updated Xcode project to c++17
* there was static check that explicitly prohibits the use of value(), so I had to change value() to valure_or() in shared code
* updated the ParserUtil to check if the property is not empty, to ensure the value's has_value() returns false when empty. 

## How Verified
Ran the relevant cards and verified.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4476)


jwoo-msft:
@paulcam206 I think you made your changes before mine. After I took your changes and merged in ParseUtil.h and ParseUtil.cpp, I'm getting source/shared/cpp/ObjectModel/ParseUtil.h:45:14: No template named 'optional' in namespace 'std'. I'm not certain what's is causing this because NumberInput.cpp and NumberInput.h have no such problems.

jwoo-msft:
closing this pr as changes are merged directly to Paul's branch

PR #4481, opened 2020-07-23
[Shared] Fix Markdown parse issue

### LINKED ISSUE #4463: [Android] Text with markdown is broken in most of the components
# Platform

* Android

# Author or host

Teams mobile

# Version of SDK

1.2.10

# Details

Sample [**Agenda**](https://adaptivecards.io/samples/Agenda.html) card, [**Expense report**](https://adaptivecards.io/samples/ExpenseReport.html) card available in public documentation is broken experience.

Here is the comparison of experience with 1.2.6(SDK version) and 1.2.10(SDK version).

| 1.2.6 | 1.2.10|
|-----------|:-----------:|
| image | image | 

`TextBlock`, `FactSet` with value enclosed with markdown support are completely broken.

Here is another example:
| 1.2.6 | 1.2.10|
|-----------|:-----------:|
| image| image| 
|  image  |image| 

Attaching the Card JSON for the same:

```
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "TextBlock",
            "text": "_Italic_"
        },
        {
            "type": "TextBlock",
            "text": "**Bold**"
        },
        {
            "type": "TextBlock",
            "text": "**~~Bold, strike-through text~~**"
        }
    ],
    "version": "1.2"
}
```

cc: @VikrantSingh01, @anbah34 ,@shbhadra

golddove:
## Related Issue
Fixes #4463 

## Description
Markdown emphasis parsing for was broken on Android devices.

## How Verified
Manual testing

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4481)


source/shared/cpp/ObjectModel/MarkDownBlockParser.cpp
```diff
@@ -263,7 +263,7 @@ bool EmphasisParser::IsLeftEmphasisDelimiter(const char ch) const
 
 bool EmphasisParser::IsRightEmphasisDelimiter(const char ch) const
 {
-    if ((ch == EOF || MarkDownBlockParser::IsSpace(ch)) && (m_lookBehind != DelimiterType::WhiteSpace) &&
+    if ((ch == EOF || ch == '\xff' || MarkDownBlockParser::IsSpace(ch)) && (m_lookBehind != DelimiterType::WhiteSpace) &&
```
jwoo-msft, line 266: why is this needed for Android? and why is it needed now?

shalinijoshi19:
This looks to be a gap in our testing; Can we add a test card for this at the very least with this change? Is it possible to cover this with existing sharedmodel unit tests? @RebeccaAnne

source/shared/cpp/ObjectModel/MarkDownBlockParser.cpp
shalinijoshi19, line 266: I think this is now a complicated enough condition to warrant a quick explanatory code comment pls!

shalinijoshi19:
> void MarkDownBlockParser::ParseBlock(std::stringstream& stream)

The Markdown parser looks like it should have unit tests for this? @RebeccaAnne, does it? 

---
Refers to: source/shared/cpp/ObjectModel/MarkDownBlockParser.cpp:10 in f6f5cac. [](commit_id = f6f5cacfd182665d7b2bb3753acfba1993cda51e, deletion_comment = False)

almedina-ms:
> This looks to be a gap in our testing; Can we add a test card for this at the very least with this change? Is it possible to cover this with existing sharedmodel unit tests? @RebeccaAnne

This is not a gap in testing, this is an issue about the device it's been deployed, in emulators everything works as expected while in devices it doesn't. As of this moment there are tests in the c++ side that verify that everything works as expected, we could add tests on Android but the tests would still yield "false positives" on emulators

golddove:
Yeah, the unit tests didn't catch it because it is non-deterministic (it happens on some devices, based on the c++ implementation).

@jwoo-msft  The bug isn't actually new, it seems to have been introduced as part of #3342 last year. But we recently brought that change from main to release with #4309. Hence, Teams is now seeing it.

As for why it works on some platforms:

C++ standard doesn't define whether 'char' is signed or unsigned. It's up to the compiler. The compiler usually makes this decision based on what the target device architecture prefers.

We can verify this by looking at CHAR_MIN (which is 0 if char is unsigned char, and -128 if its signed). This is what I see if I debug in the emulator: 
image
And when I debug on my device:
image

The issue occurs if the compiler decided 'char' is an unsigned char. So EOF (which is `-1`) assigned to unsigned char (so it becomes `255`), no longer equals EOF (which is still `-1`)... more details [here](https://stackoverflow.com/questions/1851525/c-can-native-type-char-hold-end-of-file-character).

This was just a draft PR, but I think the proper solution is to revert ch to int (unless there was a specific reason for that change?), or perhaps switching char to explicit signed char would work (but I think we would lose access to special characters if we do that? I'm not sure..). Thoughts? @paulcam206

jwoo-msft:
> Yeah, the unit tests didn't catch it because it is non-deterministic (it happens on some devices, based on the c++ implementation).
> 
> @jwoo-msft The bug isn't actually new, it seems to have been introduced as part of #3342 last year. But we recently brought that change from main to release with #4309. Hence, Teams is now seeing it.
> 
> As for why it works on some platforms:
> 
> C++ standard doesn't define whether 'char' is signed or unsigned. It's up to the compiler. The compiler usually makes this decision based on what the target device architecture prefers.
> 
> We can verify this by looking at CHAR_MIN (which is 0 if char is unsigned char, and -128 if its signed). This is what I see if I debug in the emulator:
> image
> And when I debug on my device:
> image
> 
> The issue occurs if the compiler decided 'char' is an unsigned char. So EOF (which is `-1`) assigned to unsigned char (so it becomes `255`), no longer equals EOF (which is still `-1`)... more details [here](https://stackoverflow.com/questions/1851525/c-can-native-type-char-hold-end-of-file-character).
> 
> This was just a draft PR, but I think the proper solution is to revert ch to int (unless there was a specific reason for that change?), or perhaps switching char to explicit signed char would work (but I think we would lose access to special characters if we do that? I'm not sure..). Thoughts? @paulcam206

I see. Yes, #3342 added support for non-Latin language support, and as I added new markdown list marker support '+', and '*', I noticed that #3342 was not picked up, and it's an important fix, so I added it. 
@shalinijoshi19, I think we should have a new Github issue tracking a new lexer that supports non-Latin language. I suggest we go with ANTLR for the lexer work as we will likely use ANTLR for the shared model template.  as we only want to support a subset of markdown, and ANTLR can't generate a parser for markdown language and most of the issues we are seeing with happening at the tokens and not the grammar, it would be beneficial to spend 2-3 days of dev works on a lexer generated by ANTLR based on the token rule that's already verified works with non-Latin and control characters.

paulcam206:
I think switching to `int` here is likely the correct answer :)

jwoo-msft:
fixed compiler warnings and updated IsAlphaNum function to be able to use with int.

shalinijoshi19:
Great find guys!
Wrt test coverage, question:  If markdown parsing is broken in general would we know about it today or whats a good test site that I can run if I were to make changes to it revamp the MD parser? What test suite can i run today to at least know it's not a general/widespread issue? (I understand this specidic issue is different etc)..

shalinijoshi19:
@golddove /@paulcam206, this looks like it would need to be ported back to 1.2?

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/MarkDownUnitTest.cpp
```diff
@@ -739,19 +739,19 @@ namespace AdaptiveCardsSharedModelUnitTest
             MarkDownParser parser("");
             Assert::AreEqual<bool>(false, parser.IsEscaped());
 
-            parser.TransformToHtml();
+            (void) parser.TransformToHtml();
             Assert::AreEqual<bool>(false, parser.IsEscaped());
```
shalinijoshi19, line 743: cool these are the tests I was looking for then! Tx

PR #4502, opened 2020-07-30
[WPF][InputEvolution] Fix fallback and error messages in non interactive scenarios

### LINKED ISSUE #4474: [WPF][Input evolution]SimpleFallback.json scenario card fails to render with the new WPF renderer
# Platform

* .NET WPF

# Version of SDK
main

# Details
## Script output
[FAILED]        SimpleFallback.json => Object reference not set to an instance of an object.

## Issue
Looks like it's having trouble rendering {AdaptiveCards.AdaptiveFallbackElement} type with the new changes.

almedina-ms:
## Related Issue
Fixes #4474 

## Description
This PR includes two fixes:
1. Add validation when there is no rendereable element (this means that the fallback results in dropping the element)
2. Add validation to avoid setting error messages to inputs when interactivity is disabled. Due to the nature of this property, it doesn't matter if the inputs are not linked to error messages

## How Verified
1. The fix was verified by executing the ./RunAllTests.ps1 script
2. The fix was also verified manually as can be seen below

SimpleFallback.json
image

Inputs (v1.3) with WindowsTimeline host config
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4502)


shalinijoshi19:
@almedina-ms tx! To confirm, you were able to test this with the 1.2 version of the sample(which is what had been failing? )

shalinijoshi19:
@almedina-ms there are now than test changes here.  We should update the description/title to be reflective of that(its a bit confusing right now with "Fux WPF test errors")

almedina-ms:
> @almedina-ms tx! To confirm, you were able to test this with the 1.2 version of the sample(which is what had been failing? )

The bug was found with the SimpleFallback card. As stated in the description that card works now

PR #4536, opened 2020-08-04
[TS] Live render

dclaux:
## Related Issue
No associated issue. I am opening this PR mainly so as to have the build system created a test site and for folks to be able to review the code.

## Description
This PR introduces "live rendering" of cards, as in changes to properties in the object model will cause the rendered card to update in place without requiring the full card to be re-rendered.

This change lays the groundwork for live data binding.

As a side effect, the designer is now considerably faster with larger payloads. Try it in the test site!

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4536)


dclaux:
This is obsolete.

PR #4541, opened 2020-08-05
[Android] Streamline Android publishing

golddove:
## Related Issue
No issue filed, but our current release pom's don't declare dependencies.

## Description
- Update build.gradle to use the component generated by android for publishing (refer to Android's recommendation [here](https://developer.android.com/studio/build/maven-publish-plugin).
- Setup local and ADO publishing repositories to support generating artifact to local directory or to ADO feed (or both).
- Move Android's daily pipeline to `.pipelines/android-daily.yml`

## How Verified
Needs verification. Not ready for merge.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4541)


shalinijoshi19:
@golddove is this the approach we are going with?

.pipelines/android-daily.yml
```diff
@@ -0,0 +1,57 @@
+name: $(Year:yyyy).$(Month).$(DayOfMonth)-$(rev:r)
```
almedina-ms, line 1: why is this file here?

golddove: This is the yml version of the Android daily pipeline. It will not delete the existing pipeline, but once it is checked into main, we can see if it is working as expected.

PR #4542, opened 2020-08-05
Fix empty Input number not submitting

### LINKED ISSUE #4544: [Android][Inputs] Input.Number fails validation when empty
# Platform

* Android

# Version of SDK

latest main / 2.0.0

# Details & Customer Impact
When an Input.Number field is empty, regardless of it being required or not, the end user will not be able to submit a form.
## Workaround: 
If the user sets any placeholder value, eg "0" in the form, they will be able to continue.

almedina-ms:
## Related Issue
Fixes #4544

## Description
The issue was that empty Input.Number inputs would not validate as valid if the input was empty and it was not required. This was due to how validation was being performed, inside the validation process we would parse the string into a number, if it was not a valid number then the validation marked the field as invalid, turns out, an empty string is not a well formed number.

The fix changes the parsing to check if the input value is empty, if it is and the input is not required then the validation succeeds

## How Verified
The fix was verified manually as can be seen below

inputnumberempty

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4542)

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/NumberInputHandler.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/NumberInputHandler.java
@@ -43,6 +43,12 @@ public class NumberInputHandler extends TextInputHandler
             return false;
         }
 
+        // Before performing any validation, if the input value is empty and is not required, then it's valid
+        if (numberInputValue.isEmpty() && !numberInput.GetIsRequired())
+        {
+            return true;
+        }
+
         int inputValue = 0;
         try
         {
@@ -50,15 +56,8 @@ public class NumberInputHandler extends TextInputHandler
         }
         catch (Exception ex)
         {
-            // Parsing failed, consider it invalid if it was not required otherwise it's invalid
-            if (numberInputValue.isEmpty() && !numberInput.GetIsRequired())
-            {
-                return true;
-            }
-            else
-            {
-                return false;
-            }
+            // Parsing failed, consider it invalid
+            return false;
         }
 
         boolean isValid = true;
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/NumberInputHandler.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/NumberInputHandler.java
@@ -50,8 +50,15 @@ public class NumberInputHandler extends TextInputHandler
         }
         catch (Exception ex)
         {
-            // Something failed, don't consider valid
-            return false;
+            // Parsing failed, consider it invalid if it was not required otherwise it's invalid
+            if (numberInputValue.isEmpty() && !numberInput.GetIsRequired())
+            {
+                return true;
+            }
+            else
+            {
+                return false;
+            }
         }
 
         boolean isValid = true;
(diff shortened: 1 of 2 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/NumberInputHandler.java
```diff
@@ -50,8 +50,15 @@ public boolean isValidOnSpecifics(String numberInputValue)
...
-            return false;
+            // Parsing failed, consider it invalid if it was not required otherwise it's invalid
+            if (numberInputValue.isEmpty() && !numberInput.GetIsRequired())
+            {
+                return true;
+            }
+            else
+            {
+                return false;
+            }
```
golddove, line 61: Does this need to be in the `catch`? Perhaps it makes more sense to move it above the try/catch

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/NumberInputHandler.java
```diff
@@ -50,8 +50,15 @@ public boolean isValidOnSpecifics(String numberInputValue)
         }
         catch (Exception ex)
         {
-            // Something failed, don't consider valid
-            return false;
+            // Parsing failed, consider it invalid if it was not required otherwise it's invalid
+            if (numberInputValue.isEmpty() && !numberInput.GetIsRequired())
```
golddove, line 54: Does this need to be in the catch? Perhaps it makes more sense to move it above the try/catch

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/NumberInputHandler.java
```diff
@@ -43,14 +43,20 @@ public boolean isValidOnSpecifics(String numberInputValue)
...
+
         int inputValue = 0;
         try
         {
             inputValue = Integer.parseInt(numberInputValue);
         }
         catch (Exception ex)
         {
-            // Something failed, don't consider valid
+            // Parsing failed, consider it invalid
```
shalinijoshi19, line 59: What's our logging story in Java? Ideally this would get logged to stdout or the Debug stream for debugging etc..

shalinijoshi19:
> ## Related Issue
> ## Description
> The issue was that empty Input.Number inputs would not validate as valid if the input was empty and it was not required. This was due to how validation was being performed, inside the validation process we would parse the string into a number, if it was not a valid number then the validation marked the field as invalid, turns out, an empty string is not a well formed number.
> 
> The fix changes the parsing to check if the input value is empty, if it is and the input is not required then the validation succeeds
> 
> ## How Verified
> The fix was verified manually as can be seen below
> 
> inputnumberempty
> 
> ###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4542)

This is not a shipblocker since a reasonable workaround exists. @almedina-ms can you follow up on checking thisin to main, and updating our release notes with customer impact + workaround?

shalinijoshi19:
@almedina-ms rebasing this to main ended up in conflicts here -- since this isnt a ship blocker can you now send a separate PR out with the changes against main and also include a reference to the bug?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/NumberInputHandler.java
almedina-ms: I don't think it should be logged, this will happen if an user writes something that is not a number into the input field

golddove: True, maybe the catch could be narrowed to `NumberFormatException`.

shalinijoshi19:
@almedina-ms where did we land on merging this to main?

shalinijoshi19:
this is part of another PR #4556 so closing this one out now.

Issue #4552, opened 2020-08-06
[Samples] Need to update old samples


paulcam206:
# Details

Now that 1.3 has shipped, we have input labels and input validation. Scenario sample cards created pre-1.3 should have new 1.3 versions created that take advantage of both features. 

The pages showing pre-1.3 samples will *also* be updated with a banner that explains that the newer features should be used in order to light up accessibility. (see #4807)

## Cards to update
I *think* this is the list of items that need action. I searched for `Inputs.` under the samples directory and stripped out 1.3 items, items that already had a newer version, and items under `Tests`.
* [ ] Templates/Scenarios/ActivityUpdate.template.json
* [ ] Templates/Scenarios/CalendarReminder.template.json
* [ ] Templates/Scenarios/ExpenseReport.template.json
* [X] Templates/Scenarios/FoodOrder.template.json (`FoodOrderWithValidation.template.json`)
* [ ] Templates/Scenarios/FoodOrderWithValidation.template.json
* [ ] Templates/Scenarios/InputForm.template.json
* [X] Templates/Scenarios/Inputs.template.json (`InputsWithValidation.template.json`)
* [ ] v1.0/Elements/Action.Submit.json
* [ ] v1.0/Elements/AdaptiveCard.json
* [ ] v1.0/Elements/Image.ImageBaseUrl.json
* [ ] v1.0/Elements/Input.ChoiceSet.json
* [ ] v1.0/Elements/Input.Date.json
* [ ] v1.0/Elements/Input.Number.json
* [ ] v1.0/Elements/Input.Text.Style.json
* [ ] v1.0/Elements/Input.Text.json
* [ ] v1.0/Elements/Input.Time.json
* [ ] v1.0/Elements/Input.Toggle.json
* [ ] v1.0/Scenarios/ActivityUpdate.json
* [ ] v1.0/Scenarios/CalendarReminder.json
* [X] v1.0/Scenarios/FoodOrder.json
* [ ] v1.0/Scenarios/InputForm.json
* [X] v1.0/Scenarios/Inputs.json
* [ ] v1.2/Elements/Input.Text.InlineAction.json  
* [ ] v1.2/Scenarios/ExpenseReport.json

paulcam206:
@matthidinger - care to wordsmith the banner in the description?

RebeccaAnne:
[...]

PR #4567, opened 2020-08-11
[Samples] Update some elements to use labels instead of placeholders

golddove:
## Related Issue

References #4552 
ADO 24045031

## Description
Update some v1.0 Elements to use label, and place these new cards in v1.3/ElementsPreview
When ready (i.e. hosts have upgraded to 1.3), move these cards into v1.3/Elements to activate them for #4612 .

## How Verified
Manual verificaiton

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4567)


samples/v1.3/ElementsPreview/Input.Text.json
```diff
@@ -0,0 +1,27 @@
+{
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"type": "AdaptiveCard",
+	"version": "1.3",
+	"body": [
```
RebeccaAnne, line 5: These are meant to replace/override the files currently in the v1.0/elements folder, right? I noticed that some of these new files tend to be more validation focused, where the older files showed a wider variety of the properties available on each type. For example, the v1.0\Elements\Input.Text.json file demonstrates the `maxLength`, `isMultiline`, and `value` properties. Was that change intentional? This card in particular looks like a good sample for the isRequired property, but is that the main/only thing we want to highlight in our primary Input.Text sample card on the website?

almedina-ms: agree, I think this may seem like a duplicate of https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.3/Elements/Input.Text.Label.json

shalinijoshi19:
@golddove FYI this is pending code review changes

samples/v1.3/ElementsPreview/Input.Text.json
golddove: Now reflects the existing samples.

samples/v1.3/ElementsPreview/Input.Number.json
```diff
@@ -0,0 +1,21 @@
+{
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"type": "AdaptiveCard",
+	"version": "1.3",
+	"body": [
+		{
+			"type": "Input.Number",
+			"id": "number",
+			"title": "Choose a number from 1 to 10",
```
RebeccaAnne, line 9: >title [](start = 4, length = 5)

Should be "label" not "title"

samples/v1.3/ElementsPreview/Input.Text.Style.json
```diff
@@ -0,0 +1,40 @@
...
+		{
+			"type": "Input.Text",
+			"id": "myComment",
+			"label": "Comment (style: text)",
+			"style": "text"
+		},
+		{
+			"type": "Input.Text",
+			"id": "myEmail",
+			"placeholder": "john.doe@adaptivecards.io",
```
RebeccaAnne, line 15: >placeholder [](start = 4, length = 11)

Do we still want to use placeholders in the new cards? I had thought they were being discouraged in general.

shalinijoshi19:
@golddove / @almedina-ms we should make sure the changes here reflect our POA as outlined offline in Becky's notes.. also there are some changes requested it seems.. thanks

shalinijoshi19:
We should close this out at this point since we've altered our POA on this bug

PR #4573, opened 2020-08-12
[Android] Add Base Properties Parsing APIs 

### LINKED ISSUE #4509: [Android][Dev Experience] There is no BaseActionParser class for parsing properties in BaseActionElement
# Platform

* Android

# Author or host

Dev

# Version of SDK

main

# Details

Devs creating custom action renderers have to redefine or remake the parsing for all properties that are contained in BaseActionParser, this should not be required and these properties should be easily parsed, either automatically or with an API

almedina-ms:
## Related Issue
Fixes #4509 

## Description
Adds the following methods to ease developers using our extensibility APIs to parse base properties automatically
* ```public static void deserializeBaseActionProperties(ParseContext context, JsonValue value, BaseActionElement actionElement)```
* ```public static void deserializeBaseActionPropertiesFromString(ParseContext context, String jsonString, BaseActionElement actionElement)```
* ```public static void deserializeBaseCardElementProperties(ParseContext context, JsonValue value, BaseCardElement cardElement)```
* ```public static void deserializeBaseCardElementPropertiesFromString(ParseContext context, String jsonString, BaseCardElement cardElement)```
* ```public static void deserializeBaseInputProperties(ParseContext context, JsonValue value, BaseInputElement inputElement)```
* ```public static void deserializeBaseInputPropertiesFromString(ParseContext context, String jsonString, BaseInputElement inputElement)```

## Misssing:
* Documentation
* Cast function from BaseElement to BaseCardElement
* Cast function from BaseElement to BaseActionElement

## How Verified
The fix was verified in two ways:
1. By updating the Custom elements in our sample app we can simulate how a developer would use our APIs
2. By adding unit tests we can exercise this behaviors more often

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4573)


RebeccaAnne:
>         baseCardElement->DeserializeBase<BaseCardElement>(context, json);

Can we call your new function from here instead of duplicating this code? 

---
Refers to: source/shared/cpp/ObjectModel/BaseCardElement.h:72 in b341bf0. [](commit_id = b341bf03a5a9ecc58a7af554b9d2ddc1ac05db07, deletion_comment = False)

PR #4576, opened 2020-08-12
Updates ctrl+enter checks for TextInput

### LINKED ISSUE #4575: [nodejs][Keyboard shortcut] Ctrl+Enter not working when inlineAction associated with a text.input
# Platform

* JavaScript

# Author or host

Host, using the SDK to render adaptive cards ourselves

SDK version 2.0.0  on npm

# Details

When having an `inlineAction` set against a `Text.Input`, pressing ctrl+enter with the keyboard when in the input field should trigger the inline action, right now, it is not.

This is seen in Chrome (v84.0.4147.125), haven't tried this in other browsers, but would imagine it's the same.

scottlovegrove:
## Related Issue
Fixes #4575

## Description
This ensures that the keyboard shortcut of `Ctrl+Enter` works when an inlineAction is assigned to a `Text.Input`.

## Notes
In my own project, I had to use `e.ctrlKey && e.charCode === 13 && this.inlineAction` as my condition. This PR contains the fix for this, but I'm unsure whether it is OK because apparently `charCode` is deprecated and `code `should be used, but that doesn't appear in my types as an available option 🤷‍♂️

## How Verified
I have been writing my own control to display in adaptive code and it works there, so figured I'd bring across the change to here too.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4576)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -2668,7 +2668,7 @@ export class TextInput extends Input {
         input.oninput = () => { this.valueChanged(); }
         input.onkeypress = (e: KeyboardEvent) => {
             // Ctrl+Enter pressed
-            if (e.keyCode == 10 && this.inlineAction) {
+            if (e.ctrlKey && e.charCode == 13 && this.inlineAction) {
                 this.inlineAction.execute();
             }
         }


dclaux:
Closing this PR as the fix is now in this other PR: https://github.com/microsoft/AdaptiveCards/pull/4840

PR #4577, opened 2020-08-12
[JS] Add keyboard invocation for media elements

paulcam206:
## Related Issue
Fixes VSO #24081306

## Description
Add keyboard event handlers to media play button to allow users to keyboard-invoke play.

## How Verified
* local build and test

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4577)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -2180,6 +2180,31 @@ export class Media extends CardElement {
         }
     }
 
+    private handlePlayButtonInvoke(event: UIEvent) : void
+    {
+        if (this.hostConfig.media.allowInlinePlayback) {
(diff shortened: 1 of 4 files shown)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2249,6 +2249,29 @@ export class Media extends CardElement {
                 }
             }
 
+            playButtonOuterElement.onkeypress = (e: KeyboardEvent) => {
```
dclaux, line 2252: Code for onkeypress handler is basically the same as that for onclick. I think we should move it to a helper method instead of duplicating.

Issue #4602, opened 2020-08-17
[Input Validation] Add support for focusLost validation


RebeccaAnne:
Tracking issue for support for validating inputs on focusLost.

- [ ] .NET
- [ ] Android
- [ ] iOS
- [ ] JavaScript
- [x] UWP

shalinijoshi19:
@RebeccaAnne are we still planning on shipping this for v20.09? This is merged at this point but as i understand it we are not prioritizing shipping this update.. Should this be reverted from main?

RebeccaAnne:
@shalinijoshi19 None of this work is in main. The UWP implementation is in the feature/focusLostValidation branch.

PR #4603, opened 2020-08-17
[Input Validation] Spec on Focus Lost Validation

RebeccaAnne:
## Related Issue
#4602

## Description
Spec for introducing focus lost validation. Includes host config schema changes, motivations, and description of behaviors.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4603)


schemas/host-config.json
```diff
@@ -360,6 +360,16 @@
 				},
 				"errorMessage": {
 					"$ref": "#/definitions/ErrorMessageConfig"
+				},
+				"validationBehavior":{
```
golddove, line 364: Thoughts on simply naming it `validate`?

e.g. `"validate": "onFocusLost"`

golddove, line 364: How about calling it `validationBegins` or `initialValidation`?

At first glance, the name didn't give me the impression that subsequent validation wouldn't be affected by this.

schemas/host-config.json
```diff
@@ -360,6 +360,16 @@
...
+					"type":"string",
+					"description":"Controls when inputs are validated",
+					"default":"onFocusLost",
+					"enum":[
+						"onFocusLost",
```
golddove, line 369: What about `onInputChange`? I can see hosts wanting validation to start as soon as a user starts typing, rather than when focus is lost.

Also worth considering `immediate`, where the host would start validation even before user interacts (e.g. initially empty required inputs, or a default that fails validation).

golddove (approved):

schemas/host-config.json
RebeccaAnne: We did have one respondent to our survey that chose "other" that requested basically the `onInputChange` behavior as you describe it, so it may be a reasonable option.

That said, based on reading up on this a bit using the research a previous PM did for us, I personally don't think that validating sooner than focus lost is a good idea. See for example Pitfall #1 at [https://baymard.com/blog/inline-form-validation](https://baymard.com/blog/inline-form-validation) for some discussion of this. I'm hesitant to give more options to hosts here that i think may make things worse for the user.

On the other hand, left to my own devices i'd probably remove this setting altogether and make everyone use onFocusLost, so it's possible my bias is unduly prescriptivist :)

---
In reply to: [472681978](https://github.com/microsoft/AdaptiveCards/pull/4603#discussion_r472681978) [](ancestors = 472681978)

schemas/host-config.json
RebeccaAnne: Good point. I like `initialValidation`.

---
In reply to: [472678974](https://github.com/microsoft/AdaptiveCards/pull/4603#discussion_r472678974) [](ancestors = 472678974)

specs/DesignDiscussions/FocusLostValidation.md
```diff
@@ -0,0 +1,132 @@
...
+
+#### onFocusLostWithInput
+The input is validated when it loses focus only if the user has entered a value into the input field. If this value is set we will not validate for the `isRequired` property on focus lost unless the user has typed something and subsequently removed it. 
+
+#### onSubmit
+This is the behavior that shipped with the 1.3 schema update. The inputs are validated when the submit button is pressed. No validation takes place before submit is pressed.
+
+### Subsequent Validations
+Once the field has been validated for the first time based on the behavior specified by the host config, all host config settings have the same subsequent behavior. Once the field has been marked as invalid, from that point forward it will be validated on each keystroke. This allows users to correct their errors and see when they have successfully met the validation criteria. Validation will continue to be run on each keystroke even after the error has been corrected, in order to allow users to continue to see if their new value is correct.
```
almedina-ms, line 59: is this a per field behaviour or a general validation? Take for example fields A and field B (maybe answer for B depends on answer for A), if only validation for A fails and the value has to be changed for B (now B will be invalid) it may seem odd that field A validates on each keystroke and B doesn't

almedina-ms, line 59: Also, why isn't the subsequent behaviour allowed as a starting behaviour? The very first research link makes it seem like some users may have a good experience with it
> Participant comments highlighted their strong preference for getting real-time feedback from our form:
>
>“You’d rather know about your mistakes as you go along.”

paulcam206:
@RebeccaAnne - should we go ahead and merge this, or should we just close it?

RebeccaAnne:
> @RebeccaAnne - should we go ahead and merge this, or should we just close it?

So this PR is going into a feature branch that was going to be for this feature, and that includes a UWP implementation. I'm going to go ahead and merge it, but I'll label the spec as "Draft" since we'll probably need to review it again if we actually decide to ship this at some point.

PR #4623, opened 2020-08-19
[Shared/UWP] Remove InputNecessityIndicators

### LINKED ISSUE #4622: [Shared/UWP] InputNecessityIndicators enum and property on AdaptiveCard
# Platform

* UWP
* Shared Model

# Details

An earlier version of the input validation design called for allowing card authors to indicate whether required and optional indicators should be shown by setting a property on the card. Although that was not part of the final design, that property definition and enum are still defined in the shared model and UWP, and shipped as part of our 2.0 UWP API surface.

RebeccaAnne:
## Related Issue
Fixes #4622

## Description
An earlier version of the input validation design called for allowing card authors to indicate whether required and optional indicators should be shown by setting a property on the card. Although that was not part of the final design, that property definition and enum were still defined in the shared model and UWP, and shipped as part of our 2.0 UWP API surface. This change removes this property and enum.

## How Verified
Built UWP and Shared Model solutions. Ran UWP and shared model tests.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4623)


source/android/adaptivecards/src/main/java/io/adaptivecards/objectmodel/AdaptiveCardObjectModel.java
```diff
@@ -1,6 +1,6 @@
 /* ----------------------------------------------------------------------------
  * This file was automatically generated by SWIG (http://www.swig.org).
- * Version 4.0.1
```
almedina-ms, line 3: This is ok, we should ask out team members to use swig 4.0.2

Issue #4624, opened 2020-08-19
[UWP] UWP has several unit test failures


RebeccaAnne:
# Platform

* UWP

# Details

The following UWP tests are failing in main:

- [x]  FallbackTests (Fixed by #4627)
  - ColumnFallbackAsync (crashes test host)
- [x] ObjectModelTest (Fixed by #4626)
  - ChoiceSetInput
  - DateInput 
  - TextInput
  - ToggleInput
- [x] UnitTests
  - TestAllCards (crashes test host)
    - [x] [v1.2/Elements/AdaptiveCard.BackgroundImage.DataUri.json](https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.2/Elements/AdaptiveCard.BackgroundImage.DataUri.json). Fixed by #4652
    - [x] [v1.2/Elements/Column.Fallback.json](https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.2/Elements/Column.Fallback.json). Same issue as FallbackTests failure. Fixed by #4627.
    - [x] Out of date round-trip json files causing test failures Fixed by #4645

RebeccaAnne:
All fixes are in and confirmed that UWP tests pass now in main.

PR #4626, opened 2020-08-19
[UWP] Fix UWP Object Model Tests

RebeccaAnne:
## Related Issue
Partial fix for #4624

## Description
The Object Model tests had some failures related to input validation. Specifically, it looks like properties were added to these tests without updating the expected serialized json, leading to failures.

Updated serialized json to include all properties. While I was in there, reorganized the initializers a bit to make it clearer which properties were being set and tested. Added a helper function to validate input related properties and called from input tests, and ensured all input tests were validating new properties.

## How Verified
Confirmed all tests under `ObjectModelTest` now pass.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4626)


source/uwp/UWPUnitTests/ObjectModelTests.cs
```diff
@@ -268,15 +292,18 @@ public void ColumnSet()
         {
             AdaptiveColumn column1 = new AdaptiveColumn
             {
+                // Column Properties
                 Bleed = true,
+                Width = "50px",
```
almedina-ms, line 297: width is out of order

PR #4660, opened 2020-08-26
[Site][Accessibility] Reflow during resize

golddove:
## Related Issue
ADO 23713154

## Description
When browser is resized (either by changing zoom level, or window size) to mobile size, the sample card disappears (you have to refresh the page to get a proper reflow).

## How Verified
Manual verification

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4660)


golddove:
The ADO bug referenced here was actually reported for WPF, so the bug remains unresolved.

PR #4695, opened 2020-08-28
[Samples] Update scenario samples to ensure match between templated and non-templated versions

### LINKED ISSUE #4578: [Website] Update samples to make sure template samples match non-templated versions
Website scenario samples are sometime mismatched. See https://adaptivecards.io/samples/Inputs.html with the "Use Templating" checkbox on and off.

RebeccaAnne:
## Related Issue
Fixes #4578

## Description
Updated scenario samples to match the templated and non-templated versions. 

## How Verified
Ensured all modified samples rendered as expected.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4695)


samples/v1.0/Scenarios/WeatherLarge.json
```diff
@@ -1,204 +1 @@
...
					"selectAction": {
						"type": "Action.OpenUrl",
						"title": "View Saturday",
						"url": "https://www.microsoft.com"
					}
				}
			]
		}
	]
}
```
paulcam206, line 1: Looks like the wrong line endings ended up in here :(

samples/v1.2/Scenarios/FlightDetails.json
```diff
@@ -160,7 +160,7 @@
 											"items": [
 												{
 													"type": "TextBlock",
-													"text": "November 12, 2017",
+													"text": "Sat, Mar 4, 2017",
```
almedina-ms, line 163: Shouldn't we have to update this textblocks to use {Date} formatting in the string? Seems like we're missing an opportunity here

shalinijoshi19: @almedina-ms with @RebeccaAnne out can you pick up where she left off here to shepherd this change in? Thanks!

paulcam206: I think we should probably fix the date stuff later rather than hold up the release. I'll file a new issue.

paulcam206: filed #4709

PR #4697, opened 2020-08-28
Android chevron fix cherry-pick

almedina-ms:
This is a cherry-pick of the chevron fix for Android from this pull request #4566 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4697)


shalinijoshi19:
Thanks! Curious - where's the corresponding Test/sample changes for regression testing?

almedina-ms:
> Thanks! Curious - where's the corresponding Test/sample changes for regression testing?

That's not a regression, it was a problem that always existed. The issue was that the chevron would have been showing in the wrong direction if a different show card action was "clicked" in the same action set. Unless we add interface testing then we don't have a way to add tests

PR #4735, opened 2020-09-04
[JS] Paint ToggleVisibility actions with `aria-controls` attribute

paulcam206:
## Related Issue
Fixes VSO 23940376

## Description
The `aria-controls` attribute lets screenreaders know when one element controls the visibility of another element. This change adds `aria-controls` to `ToggleVisibility` actions and reflects the `aria-expanded` property on the elements that are toggled.

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4735)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -95,6 +95,8 @@ export abstract class CardElement extends CardObject {
                 raiseElementVisibilityChangedEvent(this);
             }
         }
+
+        this._renderedElement?.setAttribute("aria-expanded", value.toString());
     }
 
     //#endregion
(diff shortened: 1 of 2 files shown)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4098,10 +4100,29 @@ export class ToggleVisibilityAction extends Action {
     // change introduced in TS 3.1 wrt d.ts generation. DO NOT CHANGE
     static readonly JsonTypeName: "Action.ToggleVisibility" = "Action.ToggleVisibility";
 
+    private updateAriaControlsAttribute() {
+        // apply aria labels to make it clear which elements this action will toggle
+        if (this.targetElements) {
+            const elementIds = Object.keys(this.targetElements);
+
+            if (elementIds.length > 0) {
+                this._renderedElement?.setAttribute("aria-controls", elementIds.join(" "));
```
dclaux, line 4109: I have made a point not to use the ? operator to ensure consuming applications will not fail at compile time if they are not using a recent enough version of Typescript. I think we should continue not using it for now, but if we are going to use it we need to at least check with Teams that it's OK with them.

dclaux: I don't think this works. `elementIds` holds the Ids of the target AC elements, not the Ids of DOM element. For this to work, we need to dynamically set the Ids of the target DOM elements to be the same as the Ids of the target AC element or simply always assign the AC Id to all rendered DOM elements in `CardElement.render()`. Given the latter has been an ask from several developers, I suggest we do that and document it.

dclaux (changes requested):

dclaux (approved):
ould you add a note in README.md to indicate that element Ids are now stamped on their associated DOM elements?

PR #4736, opened 2020-09-04
[Samples] Update FlightDetails sample to have more logical reading order

paulcam206:
## Related Issue
Fixes VSO 24096000

## Description
Update the FlightDetails sample to use a `RichTextBlock` in a single set of `ColumnSet`s so that screenreaders can give a more reasonable reading experience. Prior to this, rather than reading down a column, Narrator would read left to right and top to bottom (i.e. it would read SFO's time then JFK's time, then SFO's date, then JFK's date etc.)

## How Verified
* local build and narrator

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4736)

### DIFF
--- a/samples/v1.2/Scenarios/FlightDetails.json
+++ b/samples/v1.2/Scenarios/FlightDetails.json
@@ -27,7 +27,7 @@
 							"items": [
 								{
 									"type": "TextBlock",
-									"size": "ExtraLarge",
+									"size": "extraLarge",
 									"weight": "lighter",
 									"color": "accent",
 									"text": "Flight to JFK",
@@ -74,7 +74,7 @@
 											"items": [
 												{
 													"type": "TextBlock",
-													"size": "ExtraLarge",
+													"size": "extraLarge",
 													"weight": "lighter",
 													"text": "SFO"
 												}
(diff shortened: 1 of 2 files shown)


samples/v1.2/Scenarios/FlightDetails.json
```diff
@@ -171,47 +157,32 @@
 											"type": "Column",
 											"items": [
 												{
-													"type": "TextBlock",
-													"horizontalAlignment": "Right",
-													"text": "Sat, Mar 4, 2017",
-													"isSubtle": true,
-													"wrap": true
+													"type": "RichTextBlock",
```
RebeccaAnne, line 160: >RichTextBlock [](start = 22, length = 13)

Do we have (or plan to have) a customer facing document on how to make sure your cards are accessible? This seems like something we should capture in a tips and tricks kind of way to encourage others to avoid this pitfall.

RebeccaAnne (approved):
:shipit:

samples/v1.2/Scenarios/FlightDetails.json
paulcam206: filed #4746

Issue #4708, opened 2020-08-31
Pic2card on official website


productboard-ac:

productboard-ac:
Link to feature: https://adaptivecards.productboard.com/feature-board/planning/features/5457426

matthidinger:
@haridas feel free to reference this issue with your PR

shalinijoshi19:
@matthidinger anything else that needs to be done here before we close this out?

matthidinger:
I'll move it to 20.10 and we'll close it when @haridas has things buttoned up

haridas:
@matthidinger  Please review the PR #4893

shalinijoshi19:
@matthidinger this *may* have Accessibility implications - do we know/have we double checked with the SMEs? If not i can spin up a thread real quick..

shalinijoshi19:
@RebeccaAnne FYI around the Accessibility question.

shalinijoshi19:
@matthidinger @haridas since this is pending Accessibility Review slated for this week, I am going to have to move this out to 20.11 :(  -- The good news is that if there are no concerns raised in today's review we can pull this back in and get this checked into main && release/20.10 branches for inclusion in the 20.10 payload; Worst case we bundle this up as part of our 20.11 release next month.

shalinijoshi19:
Moving this out to 21.01 to allow time for accessibility fixes + test passes.

haridas:
The accessibility team has given **test pass** confirmation, we can proceed further on this.

RebeccaAnne:
This has been shipped.

PR #4742, opened 2020-09-08
[pic2card] Upstream release fixes and changes

haridas:
- Normalised the api response for template / non-template apis
- Dockerfile for both detr and rcnn models
- Reduced the docker image size close to 1.1GB ( Before it used to be > 2G )
- Removed the need for tf_object_detection library, which was adding lot of unnecessary dependencies for lesser gain.
- Added github actions to publish the backend to github container registry ( ghrc.io )
- Tesseract fixes to avoid mdfile false triggering special characters.

Related Issue: https://github.com/microsoft/AdaptiveCards/issues/4708

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4742)


haridas:
Tried to rerun the .NET and .iOS test builds, seems they are failing repeatedly. @matthidinger @paulcam206 could you please check it.

Vasanth-S (approved):

Issue #4731, opened 2020-09-03
[Android][Performance] Reduce the number of empty views generated


almedina-ms:
# Platform
* Android

# Author or host

dev

# Version of SDK

main

# Details

Elements such as containers or columns generate a lot of empty views (such as linear layouts) that the rendering performance as well as the debuggability of issues with this elements.

shalinijoshi19:
@almedina-ms I've moved this to our backlog for now - if you do have a fix for this available feel free to pull it out.

PR #4760, opened 2020-09-09
[Android] Fix auto column behaviour

### LINKED ISSUE #3855: [Android][inconsistency android vs web] [Columns in a column set get pushed out of the card when width set to auto]
# Platform
* Android

# Author or host
Host

# Version of SDK
1.2.6

# Details
For the attached card json.
The button in the column set is cutoff on android
image

The same card on web
image

Note : On web, no matter how large the string in the first column becomes, the second column is never pushed out of the card completely (there is a sliver of the button in the second column as seen below). Whereas on android, it looks like the column did not even render.

Card with a lot of text in the first column as rendered on web
image

P.S. : What is the expectation for the value "auto" for the column width (refer to the attached card). Is there in any documentation around the functioning of the value.

[widthAuto.json.zip](https://github.com/microsoft/AdaptiveCards/files/4341527/widthAuto.json.zip)

almedina-ms:
## Related Issue
Fixes #3855
References #4731 

## Description
Changes the rendering of ```Column``` and ```ColumnSet``` to use flexbox layout, this will make the Column rendering more consistent with how JavaScript and iOS performs it. Also, some views are removed from the ```Column``` rendering to ease the debugging process and reduce the number of empty layouts

## How Verified
The fix was verified by running the smoke test scenarios

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4760)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
```diff
@@ -214,7 +212,8 @@ private static ConstraintSet createConstraints(Context context, ImageView imageV
         ConstraintSet constraints = new ConstraintSet();
 
         // ConstraintSet requires unique id
-        if(imageView.getId() == View.NO_ID) {
+        if(imageView.getId() == View.NO_ID)
```
paulcam206, line 215: nitpick (since you fixed the same thing below):

```suggestion
        if (imageView.getId() == View.NO_ID)
```

paulcam206 (approved):
looks a lot cleaner... nice! 👍

PR #4762, opened 2020-09-09
[Samples] Update restaurant sample to read more naturally for a11y

paulcam206:
## Related Issue
Fixes VSO 23712499

## Description
The use of symbols rather than text leads to a suboptimal story for accessibility. Here we move to a more text-based model for the restaurant sample cards.

## How Verified
* Local build & narrator

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4762)

### DIFF
--- a/samples/Templates/Scenarios/Restaurant.data.json
+++ b/samples/Templates/Scenarios/Restaurant.data.json
@@ -25,5 +25,5 @@
 		"description": "Great concept and a wide selection of beers both on tap and bottled! Smaller wine selection than I wanted, but the variety of beers certainly made up for that. Although I didn't order anything, my boyfriend got a beer and he loved it. Their prices are fair too. \n\nThe concept is really awesome. It's a bar/store that you can bring outside food into. The place was pretty packed tonight. I wish we had stayed for more than one drink. I would have loved to sample everything!",
 		"author": "Blaire S."
 	}],
-	"priceRange": "$$"
+	"priceRange": "moderately priced"
 }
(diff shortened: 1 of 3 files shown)


samples/Templates/Scenarios/Restaurant.data.json
```diff
@@ -25,5 +25,5 @@
 		"description": "Great concept and a wide selection of beers both on tap and bottled! Smaller wine selection than I wanted, but the variety of beers certainly made up for that. Although I didn't order anything, my boyfriend got a beer and he loved it. Their prices are fair too. \n\nThe concept is really awesome. It's a bar/store that you can bring outside food into. The place was pretty packed tonight. I wish we had stayed for more than one drink. I would have loved to sample everything!",
 		"author": "Blaire S."
 	}],
-	"priceRange": "$$"
-}
\ No newline at end of file
+	"priceRange": "mid-priced"
```
almedina-ms, line 28: I wonder how they make google maps accessible, I've seen the $ or $$$ model work before

samples/v1.0/Scenarios/Restaurant.json
```diff
@@ -23,13 +23,13 @@
 						},
 						{
 							"type": "TextBlock",
-							"text": "4.5 ★★★★★ (176) · $$",
+							"text": "4.5 stars (176 reviews) · mid-priced",
```
almedina-ms, line 26: It would be nice to investigate if we can extend the extensibility to be able to deal with this kind of scenarios
https://stackoverflow.com/questions/55966205/star-rating-component-what-are-the-correct-accessibility-and-semantics-to-be-u

almedina-ms (approved):

samples/Templates/Scenarios/Restaurant.data.json
golddove: Well, we would need to add some mechanism for authors to specify alternative text.

Could potentially be solved if/when we figure out tooltips on all elements? Not sure.

RebeccaAnne: Yeah, i agree. We should probably have something tracking figuring out alt-text on Text elements for this kind of scenario. 

---
In reply to: [485986018](https://github.com/microsoft/AdaptiveCards/pull/4762#discussion_r485986018) [](ancestors = 485986018)

RebeccaAnne (approved):
:shipit:

samples/Templates/Scenarios/Restaurant.data.json
paulcam206: filed #4764 to track

Issue #4478, opened 2020-07-23
[Desktop vs Mobile][Platform Inconsistency] TextBlock with no text property


Gunashree-R-V:
# Platform

- [x] Android
- [x] iOS
- [x] UWP 
- [x] .Net

# Author or host

Microsoft Teams

# Version of SDK

1.2.10 [iOS & Android]

# Details

As per the [TextBlock schema](https://adaptivecards.io/explorer/TextBlock.html), **`text`** property is required and card without this must be an invalid card.
However, there is inconsistency on rendering of a card with no text property in Text Block on different platforms.

Desktop renders this invalid card. Here is the screenshot:

| Desktop| Android | 
|-----------|:-----------:|
| ![image](https://user-images.githubusercontent.com/28745768/91831824-292d3480-ec62-11ea-97f1-96f09eba4b03.png)| ![image](https://user-images.githubusercontent.com/28745768/91831936-54b01f00-ec62-11ea-9bad-5824df7ef12e.png)| 

While, mobile(Android and iOS) throws an exception stating 
`Property is required but was found empty: text`

Attaching the JSON for reference:

[EDU-Card.zip](https://github.com/microsoft/AdaptiveCards/files/4964387/EDU-Card.zip)

cc: @VikrantSingh01, @anbah34, @shbhadra

Issue #4305, opened 2020-07-07
[.NET] Excess spacing in empty text blocks


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

- [x] .NET HTML
- [x] .NET WPF

# Details
Spun off from #3702. This bug tracks fixing the spacing issue described in that bug for .NET. Specifically, if a TextBlock has an text property with an empty string, no separator or empty vertical space should be rendered for that text block. The text block should instead be silently dropped.

Issue #4565, opened 2020-08-10
[Inconsistency] Parse failures for a given property should not fail the entire card


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

* .NET HTML
* .NET WPF
* Android
* iOS
* JavaScript
* UWP

# Details

Currently the shared model platforms fail parsing of the entire card in the case that there is a parse error in any property. JavaScript on the other hand, drops only the failing property, returning a warning and continuing to parse. This issue tracks updating all platforms to behave consistently with JavaScript.

shalinijoshi19:
@jwoo-msft something to look into driving once you are back as part of the Inconsistency workstream! Thanks

shalinijoshi19:
@jwoo-msft what's the POA on this one?

PR #4765, opened 2020-09-11
Updated handling missing text property of TextBlock

jwoo-msft:
## Related Issue
reference #4478  
reference #4801
reference #4305 
## Description
Updated Android / UWP to respect the new parsing / rendering scheme.
Renderers don't fail to render the entire card. Renders make rendering decisions per element.
Renderers will try to render card after they encounter a TextBlock with no text.

## How Verified
1. A new unit test is added.
2. Regression tests are run on 1.0 TextBlocks cards

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4765)


source/shared/cpp/ObjectModel/TextElementProperties.cpp
```diff
@@ -167,9 +167,14 @@ void TextElementProperties::SetLanguage(const std::string& value)
     m_language = value;
 }
 
-void TextElementProperties::Deserialize(const ParseContext& context, const Json::Value& json)
+void TextElementProperties::Deserialize(ParseContext& context, const Json::Value& json)
 {
-    SetText(ParseUtil::GetString(json, AdaptiveCardSchemaKey::Text, true));
+    SetText(ParseUtil::GetString(json, AdaptiveCardSchemaKey::Text, false));
```
RebeccaAnne, line 172: >GetString [](start = 23, length = 9)

We actually want this behavior for all required properties going forward now, right (#4565)? Rather than adding this code here specifically in TextBlock, do we want to update GetString to not throw and to instead add a warning when a required property is missing? That probably requires some follow up in the renderers to make sure that we don't crash on missing properties. But otherwise we'll end up with all "required" properties having this block of code and never passing `true` to the ParseUtil functions.

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveTextBlockRenderer.cs
```diff
@@ -17,6 +17,11 @@ public static class AdaptiveTextBlockRenderer
     {
         public static FrameworkElement Render(AdaptiveTextBlock textBlock, AdaptiveRenderContext context)
         {
+            if (String.IsNullOrEmpty(textBlock.Text))
+            {
+                return null;
+            }
```
RebeccaAnne, line 23: The problem with this change is that it causes the higher layers of the stack to think that we don't have a TextBlock renderer. This will lead to us triggering fallback, or if there's no fallback returning the warning "No renderer for element TextBlock" which is not correct. I ran into this before which is why i ended up spinning #4305 off of #3702. We need to refactor this code somehow to be able to distinguish between a missing renderer and an element dropped for missing required properties (which i believe we decided should *not* trigger fallback).

samples/v1.0/Tests/TextBlock.Test.json
```diff
@@ -0,0 +1,55 @@
+{
```
RebeccaAnne, line 1: Can we give this file a more descriptive name?

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveTextBlockRenderer.cs
almedina-ms: I believe with the current implementation this change actually works, some code was refactored in order to determine if labels and error messages had to be rendered and as such we look for the first rendereable element with the AdaptiveRenderContext.GetRendereableElement() method.

If we look at the code in line 107, as the rendered ui element is null we're going to skip the rendering of spacing and other properties
https://github.com/microsoft/AdaptiveCards/blob/main/source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveContainerRenderer.cs

``` csharp
AdaptiveTypedElement rendereableElement = context.GetRendereableElement(cardElement);

// if there's an element that can be rendered, then render it, otherwise, skip
if (rendereableElement != null)
{
    FrameworkElement uiElement = context.Render(rendereableElement);
    if (uiElement != null)
    { // logic regarding spacing, tags, height and other stuff 
    } 
}

```

almedina-ms: I haven't tested this PR so I can be wrong

almedina-ms: sorry, I stand corrected, the warning is being raised

almedina-ms: image

almedina-ms: Fix should be done by removing duplicate code from
https://github.com/microsoft/AdaptiveCards/blob/main/source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveRenderContext.cs starting from line 245

Seems like there is some left over logic from GetRendereableElement that is duplicated, as we're precalculating the element that can be rendered we could remove the whole if in line 301

source/shared/cpp/ObjectModel/TextElementProperties.cpp
jwoo-msft: As discussed, I've created a new issue to track to update all of the required string parsing as part of [https://github.com/microsoft/AdaptiveCards/issues/4565](https://github.com/microsoft/AdaptiveCards/issues/4565).  [https://github.com/microsoft/AdaptiveCards/issues/4834](https://github.com/microsoft/AdaptiveCards/issues/4834) is the one track this particular issue.

---
In reply to: [494592127](https://github.com/microsoft/AdaptiveCards/pull/4765#discussion_r494592127) [](ancestors = 494592127)

samples/v1.0/Tests/TextBlock.Test.json
jwoo-msft: I will remove this card, and merge this card with one in the FactSet into a new card on a new PR.

---
In reply to: [494599048](https://github.com/microsoft/AdaptiveCards/pull/4765#discussion_r494599048) [](ancestors = 494599048)

almedina-ms (approved):

RebeccaAnne (approved):
:shipit:

PR #4776, opened 2020-09-15
[Android] Add support for light fonts

### LINKED ISSUE #4015: [Android][TextBlock] Font Weights are not consumed.
# Platform

* Android

# Author or host

Host - Teams Android app

# Version of SDK

1.2.6

# Issue
 

- Host config has an option to control font weight by providing metrics to it(As mentioned in the public documentation: https://docs.microsoft.com/en-us/adaptive-cards/rendering-cards/host-config#fontweightsconfig).Updating these metrics remain unaffected on client. 

- Also, the TextBlock with fontWeight=“lighter” has Italic typeface. 
The issue is illustrated in the image below:

image

However, desktop has normal typeface with lesser fontWeight.

image

Moreover it is found that each fontWeight property is mapped to Typeface as below on Android:
Lighter -> Italic
Default -> Normal
Bolder -> Bold.

On the other hand, iOS and Desktop consumes the values from hostConfig for each fontWeight and use that:

Also, attaching the relevant JSON.
[fontWeight.zip](https://github.com/microsoft/AdaptiveCards/files/4633852/fontWeight.zip)

cc: @VikrantSingh01.

almedina-ms:
## Related Issue
Fixes #4015 

## Description
Adds two fixes for consumption of light fonts:
1. In Android API 28 a new typeface constructor was made to support weights
2. For older clients, the current typeface creation process was modified to look up for fonts that can be set as light, otherwise we default to sans-serif-light to honor the request of a light font

## How Verified
The fix was verified manually using the android mobile app

| < API 28 | API 28+ |
| --- | --- |
| image | image  |

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4776)


jwoo-msft (approved):

samples/HostConfig/sample.json
```diff
@@ -14,7 +14,7 @@
 	"supportsInteractivity": true,
 	"fontTypes": {
 		"default": {
-			"fontFamily": "'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu', 'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif",
+			"fontFamily": "'Segoe UI', 'sans-serif', 'Oxygen', 'Ubuntu', 'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue'",
```
shalinijoshi19, line 17: so are we removing support for `Roboto`?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
```diff
@@ -68,7 +71,25 @@ public static void setTextSize(TextView textView, FontType type, TextSize textSi
...
-        textView.setTypeface(TextRendererUtil.getTextFormat(hostConfig, type), m_textWeightMap.get(textWeight));
+        Typeface typeface = TextRendererUtil.getTextFormat(hostConfig, type, textWeight == TextWeight.Lighter);
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P)
+        {
+            Typeface weightedTypeface = Typeface.create(typeface, (int) hostConfig.GetFontWeight(type, textWeight), false);
+            textView.setTypeface(weightedTypeface);
+        }
+        else
+        {
```
shalinijoshi19, line 82: be nice to add some sort of comments here as to why the build version check is needed.

samples/HostConfig/sample.json
almedina-ms: For some undocumented reasons, when trying to set the font as roboto the thing crashes, sans-serif is the consumer friendly name for roboto, hence the change

PR #4779, opened 2020-09-15
Allows a background image to be set for Containers and derivitives

### LINKED ISSUE #4741: [NodeJS][Column] No way to set backgroundimage on a column
# Platform

* JavaScript

# Author or host

Author, sending cards to our own custom renderer.

# Version of SDK

All versions, but seeing this with 1.2.6 and 2.3.0

# Details

The Schema and Designer at adaptivecards.io both say you can add a background image to a `Column`, however in the npm package, there is no way to actually achieve this. There is a `backgroundImage` property, but this is readonly and having looked in the source, there doesn't appear to be anywhere that's actually setting the backing field within Column or Container.

scottlovegrove:
## Related Issue
Fixes #4741 

## Description
This fix allows the `backgroundImage` property on Containers and everything that extends it. Currently this was set to be a getter only. 

## How Verified
Have added a test to `column.spec.ts`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4779)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -5286,6 +5286,10 @@ export class Container extends StylableCardElementContainer {
         return this.getValue(Container.backgroundImageProperty);
     }
 
+    set backgroundImage(value: BackgroundImage){
+        this.setValue(Container.backgroundImageProperty, value);
+    }
+
     @property(Container.verticalContentAlignmentProperty)
     verticalContentAlignment: Enums.VerticalAlignment = Enums.VerticalAlignment.Top;
 
--- a/source/nodejs/adaptivecards/src/__tests__/components/column.spec.ts
+++ b/source/nodejs/adaptivecards/src/__tests__/components/column.spec.ts
@@ -1,8 +1,17 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-import { Column } from "../../card-elements";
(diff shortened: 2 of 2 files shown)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5286,6 +5286,10 @@ export class Container extends StylableCardElementContainer {
         return this.getValue(Container.backgroundImageProperty);
     }
 
+    set backgroundImage(value: BackgroundImage){
```
dclaux, line 5289: No, the lack of setter is by design. The `backgroundImage` property is always initialized to a BackgroundImage instance automatically (if that's not the case, it might be a regression and that would need to be fixed). Developers set `container.backgroundImage.url` etc. but never `container.backgroundImage`.

dclaux (changes requested):
This change isn't necessary. The current behavior is by design. If there is something that doesn't work as it should it'd be great if we could chat about that is first.

scottlovegrove:
Ah thanks for the info. I'll verify this on Monday, and if that's the case that will certainly unblock me.

It must be said though that this is not clear, and the usage with the node sdk is not documented to suggest this (unless I've missed it somewhere).

scottlovegrove:
Have confirmed that the solution provided by @dclaux resolves my issue, so closing this PR.

Issue #4767, opened 2020-09-11
[Javascript] TypeError: Cannot read property 'isValid' of undefined


scottlovegrove:
# Platform

What platform is your issue or question related to? (Delete other platforms).
* JavaScript

# Author or host

Author

If you're an author, who are you sending cards to?

Custom renderer, but not important for this bug

# Version of SDK

2.3.0 of adaptivecards on npm

# Details

If I try and return an adaptive card as part of an API response, and the adaptive card isn't the root item, I'm getting the following error:

```
TypeError: Cannot read property 'isValid' of undefined
    at Version.compareTo (D:\doist\adaptivecards-repro\node_modules\adaptivecards\src\serialization.ts:66:44)
    at AdaptiveCard.SerializableObject.internalToJSON (D:\doist\adaptivecards-repro\node_modules\adaptivecards\src\serialization.ts:870:75)
    at AdaptiveCard.Container.internalToJSON (D:\doist\adaptivecards-repro\node_modules\adaptivecards\src\card-elements.ts:5449:29)
    at AdaptiveCard.ContainerWithActions.internalToJSON (D:\doist\adaptivecards-repro\node_modules\adaptivecards\src\card-elements.ts:6157:29)
    at AdaptiveCard.internalToJSON (D:\doist\adaptivecards-repro\node_modules\adaptivecards\src\card-elements.ts:6424:29)
    at AdaptiveCard.SerializableObject.toJSON (D:\doist\adaptivecards-repro\node_modules\adaptivecards\src\serialization.ts:913:18)
    at AdaptiveCard.CardElement.toJSON (D:\doist\adaptivecards-repro\node_modules\adaptivecards\src\card-elements.ts:326:28)
    at AdaptiveCard.toJSON (D:\doist\adaptivecards-repro\node_modules\adaptivecards\src\card-elements.ts:6548:28)
    at JSON.stringify (<anonymous>)
    at stringify (D:\doist\adaptivecards-repro\node_modules\express\lib\response.js:1123:12)
```

I've managed to create a repro project with this in. Just run `npm i && npm run start` then point to `http://localhost:4001/card` to reproduce
[adaptivecards-repro.zip](https://github.com/microsoft/AdaptiveCards/files/5210152/adaptivecards-repro.zip)

The key part to reproducing the problem is this:
```typescript
res.send({
        card: myAdaptiveCard,
    });
```
What then happens is `card` is passed through as the context of `toJSON` which results in the above error when comparing versions.

[...]

Issue #4734, opened 2020-09-04
Designer data binding error when sample data contains empty array


IntranetFactory:
Clicking on a data binding button causes an error when sample data contains any empty array:
![image](https://user-images.githubusercontent.com/7462342/92202371-dc598180-ee4c-11ea-9431-2ed5e5730a25.png)

shalinijoshi19:
@dclaux could you take a look at this pls? thanks

Issue #4483, opened 2020-07-24
Text Blocks default to false for wrap


anbah34:
# Platform

All Platforms. The issue is more relevant to Mobile - iOS and Android.

# Author or host

Are you an author (like sending something to Outlook), or a host that is rendering your own cards? If host, please specify the corresponding host (app) name.

If you're an author, who are you sending cards to?

1.2.10

What version are you using? Ex: NuGet 1.0.2, or latest main, etc...

# Details

Explain your issue. If you just have a question, please post [on StackOverflow instead

Why is it that the default value for the word wrapping is kept as false. Most times the developers test the card only on desktop and due to the large real estate the text does not get truncated but on mobile it does. I don't understand why the default value is not set to true for word wrap

![MicrosoftTeams-image-31](https://user-images.githubusercontent.com/65276405/88364696-5b18c480-cda1-11ea-834d-5ccb0e8e88b7.png)
<img width="307" alt="Screen Shot 2020-07-21 at 12 34 29 PM" src="https://user-images.githubusercontent.com/65276405/88364701-5d7b1e80-cda1-11ea-919d-d9c64dfe34ff.png">

anbah34:
@shalinijoshi19 Please see the an example of the issue that is caused with the current behavior wherin the text gets truncated unless the developer has specified word wrap to be true

cc- @VikrantSingh01

anbah34:
FYI @matthidinger, @jwoo-msft

shalinijoshi19:
We'll be keeping the behavior here, but will be updating the designer instead.

jwoo-msft:
Change for updating the designer is merged. Closing it per discussion.

PR #4791, opened 2020-09-21
[TS] Renderer and designer fixes

dclaux:
## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/4767
Fixes https://github.com/microsoft/AdaptiveCards/issues/4734
References https://github.com/microsoft/AdaptiveCards/issues/4483

## Description
- When `toJSON()` is automatically invoked via `JSON.stringify()`, it gets a parameter of type `string`. The AC renderer's implementation of `toJSON()` takes an optional parameter of type `BaseSerializationContext`. When the value of this parameter is not `undefined`, the code assumes it is of type `BaseSerializationContext` which is not the case when `toJSON()` is invoked via `JSON.stringify()`, and thus generates an exception. This PR fixes that by checking the type of the parameter `toJSON()` receives.
- The data structure generated from JSON would cause an exception when the JSON contained an empty array. The code now properly handles that case.
- New TextBlock elements created in the designer via drag/drop now have their `wrap` property set to `true` by default

## How Verified
Verified manually in adaptivecards-designer-app and [test site](https://adaptivecardsci.z5.web.core.windows.net/pr/4791)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4791)

### DIFF
--- a/source/nodejs/adaptivecards-designer/src/data.ts
+++ b/source/nodejs/adaptivecards-designer/src/data.ts
@@ -172,15 +172,17 @@ export class ArrayData extends DataType {
     generateSampleData(): any {
         let result = [];
 
-        for (let i = 1; i <= 3; i++) {
-            result.push(this.dataType.generateSampleData());
+        if (this.dataType) {
+            for (let i = 1; i <= 3; i++) {
+                result.push(this.dataType.generateSampleData());
+            }
         }
 
         return result;
     }
 
     getChildFields(): FieldDefinition[] {
-        return this.dataType.getChildFields();
+        return this.dataType ? this.dataType.getChildFields() : [];
     }
 
     qualifyFieldName(fieldName: string, fieldIsLeaf: boolean) {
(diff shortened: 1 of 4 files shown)


shalinijoshi19:
Hey @dclaux mind verifying the 3 changes also in the test site and including that little bit in the "How verified" section pls? https://adaptivecardsci.z5.web.core.windows.net/pr/4791

source/nodejs/adaptivecards-designer/src/data.ts
```diff
@@ -172,15 +172,17 @@ export class ArrayData extends DataType {
     generateSampleData(): any {
         let result = [];
 
-        for (let i = 1; i <= 3; i++) {
-            result.push(this.dataType.generateSampleData());
+        if (this.dataType) {
+            for (let i = 1; i <= 3; i++) {
```
shalinijoshi19, line 176: Nit Nit nit: Any specific reason why you prefer initializing to 1 vs 0 to fashion the loop construct? _havent checked but assuming you use this same pattern throughout the code base? If not any reason to diverge here?_ (
```suggestion
            for (let i = 0; i < 3; i++) {
```

Also just for my understanding as i start to get familiar with the code, where is the hardcoded 3 coming from again? What does it signify? Thanks again

shalinijoshi19 (approved):

source/nodejs/adaptivecards-designer/src/data.ts
dclaux: Because this is just a counter, and the index isn't used anywhere. I tend to count from 1 to 3 in real life, not from 0 to 2 :)

PR #4793, opened 2020-09-21
[Samples] Update cards to wrap long text

### LINKED ISSUE #4712: [Sample Cards] Text in sample cards truncated
# Platform

* Android
* iOS

# Issue

Newly updated Sporting Event card and Flight Update card have slightly longer text. For example, instead of **"San Francisco"**, it is **"San Francisco Airport"** after the update. Mobile platforms with narrower width such as Android and iOS would truncate the updated text. This is not what we want, we want to show all the text, so the ask is to add "wrap": true to those cards. Another way to fix this issue is making the "wrap" : true the default value as  asked in #4483.

almedina-ms:
## Related Issue
Fixes #4712

## Description
Change some textblocks to have wrap set as true to make all the content appear in screen

## How Verified
The fix was verified manually and with the UWP test app

image

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4793)


shalinijoshi19:
> ## Related Issue
> Fixes #4712
> 
> ## Description
> Change some textblocks to have wrap set as true to make all the content appear in screen
> 
> ## How Verified
> The fix was verified manually and with the UWP test app
> 
> image
> 
> image
> 
> ###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4793)

Thanks for the super quick turnaround on this @almedina-ms ; Can you verify the fix on iOS +Android where the original issue was first raised against as well? Thanks

shalinijoshi19 (approved):

Issue #4788, opened 2020-09-18
[Adaptive Card] AdaptiveCard.FromJson() not parsing adaptive card json of version 1.2


SunilsingRana:
# Adaptive Card V 1.2

What platform is your issue or question related to? (Delete other platforms).

* .NET HTML
* .NET WPF

# Author or host

Host

# Version of SDK

Adaptive card V 1.2, Nuget AdaptiveCards V 2.1.0, Bot Framework v4

 
# Details

We have setup a QnA service in which in the response I am giving the adaptive card json. And I parse the json in my bot framework code. The adaptive card json was created in V 1.2 from adaptive card designer site and below is my json payload. I am not using template here. 
Now when i use this json and use it to generate the adaptive card in my bot framework code, it gives me json parsing error.

Below is the piece of code to parse the json.
`                        string parsedJson = (JObject.Parse(response[0].Answer).ToString(Formatting.None));

                        AdaptiveCardParseResult result = AdaptiveCard.FromJson(parsedJson);`

I am getting error in **AdaptiveCard.FromJson()** method.

The error I am facing is below.

Error Message: `Current error context error is different to requested error.`
Source: Newtonsoft.Json

I further drilled down more in the issue, and found out that there are two properties from JSON which were causing the issue. The properties are **horizontalAlignment** and **verticalAlignment**. If I remove this two properties my from json payload , then it parses successfully. This two properties are newly added in v1.2 of adaptive card.

Below is my sample json that i am using. Any help or workaround to render this adaptive card will be appreciated.

[...]

PR #4800, opened 2020-09-23
Updated dotnet Additional Properties Handling

jwoo-msft:
## Related Issue
reference #4788
## Description
1. Updated JSON attributes.
 Json properties with default values were collected in additional properties by the JSON serializer.
 To unwind this behavior, we used the reflection to filter properties that ended up in the additional properties this way.
 The reflection can be very expensive if the card's depth is deep and has many children.
 These issues seemed to have been fixed with the use of IgnoreAndPopulate option for DefaultValue handling.
1. [Ignore and IgnoreAndPopulate](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_DefaultValueHandling.htm)
2. Changed Ignore to IgnoreAndPopulate and skipped the filtering.
3. Changed to have serialization error's flag set to handled since we do not surface the exception and wan to continue parsing.

## How Verified
1. New unit test is added
2. Existing unit tests were run
3. Manual verification of all Scenario cards.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4800)

### DIFF
--- a/source/dotnet/Samples/WPFVisualizer/MainWindow.xaml.cs
+++ b/source/dotnet/Samples/WPFVisualizer/MainWindow.xaml.cs
@@ -54,6 +54,9 @@ namespace WpfVisualizer
             _synth = new SpeechSynthesizer();
             _synth.SelectVoiceByHints(VoiceGender.Female, VoiceAge.Adult);
             _synth.SetOutputToDefaultAudioDevice();
+            var timer = new DispatcherTimer { Interval = TimeSpan.FromSeconds(1) };
+            timer.Tick += Timer_Tick;
+            timer.Start();
 
             foreach (var config in Directory.GetFiles("HostConfigs", "*.json"))
             {
@@ -97,6 +100,15 @@ namespace WpfVisualizer
         }
 
         public AdaptiveCardRenderer Renderer { get; set; }
+
+        private void Timer_Tick(object sender, EventArgs e)
+        {
+            if (_dirty)
+            {
+                _dirty = false;
+                RenderCard();
+            }
+        }
 
         public string CardPayload
         {
@@ -486,11 +498,6 @@ namespace WpfVisualizer
                 templateData = null;
             }
             templateDataTextBox.Text = templateData;
-        }
-
-        private void renderCard_Click(object sender, RoutedEventArgs e)
-        {
-            RenderCard();
         }
     }
 }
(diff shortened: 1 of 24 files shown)


source/dotnet/Samples/WPFVisualizer/MainWindow.xaml.cs
```diff
@@ -500,5 +487,10 @@ private void loadTemplateDataButton_Click(object sender, RoutedEventArgs e)
             }
             templateDataTextBox.Text = templateData;
         }
+
+        private void renderCard_Click(object sender, RoutedEventArgs e)
```
almedina-ms, line 491: why is this change being made, I'm understanding you're adding a Button to perform the rendering rather than doing the rendering automatically?

jwoo-msft: when you try to debug, and want to modify json, our current app will constanly render the app. If changes can be made in 1 second, this is somewhat tolerable, but if we have to make changes in multiple places, rendering takes very long time, it's very painful experience. I made changes that we make all the change we want, and just press the button to render.

---
In reply to: [494636503](https://github.com/microsoft/AdaptiveCards/pull/4800#discussion_r494636503) [](ancestors = 494636503)

source/dotnet/Test/AdaptiveCards.Test/AdaptiveCardApiTests.cs
```diff
@@ -776,9 +776,8 @@ public void BadImageWidthsAsAdditionalProperties()
 
             var result = AdaptiveCard.FromJson(payload);
 
-            // Expect one warning for each unknown property (3)
             // and one for each bad pixel size (2)
```
RebeccaAnne, line 779: >and one for each bad pixel size (2) [](start = 15, length = 35)

[nit] Could you update this comment to be a complete sentence? ("Expect one warning for each...")

RebeccaAnne (approved):
:shipit:

source/dotnet/Samples/WPFVisualizer/MainWindow.xaml.cs
almedina-ms: I understand that this can be a bother, we could provide an option to render the card automatically in the future

almedina-ms: I also believe that if we have this button and don't move the focus into the rendered card, the Accessibility team will file a bug on us, could you make that change?

almedina-ms (changes requested):
Not giving the focus to the rendered card when clicking the render button will probably make the Accessibility team file a bug on us

source/dotnet/Samples/WPFVisualizer/MainWindow.xaml.cs
jwoo-msft: i will back out this particular change, and make the change with the accessiblity change in the future PR. thanks

---
In reply to: [496179405](https://github.com/microsoft/AdaptiveCards/pull/4800#discussion_r496179405) [](ancestors = 496179405)

PR #4803, opened 2020-09-23
[JavaScript][Accessibility] Inline action img explicit role=presentation

golddove:
## Related Issue
ADO 24043835

## Description
Move inlineAction's title to img alt instead of button title

## How Verified
Manual verification

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4803)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2759,25 +2760,17 @@ export class TextInput extends Input {
                     button.classList.remove("iconOnly");
                     button.classList.add("textOnly");
 
-                    if (this.inlineAction) {
-                        button.textContent = this.inlineAction.title ? this.inlineAction.title : "Title";
-                    }
-                    else {
-                        button.textContent = "Title";
-                    }
+                    button.textContent = this.inlineAction?.title ?? "Title";
```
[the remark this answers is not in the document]
dclaux: Besides changing the syntax, what does this do that wasn't done before?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2759,25 +2760,17 @@ export class TextInput extends Input {
...
                 }
 
                 icon.src = this.inlineAction.iconUrl;
+                icon.alt = this.inlineAction.title ?? "Title";
```
dclaux, line 2766: This doesn't seem right to me:
1. Using the hard-coded "Title" string when `this.inlineAction.title` isn't set is wrong, we should change this into a resource string, which I apparently missed when I did the work recently. Could you do it? 
2. If the button does have a title, that title will be displayed, whether or not the button also has an icon. When both the icon and title are displayed, there is no point in having the alt text of the icon being the same as the button's title. Instead, the whole button + icon combination should be treated as one entity with the icon's role being "presentation"

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2759,25 +2760,17 @@ export class TextInput extends Input {
...
 
                 button.appendChild(icon);
-
-                if (this.inlineAction.title) {
-                    button.title = this.inlineAction.title;
-                }
             }
             else {
                 button.classList.add("textOnly");
-                button.textContent = this.inlineAction.title ? this.inlineAction.title : "Title";
```
dclaux, line 2780: Let's please not use the ? and ?? operators. Support for those was introduced somewhat recently in TS, and not every host supports them. If we are going to use them, we at least need to check with Teams that it's not a problem for them.

dclaux (changes requested):

source/nodejs/adaptivecards/src/card-elements.ts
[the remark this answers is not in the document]
golddove: Reverted.

source/nodejs/adaptivecards/src/strings.ts
```diff
@@ -34,4 +34,7 @@ export class Strings {
     static readonly hints = {
         dontUseWeightedAndStrecthedColumnsInSameSet: () => "It is not recommended to use weighted and stretched columns in the same ColumnSet, because in such a situation stretched columns will always get the minimum amount of space."
     }
+    static readonly fallback = {
```
dclaux, line 37: I think using the terms "defaults" for the static member and "inlineActionTitle" for the actual resource function might be better.

dclaux (changes requested):

source/nodejs/adaptivecards/src/card-elements.ts
golddove: Done - switched to resource string.

The icon already has presentation role implicitly (all children of button have this be default), according to W3C.

I removed the button's title and instead set the icon alt. So, the button's accessible name will be implicitly calculated based on the children (in this case, either the textContent or the icon's alt, whichever is present). This is recommended whenever possible, instead of explicitly setting the parent's title.

source/nodejs/adaptivecards/src/card-elements.ts
golddove: Done.

source/nodejs/adaptivecards/src/strings.ts
golddove: Renamed

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Let's take a step back here and make sure we're on the same page. Here is how things are supposed to work:
- When neither `inlineAction.iconUrl` nor `inlineAction.title` are defined, the inline action button should be displayed as a default string (i.e. "Title")
- When only `inlineAction.title` is defined, the inline action button should be displayed as text only, with the text being the value of the `title` property. No tooltip is necessary in this case.
- When only `inlineAction.iconUrl` is defined, the inline action button should be displayed as an image only. That is the case when we want to have the image's `altText` property set. But setting it to a default "Title" value will not be very helpful; for this change to make sense, we should use something else, such as "Inline action icon" (and even that isn't very helpful). On top of that, we do want the button to show a tooltip (and thus we need its `title` property set), and that tooltip should probably say "Click to execute the action associated with this input." (again, not very useful, but we have to have it, adding the tooltip is a change we made just recently for Teams)
- When both `inlineAction.title` and `inlineAction.title` are specified, the inline action button should be displayed as an image only, the button itself should have a tooltip set to the value of the `title` property, and probably should have `aria-label` set to the value of `title` as well. The image should be invisible to accessibility and its `altText` is therefore not relevant

So it seems to me we need 3 distinct resource strings to address this properly.

shalinijoshi19: For my understanding, how does this play in when we have tooltips defined as well? How do those fit in again? And are we making these same changes on other platforms ?

source/nodejs/adaptivecards/src/strings.ts
```diff
@@ -34,4 +34,7 @@ export class Strings {
     static readonly hints = {
         dontUseWeightedAndStrecthedColumnsInSameSet: () => "It is not recommended to use weighted and stretched columns in the same ColumnSet, because in such a situation stretched columns will always get the minimum amount of space."
```
shalinijoshi19, line 35: Interesting - this string -- where are we using these hints again? (_am assuming these get localized by some magic?_)

source/nodejs/adaptivecards/src/card-elements.ts
golddove: I completely forget that `title` changes the tooltip behavior - that was not the intent of this PR.

Technically, the behavior is totally correct without this PR - the presentation role is applied implicitly for children of buttons. But, the accessibility insights tool they're using doesn't seem to pick up on that (check out their attachment on the ADO bug). So, I think explicitly setting `role=presentation` should be fine, even though it is redundant.

As for the fallback text, I personally think a single term for all those cases is sufficient. Instead of "Title", I've gone ahead and put "Inline Action". So, narrator will say "Inline Action button". I don't think the default title is super important, it is not going to be descriptive no matter what default we choose.

Updated.

source/nodejs/adaptivecards/src/strings.ts
golddove: They don't currently get localized, but this is just set up to allow that. The hints are used in the designer code editor, I believe.

dclaux (approved):

source/nodejs/adaptivecards/src/strings.ts
dclaux: @shalinijoshi19 the point of these resource variables is to allow consuming applications to localize themselves. We do not ship localized SDKs.

shalinijoshi19 (approved):

source/nodejs/adaptivecards/src/card-elements.ts
shalinijoshi19: We really should be documenting this

Issue #4136, opened 2020-06-08
[iOS-Android-UWP][Parity Issue - Fact Set] [Card with empty factSet does not render on iOS but renders on other platforms]


adumrewal:
# Platform

What platform is your issue or question related to? (Delete other platforms).

- [x] Android
- [x] iOS
- [x] UWP
- [x] WPF

# Author or host

Host - Teams iOS

# Version of SDK

Latest master

# Details

Card with empty factSet does not render on iOS but renders on Android and Desktop.
` { "type": "FactSet", "facts": [ ] } `

[EmptyFactSetIssue.txt](https://github.com/microsoft/AdaptiveCards/files/4744964/EmptyFactSetIssue.txt)

shalinijoshi19:
@jwoo-msft can you take a look at this one to drive what the behavior should be here ? @dclaux / @matthidinger thoughts?

Issue #3949, opened 2020-04-21
[Android][Inconsistency between desktop and android] [Cards containing facts with empty title do not render]


shbhadra:
# Platform

- [x] Android
- [x] iOS
- [x] JS
- [x] WPF
- [x] UWP
# Author or host
Host
# Version of SDK
1.2.6
# Details
Consider the following card json, which contains certain facts with empty and null titles.
`{
        "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
        "type": "AdaptiveCard",
        "version": "1.0",
        "body": [
            {
                "type": "FactSet",
                "facts": [
                    {
                        "title": "",
                        "value": "Value 1"
                    },
                    {
                        "value": "Value 2"
                    },
                    {
                        "title": "Fact 3",
                        "value": "Value 3"
                    },
                    {
                        "title": "Fact 4",
                        "value": "Value 5"
                    }
                ]
            }
        ]
    }`

The same card as rendered on the web visualizer
![image](https://user-images.githubusercontent.com/51158485/79864396-07d85400-83f7-11ea-9747-06ad1c1c4673.png)

On the android visualizer however it doesn't render and the following error message is shown
"Property is required but was found empty: title"

shalinijoshi19:
Thanks for reporting this @shbhadra. @almedina-ms can you take a look please? @paulcam206 / @jwoo-msft FYI (Inconsistencies)

shalinijoshi19:
@jwoo-msft what's the plan for this one? ARe we able to land this in 20.09 still? I havent seen this move forward here?

PR #4809, opened 2020-09-24
Parsing & Rendering Update for FactSet

jwoo-msft:
## Related Issue
reference #3949 
reference #4136

## Description
The parsers will produce card when they encounter invalid FactSet such as missing Facts, previously parsers with shared models did not produce a card.

Renderers will make the decision of rendering the FactSet.
FactSet without Facts and Facts with nither of title or value are invalids, and the element will be dropped at the render time. 
## How Verified

1. New Test Card is added.
2. Existing relevant unit/regression tests have been ran.

ios
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4809)


samples/v1.0/Tests/FactSet.Test.json
```diff
@@ -0,0 +1,74 @@
...
+			"spacing": "None",
+			"wrap": true
+		},
+		{
+			"type": "FactSet",
+			"facts": []
+		},
+		{
+			"type": "TextBlock",
+			"text": "FactSet with facts wihout Title, without Value, wihout both, and a random property is below",
```
RebeccaAnne, line 48: >wihout [](start = 31, length = 6)

typo (without)

RebeccaAnne, line 48: >wihout [](start = 60, length = 6)

here too

samples/v1.0/Tests/FactSet.Test.json
```diff
@@ -0,0 +1,74 @@
+{
```
RebeccaAnne, line 1: Can we update the name of this file to describe what is being tested?

source/shared/cpp/ObjectModel/FactSet.cpp
```diff
@@ -45,7 +45,14 @@ std::shared_ptr<BaseCardElement> FactSetParser::Deserialize(ParseContext& contex
...
     // Parse Facts
     auto facts =
-        ParseUtil::GetElementCollectionOfSingleType<Fact>(context, value, AdaptiveCardSchemaKey::Facts, Fact::Deserialize, true);
+        ParseUtil::GetElementCollectionOfSingleType<Fact>(context, value, AdaptiveCardSchemaKey::Facts, Fact::Deserialize, false);
+
+    if (facts.empty())
+    { 
+        context.warnings.emplace_back(std::make_shared<AdaptiveCardParseWarning>(WarningStatusCode::RequiredPropertyMissing,
+                                                                                  "required property, \"fact\", is missing"));
+    }
```
RebeccaAnne, line 54: Similar to my comment in #4765, i think we want to change the behavior of the ParseUtil function rather than change each individual caller. 

That said, the change in Fact.cpp would be an exception as in that case we do need specific additional logic to ensure that at least one of the two is set.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/FactSetRenderer.java
```diff
@@ -82,8 +82,16 @@ public View render(
 
         FactVector factVector = factSet.GetFacts();
         long factVectorSize = factVector.size();
+
+        if (factVectorSize == 0)
```
almedina-ms, line 86: This seems like an extra validation that doesn't yield a lot of optimization, we could do it without this validation and just deal with the one below

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/FactSetRenderer.java
```diff
@@ -100,15 +108,19 @@ public View render(
...
+            {
+                factRow.addView(createTextView(context, RendererUtil.handleSpecialText(titleWithFormattedDates), hostConfig.GetFactSet().getTitle(), hostConfig, spacing, renderArgs.getContainerStyle()));
+                factRow.addView(createTextView(context, RendererUtil.handleSpecialText(valueWithFormattedDates), hostConfig.GetFactSet().getValue(), hostConfig, 0, renderArgs.getContainerStyle()));
+                tableLayout.addView(factRow);
+                validFacts++;
+           }
+        }
 
-            tableLayout.addView(factRow);
+        if (validFacts == 0) {
```
almedina-ms, line 122: nit: curly braces go in new line

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/FactSetRenderer.java
```diff
@@ -100,15 +108,19 @@ public View render(
...
             DateTimeParser parser = new DateTimeParser(fact.GetLanguage());
 
-            // Handle Title
             String titleWithFormattedDates = parser.GenerateString(fact.GetTitleForDateParsing());
-            factRow.addView(createTextView(context, RendererUtil.handleSpecialText(titleWithFormattedDates), hostConfig.GetFactSet().getTitle(), hostConfig, spacing, renderArgs.getContainerStyle()));
-
-            // Handle Value
             String valueWithFormattedDates = parser.GenerateString(fact.GetValueForDateParsing());
-            factRow.addView(createTextView(context, RendererUtil.handleSpecialText(valueWithFormattedDates), hostConfig.GetFactSet().getValue(), hostConfig, 0, renderArgs.getContainerStyle()));
+            if (titleWithFormattedDates.length() != 0 || valueWithFormattedDates.length() != 0)
```
almedina-ms, line 113: you can use the method titleWithFormattedDates.isEmpty() instead of length comparisson

samples/v1.0/Tests/FactSet.Test.json
jwoo-msft: sure, I will add a new card called CardsWithInvalidProperties.json that combines  TextBlock.Test.json and FactSet.Test.json, and remove the old cards. I will add the new card in a new PR.

---
In reply to: [494599248](https://github.com/microsoft/AdaptiveCards/pull/4809#discussion_r494599248) [](ancestors = 494599248)

samples/v1.0/Tests/FactSet.Test.json
jwoo-msft: thanks

---
In reply to: [494598392](https://github.com/microsoft/AdaptiveCards/pull/4809#discussion_r494598392) [](ancestors = 494598392)

source/shared/cpp/ObjectModel/FactSet.cpp
jwoo-msft: About the exception, it will be still parse warning, and renderer will drop the rendering of this FactSet.

---
In reply to: [494601157](https://github.com/microsoft/AdaptiveCards/pull/4809#discussion_r494601157) [](ancestors = 494601157)

jwoo-msft: I created a new issue that tracks array parsing similary to the PR4765(TextBlock). [https://github.com/microsoft/AdaptiveCards/issues/4837](https://github.com/microsoft/AdaptiveCards/issues/4837) 

---
In reply to: [496175681](https://github.com/microsoft/AdaptiveCards/pull/4809#discussion_r496175681) [](ancestors = 496175681,494601157)

RebeccaAnne (approved):
:shipit:

Issue #4715, opened 2020-09-01
ActionSet flyout (... actions menu)


productboard-ac:

productboard-ac:
Link to feature: https://adaptivecards.productboard.com/feature-board/planning/features/4374375

golddove:
A couple alternate proposals to think about.

## maxActions

`maxActions` tells how many actions to show before overflow. If there is insufficient space, fewer than maxActions will be shown.

So, if you want all Actions in an overflow:
```
{
	"type": "ActionSet",
	"actions": [
		{
			"type": "Action.Submit",
			"title": "View"
		},
		{
			"type": "Action.Submit",
			"title": "Edit"
		},
		{
			"type": "Action.Submit",
			"title": "Delete"
		}
	],
	"maxActions": 0
}
```
Or if you want to show the first two:
```
{
   "type": "AdaptiveCard",
   "actions": [
        {
            "type": "Action.Submit",
            "title": "Set due date"
        },
        {
            "type": "Action.OpenUrl",
            "title": "View",
            "url": "https://adaptivecards.io"
        },
        {
            "type": "Action.Submit",
            "title": "Delete"
        }
    ],
    "maxActions": 2
}
```

Alternatively, we can constrain it the other way around (call it "overflowCount" or something), where the author tells us how many actions to put in the overflow. But I would lean towards the maxActions.


## ActionSet in place of Action

Instead of an Action, the card author can put in an ActionSet. We render the set as an overflow button that shows the contained actions.

e.g.
```
{
   "type": "AdaptiveCard",
   "actions": [
        {
            "type": "Action.Submit",
            "title": "Set due date"
        },
        {
            "type": "Action.OpenUrl",
            "title": "View",
            "url": "https://adaptivecards.io"
        },
        {
            "type": "ActionSet",
            "actions": [
                {
                    "type": "Action.Submit",
                    "title": "Set time"
                },
                {
                    "type": "Action.Submit",
                    "title": "Set priority"
                },
                {
                    "type": "Action.Submit",
                    "title": "Delete"
                }
            ]
        }
    ]
}
```

If a renderer chooses to implement more advanced constructs, this option opens up the potential for:
- multiple overflows
- overflow positioning (i.e. the author may reorder the overflow before the other actions)
- nested overflows

Some real-world examples:

Multiple overflows in one action bar:
[...]

PR #4824, opened 2020-09-25
[Sample][Accessibility] Shorten action text in FlightDetails

golddove:
## Related Issue
ADO 24110322

## Description
Instead of the hacky two ActionSet fix, this is a text change to avoid overflow in small cards. The ideal underlying fix is #4715 

## How Verified
Manual verification

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4824)

### DIFF
--- a/samples/Templates/Scenarios/FlightDetails.template.json
+++ b/samples/Templates/Scenarios/FlightDetails.template.json
@@ -190,16 +190,10 @@
 											"type": "Action.Submit",
 											"title": "Check in",
 											"style": "positive"
-										}
-									],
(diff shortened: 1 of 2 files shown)


dclaux (approved):

almedina-ms (approved):

almedina-ms:
Approving even when I think removing "in calendar" removes some context for the user, we should keep in mind this when #4715 is released

PR #4828, opened 2020-09-26
[Designer][Accessibility] Click to add elements

golddove:
## Related Issue
ADO 24096354

## Description
Click (or press space/enter) to append an element to bottom of card. This is an intermediate fix, as it does not allow full keyboard replacement for drag/drop (namely, there is no support for reordering elements).

Also, switched keycode constants to use `KeyboardEvent.key` constants, as the `keyCode` is [long deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode).

## How Verified
Manual verification

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4828)


source/nodejs/adaptivecards-controls/src/constants.ts
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-export const KEY_TAB = 9;
-export const KEY_ENTER = 13;
-export const KEY_ESCAPE = 27;
-export const KEY_SPACE = 32;
-export const KEY_UP = 38;
-export const KEY_DOWN = 40;
-export const KEY_DELETE = 46;
+export const KEY_TAB = "Tab";
```
dclaux, line 3: These are exported constants. Any code that currently uses them will be broken. We have to decide whether that's OK or not.

On one hand, it is unlikely that this will affect a lot of applications; on the other hand, as a general rule we just don't introduce breaking changes unless they're absolutely necessary.

Given they are not absolutely necessary here, I think we should keep the old constants as is and introduce the new ones. We can use the `<name>Key` naming convention: `export const TabKey = "Tab";`

dclaux (changes requested):
Looks good overall, but I think I have a problem with doing this on "click". Click is a mouse event; if someone can use the mouse, they are likely to be able to do drag/drop. If they really want to use a shortcut to adding an element at the end of the card, then I think that should be on double click for two reasons:
- With single click, I think there is too much of a potential for a user that was trying to initiate a drag drop operation to end up generating a single click instead
- Many (most? all?) the designer experiences that I've used (WinForms, XAML, Delphi, PowerBuilder, Visual Basic) do this on double click

Beyond the above, I think the newly created peer should be automatically selected.

source/nodejs/adaptivecards-controls/src/constants.ts
```diff
@@ -1,9 +1,42 @@
...
+ */
 export const KEY_UP = "Up";
+/**
+ * @deprecated Use Constants.keys.down instead.
+ */
 export const KEY_DOWN = "Down";
+/**
+ * @deprecated Use Constants.keys.delete instead.
+ */
 export const KEY_DELETE = "Delete";
```
dclaux, line 30: The deprecated constants need to be static members of the Constants class, otherwise we're breaking code that refers to them as `Constants.KEY_UP` for instance.

source/nodejs/adaptivecards/src/strings.ts
```diff
@@ -30,11 +30,11 @@ export class Strings {
         elementAlreadyParented: () => "The element already belongs to another container.",
         actionAlreadyParented: () => "The action already belongs to another element.",
         elementTypeNotStandalone: (typeName: string) => `Elements of type ${typeName} cannot be used as standalone elements.`
-    };
+    } as const;
     static readonly hints = {
         dontUseWeightedAndStrecthedColumnsInSameSet: () => "It is not recommended to use weighted and stretched columns in the same ColumnSet, because in such a situation stretched columns will always get the minimum amount of space."
-    }
+    } as const;
```
dclaux, line 36: Not familiar with this - what does it do?

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -201,7 +201,9 @@ export class CardDesigner extends Designer.DesignContext {
 
                     paletteItem.onClick = (sender) => {
                         const peer = paletteItem.createPeer(this, this.designerSurface);
-                        this.designerSurface.rootPeer.tryAdd(peer);
+                        if(this.designerSurface.rootPeer.tryAdd(peer)) {
+                            this.designerSurface.setSelectedPeer(peer);
```
dclaux, line 205: `peer.isSelected = true;` works as well.

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -201,7 +201,9 @@ export class CardDesigner extends Designer.DesignContext {
 
                     paletteItem.onClick = (sender) => {
                         const peer = paletteItem.createPeer(this, this.designerSurface);
-                        this.designerSurface.rootPeer.tryAdd(peer);
+                        if(this.designerSurface.rootPeer.tryAdd(peer)) {
```
dclaux, line 204: Missing space between `if` and `(`

dclaux, line 204: I tend to put a blank line before an `if` statement unless if is the first statement after an opening `{`

source/nodejs/adaptivecards-designer/src/draggable-element.ts
```diff
@@ -125,7 +125,11 @@ export abstract class DraggableElement {
         dragSourceElement.onkeydown = (e: KeyboardEvent) => {
             if(e.key === Constants.keys.enter || e.key === Constants.keys.space) {
                 e.preventDefault();
-                this.click(null);
+                if (this.onClick) {
```
dclaux, line 128: Blank line before `if`

source/nodejs/adaptivecards-designer/src/draggable-element.ts
```diff
@@ -125,7 +125,11 @@ export abstract class DraggableElement {
...
+                    this.click(null);
+                } else if (this.onDoubleClick) {
```
dclaux, line 130: `else if` on its own line

dclaux (changes requested):

source/nodejs/adaptivecards/src/strings.ts
golddove: Applies `readonly` compile-time type to the properties in object literals (to prevent accidental object mutation).

In this case, it is equivalent to:

```
static readonly hints : {
    readonly dontUseWeightedAndStrecthedColumnsInSameSet : () => String
} = {
    dontUseWeightedAndStrecthedColumnsInSameSet: () => ...
}
```

which would be very unwieldy syntax.

source/nodejs/adaptivecards-controls/src/constants.ts
golddove: These are (and previously were) exported as top-level constants in the '/constants.ts' module, which were then exported as top-level constants in the 'adaptivecards-controls' package. So, an external library had to have used them like this:

```
import * as Controls from "adaptivecards-controls"

Controls.KEY_ENTER
```

I did not modify that. The new Constants class in the '/constants.ts' module has to be used like this:

```
import * as Controls from "adaptivecards-controls"

Controls.Constants.key.enter
```

If I move the deprecated constants into that Constants class, the old usage would be broken.

dclaux: Ignore this - my mistake, what you have works.

dclaux (approved):

PR #4850, opened 2020-09-29
[Designer] Trap keyboard for modal dialogs

paulcam206:
## Description
@golddove and I noticed a couple of issues with modal dialogs in the Designer:
* Keyboard focus wasn't returning to point of invocation when the dialog was dismissed
* Keyboard focus wasn't restricted to the dialog as it should be
* Modal dialog difficult to distinguish visually

This fixes that. Here's what modals look like now:
image

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4850)

### DIFF
--- a/source/nodejs/adaptivecards-designer/src/dialog.ts
+++ b/source/nodejs/adaptivecards-designer/src/dialog.ts
@@ -55,7 +55,6 @@ export abstract class Dialog {
 
             this._overlayElement = document.createElement("div");
             this._overlayElement.className = "acd-dialog-overlay";
-            this._overlayElement.style.backgroundColor = "rgba(0, 0, 0, 0.4)";
             this._overlayElement.onclick = (e) => {
                 // clicks on the overlay window should dismiss the dialog
                 this.close();
--- a/source/nodejs/adaptivecards-designer/src/dialog.ts
+++ b/source/nodejs/adaptivecards-designer/src/dialog.ts
@@ -28,6 +28,9 @@ export class DialogButton {
 export abstract class Dialog {
     private _overlayElement: HTMLElement;
     private _isOpen: boolean = false;
+    private _originalFocusedElement: HTMLElement;
+    private _firstFocusableElement: HTMLElement;
+    private _lastFocusableElement: HTMLElement;
 
     protected abstract renderContent(): HTMLElement;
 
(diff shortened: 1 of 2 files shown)


RebeccaAnne (approved):
:shipit:

source/nodejs/adaptivecards-designer/src/dialog.ts
```diff
@@ -121,7 +139,12 @@ export abstract class Dialog {
             this._overlayElement.appendChild(dialogFrameElement);
 
             document.body.appendChild(this._overlayElement);
-            dialogFrameElement.focus();
+
+            let focusableElements = dialogFrameElement.querySelectorAll('a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), [tabindex="0"]');
```
dclaux, line 143: Probably a corner case, but what about elements that have tabindex > 0?

dclaux (changes requested):

source/nodejs/adaptivecards-designer/src/dialog.ts
paulcam206: `>0` tabindex is broadly considered a [bad](https://karlgroves.com/2018/11/13/why-using-tabindex-values-greater-than-0-is-bad) [practice](https://web.dev/tabindex/) for [accessibility](https://dequeuniversity.com/rules/axe/4.0/tabindex) [reasons](https://adrianroselli.com/2014/11/dont-use-tabindex-greater-than-0.html)... I think we'd really want to discourage its use.

source/nodejs/adaptivecards-designer/src/dialog.ts
```diff
@@ -48,19 +51,21 @@ export abstract class Dialog {
 
     open() {
         if (!this._isOpen) {
+            this._originalFocusedElement = document.activeElement as HTMLElement;
+
             this._overlayElement = document.createElement("div");
             this._overlayElement.className = "acd-dialog-overlay";
+            this._overlayElement.style.backgroundColor = "rgba(0, 0, 0, 0.4)";
```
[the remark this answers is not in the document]
paulcam206: should be good now

dclaux (approved):

PR #4859, opened 2020-09-30
[JS] Fix arrow keys and actionset aria

paulcam206:
## Related Issue
Fixes VSO 23940376

## Description
Two, two, two fixes in one! #4828 regressed arrow keys in key handlers since the keys are referred to as `Arrow[Up|Down]`, not `[Up|Down]`. This broke keyboard accessibility in the Designer's toolbar pickers, since you couldn't arrow through the list of options. We also fix up the aria around `ActionSet`s as well as the dropdown list. Narrator now correctly reads both the expanded/collapsed state of `Action.ShowCard` buttons as well as their positions in the list of actions.

## How Verified
* local build, narrator, dev tools, NVDA

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4859)

### DIFF
--- a/source/nodejs/adaptivecards-controls/src/constants.ts
+++ b/source/nodejs/adaptivecards-controls/src/constants.ts
@@ -35,8 +35,8 @@ export class Constants {
         enter: "Enter",
         escape: "Escape",
         space: " ",
-        up: "Up",
-        down: "Down",
+        up: "ArrowUp",
+        down: "ArrowDown",
         delete: "Delete"
     } as const;
 }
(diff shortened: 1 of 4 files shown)


RebeccaAnne (approved):
:shipit:

golddove (approved):

dclaux (approved):

PR #4871, opened 2020-10-01
Fixed Memory Leak with Custom Parsing in iOS

jwoo-msft:
## Related Issue
N/A

## Description
Found a memory leak after running a profiler. JSON stream writer returned a raw memory and should have been freed after the usage. The impact is small since custom parsing is a rarely used scenario in iOS.

## How Verified
1. Memory profilers were run multiple times to ensure the fix is good.
2. Ran existing scenario cards.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4871)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOBaseActionElement.mm
```diff
@@ -54,7 +54,9 @@ - (NSData *)additionalProperty
         std::stringstream sstream;
         writer->write(blob, &sstream);
         NSString *jsonString =
-        [[NSString alloc] initWithCString:sstream.str().c_str() encoding:NSUTF8StringEncoding];
+            [[NSString alloc] initWithCString:sstream.str().c_str()
+                                     encoding:NSUTF8StringEncoding];
+        delete writer;
```
RebeccaAnne, line 59: >delete writer; [](start = 8, length = 14)

All the other changes delete the writer right after the write call. Can we do that here as well for consistency?

RebeccaAnne (approved):
:shipit:

Issue #4761, opened 2020-09-09
Adaptive Components


productboard-ac:

productboard-ac:
Link to feature: https://adaptivecards.productboard.com/feature-board/planning/features/5478079

matthidinger:
## Discussion 9/25/2020
 
1. Component lookup conventions
   1. Should we encapsulate the component name  in the `type` property or a separate fully-qualified a `name` property?
      1. **DECISION:** type will be `Component` only, use the `name` property as a fully-qualify the identifier of a component. See examples below for more.
   2. For scoping should we use (1) "namespaces" (like odata); (2) "url" (like schema.org); or (3) a bit of both
      1. **DECISION**: The `name` property will be a `string` that may optionally start with a domain and/or path qualifier. See examples below for more.
   3. Should we provide a mechanism for a default namespace at the card level? This is really only useful to reduce redundancy in cards with many components. 
      1. **DECISION**: Not right now, can always revisit later.
   4. Should we provide automatic lookup based on a data type or data shape? 
      1. **DECISION**: leave aside for now but seems like a natural evolution later
2. Where should the component data/properties go? As "addtional properties" directly on the `Component` object or in a specific attribute called `properties`
   1. **DECISION**: Components will have multiple first-class properties to make this concept work, such as `name`, `view` preference, potentially `version`, and other pieces of metadata that are required to make the Component model work. To remove risk of conflict we will put all domain-specific properties for a given component into a `properties` attribute, which will get passed along to the component to be used for binding. See examples below for more.
3. Versioning consideration
   1. **DECISION**: We will likely add a version segment to the `name`, similar to many REST APIs. E.g., `"name": "graph.microsoft.com/file/v2"`
4. What should our guidance be for vastly different UX for given data type? E.g., there are a bunch of different ways to visualize a "flight". E.g., maybe I want the "Boarding Pass" component, or maybe I want to "Flight Status" component, or even a "Check in Reminder". Should we suggest those to be separate components or different views in the same component?
   1. **DECISION**: Based on what we know now we will recommend making those separate components.
5. What about arrays of data? E.g., should there be a component for a single `User` and another component to handle multiple `Users`? 
   1. **DECISION**: As of now we are going to recommend different components for a single object vs an array of obecjts.

## Next steps

- [ ] Deploy an update to the template service that includes a component registry
- [ ] Designer updates to remotely search for components by querying the registry
- [ ] Designer update to automatically populate the property sheet from the component schema

[...]

PR #4894, opened 2020-10-08
Experimental component support

dclaux:
## Related Issue
Adaptive Components: https://github.com/microsoft/AdaptiveCards/issues/4761

## Description
This PR introduces "Adaptive Components", a new type of Adaptive Card element whose visual layout and schema is encapsulated in a single JSON file and downloaded on the fly.

No code review is required at this time.

Known issues
- This is experimental - don't expect things to work flawlessly
- Updating component properties in the JSON editor doesn't work
- If there are multiple instances of the same component on the card, updating one instance's properties in the Property Sheet updates all instances' properties
- When switching the host app or picking a different version in the version picker, the "Component" entry disappears from the toolbox
- The service backing the component registry and components themselves in slow
- It is not always clear when a component is still loading
- If using more than one instance of a component with inputs in the same card, input Ids will collide and Action.Submit will only get the input values of the last instance
- Component schemas are currently limited to a flat list of properties of type string or number only, and no constraints can be defined in terms of maximums/minimums etc.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4894)


jonmill:
Closing due to age - if this is still needed, please re-open

Issue #4830, opened 2020-09-28
[Android][Input validation error] [Input validation error UX]


dipja:
# Platform

What platform is your issue or question related to?

* Android

# Author or host

Microsoft Teams

# Version of SDK

Adaptive card SDK 2.1.0 (Adaptive card v1.3)

# Details

When there is an input validation error on the card, the SDK adds an error message to the card and changes the background color of the EditText.

The SDK gets the drawable of the EditText and applies a color filter to it assuming that the EditText would use the default drawable (`ValidatedEditText.class`). But in case the host uses a custom drawable with its EditText, the UX could be completely different. There is no place where the host can override this behavior of getting the drawable and applying the color filter to it.

Also, the color being used to highlight the EditText comes from ContainerStyle.Default -> ForegroundColor.Attention. There is no new host config to pick this color. (`TextInputRenderer.class`)

The difference between visualizer app which uses default EditText drawable and Teams android client which uses a custom drawable.

![image](https://user-images.githubusercontent.com/61967816/94392079-e84b1180-0174-11eb-932c-91e8692832a2.png)

![image](https://user-images.githubusercontent.com/61967816/94392246-5ee80f00-0175-11eb-9199-fdddf2aa5c9f.png)

dipja:
@shalinijoshi19 @dclaux We are facing this issue as we are upgrading to the latest SDK in Teams. Can you please triage and let us know how we can work around this.
FYI: @sowrabh-msft @VikrantSingh01

shalinijoshi19:
@almedina-ms any updates on this one? Thanks

PR #4895, opened 2020-10-08
[Android] Update input error visual cue rendering to support custom background drawables 

almedina-ms:
## Related Issue
Related to #4830 

## Description
Adds options to modify the rendering of the error visual cues for validated inputs. The original version of this feature only considered the default drawable which has a transparent background, due to this when tinting the background it could have unintended side effect such as tinting the whole background, such as in the example below

image

For this fix we added two new attributes called `adaptiveUsingCustomTextInputs` and `adaptiveUsingCustomChoiceSetInputs`. 
* `adaptiveUsingCustomTextInputs`: boolean flag that signalize the renderer to set the `adaptive_state_input_invalid` with a value of true instead of tinting the text input background.
* `adaptiveUsingCustomChoiceSetInputs`: boolean flag to control some internal rendering for compact choice sets

On the other hand we added a new drawable state called `adaptive_state_input_invalid` which allows customers to define background drawables which define how an invalid input would look like

Finally, we also added two new drawable selector files `adaptive_choiceset_compact_background.xml` and `adaptive_choiceset_expanded_background.xml`:
* `compact` drawable: controls the background for compact choicesets, if the `adaptiveUsingCustomTextInputs` flag is not set or set to false, the drawable is applied to a linear layout that encloses the spinner and is tinted in `attention` color.
* `expanded` drawable:  controls the background for expanded choicesets (mono and multi select) and the background for Toggle inputs. In all cases this background is tinted with the `attention` color to mimic the current JS and iOS behaviours.

## How Verified
The fix was verified by adding the previously mentioned properties into our app theme (which will be modified for each of the examples below)
```
<item name="adaptiveUsingCustomTextInputs">true</item>
<item name="adaptiveUsingCustomChoiceSetInputs">true</item>
<item name="android:editTextBackground">@drawable/layered_edit_text_styles</item>
```

For reference, layered_edit_text_styles is a selector which only defines a drawable for invalid state and a default drawable
``` 
<selector xmlns:android="http://schemas.android.com/apk/res/android" xmlns:adaptivecards="http://schemas.android.com/apk/res-auto" >
    <item adaptivecards:adaptive_state_input_invalid="true" android:drawable="@drawable/layered_edit_text_invalid" />
    <item android:drawable="@drawable/layered_edit_text" />
</selector>
```

Example 1: No custom inputs
```
<item name="usingCustomInputs">false</item>
<!-- <item name="android:editTextBackground">@drawable/layered_edit_text_styles</item> -->
```

image

Example 2: Having a custom drawable, but stating that is not using custom inputs
```
<item name="usingCustomInputs">false</item>
<item name="android:editTextBackground">@drawable/layered_edit_text_styles</item>
```

image
> NOTE: the image is tinted, as it has a semi transparent background color it tints it in a lighter shade of Attention color

Example 3: Having a custom drawable
```
<item name="usingCustomInputs">true</item>
<item name="android:editTextBackground">@drawable/layered_edit_text_styles</item>
```

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4895)


source/android/adaptivecards/src/main/res/values/attrs.xml
```diff
@@ -7,4 +7,9 @@
     <attr name="adaptiveShowCardAction" format="reference" />
     <style name="adaptiveInlineAction" format="reference" />
     <style name="adaptiveInlineActionImage" format="reference" />
+
+    <attr name="usingCustomInputs" format="boolean" />
```
golddove, line 11: We should try to avoid requiring this flag if at all possible, not a great host developer experience.

source/android/adaptivecards/src/main/res/values/attrs.xml
```diff
@@ -7,4 +7,9 @@
...
+    <declare-styleable name="custom_input_state">
+        <attr name="state_input_invalid" format="boolean" />
```
golddove, line 13: How about "state_error" to keep it generic?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/ValidatedEditText.java
```diff
@@ -2,40 +2,102 @@
...
-            // Change border color to original color
-            getBackground().setColorFilter(null);
+            // If using custom inputs, set the selector state for state_input_invalid as true
+            setInputInvalid(!isValid);
         }
         else
         {
-            // Change border to red
-            getBackground().setColorFilter(m_errorColor, PorterDuff.Mode.SRC_ATOP);
+            // If not using custom inputs, tint the whole input with attention color
```
golddove, line 63: Instead of this conditional, would it be possible to always set the custom state, and include a default drawable in our sdk for EditText (just like the sample you created for the sample app here)?

That drawable would then be replaced by the host, if desired. This approach would allow us to remove the extra boolean flag.

golddove (changes requested):
This is so cool! The custom state you've implemented here seems like a great pattern for Android native styling going forward.

source/android/adaptivecards/src/main/res/values/attrs.xml
almedina-ms: yes, I agree requiring that flag may not be the best dev experience but there's no way to know if an external developer has set an application wide style or drawable, this issue becomes even more evident with the spinners which have to deal with greater complications

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/ValidatedEditText.java
almedina-ms: I would prefer to avoid having custom views for the controls as they may not match the remainder of the user application, if possible I would advice against modifying an application wide type of view just for our product

source/android/adaptivecards/src/main/res/values/attrs.xml
almedina-ms: I like the idea but I'm having a tough time thinking of other scenarios where we may use the error state

PR #4897, opened 2020-10-09
Added placeholder support

### LINKED ISSUE #4665: [iOS][Image Rendering][ResourceResolvers] Card elongation
# Platform

* iOS

# Version

v1.2.11

# Details

Cards seems to get long empty spaces when image is set via ResourceResolvers. This sporadically happens on Simulators, but can be frequently noticed on Real Devices in Release Config. (Our Release config is mostly similar to the default Release config)

[payloads.zip](https://github.com/microsoft/AdaptiveCards/files/5129975/payloads.zip)
[card-images.zip](https://github.com/microsoft/AdaptiveCards/files/5129977/card-images.zip)

IMG_3087
IMG_3088

### Steps to Reproduce
This happens in Release mode and on real devices
- Configure a tableView with cards in every cell. (Sample payloads provided)
- Populate images using **ResourceResolvers**

This is noticed on cards that has images contributing to the height of the card. Cards with fixed height don't seem to have this issue.

### Visual Debugger Screenshot
Screenshot 2020-08-26 at 3 07 54 PM

### LINKED ISSUE #4682: [iOS][Resource Resolver] Add Placeholder Image Support For Resource Resolver
**What platform is your feature request related to?**

- [ ] iOS
- [ ] Dev Spec

We don't currently support placeholder images while resource resolver is resolving images. Once we get the image through the resource resolver, we can not update the placeholder image with the resolved images. This is by design. However, being able to configure a placeholder image is a real use case. 

If the placeholder image is in the cache, we can draw the card quickly, and when the image is ready for replacement. We don't have to redraw the entire card, instead, it becomes a possibility that Adaptive Card re-draws only the image portion. Please note that the performance benefit will be lost if we do not know the size of the real image, therefore knowing the size of the real image at the time of acquiring a placeholder image is a requirement for this feature.

This feature will be implemented for iOS first since iOS re-draws the entire card in such events, and there are no such requests from the customers of other platforms. Dev Spec task is for review purposes, and for the future consistency consideration. Once the spec is drawn and agreed, we can refer back to the spec, once there are requests from customers for the other platforms.

**Related Issues**
 #4667, #4650, #4585

jwoo-msft:
## Related Issue
Fixed #4665 and Fixed #4682 

## Description
* Added support for placeholder image.
* Fixed card elongation issue.
* refactored code

## How Verified
1. Existing relevant unit is ran
2. Created cards with cashed images.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4897)


shalinijoshi19:
Out of curiosity, is this something that was missed on iOS but is available on the pther platforms?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRIBaseCardElementRenderer.h
```diff
@@ -25,4 +25,5 @@
 @protocol ACRIKVONotificationHandler
 
 - (void)configUpdateForUIImageView:(ACOBaseCardElement *)acoElem config:(ACOHostConfig *)acoConfig image:(UIImage *)image imageView:(UIImageView *)imageView;
+
```
shalinijoshi19, line 28: Nit; looks like we have a stray newline added here-- probably just revert changes to this file altogether.

shalinijoshi19:
Nit/typo: *cached

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRImageRenderer.mm
```diff
@@ -45,16 +45,33 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
     BOOL isAspectRatioNeeded = !(pixelWidth && pixelHeight);
     CGSize cgsize = [acoConfig getImageSize:imgElem->GetImageSize()];
 
-    NSString *key = [NSString stringWithCString:imgElem->GetUrl().c_str() encoding:[NSString defaultCStringEncoding]];
+    NSString *number = [[NSNumber numberWithUnsignedLongLong:(unsigned long long)(elem.get())] stringValue];
+    NSString *urlString = [NSString stringWithCString:imgElem->GetUrl().c_str() encoding:[NSString defaultCStringEncoding]];
```
shalinijoshi19, line 49: Ask: Could you try to comment the code a bit to explain what is going on here at all? Thanks!

jwoo-msft:
> Out of curiosity, is this something that was missed on iOS but is available on the pther platforms?

i don't believe we support placeholder images while the image is loading on other platforms. The usage here is unique to Cisco. They know exactly the size of the image they are using before rendering the card, and they don't want to dynamically size their card because then they have to refresh the layout on the super view of an adaptivecard view.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRImageRenderer.mm
jwoo-msft: sure

jwoo-msft: sure

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRMediaRenderer.mm
```diff
@@ -43,26 +43,37 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
...
-    CGFloat heightToWidthRatio = 0.0f;
     ACRContentHoldingUIView *contentholdingview = nil;
 
     // if poster is available, restrict the image size to the width of superview, and adjust the height accordingly
     if (img) {
-        view = [[UIImageView alloc] initWithImage:img];
-
-        if (img.size.width > 0) {
-            heightToWidthRatio = img.size.height / img.size.width;
+        contentholdingview = (ACRContentHoldingUIView *)[rootView getImageView:mediaKey];
```
shalinijoshi19, line 65: >(ACRContentHoldingUIView *)[ [](start = 29, length = 28)

Curious: Is there something like a safe casting in ObjectiveC? Do we have the right level of compiler flags enabled to catch potential issues?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRMediaRenderer.mm
```diff
@@ -43,26 +43,37 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
...
+        if (contentholdingview) {
+            view = contentholdingview.subviews[0];
```
shalinijoshi19, line 67: >ews[0]; [](start = 43, length = 7)

Is there ever a chance that the subviews array is empty so that accessing the first element will result in an error/av? Should we be checking for it upfront or adding an assert condition before trying to index into it?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRMediaRenderer.mm
```diff
@@ -81,16 +92,16 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
     view.contentMode = UIViewContentModeScaleAspectFill;
     contentholdingview.isMediaType = YES;
 
+
     // process play icon image
-    NSString *piikey = [NSString stringWithCString:[acoConfig getHostConfig] -> GetMedia().playButton.c_str() encoding:[NSString defaultCStringEncoding]];
-    UIImage *playIconImage = imageViewMap[piikey];
+    NSString *playIconKey = makeKeyForImage(acoConfig, @"media-playicon-image", pieces);
```
shalinijoshi19, line 97: >media-playicon-image [](start = 57, length = 20)

Avoid hard coded literals in the code - where is this coming from? Also is this going to have potential localization impact? SHould this be out in its own const strings class for easy localizability potentially?

shalinijoshi19:
Is this going to cause our renderers to be inconsistent at this point? 

---
In reply to: [713932386](https://github.com/microsoft/AdaptiveCards/pull/4897#issuecomment-713932386) [](ancestors = 713932386)

shalinijoshi19 (approved):
:shipit:

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRMediaRenderer.mm
jwoo-msft: The ACRContentHoldingUIView is a wrapper view to hold a subview such as UIImageView. The content view is created when there is an image view to add to the image view map. before the image view is added to the map, the view is added to the content view.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRMediaRenderer.mm
jwoo-msft: there will be no localization impact. the strings are only used internally.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRMediaRenderer.mm
jwoo-msft: Objective-C is a superset of c, but there are no other ways of casting.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRMediaRenderer.mm
shalinijoshi19: In general bad practice to hard code string literals - whether they are to be localized or not.

---
In reply to: [509830950](https://github.com/microsoft/AdaptiveCards/pull/4897#discussion_r509830950) [](ancestors = 509830950)

PR #4939, opened 2020-10-20
[Localization] Draft spec on areas for localization.

RebeccaAnne:
## Description
Draft spec covering areas for possible localization work.


specs/DesignDiscussions/Localization.md
```diff
@@ -0,0 +1,109 @@
+# Localization in Adaptive Cards
+
+Currently Adaptive Cards has minimal localization support. This document covers areas where localization may be necessary. Currently this document does not cover detailed solutions to these scenarios, though those solutions will be added here as work progresses.
+
+## Current Support - Date/Time strings
+Currently, there is only one scenario in which we support localization. Our TextBlocks and other string elements support displaying [Dates and Times](https://docs.microsoft.com/en-us/adaptive-cards/authoring-cards/text-features#datetime-formatting-and-localization). This allows the card author to specifiy a date in [RFC 3389](https://tools.ietf.org/html/rfc3339) format. We then display the time in the timezone of the host's device, and in the language specified by the card author. For example:
```
shalinijoshi19, line 6: Does it make sense to explicitly list out/confirm the "other string elements" referenced for completeness?

specs/DesignDiscussions/Localization.md
```diff
@@ -0,0 +1,109 @@
+# Localization in Adaptive Cards
+
+Currently Adaptive Cards has minimal localization support. This document covers areas where localization may be necessary. Currently this document does not cover detailed solutions to these scenarios, though those solutions will be added here as work progresses.
```
shalinijoshi19, line 3: Given the ever-expanding scope of "Adaptive Cards" do we need to clarify the scope of the discussion here? For eg it doesn't cover the broader AdaptiveCards ecosystem at this point (adapticecards.io website, designer, CMS, Templating SDKs etc) but focused on our renderer SDKs? Maybe like an explicit call out of hte "Goals/Non-Goals" and a clear articulation of  "Out of scope" aspects? Alternatively/Eventually we could have a parallel doc that might talk more holistically about localizability + localization of AC including our ecosystem and offerings beyond the renderers ?

RebeccaAnne: This document does reference all the aspects of the ecosystem you call out as you can see below. Admittedly there may be more detail on the renderer issues in this early draft, but I envision the "Documentation and Tooling" and "Components/Templating" sections below to be as relevant as the other sections, and as likely to be drilled into in the future as we prioritize this work. The goal of this early draft is breadth rather than depth, to call out all areas where localization may be needed. We can then decide if and where we want to expend resources to dig in further. 

---
In reply to: [509283708](https://github.com/microsoft/AdaptiveCards/pull/4939#discussion_r509283708) [](ancestors = 509283708)

specs/DesignDiscussions/Localization.md
RebeccaAnne: Sure.

---
In reply to: [509279373](https://github.com/microsoft/AdaptiveCards/pull/4939#discussion_r509279373) [](ancestors = 509279373)

jonmill:
@RebeccaAnne - is this PR still wanted or should it be closed?

PR #4942, opened 2020-10-20
[Android] Update input error visual cue rendering to support custom background drawables

### LINKED ISSUE #4830: [Android][Input validation error] [Input validation error UX]

almedina-ms:
An accident happened and #4895 got merged, redoing the PR
Fixes #4830 

## Description
Adds options to modify the rendering of the error visual cues for validated inputs. The original version of this feature only considered the default drawable which has a transparent background, due to this when tinting the background it could have unintended side effect such as tinting the whole background, such as in the example below

image

For this fix we added two new attributes called `adaptiveUsingCustomTextInputs` and `adaptiveUsingCustomChoiceSetInputs`. 
* `adaptiveUsingCustomTextInputs`: boolean flag that signalize the renderer to set the `state_error` with a value of true instead of tinting the text input background.
* `adaptiveUsingCustomChoiceSetInputs`: boolean flag to control some internal rendering for compact choice sets. If the value is set to `true` but no background is provided results may vary

On the other hand we added a new drawable state called `state_error` which allows customers to define background drawables which define how an invalid input would look like

Finally, we also added two new drawable selector files `adaptive_choiceset_compact_background.xml` and `adaptive_choiceset_expanded_background.xml`:
* `compact` drawable: controls the background for compact choicesets, if the `adaptiveUsingCustomTextInputs` flag is not set or set to false, the drawable is applied to a linear layout that encloses the spinner and is tinted in `attention` color.
* `expanded` drawable:  controls the background for expanded choicesets (mono and multi select) and the background for Toggle inputs. In all cases this background is tinted with the `attention` color to mimic the current JS and iOS behaviours.

## How Verified
The fix was verified by adding the previously mentioned properties into our app theme (which will be modified for each of the examples below)
```
<item name="adaptiveUsingCustomTextInputs">true</item>
<item name="adaptiveUsingCustomChoiceSetInputs">true</item>
<item name="android:editTextBackground">@drawable/layered_edit_text_styles</item>
```

For reference, layered_edit_text_styles is a selector which only defines a drawable for invalid state and a default drawable
``` 
<selector xmlns:android="http://schemas.android.com/apk/res/android" xmlns:adaptivecards="http://schemas.android.com/apk/res-auto" >
    <item adaptivecards:state_error="true" android:drawable="@drawable/layered_edit_text_invalid" />
    <item android:drawable="@drawable/layered_edit_text" />
</selector>
```

### Example 1: No custom inputs
```
<item name="adaptiveUsingCustomTextInputs">false</item>
<!-- <item name="android:editTextBackground">@drawable/layered_edit_text_styles</item> -->
```

image

### Example 2: Having a custom drawable, but stating that is not using custom inputs
```
<item name="adaptiveUsingCustomTextInputs">false</item>
<item name="android:editTextBackground">@drawable/layered_edit_text_styles</item>
```

image
> NOTE: the image is tinted, as it has a semi transparent background color it tints it in a lighter shade of Attention color

### Example 3: Having a custom drawable
```
<item name="adaptiveUsingCustomTextInputs">true</item>
<item name="android:editTextBackground">@drawable/layered_edit_text_styles</item>
```
image

### Example 4: Default behavior for choiceset and toggle input validations

```
<item name="adaptiveUsingCustomChoiceSetInputs">false</item>
```

Also, the default behaviour for choiceset inputs validation has changed to show a border around the inputs 

image

Toggle Input has changed too

image

And the default behaviour of compact choice set is as follows

image

### Example 5: Overriding a compact choiceset

styles.xml
```
<item name="adaptiveUsingCustomChoiceSetInputs">true</item>
```

The default background drawable for spinners can be updated by creating a file called `adaptive_choiceset_compact_background.xml` in the drawables folder, in this file an `state_error` flag can be set in the selector to define how the compact choice set should look like

adaptive_choiceset_compact_background.xml in drawables folder
```
<selector xmlns:android="http://schemas.android.com/apk/res/android" xmlns:adaptivecards="http://schemas.android.com/apk/res-auto">
    <item adaptivecards:state_error="true">
        <layer-list android:opacity="transparent">
            <item android:gravity="left" android:width="300dp" android:end="300dp">
                <shape>
                    <solid android:color="@color/colorWhite">
                    </solid>
                </shape>
            </item>
            <item android:gravity="left" android:width="100dp" android:start="300dp">
                <shape>
                    <solid android:color="@color/colorAccent">
                    </solid> 
                </shape>
            </item>
            <item android:width="100dp" android:start="300dp" android:height="40dp">
                <bitmap android:src="@drawable/cat1" android:gravity="fill"/>
            </item>
        </layer-list>
    </item>
</selector>
```

image

### Example 6: Overriding the rendering of expanded choicesets

The rendering for the expanded choice set and toggle inputs can also be modified (in this case no flag is required), the only consideration is that the border is always tinted with the `attention` color to keep consistency. As can be seen in the image below, the corners can be changed to be more rounded or change the width of the enclosing rectangle

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4942)


shalinijoshi19:
@almedina-ms this is running into asome Android build failures - is this needed for 20.10 Teams unblocking?

shalinijoshi19:
@jwoo-msft is this actually going into 20.10?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/customcontrols/IValidatedInputView.java
```diff
@@ -0,0 +1,6 @@
+package io.adaptivecards.renderer.input.customcontrols;
+
+public interface IValidatedInputView
+{
```
shalinijoshi19, line 4: For new files please add the Microsoft copyright header that other platform source files already include.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/customcontrols/ValidatedCheckBox.java
```diff
@@ -0,0 +1,72 @@
+package io.adaptivecards.renderer.input.customcontrols;
+
+import android.content.Context;
+import android.graphics.Color;
+import android.graphics.PorterDuff;
+import android.support.v4.content.ContextCompat;
+import android.widget.CheckBox;
+import android.widget.LinearLayout;
+
+import io.adaptivecards.R;
```
shalinijoshi19, line 10: >.R; [](start = 23, length = 3)

This is a very interesting namespace! What gives?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/customcontrols/ValidatedCheckBox.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class ValidatedCheckBox extends CheckBox implements IValidatedInputView
+{
+
+    public ValidatedCheckBox(Context context)
```
shalinijoshi19, line 15: >ValidatedCheckBox [](start = 11, length = 17)

how come we are selective about where we are adding documentation headers in this file? whats your criteria?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/customcontrols/ValidatedCheckBox.java
```diff
@@ -0,0 +1,72 @@
...
+            mergeDrawableStates(drawableState, STATE_INPUT_INVALID);
+        }
+
+        return drawableState;
+    }
+
+    private boolean m_isInvalid = false;
+    private int m_errorColor = Color.TRANSPARENT;
+
+    private static final int[] STATE_INPUT_INVALID = {R.attr.adaptive_state_input_invalid};
```
shalinijoshi19, line 71: >R [](start = 54, length = 1)

really not liking the "R" in here -- what does it denote?

source/android/mobile/src/main/res/drawable/layered_edit_text_styles.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android" xmlns:adaptivecards="http://schemas.android.com/apk/res-auto" >
+    <item adaptivecards:adaptive_state_input_invalid="true" android:drawable="@drawable/layered_edit_text_invalid" />
```
golddove, line 3: Consider using `adaptivecards:state_error`. We can use it in the future for non-inputs that can be in an error state (e.g. an Action, an entire AdaptiveCard).

There is no need to prefix `adaptive_` because it is already restricted to our namespace.

Android's existing states are in the following formats: "state_selected", "state_pressed", etc..

golddove (approved):
I still think we should try to detect custom input rather than requiring a flag, but this is something we can revisit in the future, as it wouldn't be a breaking change.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/customcontrols/ValidatedCheckBox.java
almedina-ms: R is the name of the container that includes drawables, color definitions, attribute definitions and other resources

https://developer.android.com/reference/android/R

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/customcontrols/ValidatedCheckBox.java
almedina-ms: as mentioned in a comment above, that's the android way of naming one of their APIs

https://developer.android.com/reference/android/R

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/customcontrols/IValidatedInputView.java
almedina-ms: Fixed

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/customcontrols/ValidatedCheckBox.java
almedina-ms: Added documentation headers for all new elements

source/android/mobile/src/main/res/drawable/layered_edit_text_styles.xml
almedina-ms: changed the name to use state_error

PR #4943, opened 2020-10-21
Added Event Handler Registration

### LINKED ISSUE #4888: [iOS] [Input Validation] Send notification with button reference to host on submit action
# Platform

**What platform is your feature request related to?**

* iOS

**Is your feature request related to a problem? Please describe.**

Scenario - Whenever user clicks on action.submit button, host is disabling the submit button for 3 sec (on submit action two events are triggered - button state change event and SDK send event). 
With this ACv1.3 input validation feature, if there are errors on the card then we don't want button to be disabled for 3 seconds. 
Our ask is to notify host with button ref on submit action when input validation fails. We do not have provision to change the button state when user clicks on submit button and there are errors on the card.

**Describe the solution you'd like**
There is one event `didChangeVisibility` that is triggered to change the button visibility (in case of Show card), we are expecting something like this for submit action to change the button state when input validation fails. This is a bad user experience if we disabled the button when input fields are not valid.

jwoo-msft:
## Related Issue
Fixed #4888

## Description
#4888 asked for sending a ref to a button, so it can be configured on input submission. Our team's general consensus is that we need to take a long term view and review other renderers for us to change the behaviors of the existing action delegates that are common to other renderers. This PR adds a registration for event handlers aka targets in iOS builders. Users can register event handler builders as they would with the renderer.  When implemented the builder interface, the builder builds the event handler and can attach the handler to the associated button. This change also includes sample code for demonstration purposes. The demo code will turn the submit action button to green when clicked when custom rendering is enabled in the sample app.
## How Verified
1. Unit tests are run.
2. Cards with actions were ran
3. Added sample code to the sample app to verify that users can register a target builder.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4943)

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRRegistration.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRRegistration.mm
@@ -7,6 +7,7 @@
 
 #import "ACRRegistration.h"
 #import "ACOBaseActionElement.h"
+#import "ACOHostConfigPrivate.h"
 #import "ACOParseContextPrivate.h"
 #import "ACRActionOpenURLRenderer.h"
 #import "ACRActionSetRenderer.h"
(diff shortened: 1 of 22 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRRegistration.mm
```diff
@@ -280,4 +281,142 @@ - (NSString *)getFeatureVersion:(NSString *)featureName
...
+                showcard : [ACRShowCardTargetBuilder getInstance],
+                toggle : [ACRToggleVisibilityTargetBuilder getInstance]
+            };
+            break;
+        case ACRSelectAction:
+            _builders = @{
+                openUrl : [ACRAggregateTargetBuilder getInstance],
+                submit : [ACRAggregateTargetBuilder getInstance],
+                toggle : [ACRToggleVisibilityTargetBuilder getInstance],
+                unknown : [ACRUnknownActionTargetBuilder getInstance]
```
RebeccaAnne, line 324: >unknown : [ACRUnknownActionTargetBuilder getInstance] [](start = 16, length = 53)

Does this represent custom actions? Why is it only here for SelectAction but not for the case above?

RebeccaAnne:
Is there an issue tracking making similar changes on other platforms?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRRegistration.mm
jwoo-msft: @RebeccaAnne , let me test it out the scenario and get back

jwoo-msft:
> Is there an issue tracking making similar changes on other platforms?

No, there is no issue tracking for other platforms. I will create one

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRRegistration.mm
jwoo-msft: @RebeccaAnne, I checked the sample, and the reason for not having unknown action handler registration is that it is not needed. If you have a custom action, you need a custom renderer, and you can register the action handler there. The select action, on the other hand, is different since they may want to have a custom action type, but if the users use a default renderer such as a TextBlock renderer,  the renderer has to register the custom action handler, so it's only there in the select action. The users can still register their action handler builders if they wish.

RebeccaAnne (approved):
:shipit:

jwoo-msft:
cc. @RebeccaAnne 
https://github.com/microsoft/AdaptiveCards/issues/4975

Issue #4912, opened 2020-10-12
[All][Accessibility] [Lack of accessibility support for error messages with the new Input feature]


jwoo-msft:
# All platforms
- [x]  Android (#4958)
- [x] iOS (#4966)
- [x] UWP
- [x] .NET WPF
- [x] JS

# Host

Microsoft Teams

# Version of SDK

SDK version 2.1.0 (ACv1.3)

# Details
reference #4913 
When there are errors on the cards after pressing submit, failure to submit due to errors is not called to users. The user also not aware if there are multiple errors on the card.  
Proposed fix: Call out "There are errors within the card" 
Link to [video](https://msit.microsoftstream.com/video/af0ba1ff-0400-9fb2-d848-f1eb08a2e1e4)

jwoo-msft:
@almedina-ms only for Android & 20.10 is TBD

shiladityasaha:
@shalinijoshi19 @jwoo-msft this is a ring blocker on all platforms for us. Can it be picked up for other platforms too?

shalinijoshi19:
Current status : P0's for Teams-impacting platforms needed for 20.10
[P0] - Android: @golddove working on changes.
[P0] - JS: @dclaux are you able to help out on this for JS? 
[P0] - iOS: @jwoo-msft is this on your radar? 
[P1] - UWP: @RebeccaAnne will need your help on this one and WPF to check out current state and changes that might be needed here 
[P1] - WPF: @RebeccaAnne

RebeccaAnne:
Per discussion the other day, I believe we decided that moving the focus to the first error input and reading out that input's error would be our approach for this bug. @golddove can you confirm whether that solution is the official POR here? Is there another  issue tracking that other than this one?

jwoo-msft:
@shalinijoshi19 yes, this is the issue I will be tackling on Today.

jwoo-msft:
@shalinijoshi19  @RebeccaAnne @golddove 
There has been a concern that error message can not be controlled by Teams or host app.

shalinijoshi19:
The plan of action here is to get into a debugging session tomorrow AM with the India team - @golddove @RebeccaAnne 
 are you able to get on a call tomorrow pls? Ideally we'd like to go over the precise scenario, and go over our proposal to make sure it will work.

dclaux:
> Per discussion the other day, I believe we decided that moving the focus to the first error input and reading out that input's error would be our approach for this bug. 

[...]

PR #4958, opened 2020-10-22
[Android][Accessibility] Announce label, errorMessage with inputs

golddove:
## Related Issue
Android part of #4912 

Also fixed ADO 28382490

## Description
Set inputs' accessible text to Input element's `label` + `errorMessage` when input has been invalidated (but only `label` when input is valid).

On submit, move keyboard focus, accessibility focus, and scroll position to first invalid input. So, above accessible text is spoken for first invalid input on submit.

## How Verified
Video:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4958)

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/StretchableInputLayout.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/StretchableInputLayout.java
@@ -18,15 +18,10 @@ public class StretchableInputLayout extends StretchableElementLayout
     private TextView m_label = null;
     private View m_inputView = null;
     private TextView m_errorMessage = null;
-//    private TextView m_errorLabel;
-    private boolean isValid = true;
 
     public StretchableInputLayout(Context context, boolean mustStretch)
     {
         super(context, mustStretch);
-
-        // Using setContentDescription on EditText is discouraged, so using hidden TextView instead
-//        m_errorLabel = new TextView(context);
     }
 
     public View getLabel()
@@ -49,9 +44,9 @@ public class StretchableInputLayout extends StretchableElementLayout
     {
         addView(input);
         m_inputView = input;
-//        if(m_inputView.getId() == View.NO_ID) {
-//            m_inputView.setId(View.generateViewId());
-//        }
+        if(m_inputView.getId() == View.NO_ID) {
+            m_inputView.setId(View.generateViewId());
+        }
         m_label.setLabelFor(m_inputView.getId());
     }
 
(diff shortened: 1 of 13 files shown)


shalinijoshi19:
> // Licensed under the MIT License.

@golddove as you are in these files can you update them with this copyright header where they are missing? thanks! 

---
Refers to: source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java:2 in cc78084. [](commit_id = cc78084cdbbfb89f5492918b011f055b3a84574d, deletion_comment = True)

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/StretchableInputLayout.java
```diff
@@ -1,13 +1,24 @@
 package io.adaptivecards.renderer.layout;
 
 import android.content.Context;
+import android.graphics.PorterDuff;
 import android.view.View;
+import android.view.ViewGroup;
+import android.widget.LinearLayout;
```
shalinijoshi19, line 7: This one for eg; Could you update with the copyright header here while you are making changes to the file? Thanks

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/StretchableInputLayout.java
```diff
@@ -33,20 +44,38 @@ public void setInputView(View input)
...
     {
         return m_errorMessage;
     }
 
-    public void setErrorMessage(View errorMessage)
+    public void setErrorMessage(TextView errorMessage)
     {
         addView(errorMessage);
         m_errorMessage = errorMessage;
+        m_errorMessage.setImportantForAccessibility(IMPORTANT_FOR_ACCESSIBILITY_NO);
```
shalinijoshi19, line 62: >IMPORTANT_FOR_ACCESSIBILITY_NO [](start = 52, length = 30)

What does this signify? Does setting it to NO cause readers to skip reading the error message? Is that the behavior we want ?

shalinijoshi19:
Could you specify (for later regression testing) what scenario/sample card you used to validate the e2e behavior here?

shalinijoshi19 (approved):
:shipit:

shalinijoshi19:
@jwoo-msft is added to the review. #Closed

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/StretchableInputLayout.java
golddove: This is to avoid announcing the error multiple times.

Issue #4967, opened 2020-10-26
[Templating] Complex expressions are not handled properly


dclaux:
# Platform

- [x] JavaScript
- [ ] .NET

# Version of SDK

All versions since release

# Details

The templating engine fails to correctly interpret complex expressions such as this one:
```
"targetElements": "${foreach(foreach(indicesAndValues(LineItems), x, concat('cardContent', x.index)), y, json(concat('{ \"elementId\": \"', y, '\", \"isVisible\": true }')))}"
```

The reason for this is the code mistakenly interprets the `}` character after `\"isVisible\": true` as being the end marker for the expression when it is actually a character within a string within the expression.

shalinijoshi19:
@dclaux / @jwoo-msft / @RebeccaAnne this is currently in 20.11; Based on @dclaux 's offline update this looks to be in the works for 20.10 potentially - somethings to consider before actually considering it for 20.10 though: 
- Will it have enough testing coverage to be able to land this wihtout regressions potentially @dclaux ? 
- @jwoo-msft is this something that needs to be addressed in .NET as well? If so we should update the bug title/description to include .NET also.
- If Yes to 2 above we should look to releasing this next sprint instead with greater testing coverage and confidence. Even an expedited release if needed to account for the holidays in Nov.

dclaux:
Reopening until the bug is fixed in .NET as well.

Stevenic:
@dclaux FYI... You can simplify your expression:

```
"targetElements": "${foreach(foreach(indicesAndValues(LineItems), x, concat('cardContent', x.index)), y, {elementId: y, isVisible: true})))}"
```

That's the same thing as using `json()` to parse a JSON based string but way easier to write :)

shalinijoshi19:
@jwoo-msft i've moved this out to 20.12 for now but do you have cycles to look into this in 20.11? @RebeccaAnne for FYI in case you are able to load balance if the Approvals team comes back wtih requests to accelerate this delivery in .NET

golddove:
@RebeccaAnne @jwoo-msft either `Status-Fixed` label should be removed, or the .NET part should be split into a new issue

PR #4971, opened 2020-10-26
Fix complex expression parsing

dclaux:
## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/4967

## Description
In some cases, a complex binding expression might use the `}` character within a string literal within the expression. In those cases, the template engine would mistakenly identify the character as the end marker of an `${ }` expression block, resulting in the expression not parsing nor evaluating.

This fixes that problem by solely relying on the Adaptive Expression library's logic to do the string interpolation.

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4971)


source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -240,17 +228,25 @@ export class Template {
...
+                if (parsedExpression.children.length === 1 && !(parsedExpression.children[0] instanceof AEL.Constant)) {
+                    // The concat contains a single child that isn't a constant, thus the original
+                    // string was a single expression. When evaluated, we want it to produce the type
+                    // of that single expression
+                    return parsedExpression.children[0];
+                }
+
+                // Otherwise, we want the expression to produce a string
+                return parsedExpression;
+            }
```
golddove, line 245: If `parsedExpression.type` is not `concat`, do we want to return the unparsed string?

golddove (approved):

source/nodejs/adaptivecards-templating/src/template-engine.ts
dclaux: That's what the code does.

PR #4986, opened 2020-10-28
[Android] Update Cpp flags in build.gradle

### LINKED ISSUE #4972: [Android][UnsatisfiedLinkError] [dlopen failed for libadaptivecards-native-lib.so file]
# Platform

* Android

# Author or host

Microsoft Teams

# Version of SDK

Adaptive card sdk version 2.2.0

# Details

Upgrading to the adaptive card sdk version from 2.1.0 to 2.2.0 is leading to this issue. When a call is made to the adaptive card native library there is an error which says failed to dlopen the libadaptivecards-native-lib.so file and the app crashes.
The error observed is as below:

`java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "_Unwind_Resume" referenced by "/data/app/com.microsoft.skype.teams.dev-GgORJFzm38OEBdFRaSYDzQ==/lib/arm64/libadaptivecards-native-lib.so"...`

This error is NOT observed while upgrading to 2.1.0 and something which has been introduced with 2.2.0. 

Additional Info: Teams uses the adaptive card sdk by downloading the maven package through [maven feeds and artifacts](https://docs.microsoft.com/en-us/azure/devops/artifacts/get-started-maven?view=azure-devops) offered by azure devops.

almedina-ms:
## Related Issue
Tries to fix #4972

## Description
Updates the cpp flags trying the solution found in this [stackoverflow post](https://stackoverflow.com/questions/10419801/undefined-reference-to-unwind-resume-and-gxx-personality-v0)

## How Verified
We have to wait 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4986)

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -9,12 +9,12 @@ assignees: dclaux, matthidinger, shalinijoshi19
 
 **What platform is your feature request related to? (Delete other platforms).**
 
-* .NET HTML
-* .NET WPF
-* Android
-* iOS
-* JavaScript
-* UWP
+- [ ] .NET HTML
+- [ ] .NET WPF
+- [ ] Android
+- [ ] iOS
+- [ ] JavaScript
+- [ ] UWP
 
 **Is your feature request related to a problem? Please describe.**
 A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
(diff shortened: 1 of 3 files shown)


jwoo-msft (approved):

shalinijoshi19:
@almedina-ms does this need to be merged in main to figure out if its working or not ?

PR #4987, opened 2020-10-28
[Android] Switch pipelines to use Windows hosted pools

### LINKED ISSUE #4972: [Android][UnsatisfiedLinkError] [dlopen failed for libadaptivecards-native-lib.so file]

golddove:
## Related Issue
Fixes #4972

## Description
Use Windows images to run pipelines to see if it solves Teams' issue

Also commented out the Get NDK task, as it is causing issues. Pipelines should now use the default NDK from the system image.

Will need to investigate further to see what the constraints are with NDK versions we can use/upgrade going forward.

## How Verified
Manual verification with Teams

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4987)

### DIFF
--- a/.pipelines/android-daily.yml
+++ b/.pipelines/android-daily.yml
@@ -42,7 +42,7 @@ steps:
    #!/usr/bin/env bash
    
    unset ANDROID_NDK_HOME
-   sudo ${ANDROID_HOME}/tools/bin/sdkmanager --verbose --install "ndk;21.0.6113669"
+   ${ANDROID_HOME}/tools/bin/sdkmanager --verbose --install "ndk;21.0.6113669"
   displayName: 'Get NDK (TODO: Remove this step)'
 
 - task: Gradle@2
--- a/.pipelines/android-daily.yml
+++ b/.pipelines/android-daily.yml
@@ -47,14 +47,14 @@ steps:
 - task: Gradle@2
   displayName: 'gradlew adaptivecards:build'
   inputs:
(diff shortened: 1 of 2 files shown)


shalinijoshi19:
@golddove does this need to be merged into main to see if it will work?

almedina-ms (approved):

PR #4561, opened 2020-08-10
[Samples] Update some of the samples to use v1.3 features

almedina-ms:
## Related Issue
References #4552 

ADO 23713386
ADO 23703188
ADO 23702956
ADO 23702320

## Description
Updates some of the samples that contain Inputs to use the `label`, `errorMessage`, `isRequired`, and the validation properties to update users into our v1.3 story

Waiting on the message to update older cards

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4561)

PR #4992, opened 2020-10-29
[Samples][Accessibility] Scenario samples with validation

golddove:
## Related Issue
ADO 28383590

## Description
Reviving part of @almedina-ms's unmerged PR #4561

These scenarios use schema 1.3 labels.

Need to show the new cards with labels/validation in the Android visualizer for accessibility testing. So, updated the Android visualizer to surface these new cards instead,

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/4992)


shalinijoshi19:
@golddove i agree this one doesnt need to be pulled into release/20.10 since it's a Visualizer change for Android and the samples it surfaces, but eventually will find itself in next month's release automatically.  Any concerns with that?

shalinijoshi19 (approved):

golddove:
That should be good

Issue #4470, opened 2020-07-22
[Javascript][Template SDK] [Template expander dislikes backtick markdown notation]


jpjpjp:
# Platform

* JavaScript

# Author or host

Author sending cards to Webex Teams

# Version of SDK

1.0.0

# Details

I'm updating a previously working application that was using an alpha version of the template lib.   I've updated my card JSON to include the dollar sign before all {templates}, and for the most part everything is working fine.   

I am getting an error when I call template.expand on the following JSON which includes markdown with the backtick element:
```json
{
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "TextBlock",
      "text": "This bot's [README](${appSourceUrl}/blob/master/README.md) talks more about how `it` works.",
      "wrap": true
    }
  ]
}
```

If I attempt to execute the following:
```js
let template = new ACData.Template(thatCardAbove);
let expanded = template.expand({
   $root: {
      appSourceUrl: "https://github.com/somerepo"
   }
});
```
The error message is: Error generating cards: syntax error at line 1:92 mismatched input 'it' expecting <EOF>

SDK Version 0.1.1-alpha.0 had no problem with this same JSON (without the dollar sign).

RebeccaAnne:
Please reopen if this is still an issue with the latest packages.

PR #5000, opened 2020-10-30
[Javascript][Template SDK] Add escape operator before string interpolation

Danieladu:
## Related Issue
Fix: https://github.com/microsoft/AdaptiveCards/issues/4470

## Description
The original logic of parsing template into Expression is simply wrapping the template with the pair of the backticks.
Which would bring in some special characters by mistake. 
This PR aims to add some escape conversions for the  backticks ``` ` ```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5000)

### DIFF
--- a/source/nodejs/adaptivecards-templating/src/template-engine.ts
+++ b/source/nodejs/adaptivecards-templating/src/template-engine.ts
@@ -84,24 +84,24 @@ export class GlobalSettings {
      * and that field is undefined or null. By default, expression evaluation will substitute an undefined
      * field with its binding expression (e.g. `${field}`). This callback makes it possible to customize that
      * behavior.
-     *
+     * 
      * **Example**
      * Given this data object:
-     *
+     * 
      * ```json
      * {
      *     firstName: "David"
      * }
      * ```
-     *
+     * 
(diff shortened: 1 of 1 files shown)


Danieladu:
Another solution is to replace the "Manual stitching"  with `StringExpression`. This class takes some builtin escape operators.

boydc2014:
@dclaux can you help take a look at this?

dclaux:
@Danieladu update:
- The \ issue is indeed a bug in Adaptive Expressions. It's been fixed in 4.10.2. We will have to issue a new version of adaptivecards-templating that uses that
- The ` issue is also a bug in Adaptive Expressions. There is a PR for the fix here: https://github.com/microsoft/botbuilder-js/pull/2983

So all in all this PR will not be necessary.

Danieladu:
Thanks @dclaux.  Actually the fix is in StringExpression which I recommended to use. But the way you guys use in code is manually concat, that should escape the special characters manually.

Danieladu:
Sorry for my understanding of the backslash issue before. We have made a change in 4.10 so it may work now. This pr just try to fix the backticks ` escape issue.

PR #5002, opened 2020-10-30
[NodeJS] Adds getActionById implementation to ActionSet

### LINKED ISSUE #5001: [NodeJS] ActionSet.getActionById() always returns undefined
# Platform

- [ ] JavaScript

# Author or host

Author

# Version of SDK

2.4.0

# Details

I'm trying to use `getActionById()` an an `ActionSet` and it is always returning undefined. When I look at the source code, I can see that AdaptiveCard extends `ContainerWithActions` which has a fully implemented instance of this method, but when I look at the ActionSet code, I can see this method has been missed entirely.

scottlovegrove:
## Related Issue
Fixes #5001 

## Description
Adds an implementation of `getActionById()` to ActionSet

When calling `ActionSet.getActionById()` it is always returning `undefined` because it is using the base implementation and not a correct one.

## How Verified
Have added a test for this.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5002)


scottlovegrove:
@matthidinger is there someone I can poke for this PR so it doesn't get closed as inactive?

dclaux (approved):
Good catch, and thanks for the fix!

PR #5008, opened 2020-10-30
Updated Target Builder Director Interface

### LINKED ISSUE #5025: [iOS][Custom Action] [ Can not send additional properties to action delegates in case of custom action]
# Platform

- [ ] iOS

# Author or host
Teams

# Version of SDK
2.3.0

# Details
Currently we can not build target with action element. It would be good to have build target function for `BaseActionElement` so that we can send extra properties in case of custom action to the delegate function.

jwoo-msft:
## Related Issue && Description
Fixed #5025 
The target builder director's interface exposes C++ interfaces, and these changes wrap the interface with the objective-c version.

## How Verified
1. Existing Unit tests & tested with sample cards for regression
2. Manual scenario verification if any; ***Do include .gif's or screenshots of the testing you performed here if you think that it 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5008)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,6 +12,9 @@ This information lives with the code and will help future readers (***including 
 
 At the very least please describe the issue you are addressing, and what your fix entails. 
 
+## Sample Card
+If appropriate, please include a link to a card in one of the samples directories that can be used to validate this change. This can be an existing card or a card added with this PR.
+
 ## How Verified
 How you verified the fix, including one or all of the following: 
 1. New unit tests that were added if any. If none were added please add a quick line explaining why not.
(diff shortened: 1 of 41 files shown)


shalinijoshi19:
@jwoo-msft there seems to be sort of an issue with the iOS/UWP CI pipelines - @golddove ideas on why we are seeing duplicate pipelines run (nightly and CI) both?

jwoo-msft:
> @jwoo-msft there seems to be sort of an issue with the iOS/UWP CI pipelines - @golddove ideas on why we are seeing duplicate pipelines run (nightly and CI) both?

the build pipeline result here ran 6 days ago, it seems fine now.

golddove (approved):

PR #5044, opened 2020-11-04
[UWP] Automatic tooltip on select action

### LINKED ISSUE #4700: [UWP] Automatic tooltip for selectAction title
UWP Implementation for #4694

RebeccaAnne:
## Related Issue
Fixes #4700

## Description
Set the action title as a tooltip on select actions.

## Sample Card
[Column.SelectAction.json](https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.0/Elements/Column.SelectAction.json)

## How Verified
Confirmed tooltip appeared for select actions in the Column.SelectAction card.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5044)


source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -677,21 +676,44 @@ namespace AdaptiveNamespace::ActionHelpers
...
         {
-            // If we have an action, use the title for the AutomationProperties.Name
+            // If we have an action, use the title for the AutomationProperties.Name and tooltip
             HString title;
             THROW_IF_FAILED(action->get_Title(title.GetAddressOf()));
 
-            ComPtr<IDependencyObject> buttonAsDependencyObject;
-            THROW_IF_FAILED(button.As(&buttonAsDependencyObject));
+            if (title.IsValid())
+            {
```
shalinijoshi19, line 684: Where does the tooltip show up by default? Where/How do we configure that and is that controllable by the host?

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -677,21 +676,44 @@ namespace AdaptiveNamespace::ActionHelpers
...
-                GetActivationFactory(HStringReference(RuntimeClass_Windows_UI_Xaml_Automation_AutomationProperties).Get(),
-                                     &automationPropertiesStatics));
+                THROW_IF_FAILED(automationPropertiesStatics->SetName(buttonAsDependencyObject.Get(), title.Get()));
 
-            THROW_IF_FAILED(automationPropertiesStatics->SetName(buttonAsDependencyObject.Get(), title.Get()));
+                WireButtonClickToAction(button.Get(), action, renderContext);
 
-            WireButtonClickToAction(button.Get(), action, renderContext);
+                // Also use the title as the tooltip
+                ComPtr<IToolTip> toolTip =
```
shalinijoshi19, line 699: Are there any accessibility settings/configuration needed around tooltips? Also is there any way to suppress tooltips instead of always displaying it? Based on recommendations here [https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/tooltips#recommendations](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/tooltips#recommendations) and in general Accessibility concerns looks like tooltips (like placeholders) could be prone to abuse/misuse etc ?

RebeccaAnne: This scenario is using the title as a tooltip on a select action. As you can see in the code above, we're already using that title as the accessibility text for the button that is the select action. This tooltip is designed to provide more information to sighted users as to what will happen when they click the select action. For suppression, were you thinking for the card author or the host? In what scenario would they want to suppress it? 

---
In reply to: [523179028](https://github.com/microsoft/AdaptiveCards/pull/5044#discussion_r523179028) [](ancestors = 523179028)

source/uwp/Renderer/lib/ActionHelpers.cpp
RebeccaAnne: By default, a ToolTip is displayed centered above the pointer. ([https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/tooltips#create-a-tooltip](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/tooltips#create-a-tooltip)). There is some placement configuration available in XAML, but i don't think it's something that makes sense for the host to control, as presumably any positioning concerns would be card specific. I suppose we could expose positioning configuration to the card schema, but i'm not convinced that's necessary (and would be a schema change, which this currently isn't). 

---
In reply to: [523177305](https://github.com/microsoft/AdaptiveCards/pull/5044#discussion_r523177305) [](ancestors = 523177305)

source/uwp/Renderer/lib/ActionHelpers.cpp
RebeccaAnne: For what it's worth on the accessibility question, i just tested it out using keyboard nav and narrator. When you navigate to the select action using keyboard navigation the tooltip does show, and if narrator is on the title text is read out a single time. That seems like the right behavior to me.

---
In reply to: [525346775](https://github.com/microsoft/AdaptiveCards/pull/5044#discussion_r525346775) [](ancestors = 525346775,523179028)

paulcam206 (approved):

PR #5066, opened 2020-11-10
Fixed Stock Accesibility Symbol

jwoo-msft:
## Related Issue
Fixed https://microsoft.visualstudio.com/OS/_queries/edit/29467497

## Description
When the Up arrow symbol is positioned as in the below image, the voiceover will read the symbol in a foreign language. Changed the order of the symbol for the voiceover to correctly detect the symbol.
**Image with the voiceover issue**
IMG_75C0E6F9DBD3-1
**Image after the fix**
IMG_84FDEDCA90D8-1

## Sample Card
https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.0/Scenarios/StockUpdate.json

## How Verified
1. Ran the voiceover on the stock update card.
2. Existing relevant unit/regression tests that you ran: N/A
3. Manual scenario verification if any;  Turn on voiceover, and click on the label, and make sure no foreign language is heard.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5066)


dclaux:
Is the goal of this fix only to "check the box" for our accessibility assessment? Because it's not really a solution to the actual problem; nothing prevents a card author to put the symbol at the beginning of the text.

Also, what platform is this reproing on? In HTML, the symbol is not read by narrator at all as far as I can tell.

jwoo-msft:
> Is the goal of this fix only to "check the box" for our accessibility assessment? Because it's not really a solution to the actual problem; nothing prevents a card author to put the symbol at the beginning of the text.
> 
> Also, what platform is this reproing on? In HTML, the symbol is not read by narrator at all as far as I can tell.

Thanks for your comment. This is for iOS platform. Another solution will be having the AdaptiveCards library to set the voiceover language. Currently, the language is auto-detected; thus the problem. 
I modeled after the below UI.
Screen Shot 2020-11-16 at 10 58 00 AM

dclaux:
I think that's what this fix should be about - explicitly setting the voiceover language instead of relying on auto-detection.

You could use the value of the `lang` property as a hint if it is specified, and otherwise fallback to the system's configured locale.

Again, if our goal is simply to pass the accessibility assessment on the web site, what you did is good enough. It is not a "fix" though, it's a workaround for one particular sample.

shalinijoshi19:
@dclaux  this isn't impacting the website but the iOS behavior of the sample; I agree with you here -- this seems very odd and @jwoo-msft we should follow up with the accessibility team here if needed.
It's curious that it's treating the symbol differently as text and not a symbol, and just the placement seems to make it less paranoid, and its specific to iOS -  Would we run into issues when we have the locale set to "that non-english 'auto-detected' language" in this case? What is the behavior for this same sample on Android/UWP and why doesn't it repro on there?

shalinijoshi19:
Tracking this as an Apple platform bug.

PR #5070, opened 2020-11-11
Added correct accessibility roles for select actions

jwoo-msft:
## Related Issue
ADO https://microsoft.visualstudio.com/OS/_queries/edit/29254830/?triage=true

## Description
Selection Actions on Image doesn't announce their accessibility roles, the fix adds the correct roles.

## Sample Card
https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.0/Elements/Image.SelectAction.json
## How Verified
1. New unit tests that were added if any. The issue is reproducible with the sample card above.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5070)


dclaux (commented):
How does this work with custom actions implemented by developers? Why is the action itself not providing its UI accessibility trait?

jwoo-msft:
> How does this work with custom actions implemented by developers? Why is the action itself not providing its UI accessibility trait?

@dclaux,
If you are referring the action to the one in the code below,
`            if (action.type == ACRSubmit || action.type == ACRToggleVisibility) {
`
that would work for custom action as well. Since that will be part for public interface change, do you want me to bring this to the wider discussion? 
My proposal will be adding a platform specific accessibility role tag to Adaptive Actions elements such Action.OpenURL, Action.Submit, and etc.
`                recipientView.accessibilityTraits |= action.accessibilityRole;
`

dclaux:
@jwoo-msft not sure I understand. Take `Action.Execute`; its accessibility trait should be "button". But with your code, its accessibility trait will simply not be set at all, because it doesn't meet any of the conditions in your `if...else` statement. And even if it did meet the condition, how would you guarantee that a custom action always gets the appropriate trait, unless the custom action itself "says" what trait it should have?

In JS, the base `Action` class declares a `getAriaRole` method that custom action classes can override.

jwoo-msft:
> @jwoo-msft not sure I understand. Take `Action.Execute`; its accessibility trait should be "button". But with your code, its accessibility trait will simply not be set at all, because it doesn't meet any of the conditions in your `if...else` statement. And even if it did meet the condition, how would you guarantee that a custom action always gets the appropriate trait, unless the custom action itself "says" what trait it should have?
> 
> In JS, the base `Action` class declares a `getAriaRole` method that custom action classes can override.

@dclaux,
Thanks, David.
I forgot to mention that I appreciate your suggestion because I wasn't sure what you were suggesting initially. I agree with you on the change, and also I think we should make analogous changes to all platforms for consistency, and I believe that the role should be left as platform specific, e.g getAriaRole for JS, getAccessibilityTrait for iOS.

dclaux (approved):

Issue #5092, opened 2020-11-18
[JavaScript][Extensibility] [Min/max limit not set on number input]


aroraketan:
# Platform

What platform is your issue or question related to? (Delete other platforms).

- [ ] JavaScript

# Author or host

Host: Microsoft Teams (multi-window).

# Version of SDK

JS v2.5.0.

# Details

On a custom number type input, if the limit `min`/`max` is set to `0`, it is set as `undefined` and hence the range validation is not effective.

shalinijoshi19:
@dclaux are you able to take a look at this real quick? @golddove who;d been investigating with the Teams dev.

dclaux:
@aroraketan given that this cannot be reproduced with the default implementation of `NumberInput`, and assuming that your custom `NumberInput`:
- Extends the base `NumberInput`
- Replaces its rendering by overriding `internalRender`

My conclusion would be that your custom control code doesn't properly set the min and max values on your replacement control. Please check that and let me know.

dclaux:
Scratch the above - there is indeed a bug in the base `NumberInput` implementation whereas if `min` or `max` are set to 0 they are not set on the input control.

I will fix today.

PR #5093, opened 2020-11-18
[TS] Input.Number bug fixes

dclaux:
## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/5092

## Description
When `min` or `max` where set  to 0 in the AC payload, the code would treat them as `undefined` and thus not stamp them onto the generated `<input>` tag. This was due to the use of statements like `if (!min)` which returns true whether `min` is `undefined` or equal to 0.

## Sample Card
```json
{
    "type": "AdaptiveCard",
    "version": "1.3",
    "body": [
        {
            "type": "Input.Number",
            "min": 0,
            "isRequired": true,
            "errorMessage": "Error",
            "id": "dss"
        }
    ],
    "actions": [
        {
            "type": "Action.Submit",
            "title": "Action.Submit"
        }
    ]
}
```

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5093)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3345,11 +3345,11 @@ export class NumberInput extends Input {
         this._numberInputElement = document.createElement("input");
         this._numberInputElement.setAttribute("type", "number");
 
-        if (this.min) {
```
golddove, line 3348: Feature request for typescript: this should compile down to a null/undefined check. Truthy/falsy values only cause pain, with very little gain.

golddove (approved):

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: So true

Issue #5100, opened 2020-11-20
[JS][Templating] Template engine broken after change in AEL


dclaux:
# Platform

- JavaScript

# Description

As of version 4.10, AEL has introduced a change in the way it generates a root `concat` expression. Namely, it now always prepend an empty string constant.

This unfortunately break our templating logic.

This is a high priority issue as Search depends on it.

Because of this issue, the AEL team is now committing to providing a built-in solution to something that we've so far had to build manually in our templating engine SDKs.

jwoo-msft:
@dclaux, I checked .NET. the issue isn't reproducible.
![image](https://user-images.githubusercontent.com/4112696/99747098-0ce5aa80-2a8f-11eb-8965-bd848520dccf.png)

dclaux:
Great

RebeccaAnne:
This is currently repro-ing on the latest bits so reactivating.

RebeccaAnne:
False alarm. This is in fact working as expected so re-closing.

PR #5101, opened 2020-11-20
Fix break caused by AEL behavioral change in v4.10

dclaux:
## Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/5100

## Description
Works around a behavioral change in AEL introduced in version 4.10 which causes single expressions such as `${age}` to be generated as `concat` and produce strings instead of the actual type of the expression.

## Test card
```
{
    "type": "AdaptiveCard",
    "$data": {
        "title": "Not empty"
    },
    "body": [
        {
            "type": "TextBlock",
            "size": "Medium",
            "weight": "Bolder",
            "text": "Test card"
        },
        {
            "type": "TextBlock",
            "text": "Should be visible only if title isn't empty",
            "wrap": true,
            "$when": "${title != ''}"
        }
    ],
    "version": "1.3"
}
```

Without this fix, the "$when" clause will incorrectly evaluate to the string "true". With the fix, it will correctly evaluate to the boolean value `true`.

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5101)

### DIFF
--- a/source/nodejs/adaptivecards-designer-app/package-lock.json
+++ b/source/nodejs/adaptivecards-designer-app/package-lock.json
@@ -45,13 +45,13 @@
 		},
 		"@microsoft/recognizers-text-data-types-timex-expression": {
 			"version": "1.1.4",
-			"resolved": "https://onedrive.pkgs.visualstudio.com/_packaging/odsp-npm/npm/registry/@microsoft/recognizers-text-data-types-timex-expression/-/recognizers-text-data-types-timex-expression-1.1.4.tgz",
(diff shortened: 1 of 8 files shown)


jwoo-msft (approved):

Danieladu:
It looks OK. If change back to the original model is more convenient,  we will also consider it.

PR #5128, opened 2020-12-03
[JS] Ingest marked-schema package

paulcam206:
## Description
While addressing some package updates, `marked-schema` showed up as needing updating. As it's now pretty tightly coupled with our `spec-generator` package, it makes more sense to just ingest the codebase into our repo.

## How Verified
* local build

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5128)


source/nodejs/marked-schema/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
paulcam206, line 1: @matthidinger - do we have any concerns with just bringing this package in with its existing license?

PR #5139, opened 2020-12-07
[Shared] Add associatedInputs property to Action.Submit

golddove:
## Related Issue
Shared model part of #5078

## Description
Add `associatedInputs` property to `Action.Submit` on shared model with possible values of `Auto` and `None`. Default is `Auto`.

## How Verified
(pending implementation on consuming renderers)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5139)


source/shared/cpp/ObjectModel/SubmitAction.h
```diff
@@ -25,6 +25,10 @@ namespace AdaptiveSharedNamespace
 
         virtual bool GetIgnoreInputValidation() const;
         virtual void SetIgnoreInputValidation(const bool value);
+        #pragma deprecated(GetIgnoreInputValidation, SetIgnoreInputValidation)
```
RebeccaAnne, line 28: I think we can just remove this. It was never hooked up to anything so i would be surprised if anyone had a dependency on it. That said, UWP has a dependency on it :) so the build will fail there. You can either remove it from there too or I'd be fine with just checking it in broken to the feature branch and i can fix it when i do UWP.

RebeccaAnne (approved):
:shipit:

source/shared/cpp/ObjectModel/SubmitAction.h
golddove: Removed, and updated tests

PR #5141, opened 2020-12-07
[UWP] Add support for associatedInputs none value

RebeccaAnne:
## Related Issue
UWP implementation for #5078

## Description
Consumed the associatedInputs property in UWP from the shared model. Updated the submit code to not perform validation and to return an empty inputs object when none is set.

## Sample Card
New card added with this change at samples\v1.3\Elements\Action.Submit.AssociatedInputs.json

## How Verified
Updated UWP object model tests and ensured that they still run and pass. Manual validation of scenario using new sample card.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5141)

### DIFF
--- a/samples/v1.3/Elements/Action.Submit.AssociatedInputs.json
+++ b/samples/v1.3/Elements/Action.Submit.AssociatedInputs.json
@@ -1,0 +1,25 @@
+{
+  "type": "AdaptiveCard",
+  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+  "version": "1.3",
+  "body": [
+    {
+      "type": "Input.Text",
+      "id": "name",
+      "label": "Please enter your name:",
+      "isRequired": true,
+      "errorMessage": "Name is required"
+    }
+  ],
+  "actions": [
+    {
+      "type": "Action.Submit",
+      "title": "Cancel",
+      "associatedInputs": "none"
+    },
+    {
+      "type": "Action.Submit",
+      "title": "Submit"
+    }
+  ]
+}
--- a/samples/v1.3/Elements/Action.Submit.AssociatedInputs.json
+++ b/samples/v1.3/Elements/Action.Submit.AssociatedInputs.json
@@ -1,25 +1,25 @@
 {
-  "type": "AdaptiveCard",
-  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
-  "version": "1.3",
-  "body": [
-    {
-      "type": "Input.Text",
-      "id": "name",
-      "label": "Please enter your name:",
-      "isRequired": true,
-      "errorMessage": "Name is required"
-    }
-  ],
-  "actions": [
-    {
-      "type": "Action.Submit",
-      "title": "Cancel",
-      "associatedInputs": "none"
-    },
-    {
-      "type": "Action.Submit",
-      "title": "Submit"
-    }
-  ]
+	"type": "AdaptiveCard",
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"version": "1.3",
+	"body": [
+		{
+			"type": "Input.Text",
+			"id": "name",
+			"label": "Please enter your name:",
+			"isRequired": true,
+			"errorMessage": "Name is required"
+		}
+	],
+	"actions": [
+		{
+			"type": "Action.Submit",
+			"title": "Cancel",
+			"associatedInputs": "none"
+		},
+		{
+			"type": "Action.Submit",
+			"title": "Submit"
+		}
+	]
 }
--- a/source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
+++ b/source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
@@ -358,7 +358,8 @@ AdaptiveNamespaceStart
 
     [version(NTDDI_WIN10_RS1)]
 #ifdef ADAPTIVE_CARDS_WINDOWS
-    [contract(InternalContract, 1)][internal]
+    [contract(InternalContract, 1)]
+    [internal]
 #endif
     typedef[v1_enum] enum IsVisible {
         IsVisibleToggle = 0,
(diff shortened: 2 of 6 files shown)


golddove (approved):

samples/v1.3/Elements/Action.Submit.AssociatedInputs.json
```diff
@@ -0,0 +1,25 @@
+{
+  "type": "AdaptiveCard",
```
paulcam206, line 2: you'll need to reconcile the whitespace in this version (spaces) vs. the .NET version (tabs) :)

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -389,6 +389,15 @@ AdaptiveNamespaceStart
         All = Left | Right | Up | Down
     } BleedDirection;
 
+    [version(NTDDI_WIN10_RS1)]
+#ifdef ADAPTIVE_CARDS_WINDOWS
+    [contract(InternalContract, 1)][internal]
```
paulcam206, line 394: not an issue with this PR in particular, but I thought `[internal]` was supposed to be on its own line? looks like we're inconsistent with it...

own line:
https://github.com/microsoft/AdaptiveCards/blob/58e35dbe6440432ed8c5d5253961d2734392cca1/source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl#L33-L45

just do whatever:
https://github.com/microsoft/AdaptiveCards/blob/58e35dbe6440432ed8c5d5253961d2734392cca1/source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl#L359-L367

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -1519,7 +1528,7 @@ AdaptiveNamespaceStart
         AdaptiveSubmitAction();
 
         Windows.Data.Json.JsonValue DataJson;
-        Boolean IgnoreInputValidation;
```
paulcam206, line 1522: was this ever wired up, or just in the object model?

source/uwp/Renderer/lib/RenderedAdaptiveCard.cpp
```diff
@@ -294,22 +290,41 @@ namespace AdaptiveNamespace
             ComPtr<IAdaptiveSubmitAction> submitAction;
             RETURN_IF_FAILED(localActionElement.As(&submitAction));
 
+            ABI::AdaptiveNamespace::AssociatedInputs associatedInputs;
+            submitAction->get_AssociatedInputs(&associatedInputs);
```
paulcam206, line 294: not sure if there's much we can do about the nomenclature at this point, but is it a bit weird that you `get_AssociatedInputs()` to retrieve an enum value that tells you whether this instance might have associated inputs, but `get_UserInputs()` to actually maybe get the inputs (if present)?

paulcam206 (approved):
approve with comments 👍

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
RebeccaAnne: I'll move the ones that are not on their own line. For what it's worth, the autoformatting tries to move it to share the line, but it also makes a lot of other questionable changes. At some point we might want to figure out if there's a way to make the auto-formatting for .idl consistent with what we want (or just accept what it does so we're not fighting it).

---
In reply to: [537960705](https://github.com/microsoft/AdaptiveCards/pull/5141#discussion_r537960705) [](ancestors = 537960705)

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
RebeccaAnne: Just object model, it never did anything.

---
In reply to: [537961179](https://github.com/microsoft/AdaptiveCards/pull/5141#discussion_r537961179) [](ancestors = 537961179)

source/uwp/Renderer/lib/RenderedAdaptiveCard.cpp
RebeccaAnne: Ugh. Yes, that's horrible. Like you say though, i'm not sure what we can do about it now. I think we want to stick with the associated inputs naming for the new schema prop (which may someday support a list of inputs) and the API should match. But it's also probably not worth the breaking change to change UserInputs. Hopefully the fact that associatedInputs matches the property in the schema will help somewhat. Also someday we should have some API docs :)

---
In reply to: [537962522](https://github.com/microsoft/AdaptiveCards/pull/5141#discussion_r537962522) [](ancestors = 537962522)

golddove: Does the naming have to match the schema property? The getters/setters can say `AssociatedInputIds`, perhaps

RebeccaAnne: That seems worse for right now, because you're not getting a list of Ids, you're only getting "none" or "auto". I also think matching makes it clearer because the devs can see the mapping to the schema props.

---
In reply to: [538626638](https://github.com/microsoft/AdaptiveCards/pull/5141#discussion_r538626638) [](ancestors = 538626638)

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
paulcam206: filed #5144 to track this -- thanks!

source/uwp/Renderer/lib/RenderedAdaptiveCard.cpp
paulcam206: I agree with you @RebeccaAnne -- naming should follow the schema to the furthest extent possible (especially since we don't have solid API docs) :)

PR #5155, opened 2020-12-10
Feature/associated inputs

### LINKED ISSUE #5078: [UX] Unable to have two SubmitAction Buttons *and* validation

RebeccaAnne:
## Related Issue
Fixes #5078

## Description
Merges feature branch containing all platforms implementations of asssociatedInputs none to main

## Sample Card
Sample card added as samples\v1.3\Elements\Action.Submit.AssociatedInputs.json

## How Verified
See individual PRs for per platform verification performed

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5155)


golddove:
@RebeccaAnne @paulcam206 what was the resolution about the XML doc files? Do we want to commit and start tracking them?

RebeccaAnne:
> @RebeccaAnne @paulcam206 what was the resolution about the XML doc files? Do we want to commit and start tracking them?

I can delete them if we don't want them. But if we don't want them we should also add them to .gitignore because right now everytime I build .NET they show up and I have to delete them. Or, alternatively, if they're supposed to get deleted by the build process we should figure out why they don't (at least locally for me).

RebeccaAnne:
I went ahead and deleted them and logged #5157 to figure out what we want to happen to these.

golddove (approved):

paulcam206 (approved):

paulcam206:
you can ignore the failing `Generate Specs` check -- it's failing to push changes to the branch due to branch protection rules

PR #5169, opened 2020-12-11
[UWP][NET] Introduce correct package EULAs

paulcam206:
## Description
Update EULAs for binary .NET and UWP packages. Text versions of the EULAs can be found in the `source` directory. Note that these EULAs do not govern our source code which, of course, remains under the MIT license. Instead, the EULAs outline rights and responsibilities for our published binary packages.

It's probably worth noting that nuget is picky about where it pulls files from, hence the various copies of the EULAs in different dirs.

## How Verified
Nuget Package Explorer tool from the Windows Store.


jwoo-msft (approved):

Kasbr305:
Hi

PR #5232, opened 2021-01-06
[NET] Consume Microsoft.MarkedNet package

paulcam206:
## Description
Our HTML and WPF renderers both rely on the `Microsoft.MarkedNet` nuget package, which is a fork of `MarkedNet`. The fork was created in order to ship it as a nuget package as well as fix some bugs that were difficult to upstream. The package hasn't changed since 2017 and was never updated to be signed SHA256. Rather than ship a new packaging-changes-only version, we're inboxing the library.

## How Verified
* local build and tests

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5232)


matthidinger (approved):
I didn't try building it, but the csproj files look good to me

RebeccaAnne (approved):

PR #5254, opened 2021-01-11
[Site] Make Feedback button appear as link for accessibility

paulcam206:
## Related Issue
Fixes VSO 30854543

## Description
Our feedback button behaves like a link (navigates to another site), so we need to make sure it looks like a link to accessibility.

image
(note that the github button already appears as a link)

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5254)


source/nodejs/adaptivecards-site/themes/adaptivecards/layout/_partial/header.ejs
```diff
@@ -442,7 +442,7 @@
         <div class="action-bar">
             <ul class="action-list">
                 <li>
-                    <a role="button" id="feedback-button" class="button" data-bi-name="feedback" tabindex="0" focusable="true">
```
RebeccaAnne, line 445: For my own learning, why don't we need to assign a role of "link" or something here? I assume that's a default, but where is it coming from?

RebeccaAnne (approved):

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/_partial/header.ejs
paulcam206: yeah -- some element types have an implied role (`a` is `link`, `button` is, well... `button), but you can use the `role` attribute to override it. usually you see explicit role declarations on things like a `div` that has a click handler :)

RebeccaAnne: Ah, got it. The class="button" threw me as I was wondering why a button defaulted to link role and I missed that we were in a link element here.

paulcam206: I forgot to point out that for purposes of accessibility, if a click action causes a navigation to occur (or a new tab to appear), it should be marked as `link`, even if it's visually a button. The opposite is true in that something that looks like a link should me marked as `button` if it simply performs some action on the current page.

PR #5256, opened 2021-01-11
[Designer] Collapse toolbox on ESC

paulcam206:
## Related Issue
Fixes VSO 30863472

## Description
The expand/collapse widget in the designer should respect an `ESC` keypress to trigger collapsing a toolbox:

image

## How Verified
* local build, devtools

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5256)


dclaux:
Why?

paulcam206:
it's a fair question -- the bug has details. I'm not sure I necessarily agree with the requirement, but I figure that the change is simple and innocuous enough that it's easier to take the change rather than argue about it

dclaux:
I must say this "bug" doesn't make much sense to me. It's a nit, not an actual issue (let alone an accessibility issue), pretending that it would make users more efficient without demonstrating it would.

In the tree view in particular, ESC would be a lot more useful if it selected the parent of the currently selected element (which for that matter is already implemented on the design surface itself). This would allow for much faster and predictable navigation.

The code change itself is fine, but I really don't think it's necessary.

paulcam206:
> I must say this "bug" doesn't make much sense to me. It's a nit, not an actual issue (let alone an accessibility issue), pretending that it would make users more efficient without demonstrating it would.
> 
> In the tree view in particular, ESC would be a lot more useful if it selected the parent of the currently selected element (which for that matter is already implemented on the design surface itself). This would allow for much faster and predictable navigation.
> 
> The code change itself is fine, but I really don't think it's necessary.

feel free to push back in the bug -- I just don't think the argument is worth the effort

dclaux:
>  I just don't think the argument is worth the effort
Agreed...

RebeccaAnne (approved):

PR #5275, opened 2021-01-14
Added the extensibility for the custom text block, Image and ActionButton component

IbrahimSulai:
Added the extensibility for the custom text block component

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5275)

### DIFF
--- a/source/community/reactnative/src/visualizer/renderer.js
+++ b/source/community/reactnative/src/visualizer/renderer.js
@@ -129,7 +129,7 @@ export default class Renderer extends React.Component {
         //Register Custom Action Button Components
         Registry.getManager().registerComponent('CustomActionButton', CustomActionButtonRenderer);
 
-        //Register Custom Action Button Components
+        //Register Custom Image Components
         Registry.getManager().registerComponent('CustomImage', CustomImageRenderer);
 
         //Register Custom Actions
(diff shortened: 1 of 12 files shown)


vivekvijayakrishnan (approved):
LGTM

source/community/reactnative/src/visualizer/renderer.js
```diff
@@ -169,7 +181,7 @@ export default class Renderer extends React.Component {
                 ],
                 { cancelable: false }
             )
-        } else if (actionObject.type === "Action.OpenUrl" && !Utils.isNullOrEmpty(actionObject.url)) {
+        } else if (actionObject.type === "Action.OpenUrl" && !Utils.isNullOrEmpty(actionObject.url) || actionObject.type === "CustomActionButton") {
```
vivekvijayakrishnan, line 184: `CustomActionButton` can be anything such as `URL / submit or showCard`. we can just alert it with the payload details. Please use **_`alertAction`_** method. I think we are already calling **_`alertAction`_** by default if it's not `Submit/URL`. So you can remove this condition.

vivekvijayakrishnan (changes requested):

vivekvijayakrishnan (approved):
LGTM

source/community/reactnative/src/visualizer/renderer.js
IbrahimSulai: Resolved

pragadeeshk (approved):
LGTM

Issue #5302, opened 2021-01-21
[Adaptive Card-UI][Text Block] Ellipses do not appear in Text Block when text length exceeds the limit set by "maxLines" and "wrap" = "true"


riagarwa-ms:
1. Ellipses in TextBlock of Adaptive Card do not show up if text is truncated when "maxLines" and "wrap" values are set. Example in the below screenshot "maxLines" is set to 2 and "wrap" is set to true.
![image](https://user-images.githubusercontent.com/64960865/105281942-34faa280-5bd3-11eb-8bd3-4bff91becfed.png)

2. However, if "maxlines" and "wrap" values are not set. The text gets truncated and ellipses show up.
Below is the screenshot
![image](https://user-images.githubusercontent.com/64960865/105282026-6c694f00-5bd3-11eb-9e19-5399d4c921d9.png)

How can we get the ellipses enabled when "maxLines" value is set and "wrap" is enabled?

riagarwa-ms:
The screenshot is taken from Teams client. (An app added in Teams to handle link unfurling request). Here is the json for the Text block: {
                      "type": "TextBlock",
                      "text": "Now that we have defined the main rules and features of the format, we need to produce a schema and publish it to GitHub The schema will be the starting point of our reference documentation.",
                      "fontType": "Default",
                      "size": "Large",
                      "spacing": "small",
                      "maxLines": "2",
                      "wrap": true
                    }

Attaching the complete json payload.
[jsonPayload.txt](https://github.com/microsoft/AdaptiveCards/files/5850515/jsonPayload.txt)

dclaux:
This is a known limitation and something we don't really have a plan to fix it for now. The fix is not straightforward.

dclaux:
I found an acceptable solution to this that also happens to address another bug. Will fix.

Issue #5249, opened 2021-01-08
[JavaScript][Input.ChoiceSet] [If no choices are selected on a multi-selection ChoiceSet a JavaScript error is raised when clicking on a Action.Submit button]


bbullock-trustquay:
# Platform

What platform is your issue or question related to? (Delete other platforms).

- [x] JavaScript

# Author or host

I'm both an author and a host. I author the Adaptive Card template and data JSON in a C# .NET desktop application and render it in a web application.

# Version of SDK

C# .NET - NuGet 2.4.0
JavaScript - NPM 2.6.0

# Details

Please note, the issue I'm about to describe also occurs in the Adaptive Cards Designer (https://adaptivecards.io/designer/).

When authoring a ChoiceSet of A, B and C with default values 1, 2, and 3. If no selection is made and I click on an Action.Submit button a JavaScript error is thrown.

The error message is:
core.js:5967 ERROR TypeError: Cannot read property 'toString' of undefined
    at SubmitAction.push.ABWh.SubmitAction.internalPrepareForExecution (card-elements.js:3366)
    at SubmitAction.push.ABWh.Action.prepareForExecution (card-elements.js:3247)
    at raiseExecuteActionEvent (card-elements.js:5225)
    at SubmitAction.push.ABWh.Action.raiseExecuteActionEvent (card-elements.js:3167)
    at SubmitAction.push.ABWh.Action.execute (card-elements.js:3238)
    at ActionButton.actionButton.onClick (card-elements.js:4024)
    at ActionButton.push.ABWh.ActionButton.click (card-elements.js:3111)
    at HTMLButtonElement.action.renderedElement.onclick [as __zone_symbol__ON_PROPERTYclick] (card-elements.js:3104)
    at HTMLButtonElement.wrapFn (zone-evergreen.js:1218)
    at ZoneDelegate.invokeTask (zone-evergreen.js:399)

A screenshot of the error occurring in the Adaptive Cards Designer can be seen below:
![image](https://user-images.githubusercontent.com/69152251/104039258-0c83a900-51ce-11eb-90f8-bf668e8c7303.png)

shalinijoshi19:
@dclaux could you please take a look at this?

bbullock-trustquay:
When will this change be available in NPM please?

dclaux:
In the 21.01 release which is probably happening today, at worst early next week.

dclaux:
[...]

Issue #4847, opened 2020-09-29
[JS] Markdown headings clipped


paulcam206:
Using markdown to generate headings leads to clipped text.

Sample:
```json
{
    "type": "AdaptiveCard",
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "version": "1.0",
    "body": [
        {
            "type": "TextBlock",
            "text": "# Markdown Heading"
        }
    ]
}
```

Here's what it looks like in the designer (note that this repros across hostconfigs regardless of whether in preview mode or not):

![image](https://user-images.githubusercontent.com/16614499/94601331-34b35180-0248-11eb-994b-7f6812ad2009.png)

paulcam206:
It appears that this is due to the `line-height` property. I'm guessing this calculation doesn't account for the possibility of header tags?

https://github.com/microsoft/AdaptiveCards/blob/e736ed7dfca1949ac76a499e205b2d61997d7a65/source/nodejs/adaptivecards/src/card-elements.ts#L1077-L1104

shalinijoshi19:
@dclaux do you mind taking a quick stab at this to understand what is going on here (pls also see @paulcam206 's comments above).. ?

dclaux:
@shalinijoshi19 the solution to this is to use the non-standard `-webkit-line-clamp` CSS property which was not even an option back when the JS renderer was initially implemented. That is why the renderer uses the same trick basically every other Web developer uses, which is to explicitly set `line-height` and impose a `max-height`, which only works well when the text uses a single font size, which isn't true when using Markdown headers. Also, `-webkit-line-clamp` isn't supported in IE, so the trick currently used would have to remain as a fallback for IE.

I do not think this is a high priority issue though, or even worth fixing at all.

Do the other renderers handle this case properly?

PR #5307, opened 2021-01-21
Fix exception with multiselect ChoiceSet, ellipsis with maxLines in TextBlock and truncated Markdown headers

dclaux:
## Related Issue
- Fixes https://github.com/microsoft/AdaptiveCards/issues/5302
- Fixes https://github.com/microsoft/AdaptiveCards/issues/4847
- Fixes https://github.com/microsoft/AdaptiveCards/issues/5249
- Fixes https://github.com/microsoft/AdaptiveCards/issues/5315

## Description
- Code was attempting to call `toString()` on `undefined` when a ChoiceSet input had no selection
- `TextBlock.maxLines` is now handled using the non-standard `--webkit-line-clamp` CSS property. While non-standard, this property is supported across all browsers, except IE. In IE, the renderer will fall back to the old behavior. This change fixes two bugs:
  - Because line clamping takes care of truncation, the renderer doesn't need to specify explicit line heights anymore, which leads to Markdown headings not being cut off anymore
  - Truncated `TextBlock.text` now always has an ellipsis, even if it wraps

## Sample Card
```json
{
    "type": "AdaptiveCard",
    "version": "1.3",
    "body": [
        {
            "type": "TextBlock",
            "text": "# Lorem ipsum dolor sit amet,\n## consectetur adipiscing elit,\nsed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.",
            "wrap": true,
            "maxLines": 5
        },
        {
            "type": "Input.ChoiceSet",
            "choices": [
                {
                    "title": "Choice 1",
                    "value": "Choice 1"
                },
                {
                    "title": "Choice 2",
                    "value": "Choice 2"
                }
            ],
            "id": "ds",
            "isMultiSelect": true
        }
    ],
    "actions": [
        {
            "type": "Action.Submit",
            "title": "Action.Submit"
        }
    ]
}
```

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5307)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -1015,8 +1015,20 @@ export class TextBlock extends BaseTextBlock {
                 element.style.wordWrap = "break-word";
 
                 if (this.maxLines && this.maxLines > 0) {
-                    element.style.maxHeight = (this._computedLineHeight * this.maxLines) + "px";
                     element.style.overflow = "hidden";
+
+                    if (Utils.isInternetExplorer() || !GlobalSettings.useWebkitLineClamp) {
+                        element.style.maxHeight = (this._computedLineHeight * this.maxLines) + "px";
+                    }
+                    else {
+                        // While non standard, --webkit-line-clamp works in every browser (except IE)
+                        // and is a great solution to support the maxLines feature with ellipsis
+                        // truncation. With --webkit-line-clamp there is need to use explicit line heights
+                        element.style.removeProperty("line-height");
+                        element.style.display = "-webkit-box";
+                        element.style.webkitBoxOrient = "vertical";
+                        element.style.webkitLineClamp = this.maxLines.toString();
+                    }
                 }
             }
             else {
@@ -3995,7 +4007,7 @@ export class SubmitAction extends Action {
             for (let key of Object.keys(inputs)) {
                 let input = inputs[key];
 
-                if (input.id) {
+                if (input.id && input.isSet()) {
                     this._processedData[input.id] = typeof input.value === "string" ? input.value : input.value.toString();
                 }
             }
--- a/source/nodejs/adaptivecards-designer-app/package-lock.json
+++ b/source/nodejs/adaptivecards-designer-app/package-lock.json
@@ -576,7 +576,7 @@
 				},
 				"util": {
 					"version": "0.10.3",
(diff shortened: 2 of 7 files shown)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -1015,8 +1015,20 @@ export class TextBlock extends BaseTextBlock {
...
+                    if (Utils.isInternetExplorer() || !GlobalSettings.useWebkitLineClamp) {
+                        element.style.maxHeight = (this._computedLineHeight * this.maxLines) + "px";
+                    }
+                    else {
+                        // While non standard, --webkit-line-clamp works in every browser (except IE)
+                        // and is a great solution to support the maxLines feature with ellipsis
+                        // truncation. With --webkit-line-clamp there is need to use explicit line heights
+                        element.style.removeProperty("line-height");
+                        element.style.display = "-webkit-box";
+                        element.style.webkitBoxOrient = "vertical";
```
paulcam206, line 1029: looks like box orient [doesn't work by default](https://caniuse.com/?search=-webkit-box-orient) in Firefox? is that something we should be concerned about?

paulcam206 (approved):
looks good (assuming we're cool with not everything working in Firefox)

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: I've tested the change in Edge, Chrome, Firefox, Opera and Brave and it worked across the board.

dclaux: Ah - and now I remember what happened when I looked at the DOM in Firefox. It actually automatically replaced `--webkit-box-orient` with `-moz-box-orient`:
image

So the question is, how did that happen?
- Is that something new in more recent versions of Firefox?
- Is it TypeScript magic?

I'm going to assume the latter, but I can't find any info.

PR #5323, opened 2021-01-27
[Shared] Validate floating-point values for Input.Number

### LINKED ISSUE #5278: [Teams][Input Validation] Input.Number floating point support
# Platform

- [ ] Android
- [ ] SharedModel 
- [ ] iOS

See also #5308(.NET) and #5309(UWP)

# Author or host

Host - Teams

# Version of SDK

20.12 releases

# Details

There is platform inconsistency on acceptance of floating-point values for Input.Number.

* JS renderer accepts floating-point in the card payload (i.e. Input.Number.min, Input.Number.max, Input.Number.value) and accepts floating-point user input as valid.
* iOS renderer does not accept floating-point in card payload, but accepts floating-point user input
* Android renderer does not accept floating-point at all
* UWP renderer does not accept floating-point in card payload, but accepts floating-point user input. Additionally, min/max comparisons can be incorrect due to flooring.
* .NET Xceed renderer does not accept floating-point at all

This needs to be made consistent. The correct solution seems to be to match the JS renderer on all platforms.

### LINKED ISSUE #5309: [UWP][Input Validation] Input.Number floating point support
Spun off from #5278

# Platform

- [ ] UWP

# Details

There is platform inconsistency on acceptance of floating-point values for Input.Number.

- JS renderer accepts floating-point in the card payload (i.e. Input.Number.min, Input.Number.max, Input.Number.value) and accepts floating-point user input as valid.
- UWP renderer does not accept floating-point in card payload, but accepts floating-point user input. Additionally, min/max comparisons can be incorrect due to flooring.
- 

This needs to be made consistent. The correct solution seems to be to match the JS renderer on all platforms.

golddove:
## Related Issue
Fixes #5278 and fixes #5309

## Description
Update shared model to support floating-point values for min, max, and value properties on Input.Number element. Update Android library accordingly and perform proper validation.

## Sample Card
`v1.3/Tests/Input.Number.Float`

## How Verified
Updated existing automated Android tests to support double.
Added test card with floating min, max, value for manual testing: 
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5323)


paulcam206 (approved):

golddove:
Update: iOS reinterprets the double as int, so need to make the changes on iOS (and likely UWP, too) to avoid regressions. Moving out so we can ship the fix for all platforms at once

jwoo-msft (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTextInputHandler.h
```diff
@@ -22,9 +22,9 @@
 
 @interface ACRNumberInputHandler : ACRTextInputHandler
 
-@property NSInteger min;
+@property NSNumber *min;
```
jwoo-msft, line 25: why not CGFloat?

golddove: Looks like CGFloat is typedef for `float` on 32-bit arch. Wouldn't that mean we truncate the value coming from the shared model?

golddove: I could use a plain double, though

jwoo-msft: double is also fine.

jwoo-msft (approved):

RebeccaAnne (approved):

PR #5357, opened 2021-02-04
[.NET] Update XCeed number input control to support non-integer values

### LINKED ISSUE #5308: [.NET][Input Validation] Input.Number floating point support
Spun off from #5278

# Platform

- [ ] .NET XCeed

# Details

There is platform inconsistency on acceptance of floating-point values for Input.Number.

JS renderer accepts floating-point in the card payload (i.e. Input.Number.min, Input.Number.max, Input.Number.value) and accepts floating-point user input as valid.

.NET Xceed renderer does not accept floating-point at all

This needs to be made consistent. The correct solution seems to be to match the JS renderer on all platforms. This bug tracks the .NET XCeed change

RebeccaAnne:
## Related Issue
Fixes #5308

## Description
Updated the XCeed number input renderer to use DoubleUpDown rather than IntegerUpDown in order to support non-integer values.

Also update the code to set the element max and min on the control in order to allow the control to enforce those properties.

## Sample Card
Sample card to be added in #5323

## How Verified
Verified manually that decimal values are supported and that min and max are enforced as expected.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5357)


paulcam206 (approved):

jwoo-msft:
@RebeccaAnne FWIW, iOS CI failed because the shared module failed to parse the card.

RebeccaAnne:
> @RebeccaAnne FWIW, iOS CI failed because the shared module failed to parse the card.

Yeah, I think we need to wait for #5278 before adding a sample. Looks like @golddove has a sample card as part of his outstanding PR for that (#5323), so I'll just pull the sample card from here for now.

Issue #5356, opened 2021-02-04
[Input Labels] Support for labels to the side of inputs 


RebeccaAnne:
# Platform

All

# Details

We don't currently support a way for a label to appear to the side of an input. This came up during accessibility work for the designer, where the Element Properties are built using adaptive card elements. Because we cannot put the label to the side of the input as is done in the designer, there is no way using standard adaptive card properties to correctly label the inputs for accessibility purposes. I worked around it for the purpose of fixing the designer bug, but we should support this type of layout for our customers as well, given that since we use it ourselves its clearly a useful scenario.

![image](https://user-images.githubusercontent.com/9091833/106928853-82453c80-66c8-11eb-97a7-a873a828ed8c.png)

PR #5375, opened 2021-02-10
[Designer] Update peer button name from "X" to "Remove"

paulcam206:
## Related Issue
Fixes VSO #30909723

## Description
Simple change - button was titled `X`, but it should be more descriptive. I chose `Remove`.

## How Verified
* local build, narrator

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5375)

### DIFF
--- a/source/nodejs/adaptivecards-designer/src/designer-peers.ts
+++ b/source/nodejs/adaptivecards-designer/src/designer-peers.ts
@@ -782,7 +782,7 @@ class NameValuePairPropertyEditor extends PropertySheetEntry {
                 valueColumn.addItem(textInput);
 
                 let removeAction = new Adaptive.SubmitAction();
-                removeAction.title = "X";
+                removeAction.title = "Remove";
                 removeAction.onExecute = (sender) => {
                     nameValuePairs.splice(i, 1);
 
--- a/source/nodejs/adaptivecards-designer/src/designer-peers.ts
+++ b/source/nodejs/adaptivecards-designer/src/designer-peers.ts
@@ -782,7 +782,8 @@ class NameValuePairPropertyEditor extends PropertySheetEntry {
                 valueColumn.addItem(textInput);
 
                 let removeAction = new Adaptive.SubmitAction();
-                removeAction.title = "Remove";
+                removeAction.title = "X";
+                removeAction.accessibleTitle = "Remove";
                 removeAction.onExecute = (sender) => {
                     nameValuePairs.splice(i, 1);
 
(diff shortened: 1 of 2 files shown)


paulcam206:
ah - need to update role as well

dclaux (changes requested):
I don't agree with this. The label doesn't need to be anything but an X, which is expressive enough and a known paradigm for "delete" or "remove" across many applications. Also, it is short which is preferable in this case where the horizontal space available to edit the key/value pair is limited.

If anything, the change should be:
- Make sure there's an accessible label on that button so that narrator doesn't just read X
- Allow for the text in that button to be configurable by the host app, by adding it (along with the accessible label mentioned above) to the resource strings

dclaux:
To add to my comment above. If there was true justification in changing this "X" into "Remove", then there would also be justfified to change the "X" to close a window or dialog into "Close" - but that's not the case.

paulcam206:
> To add to my comment above. If there was a real justification for changing this "X" into "Remove", then it would also be justified to change the "X" to close a window or dialog into "Close" - but that's not the case.

The alternative is to provide a means to paint `aria-label` independent of an `Action.Submit`'s `title`, which is probably spec-level work?

RebeccaAnne:
> > To add to my comment above. If there was a real justification for changing this "X" into "Remove", then it would also be justified to change the "X" to close a window or dialog into "Close" - but that's not the case.
> 
> The alternative is to provide a means to paint `aria-label` independent of an `Action.Submit`'s `title`, which is probably spec-level work?

You could try to add something like this just in the designer and then log a bug for the "real work", sort of like what I did in #5277 (and logged #5356). Not the best approach maybe, but you might be able to decouple the designer accessibility work from the renderer feature.

dclaux:
@paulcam206 just add an `accessibleLabel` property to the Action class and update the render logic to use it as the `aria-label`. Not talking about adding a schema property, just an internal public field that can only be used in code. This has been done multiple times already, like to disable Markdown processing on TextBlock via the `useMarkdown` property.

RebeccaAnne:
> @paulcam206 just add an `accessibleTitle` property to the Action class and update the render logic to use it as the `aria-label`. Not talking about adding a schema property, just an internal public field that can only be used in code. This has been done multiple times already, like to disable Markdown processing on TextBlock via the `useMarkdown` property.

We probably still should log something to put it in the schema eventually though.

jwoo-msft (approved):

paulcam206:
> > @paulcam206 just add an `accessibleTitle` property to the Action class and update the render logic to use it as the `aria-label`. Not talking about adding a schema property, just an internal public field that can only be used in code. This has been done multiple times already, like to disable Markdown processing on TextBlock via the `useMarkdown` property.
> 
> We probably still should log something to put it in the schema eventually though.

done! (#5383)

dclaux (approved):

PR #5394, opened 2021-02-17
[JS] Fix some accessibility issues

dclaux:
## Related Issue
- Partially fixes ADO 30961076
- Fixes ADO 30868015
- Fixes ADO 30926701

### ADO 30961076
- The label of the input is now read by narrator when the input gets focus.
  - Note that somehow this bug seemed to already be fixed. However, I found that the fix could be streamlined and more tightly baked into the SDK, so I tweaked it.
- Narrator will however also read the placeholder text ("not set") even when the input has a valid value. This appears to be a Narrator bug.

### ADO 30868015
- It is unclear that the CTRL+SHIFT+F10 keyboard shortcut is actually supposed to force showing the tooltip of a focused element. However showing the tooltips for those buttons on keyboard navigation is really to compensate for the fact the buttons do not have a label. Given the cost of showing tooltips on focus (this isn't a standard HTML5 behavior and requires implementing custom tooltips) this fix actually simply adds "Undo" and "Redo" labels to the buttons.

### ADO 30926701
- The toolbar will now wrap when horizontal space isn't enough to show all commands.
- Note that overlapping and/or truncation can still happen when the width of the window is very narrow.

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5394)

### DIFF
--- a/source/nodejs/adaptivecards-designer/src/card-designer.ts
+++ b/source/nodejs/adaptivecards-designer/src/card-designer.ts
@@ -670,7 +670,6 @@ export class CardDesigner extends Designer.DesignContext {
             this._hostContainerChoicePicker = new ToolbarChoicePicker(CardDesigner.ToolbarCommands.HostAppPicker);
             this._hostContainerChoicePicker.separator = true;
             this._hostContainerChoicePicker.label = "Select host app:"
-            this._hostContainerChoicePicker.width = 350;
 
             for (let i = 0; i < this._hostContainers.length; i++) {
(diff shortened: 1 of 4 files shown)


paulcam206 (approved):

paulcam206:
I re-ran the Android job -- there were some agent pool capacity issues should now be resolved

PR #5405, opened 2021-02-18
[Designer][Accessibility] Group Card Elements sections with label for accessibility

RebeccaAnne:
## Related Issue
Fixes ADO [30898371](https://microsoft.visualstudio.com/OS/_workitems/edit/30898371/)

## Description
Group elements for each element category (Containers, Elements, Inputs) in a div with a label so that the group heading is read out by narrator when that section is entered via tab navigation.

## How Verified
Confirmed expected behavior locally using narrator.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5405)

### DIFF
--- a/source/nodejs/adaptivecards-designer/src/card-designer.ts
+++ b/source/nodejs/adaptivecards-designer/src/card-designer.ts
@@ -675,7 +675,6 @@ export class CardDesigner extends Designer.DesignContext {
             this._hostContainerChoicePicker = new ToolbarChoicePicker(CardDesigner.ToolbarCommands.HostAppPicker);
             this._hostContainerChoicePicker.separator = true;
             this._hostContainerChoicePicker.label = "Select host app:"
-            this._hostContainerChoicePicker.width = 350;
 
             for (let i = 0; i < this._hostContainers.length; i++) {
                 this._hostContainerChoicePicker.choices.push(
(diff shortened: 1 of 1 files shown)


source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -670,6 +675,7 @@ export class CardDesigner extends Designer.DesignContext {
             this._hostContainerChoicePicker = new ToolbarChoicePicker(CardDesigner.ToolbarCommands.HostAppPicker);
             this._hostContainerChoicePicker.separator = true;
             this._hostContainerChoicePicker.label = "Select host app:"
+            this._hostContainerChoicePicker.width = 350;
```
paulcam206, line 678: looks like an incorrect merge (here and below)?

paulcam206 (changes requested):

source/nodejs/adaptivecards-designer/src/card-designer.ts
RebeccaAnne: Indeed. Not sure where those came from. Fixed.

paulcam206 (approved):

PR #5408, opened 2021-02-19
[Site] Mark current header page with aria-current

paulcam206:
## Related Issue
Fixes VSO 30961098

## Description
There's no indication on the site header of which option is currently being viewed:

image

This change adds `aria-current="page"` to the currently-viewed page. Note that there's no visual indication. It appears that Narrator doesn't look at this property, but NVDA (and presumably others) do.

## How Verified
* local build, devtools, NVDA

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5408)


RebeccaAnne:
If this doesn't fix the issue for narrator do we need another fix that does? Or is that more considered a bug in narrator itself?

paulcam206:
follow from our contact is that they will follow up with Narrator folks, so we're good to take this fix.

RebeccaAnne (approved):

PR #5417, opened 2021-02-23
[Site] Emit status messages on copy button invoke

paulcam206:
## Related Issue
Fixes VSO 30959686

## Description
There's no audible or visual indication that invoking the copy button on the samples page has performed any action. This fixes that.

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5417)

### DIFF
--- a/source/nodejs/adaptivecards-site/themes/adaptivecards/source/js/script.js
+++ b/source/nodejs/adaptivecards-site/themes/adaptivecards/source/js/script.js
@@ -478,14 +478,24 @@ $(function () {
 	$("button.copy-data").click(function (e) {
 		var content = $(this).parent().next("pre").text();
 		copyToClipboard(content);
-		$(this).append("<span role='status' aria-label='Sample data copied successfully' id='dataCopySuccess' />");
+
+        if ($("#dataCopySuccess").length === 0) {
+		    $(this).append("<span role='status' aria-label='Sample data copied successfully' id='dataCopySuccess'> <i class='fas fa-check'></i></span>");
+            setTimeout(() => { $("#dataCopySuccess").remove(); }, 2000);
+        }
+
 		$(this).focus();
 	});
 
 	$("button.copy-code").click(function (e) {
 		var content = $(this).parent().next("pre").text();
 		copyToClipboard(content);
-		$(this).append("<span role='status' aria-label='Sample JSON copied successfully' id='copySuccess' />");
+
+        if ($("#copySuccess").length === 0) {
+		    $(this).append("<span role='status' aria-label='Sample JSON copied successfully' id='copySuccess'> <i class='fas fa-check'></i></span>");
+            setTimeout(() => { $("#copySuccess").remove(); }, 2000);
+        }
+
 		$(this).focus();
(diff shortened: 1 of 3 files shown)


source/nodejs/adaptivecards-site/themes/adaptivecards/source/js/script.js
```diff
@@ -475,10 +475,18 @@ $(function () {
 
 	});
 
+	$("button.copy-data").click(function (e) {
+		var content = $(this).parent().next("pre").text();
+		copyToClipboard(content);
+		$(this).append("<span role='status' aria-label='Sample data copied successfully' id='dataCopySuccess' />");
+		$(this).focus();
```
RebeccaAnne, line 482: >$(this).focus() [](start = 2, length = 15)

How does this effect where the keyboard focus lands after invoking? Playing around with both your version and the shipped site the keyboard focus wanders off someplace weird and inconsistent for me in both cases, so it doesn't look like this makes things any more broken than the already are (pretty broken). That said, what the is expectation with how the keyboard focus finds its way back since this line explicitly moves the focus to this span?

source/nodejs/adaptivecards-site/themes/adaptivecards/source/js/script.js
```diff
@@ -475,10 +475,18 @@ $(function () {
 
 	});
 
+	$("button.copy-data").click(function (e) {
+		var content = $(this).parent().next("pre").text();
+		copyToClipboard(content);
+		$(this).append("<span role='status' aria-label='Sample data copied successfully' id='dataCopySuccess' />");
```
RebeccaAnne, line 481: >$(this).append("<span role='status' aria-label='Sample data copied successfully' id='dataCopySuccess' />"); [](start = 2, length = 107)

Does this result in a new span being added for each button click in such a way that they pile up?

paulcam206: it does, but it's not really a concern -- there's no real end user impact... the message is read only when the element is added to the DOM.

source/nodejs/adaptivecards-site/themes/adaptivecards/source/js/script.js
paulcam206: the shipped site has a bug with templated samples due to both copy buttons having the `copy-code` class. I noticed that and fixed it in this change. 

as for the `focus()` call itself, we're re-focusing `$(this)`, which is `$("button.copy-data")` or `$("button.copy-code")`, and not the newly-added child.

RebeccaAnne: I'm still seeing weirdness here, the narrator focus often wanders off to seemingly random places after clicking copy. That said, I'm seeing it on the shipped site too, so it doesn't seem to be related to this change.

RebeccaAnne (approved):
:shipit:

PR #5424, opened 2021-02-24
[Designer][Accessibility] Put open showcard next in tab order

golddove:
## Related Issue
[ADO 30866987](https://microsoft.visualstudio.com/OS/_queries/edit/30866987/)

## Description
For keyboard users, an ActionSet with multiple Action.ShowCard's could cause an issue with tab order: after a user selects a ShowCard to activate, they may have to tab through several other Actions in the ActionSet before they reach the newly revealed content. This temporarily removes those Actions from the taborder after a card is revealed (then resets them after tabbing).

Seems to fall under this WCAG failure condition:
[`F85: Failure of Success Criterion 2.4.3 due to using dialogs or menus that are not adjacent to their trigger control in the sequential navigation order`](https://www.w3.org/TR/WCAG20-TECHS/F85.html)

## Sample Card
`Scenarios/FoodOrder`

## How Verified
Manual verification

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5424)


dclaux:
I don't believe this is the right way to handle this.

Currently, one navigates from one button to another in a set using the Tab key, thus indeed a user needs to tab through all buttons before they can land in the open ShowCard.

We should change so that the navigation in the button strip is done via arrow keys. The full button strip would be a tab stop, but the next tab would go to to the next focusable element in the card, namely the open card.

dclaux:
We probably want to have a design conversation about this.

dclaux (changes requested):
I don't believe this is the right way to handle the requirement. Let's have a design discussion.

dclaux (approved):

PR #5425, opened 2021-02-24
[Designer][Accessibility] Allow disabled button focus in toolbar

golddove:
## Related Issue
[ADO 30909987](https://microsoft.visualstudio.com/OS/_queries/edit/30909987)

## Description
When a user uses the Undo/Redo buttons, they may become disabled after the action. This causes the focus to be lost. During investigation, some recommendations were found regarding focusable disabled buttons in certain scenario (e.g. toolbar). 

ARIA-practices [guidance ](https://www.w3.org/TR/wai-aria-practices/#kbd_disabled_controls) and [example](https://www.w3.org/TR/wai-aria-practices/examples/toolbar/toolbar.html) (note the cut/copy/paste focusable buttons, even in disabled state). This allows screen reader users to be aware of these disabled buttons.

## Sample Card
N/A

## How Verified
Manual verification

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5425)


source/nodejs/adaptivecards-designer/src/adaptivecards-designer.css
```diff
@@ -788,9 +788,10 @@
     border: 1px solid black;
 }
 
-.acd-toolbar-button:disabled {
+.acd-toolbar-button.acd-toolbar-button-disabled {
```
dclaux, line 791: This change will break apps that are customizing the existing styles. While I don't disagree that it might have been better to name the class as you have it here, I don't think this change is necessary.

dclaux (changes requested):
Looks good, but I don't think we should rename the CSS class.

source/nodejs/adaptivecards-designer/src/adaptivecards-designer.css
golddove: Using the `:disabled` state prevents focus.

dclaux: Oh sorry, my mistake - I didn't see the ":" and thought it was a "." You're introducing a new class.

Got it, looks good.

dclaux (approved):

paulcam206 (approved):

PR #5440, opened 2021-02-26
Fixed #5303

### LINKED ISSUE #5303: [iOS][Rendering] There's no way to get additional properties from ACOAdaptiveCard
# Platform

- [ ] iOS

# Author or host

Host

# Version of SDK

2.4.6

# Details

We have a need to override some properties at the card level that are normally at the host config level. For example our host config file might contain:

```
    "mycompany.contentLightColor": "FFFFFFFF",
    "mycompany.contentDarkColor": "00263EFF",
```

But for some cards, we want to override those with different colors. In the rendering hooks for the UI elements we can get a reference to the `ACOAdaptiveCard` via `ACRView.card` but the problem is that `ACOAdaptiveCard` doesn't expose any of the properties that we've included at the top level of the card.

jwoo-msft:
## Related Issue
Fixed #5303 

## Description
Added additional properties support at the card level for Shared model & iOS

## Sample Card
```json
 {
    "type": "AdaptiveCard",
    "version": "1.3",
    "card.cornerRadius" : 50.0,
    "body": [
        {
            "type": "Container",
            "items": [
                {
                    "type": "TextBlock",
                    "text": "Test Card",
                    "wrap": true
                }
            ],
            "my.cornerRadius" : 20.0
        }
    ]
}
```
## How Verified
1. Added a new Unit Test to iOS & Shared models.
2. Ran existing unit tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5440)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
```diff
@@ -111,4 +111,24 @@ - (void)setCard:(std::shared_ptr<AdaptiveCard> const &)card
...
+    if (_adaptiveCard) {
+        Json::Value blob = _adaptiveCard->GetAdditionalProperties();
+        if (blob.empty()) {
+            return nil;
+        }
+        Json::StreamWriterBuilder streamWriterBuilder;
+        auto writer = streamWriterBuilder.newStreamWriter();
+        std::stringstream sstream;
+        writer->write(blob, &sstream);
+        delete writer;
```
paulcam206, line 125: If you use `std::unique_ptr<>`, its lifetime is automatically managed for you :)

```suggestion
        std::unique_ptr<StreamWriter> writer(streamWriterBuilder.newStreamWriter());
        std::stringstream sstream;
        writer->write(blob, &sstream);
```

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
```diff
@@ -228,6 +233,10 @@ std::shared_ptr<ParseResult> AdaptiveCard::Deserialize(const Json::Value& json,
     // Parse optional selectAction
     result->SetSelectAction(ParseUtil::GetAction(context, json, AdaptiveCardSchemaKey::SelectAction, false));
 
+    Json::Value addtionalProperties;
```
paulcam206, line 236: typo -- should be `additionalProperties`

source/shared/cpp/ObjectModel/Enums.cpp
```diff
@@ -128,6 +128,7 @@ namespace AdaptiveSharedNamespace
             {AdaptiveCardSchemaKey::Requires, "requires"},
             {AdaptiveCardSchemaKey::RichTextBlock, "richTextBlock"},
             {AdaptiveCardSchemaKey::Right, "right"},
+            {AdaptiveCardSchemaKey::Schema, "$schema"},
```
paulcam206, line 131: have you verified that the existing shared model tests pass? also, we should probably have a card-level additional properties test...

paulcam206 (changes requested):

source/shared/cpp/ObjectModel/Enums.cpp
jwoo-msft: The shared model tests are passed & new tests for card level additional properties are added to  `AdditionalPropertiesTest.cpp`

paulcam206 (approved):

PR #5449, opened 2021-03-02
[JS] Action.mode designer support + promote secondary action to primary on execute

dclaux:
## Related Issue
Finalizes implementation of https://github.com/microsoft/AdaptiveCards/issues/4715

## Description
- Adds designer support for schema "1.5 (Preview)"
- Adds designer support for `Action.mode`
- Adds secondary action promotion to primary on execution. It works this way:
  - Actions are only promoted if there is at least one spot they can be promoted to. That isn't the case when all actions in a set are secondary (and thus only the overflow button is displayed)
  - When promoted, an action replaces the last primary action (the one right before the overflow button) and the action it replaces is moved to the overflow menu at the old spot of the newly promoted action
  - Secondary ShowCard actions are always promoted
  - Secondary non-ShowCard actions are never promoted
  - Primary non-ShowCard actions that happen to have been demoted are promoted again when clicked

## Sample Card
```json
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "TextBlock",
            "size": "Medium",
            "weight": "Bolder",
            "text": "ActionSet overflow test card"
        }
    ],
    "actions": [
        {
            "type": "Action.OpenUrl",
            "title": "Primary OpenUrl",
            "url": "http://adaptivecards.io"
        },
        {
            "type": "Action.OpenUrl",
            "title": "Secondary OpenUrl",
            "url": "http://adaptivecards.io",
            "mode": "secondary"
        },
        {
            "type": "Action.ShowCard",
            "title": "Secondary Action.ShowCard #1",
            "card": {
                "type": "AdaptiveCard",
                "body": [
                    {
                        "type": "TextBlock",
                        "text": "ShowCard #1",
                        "size": "ExtraLarge",
                        "horizontalAlignment": "Center"
                    }
                ]
            },
            "mode": "secondary"
        },
        {
            "type": "Action.ShowCard",
            "title": "Secondary Action.ShowCard #2",
            "card": {
                "type": "AdaptiveCard",
                "body": [
                    {
                        "type": "TextBlock",
                        "text": "ShowCard #2",
                        "size": "ExtraLarge",
                        "horizontalAlignment": "Center"
                    }
                ]
            },
            "mode": "secondary"
        }
    ],
    "version": "1.5"
}
```

## How Verified

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5449)

### DIFF
--- a/source/nodejs/adaptivecards-designer-app/dist/containers/bf-image-container.css
+++ b/source/nodejs/adaptivecards-designer-app/dist/containers/bf-image-container.css
@@ -154,3 +154,102 @@
 .ac-inlineActionButton:active {
     background-color: #CCCCCC;
 }
+
+/* Popup menu */
+
+.ac-ctrl-overlay {
+    position: absolute;
+    left: 0;
+    top: 0;
+    z-index: 10000;
+}
+  
+@keyframes fadeIn {
+    0% {
+        opacity: 0;
+    }
+}
+  
+@keyframes slideTopToBottom {
+    0% {
+        transform: translate(0, -20px);
+    }
+}
+  
+@keyframes slideBottomToTop {
+    0% {
+        transform: translate(0, 20px);
+    }
+}
+  
+@keyframes slideLeftToRight {
+    0% {
+        transform: translate(-20px, 0);
+    }
+}
+  
+@keyframes slideRightToLeft {
+    0% {
+        transform: translate(20px, 0);
+    }
+}
+  
+.ac-ctrl-slide {
+    animation-duration: 0.5s;
+    animation-timing-function: cubic-bezier(0.1, 0.9, 0.2, 1);
+}
+  
+.ac-ctrl-slide.ac-ctrl-slideTopToBottom {
+    animation-name: fadeIn, slideTopToBottom;
+}
+  
+.ac-ctrl-slide.ac-ctrl-slideBottomToTop {
+    animation-name: fadeIn, slideBottomToTop;
+}
+  
+.ac-ctrl-slide.ac-ctrl-slideLeftToRight {
+    animation-name: fadeIn, slideLeftToRight;
+}
+  
+.ac-ctrl-slide.ac-ctrl-slideRightToLeft {
+    animation-name: fadeIn, slideRightToLeft;
+}
+  
+.ac-ctrl {
+    box-sizing: border-box;
+    font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
+    font-size: 14px;
+    font-weight: normal;
+}
+  
+.ac-ctrl-dropdown-item {
+    width: 100%;
+    text-overflow: ellipsis;
+    overflow: hidden;
+    white-space: nowrap;
+    display: block;
+    padding: 0px 12px 0px 12px;
+    line-height: 30px;
+    cursor: pointer;
+}
+  
+.ac-ctrl-dropdown-item:hover {
+    background-color: #EEEEEE;
+}
+  
+.ac-ctrl-dropdown-item:focus {
+    outline: 0;
+    background-color: #CCCCCC;
+}
+  
+.ac-ctrl-popup-container {
+    border: 1px solid #EEEEEE;
+    background-color: white;
+    position: absolute;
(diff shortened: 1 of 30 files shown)


golddove:
> Actions are only promoted is there is at least one spot they can be promoted to. That isn't the case when all actions in a set are secondary (and thus only the overflow button is displayed)

In this case, what happens when a ShowCard is opened?

dclaux:
No button is shown. While not ideal, it's going to be a corner case, and there's no real confusion because there are no other buttons anyway.

source/nodejs/adaptivecards-designer-app/dist/containers/bf-image-container.css
```diff
@@ -153,4 +153,103 @@
...
+    background-color: #CCCCCC;
+}
+  
+.ac-ctrl-popup-container {
+    border: 1px solid #EEEEEE;
+    background-color: white;
+    position: absolute;
+    box-shadow: 0 0 15px -5px rgba(0, 0, 0, 0.4);
+    overflow-y: auto;
+}
```
paulcam206, line 251: should double-check that these classes have good behavior in High Contrast (here and elsewhere)

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3950,6 +3978,10 @@ export abstract class Action extends CardObject {
         return this.internalValidateInputs(this.getReferencedInputs());
     }
 
+    get shouldPromoteAsPrimaryOnExecute(): boolean {
+        return this.mode === "primary";
+    }
+
     get isPrimary(): boolean {
```
paulcam206, line 3985: is this getter still accurately named? should we be checking `mode` here? (entirely possible that I'm missing something)

source/nodejs/adaptivecards/src/controls/constants.ts
```diff
@@ -1,34 +1,5 @@
...
-export const KEY_UP = 38;
-/**
- * @deprecated Use Constants.keys.down instead.
- */
-export const KEY_DOWN = 40;
-/**
- * @deprecated Use Constants.keys.delete instead.
- */
-export const KEY_DELETE = 46;
-
```
paulcam206, line 31: breaking change?

paulcam206 (changes requested):

source/nodejs/adaptivecards/src/card-elements.ts
paulcam206: (in case it's not clear, I'm referring to `isPrimary()`)

source/nodejs/adaptivecards-designer-app/dist/containers/bf-image-container.css
dclaux: No effort was made here. These classes are copied straight from the designer (with a different prefix)

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: `isPrimary` is something that's existed in the JS API for a while, and is unrelated. Outlook uses it, I can't rename it unfortunately. So it's not great, but I have no better option other than to suggest different names for the mode values. Maybe "main" and "secondary"

source/nodejs/adaptivecards/src/controls/constants.ts
dclaux: No. Thiese constants still exist in the adaptivecards-controls library. They didn't need to exist here.

source/nodejs/adaptivecards/src/card-elements.ts
paulcam206: perhaps just some docs on `isPrimary()` to clarify its purpose?

source/nodejs/adaptivecards/src/controls/constants.ts
dclaux: To be clearer, the PopupMenu class and everything related to was brought into the adaptivecards library. The controls library keeps its own copy.

paulcam206 (approved):

PR #5540, opened 2021-03-12
[dotnet] Added AEL simple object memory interface

### LINKED ISSUE #5518: [.Net] [Unicode literals are escaped in if statements]
# Platform

- [ ] .NET

# Host

The host is a bot framework solution sending messages to Teams and emails to Outlook using the appropriate bot framework channels.

# Version of SDK

AdaptiveCards.Templating v1.1.0

# Details

I have the following Textblock in a template (it happens with AdaptiveCard version 1.0 - 1.2, haven't tried 1.3). Note the if statement, where the return values contain unicode literals..

```
{
      "type": "TextBlock",
      "text": "${if(IntTemplateProperty >= 0, IntTemplateProperty + ' % \u25b2', IntTemplateProperty + '%  \u25bc')}"
}
```

When I use the AdaptiveCards.Templating.AdaptiveCardTemplate.Expand method to populate the text property with the code below, it escapes the unicode literal string and sets text property *5% \\\\u25b2*. This means that when the card is sent (I've tried to Outlook and Teams) the value displays as *5% \\u25b2* rather than *5% ▲* as I would expect.

```
var model = new
  {
      IntTemplateProperty = 5
  };
var template = new AdaptiveCardTemplate(scenarioTemplate.Template);
var card = template.Expand(model);
```

But when you use the unicode literal as a static value, as in the example below, it displays correctly.

```
{
      "type": "TextBlock",
      "text": "5% \u25b2"
}
```

In the template I am using the literal ( *\u25b2* ) rather than the characters it self ( *▲* ) because the later doesn't render when sent to Outlook via bot framework email channel.

jwoo-msft:
## Related Issue
Fixed #5518 

## Description
Expansion of the template is done in the context of serialized json payload,
However, the memory that bounds to the template are in deserialized form.
In the previous implementation, we serialized the expanded string returned from AEL. This can lead to double serialization as observed in #5518.  Only data that is bound to the template is serialized, and the template portion is left untouched in the new implementation.

## How Verified
1. A new unit test is added.
2. Unit tests are run & verified.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5540)


source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateSimpleObjectMemory.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    /// <summary>
+    /// Implements AEL's IMemory interface
+    /// </summary>
+    public class AdaptiveCardsTemplateSimpleObjectMemory : IMemory 
+    {
+        private SimpleObjectMemory simpleObjectMemory;
+        /// <summary>
+        /// Simple implementation of AEL.Memory.IMemory
+        /// It serves as an entry points to additional customazation and optimazation
```
paulcam206, line 17: minor nit

```suggestion
        /// It serves as an entry points to additional customization and optimization
```

paulcam206 (approved):

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateSimpleObjectMemory.cs
jwoo-msft: Thanks, Paul. PR got auto merged. I will make this change.

PR #5581, opened 2021-03-25
[Android] Improve right-to-left locale support, add rtl override flag support

### LINKED ISSUE #5490: [Android] Renderer support for RTL property

golddove:
## Related Issue
Fixes #5490 

## Description
Fixes existing issues with right-to-left support, and respects the `rtl` flag (if present) to override inherited layout direction.

## Outstanding concerns

ImageSet ordering of its images doesn't respect RTL. Fix pending...

By default, the Android platform seems to prefer using the layout direction (typically set by the user's device language) to affect alignment and ordering of everything except text. For text, the platform seems to prefer inferring the text alignment and ordering based on the content (using the Unicode first strong directional character algorithm), allowing some text to be right aligned and other to be left aligned.

This PR partially overrides that default to set text alignment based on the layout direction (derived from device language, or overridden by `rtl` property), but character/word ordering continues to be determined by Unicode heuristics.

This seems to be closer to JS behavior, but may not be what is ultimately desired by users (especially in the common case of a plain English card shown to a bilingual user who has their device set to a right-to-left, where they may expect the default platform heuristics to apply). 

## Sample Card
To be added

## How Verified
Switched test device to RTL.
image
Modified some existing sample cards to force LTR on certain containers.
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5581)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/InputUtil.java
```diff
@@ -53,6 +53,7 @@ public static TextView RenderInputLabel(String label, boolean isRequired, Contex
 
         TextView labelView = new TextView(context);
         labelView.setText(paragraph);
+        labelView.setTextAlignment(View.TEXT_ALIGNMENT_VIEW_START);
```
jwoo-msft, line 56: questions. why is the text alignment set without consulting the rtl?

jwoo-msft: i found my answer looking at the changes

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ContainerRenderer.java
```diff
@@ -136,6 +137,7 @@ public static void ApplyBleed(CollectionTypeElement collectionElement, ViewGroup
         {
             int padding = Util.dpToPixels(context, hostConfig.GetSpacing().getPaddingSpacing());
             ViewGroup.MarginLayoutParams layoutParams = (ViewGroup.MarginLayoutParams) collectionElementView.getLayoutParams();
+            // TODO: Check RTL support
```
jwoo-msft, line 140: layoutParam doesn't have the direction neutral params for setting direction like view_start?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
```diff
@@ -221,6 +221,7 @@ private static ConstraintSet createConstraints(Context context, ImageView imageV
         // Start with base constraints from XML layout
         constraints.clone(context, R.layout.image_constraint_layout);
 
+        // TODO: Check RTL support
```
jwoo-msft, line 224: you may want to create an issue and include the # in this comment to track.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ColumnRenderer.java
```diff
@@ -178,6 +178,7 @@ public View render(
 
         ContainerRenderer.ApplyPadding(styleForThis, renderArgs.getContainerStyle(), columnLayout, context, hostConfig);
         ContainerRenderer.ApplyBleed(column, columnLayout, context, hostConfig);
+        BaseCardElementRenderer.applyRtl(column.GetRtl(), columnLayout);
```
jwoo-msft, line 181: I don't see codes percolating parent's rtl attribute down to children.

jwoo-msft (changes requested):

jwoo-msft (commented):
looks good

paulcam206 (approved):

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ContainerRenderer.java
golddove: Checked, works

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
golddove: Fixed

PR #5589, opened 2021-03-26
[JS] Handle explicit rtl: false

golddove:
## Related Issue
Untracked. JS is only respecting explicit rtl: false if a parent Element has rtl: true.

## Description
This explicitly sets or unsets the dir property. Additionally, allows the browser to set text alignment based on the inherited layout direction, rather than using the card tree to determine alignment.

## Sample Card
`v1.5\Tests\AdaptiveCard.Rtl.unset`
`v1.5\Tests\AdaptiveCard.Rtl.True`
`v1.5\Tests\AdaptiveCard.Rtl.False`

## How Verified

`v1.5\Tests\AdaptiveCard.Rtl.unset`, `document.dir=ltr`:
image

`v1.5\Tests\AdaptiveCard.Rtl.unset`, `document.dir=rtl`:
image

`v1.5\Tests\AdaptiveCard.Rtl.True`, `document.dir=ltr`:
image

`v1.5\Tests\AdaptiveCard.Rtl.True`, `document.dir=rtl`:
image

`v1.5\Tests\AdaptiveCard.Rtl.False`, `document.dir=ltr`:
image

`v1.5\Tests\AdaptiveCard.Rtl.False`, `document.dir=rtl`:
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5589)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -5462,12 +5462,8 @@ export class Container extends StylableCardElementContainer {
 
         let element = document.createElement("div");
 
-        if (this.rtl === undefined) {
-            element.removeAttribute("dir");
-        } else if (this.rtl) {
-            element.dir = "rtl";
-        } else {
-            element.dir = "ltr";
+        if (this.rtl !== undefined) {
+            element.dir = this.rtl ? "rtl" : "ltr";
         }
 
         element.classList.add(hostConfig.makeCssClassName("ac-container"));
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -1082,18 +1082,15 @@ export class TextBlock extends BaseTextBlock {
     applyStylesTo(targetElement: HTMLElement) {
         super.applyStylesTo(targetElement);
 
-        let parentContainer = this.getParentContainer();
-        let isRtl = parentContainer ? parentContainer.isRtl() : false;
-
         switch (this.horizontalAlignment) {
             case Enums.HorizontalAlignment.Center:
                 targetElement.style.textAlign = "center";
                 break;
             case Enums.HorizontalAlignment.Right:
-                targetElement.style.textAlign = isRtl ? "left" : "right";
+                targetElement.style.textAlign = "end";
                 break;
             default:
-                targetElement.style.textAlign = isRtl ? "right" : "left";
+                targetElement.style.textAlign = "start";
                 break;
         }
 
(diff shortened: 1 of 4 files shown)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -1082,18 +1082,15 @@ export class TextBlock extends BaseTextBlock {
...
-        let parentContainer = this.getParentContainer();
-        let isRtl = parentContainer ? parentContainer.isRtl() : false;
-
         switch (this.horizontalAlignment) {
             case Enums.HorizontalAlignment.Center:
                 targetElement.style.textAlign = "center";
                 break;
             case Enums.HorizontalAlignment.Right:
-                targetElement.style.textAlign = isRtl ? "left" : "right";
+                targetElement.style.textAlign = "end";
```
dclaux, line 1090: That's not supported in IE... but maybe at this point we shouldn't worry anymore.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5468,8 +5462,12 @@ export class Container extends StylableCardElementContainer {
 
         let element = document.createElement("div");
 
-        if (this.rtl !== undefined && this.rtl) {
+        if (this.rtl === undefined) {
+            element.removeAttribute("dir");
```
dclaux, line 5466: Explicitly removing the attribute shouldn't be necessary, the attribute isn't set by default on newly created elements.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5468,8 +5462,12 @@ export class Container extends StylableCardElementContainer {
...
+        } else if (this.rtl) {
```
dclaux, line 5467: Style: else on new line

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5468,8 +5462,12 @@ export class Container extends StylableCardElementContainer {
...
             element.dir = "rtl";
+        } else {
```
dclaux, line 5469: Style: else on new line

dclaux (changes requested):

jwoo-msft (approved):

source/nodejs/adaptivecards/src/card-elements.ts
golddove: The problem is that the root card can be unset, so I believe we'd have to compute directionality by looking at the card's ancestors all the way up to `document.dir`.

source/nodejs/adaptivecards/src/card-elements.ts
golddove: Fixed

source/nodejs/adaptivecards/src/card-elements.ts
golddove: Refactored

source/nodejs/adaptivecards/src/card-elements.ts
golddove: Refactored

dclaux (approved):

PR #5605, opened 2021-03-31
HostConfig, StyleConfig and ThemeConfig at Card level

vivekvijayakrishnan:
HostConfig, StyleConfig, and ThemeConfig are handled at the Adaptive Card level instead of for a particular host application. This enables developers to use different configs for each card rather than one set of configs for a host.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5605)

### DIFF
--- a/source/community/reactnative/src/components/actions/select-Action.js
+++ b/source/community/reactnative/src/components/actions/select-Action.js
@@ -12,7 +12,6 @@ import {
 } from '../../utils/context';
 import * as Constants from '../../utils/constants';
 import * as Utils from '../../utils/util';
-import { HostConfigManager } from '../../utils/host-config';
 
 export class SelectAction extends React.Component {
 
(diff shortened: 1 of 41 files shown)


IbrahimSulai (approved):
LGTM

source/community/reactnative/src/components/actions/select-Action.js
```diff
@@ -58,7 +57,7 @@ export class SelectAction extends React.Component {
 	}
 
 	render() {
-		if (HostConfigManager.getHostConfig().supportsInteractivity === false) {
+		if (!this.props.configManager.hostConfig.supportsInteractivity) {
```
pragadeeshk, line 60: Please confirm supportsInteractivity is not true by default. In such case, this would fail.

vivekvijayakrishnan: I think **`supportsInteractivity`** should be true by default, only then the `SelectAction` component will get rendered. If it is false then we won't be using the select action and it will return `null`.

pragadeeshk: Ok. As you said, i just checked supportsInteractivity is set to true on undefined case. This looks good to me now.

pragadeeshk:
LGTM too

pragadeeshk (approved):

PR #5612, opened 2021-04-01
[UWP][Shared] Added host config and UWP renderer support for heading style

### LINKED ISSUE #5477: [UWP] Renderer support for TextBlock headings

### LINKED ISSUE #5478: [.NET] Parser support for TextBlock headings

### LINKED ISSUE #5479: [WPF] Renderer support for TextBlock headings
Narrator supports navigation by headings. See [support article](https://support.microsoft.com/en-us/windows/chapter-5-navigation-68941680-3245-6ef5-5012-0674b8b6fc59) that describes how to cycle through navigation modes.

Text with `style: heading` receive the appropriate accessibility flags on other platforms to support this mode of navigation, including UWP XAML. However, the .NET WPF renderer doesn't seem to offer this support (see details in #5626).

Consult with accessibility team to determine if there is a workaround, or close this as external.

RebeccaAnne:
## Related Issue
Fixes #5477 

## Description

- Added Heading level support to host config in shared model and uwp
- Added uwp renderer support by setting the appropriate heading level for textblocks with heading style

## Sample Card
samples\v1.5\Elements\TextBlock.Style.json

## How Verified

- Ran UWP and Shared model unit tests.
- Confirmed narrator read appropriate heading level using card referenced above.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5612)


source/uwp/Renderer/lib/AdaptiveTextBlockRenderer.cpp
```diff
@@ -89,4 +115,52 @@ namespace AdaptiveNamespace
             jsonObject, elementParserRegistration, actionParserRegistration, adaptiveWarnings, element);
     }
     CATCH_RETURN;
+
+    HRESULT AdaptiveTextBlockRenderer::GetHeadingLevelFromContext(ABI::AdaptiveNamespace::IAdaptiveRenderContext* renderContext,
+                                                                  ABI::Windows::UI::Xaml::Automation::Peers::AutomationHeadingLevel* headingLevel)
+    {
+        ComPtr<IAdaptiveHostConfig> hostConfig;
+        RETURN_IF_FAILED(renderContext->get_HostConfig(&hostConfig));
```
jwoo-msft, line 123: nit: have you considered moving this method to RenderContext since it owns the host config, and we don't have to pass in the render context?

source/uwp/Renderer/lib/AdaptiveTextBlockRenderer.cpp
```diff
@@ -89,4 +115,52 @@ namespace AdaptiveNamespace
...
+            break;
+        case 7:
+            *headingLevel = AutomationHeadingLevel_Level7;
+            break;
+        case 8:
+            *headingLevel = AutomationHeadingLevel_Level8;
+            break;
+        case 9:
+        default:
+            *headingLevel = AutomationHeadingLevel_Level9;
```
jwoo-msft, line 160: nit: from the discussion, the reasonable default value seems to be level 2.

jwoo-msft (approved):

source/uwp/Renderer/lib/AdaptiveTextBlockRenderer.cpp
RebeccaAnne: Hmm.... So the default coming from the host config definition will be 2, so if they don't put anything that's what they'll get. But this is the scenario where someone has explicitly put like 11 or something in their host config. My thinking is if they put a really big number in we should use the biggest one we have available.

source/uwp/Renderer/lib/AdaptiveTextBlockRenderer.cpp
RebeccaAnne: That's not a bad idea, except the IAdaptiveRenderContext interface is public. I don't really want to put this there i don't think since it feels like an internal implementation detail rather than something we want to commit to the public API. I could put it on the implementation class only and use "Peek" to get at it, but at that point i think it's probably more trouble than it's worth.

---
In reply to: [606039737](https://github.com/microsoft/AdaptiveCards/pull/5612#discussion_r606039737) [](ancestors = 606039737)

PR #5616, opened 2021-04-05
[Schema] Add RTL property to schema

### LINKED ISSUE #5614: [RTL][Schema] Update the schema to reflect the new RTL Property

RebeccaAnne:
## Related Issue
Fixed #5614

## Description
Updated schema source to include new rtl property on columns and containers and regenerated specs and schema.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5616)

### DIFF
--- a/schemas/src/elements/Column.json
+++ b/schemas/src/elements/Column.json
@@ -1,73 +1,78 @@
 {
-  "$schema": "https://raw.githubusercontent.com/microsoft/AdaptiveCards/6f39aedce45864ae1067ed44a5551dc973790bb5/source/nodejs/typed-schema/schema/lib/Type.json",
-  "extends": "ToggleableItem",
-  "description": "Defines a container that is part of a ColumnSet.",
-  "properties": {
-    "items": {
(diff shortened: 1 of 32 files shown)


schemas/src/elements/Column.json
```diff
@@ -1,73 +1,78 @@
...
+				"50px"
+			],
+			"version": "1.2",
+			"features": [
+				2293
+			]
+		},
+		"rtl": {
+			"type": "boolean",
+			"description": "When `true` content in this column should be presented right to left.",
```
paulcam206, line 47: Is it worth calling out the difference between `false` and "not set"?

paulcam206 (approved):
only one minor piece of feedback

schemas/src/elements/Column.json
RebeccaAnne: Good point. And this should probably be boolean? rather than just boolean. Added updated text.

PR #5622, opened 2021-04-07
Handled the padding value as zero in the host configuration and Added…

IbrahimSulai:
This PR contains the below changes:
1. Handled the padding value as zero in the host configuration
2. Added the condition to avoid separator for the first element

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5622)


vivekvijayakrishnan (approved):
LGTM

PR #5669, opened 2021-04-14
[Site] Globalization support for adaptivecards.io

### LINKED ISSUE #5649: [Site] Glob support
Tracking item for glob/loc support for the website

paulcam206:
## Related Issue
Fixes #5649 

## Description
Adds globalization support for our website. Note that this does not add localization to the site, that will be a future endeavour. Also note that for the sake of compartmentalizing changes, this PR does not address Designer globalization -- that will come in a PR later this week.

### Screenshots
*Site depicted with localized strings prepended with `"(en)"` -- prepended text not actually checked in here*

image

image

image

image

### How to localize content
Because of how our site is generated, it's a little complicated to localize strings -- where you make changes depends on what portion of the site you need to affect:
* Schema-generated content is generated by `ac-typed-schema` and has localizable strings in `source/nodejs/ac-typed-schema/src/markdown/languages/[locale].json`
  * Chosen because the `i18n` node module is a popular option and supports sprintf-style strings
  * Should require no manual intervention at the code level
  * JSON files automatically generated/used at build time
  * Standard loc structure:
```json
{
    "This is my string": "This is the string that is applicable to the current file's locale"
}
```
* Broader site content like the home page, header/footer, and layouts 
  * Couldn't use hexo's builtin glob support (via `hexo-i18n`), because it's not readily available for use in themes, which is what we use to generate our site
  * Didn't want to use `i18n` module because it would duplicate work between our site and our theme
  * Localized strings go into `source/nodejs/adaptivecards-site/pages/_data/[something].yml`
  * Hand-rolled yaml structure:
```yaml
en:
  string_id: The string to use
  some_object:
    - free-form-field: blerg
      thing: something
    - free-form-field: blarg
      thing: some other thing
sp:
  string_id: The string to use (en Español)
  # etc.
```
  * Reference in `ejs` files:
    * If in hexo layouts, reference e.g. `site.data.footer.en.some_string`
    * If in adaptivecards theme, reference e.g. `locals.data.footer.en.some_thing`

### Work left to localize
* Need to generate outputs for all languages (currently just emits `en`, though I've tested manually changing the locale)
* Need to investigate automatic language redirection on the site (probably through a change to the [web.config](https://github.com/microsoft/AdaptiveCards/blob/main/source/nodejs/adaptivecards-site/themes/adaptivecards/source/web.config))
* Unclear if we need to support localized blog posts. Shouldn't be difficult to do, but will require a little work
* Need localized samples in order to stand up a localized Samples tab (localized beyond just boilerplate content)

## How Verified
* local build

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5669)


source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
```diff
@@ -1,15 +1,46 @@
...
 import { SchemaLiteral } from "../SchemaLiteral";
 import { SchemaProperty } from "../SchemaProperty";
 import { SchemaPropertyType } from "../SchemaPropertyType";
+import * as path from "path";
+import * as console from "console";
 import * as mdTable from "markdown-table";
 import * as style from "./style";
 
+class MarkdownConfig {
+    i18n : I18n;
```
jonmill, line 17: Should this be `private`?

source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
```diff
@@ -1,15 +1,46 @@
...
+    constructor() {
+        this.i18n = new I18n();
+        this.i18n.configure({
+            locales: ['en', 'de', 'sp'],
+            directory: path.join(__dirname, '../../src/markdown/languages'),
+            syncFiles: true
+        });
+    }
+
+    _locale : string = "en";
```
jonmill, line 28: Should this be declared up top with the other variable? Also, can it be initialized by the URL or default browser language?

source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
```diff
@@ -21,7 +52,7 @@ export function createPropertiesSummary(classDefinition: SchemaClass, knownTypes
 		properties = sortProperties(properties);
 
 		if (includeVersion && defined(elementVersion) && elementVersion != "1.0") {
-			md += "**Introduced in version " + elementVersion + "**\n\n";
+			md += __("**Introduced in version %s**\n\n", elementVersion);
```
jonmill, line 55: In the future, this hardcoded string (and all others) will be replaced by localized resources right? But that is out-of-scope for this PR

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/_partial/header.ejs
```diff
@@ -51,43 +51,43 @@
                                 <nav id="uhf-g-nav" class="c-uhfh-gnav" role="presentation" data-m="{&quot;cN&quot;:&quot;Category nav_cont&quot;,&quot;cT&quot;:&quot;Container&quot;,&quot;id&quot;:&quot;c8c2m1r1a1&quot;,&quot;sN&quot;:8,&quot;aN&quot;:&quot;c2m1r1a1&quot;}">
                                     <ul class="js-paddle-items">
                                         <li class="single-link js-nav-menu x-hidden-none-mobile-vp">
-                                            <a class="c-uhf-nav-link" href="<%- url_for('/') %>" data-m="{&quot;cN&quot;:&quot;CatNav_Home_nav&quot;,&quot;id&quot;:&quot;n1c8c2m1r1a1&quot;,&quot;sN&quot;:1,&quot;aN&quot;:&quot;c8c2m1r1a1&quot;}">Home</a>
+                                            <a class="c-uhf-nav-link" href="<%- url_for('/') %>" data-m="{&quot;cN&quot;:&quot;CatNav_Home_nav&quot;,&quot;id&quot;:&quot;n1c8c2m1r1a1&quot;,&quot;sN&quot;:1,&quot;aN&quot;:&quot;c8c2m1r1a1&quot;}"><%= site.data.banner.en.home %></a>
```
jonmill, line 54: I suppose the same question applies here, just to be safe - this hardcoding of `en` will be replaced later when we have localized strings...is that correct?

jonmill (changes requested):
Just a couple of context-clarifying questions from me, looks great!

source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
paulcam206: need access to allow translation to be called from `adaptivecards-site` -- it's probably worth noting that this is a private module that's only used during generation of site content. :)

source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
paulcam206: should be private as well -- I'll fix it up :)

source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
paulcam206: the way this library works is that you use a string literal that's the key for the translation backend. effectively, you write your hardcoded strings in your native language and use those as the default translations. under the hood, the library will use the locale-appropriate string as specified.

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/_partial/header.ejs
paulcam206: that's correct -- as noted in the description, there's still a little work left to do per-locale generation/detection. I didn't want to ship that on the site publically, because it'd be weird if we had e.g. https://adaptivecards.io/en-us/explorer, but no https://adaptivecards.io/es-es/explorer

source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
paulcam206: perhaps private with a public getter would be more appropriate :)

source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
jonmill: Cool, thanks for clarifying!

jonmill (approved):
LGTM!

PR #5685, opened 2021-04-17
[Designer] Add i18n support to designer

paulcam206:
## Description
Add i18n support to designer. Not complete yet (was going through files alphabetically and stopped at `designer-peers.ts`).

@dclaux - before I got too far, I wanted to see what you thought about this approach. I think I want to restructure the translation file a bit, but haven't done that yet. Also, I know that this isn't the framework we talked about, but `i18n` isn't suited for browser apps. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5685)

### DIFF
--- a/source/nodejs/adaptivecards-designer/assets/languages/en/designer.json
+++ b/source/nodejs/adaptivecards-designer/assets/languages/en/designer.json
@@ -1,0 +1,83 @@
+{
+	"cardElements": "CARD ELEMENTS",
+	"cardEditor": "CARD PAYLOAD EDITOR",
+	"cardPreview": "Card Preview",
+	"sampleEditor": "SAMPLE DATA EDITOR",
+	"sampleData": "Sample data",
+	"elementProperties": "ELEMENT PROPERTIES",
+	"ellipsis": "...",
+	"dataBinding": "Data Binding",
+	"cardStructure": "CARD STRUCTURE",
+	"dataStructure": "DATA STRUCTURE",
+	"loadingEditor": "Loading editor...",
+	"expandCollapseAria": "{{action}} {{label}} element",
+	"collapse": "collapse",
+	"expand": "expand",
+	"remove": "Remove",
+	"dragElement": "Drag to move this element",
+	"returnToDesign": "Return to Design mode",
+	"switchToPreview": "Switch to Preview mode",
+	"noStructure": "The Card Structure isn't available in Preview mode.",
+	"noSelection": "**Nothing is selected**",
+	"selectAnElement": "Select an element in the card to modify its properties.",
+	"clickToSelect": "Click to select this element",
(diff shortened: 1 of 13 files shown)


dclaux:
I've been thinking more about this, and my conviction is that the designer should not take a direct dependency on i18n nor any other external localization library for that matter. I strongly believe the designer, as a component, should:
- Continue to provide default, out of the box strings (the English strings we already have) without the developer having to do anything including hosting a file
- But at the same time also provide a way for host applications to plug in their preferred localization system

Overall I'm OK with the resource JSON as you have it even though neither the string identifiers nor the parameters are discoverable via IntelliSense. I think it's acceptable in the name of flexibility.

I think we should have a version of that JSON built into our code, and we should have our own built-in logic to read string values from it. Converting a dotted path (say "newCard.dialog.title") into a field accessor isn't difficult, nor is the logic for simple parameter substitution like "{{action}}" (we already do it in the renderer for DATE and TIME functions.)

Beyond that, we can allow a host application to provide its own "localization provider" so they can use their preferred system or the one they already use for the rest of their app.

So I suggest we do something along these lines:

```typescript
export var strings = {
	"cardElements": "CARD ELEMENTS",
	"cardEditor": "CARD PAYLOAD EDITOR",
	"cardPreview": "Card Preview",
	"sampleEditor": "SAMPLE DATA EDITOR",
	"sampleData": "Sample data",
	"elementProperties": "ELEMENT PROPERTIES",
	"ellipsis": "...",
        etc.
};

export abstract class LocalizationProvider {
    abstract localize(identifier: string, parameters: object): string;
}

export var localizationProvider?: LocalizationProvider;

export function localizeString(identifier: string, parameters: object): string {
    // If the host app has defined a custom localization provider, use it
    if (localizationProvider) {
        return localizationProvider.localize(identifier, parameters);
    }

    // Otherwise, use our built-in localization
    return <run our own logic to read the value from the global strings variable>;
}
```

The designer package would not have any dependency on i18n (not even external). The designer code would call `localizeString` exclusively:
```typescript
new containers.WebChatContainer(localizeString("hostContainer.names.webChat"), "containers/webchat-container.css"),
```

For the site, if you deem it practical, you'd simply implement an i18n provider:

```typescript
class i18nProvider extends LocalizationProvider {
    localize(identifier: string, parameters: object): string {
        return __(identifier, parameters);
    }
}

localizationProvider = new i18nProvider();
```

Only the site code would have a dependency on i18n.

What do you think?

jonmill:
Bump to reset staleness

jonmill:
@paulcam206 - is this PR still desired or should it be closed?

paulcam206:
this should probably be resurrected in the near future, but the PR can be closed for now

PR #5699, opened 2021-04-20
[iOS] Added workaround for apple's bug that turns superview's visibility to hidden

### LINKED ISSUE #5234: [iOS][is Visible] [is visibility property is broken for text block]
# Platform

- [ ] iOS

# Author or host

 Microsoft Teams

# Version of SDK

2.4.5 (latest)

# Details

If text block is the first element and isVisible is set to false then card won't get rendered in iOS. Ideally we should not block rendering if isVisible if false for the first element. 
Verified the same for desktop card and rest of the card elements rendered fine.
```json
{
   "type":"AdaptiveCard",
   "version":"1.0",
   "body":[
      {
               "type":"TextBlock",
               "text":"Hlo1",
               "id":"id11",
               "color":"Dark",
               "wrap":true,
               "isVisible":false
      },
      {
               "type":"TextBlock",
               "text":"Hi",
               "id":"id22",
               "color":"Dark",
               "wrap":true,
               "isVisible":true
      }
   ],
   "schema":null,
   "$schema":"http://adaptivecards.io/schemas/adaptive-card.json",
   "minHeight":"74px"
}
```

jwoo-msft:
## Related Issue
Fixed #5234 

## Description
If the visibility of a first view to be added to UIStackView hidden, its superview's visibility is set hidden regardless of the visibility of the rest of the subviews added later by the system. This PR adds a workaround where it toggles back the first view's visibility if the visibility is set to hidden.
## Sample Card
```
{
    "type":"AdaptiveCard",
    "version":"1.0",
    "body":[
        {
            "type":"TextBlock",
            "text":"Hlo1",
            "id":"id11",
            "color":"Dark",
            "wrap":true,
            "isVisible":false
        },
        {
            "type":"TextBlock",
            "text":"Hi",
            "id":"id22",
            "color":"Dark",
            "wrap":true,
            "isVisible":true
        },
        {
            "type":"TextBlock",
            "text":"Hi",
            "id":"id23",
            "color":"Dark",
            "wrap":true,
            "isVisible":true
        }
    ],
    "schema":null,
    "$schema":"http://adaptivecards.io/schemas/adaptive-card.json",
    "minHeight":"74px"
}

```
### Before
image

### After
image

Visually verified fix.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5699)


RebeccaAnne (approved):
:shipit:

RebeccaAnne:
Do we know if Apple is aware of this issue? Is there any way to report it to them?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRContentStackView.mm
```diff
@@ -578,4 +578,12 @@ - (void)dealloc
     }
 }
 
+- (void)toggleVisibilityOfFirstView
+{
+    if ([_stackView.subviews count] and _stackView.subviews[0].hidden) {
+        _stackView.subviews[0].hidden = NO;
+        _stackView.subviews[0].hidden = YES;
+    }
```
jonmill, line 586: This seems...hacky. Why is it necessary to toggle it this way? Is there no alternative?

jonmill (changes requested):
One quick question for you

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRContentStackView.mm
jwoo-msft: It appears to be an Apple bug. iOS has a UIStackView which is similar to a Grid. If a view is added to the empty UIStackView for layout, The UIStackView will become invisible if the view is invisible. This is expected behavior since all of its subviews are invisible although the subviews count is 1. However, what is not expected is that when we subsequently add a visible view to the UIStackView, the UIStackView won't turn its visibility visible. We could have made UIStackView visible but it requires us to check all the subviews of the UIStackView to make sure that there is at least one subview that is visible. I feel that this is the job of the system. Although it looks hacky, I think letting the system decide the visibility looked better alternative.

jonmill (approved):
LGTM, thanks for clarifying!

PR #5704, opened 2021-04-21
[Android] Always apply configured Actions.IconSize

### LINKED ISSUE #5567: [Android][iconSize is not honoured for iconUrl using resource resolver]
# Platform

- [ ] Android

# Author or host

Host - Cisco WebEx

Cisco Webex

# Version of SDK

latest main

# Details

The `iconSize` host config field for `iconUrl` (inline) is not being honoured when resource resolver is used.

golddove:
## Related Issue
Fixes #5567 

## Description
Fixes Android renderer to match JS behavior of applying HostConfig.ActionsConfig.IconSize always, not just when placement is above title. Previously, we were setting image height to match text height when placement is left of title.

However, this change results in a poor default (see screenshot below), because the default is arbitrarily set to 30 in the shared model. So, any host currently using `iconPlacement: LeftOfTitle` will need to newly configure their `iconSize` to restore the previous size.

If this is not a default we want, we could remove the arbitrary default in the object model (and make it optional), to allow the renderer to use a good default (i.e. icon height matches text height).

## Sample Card
`v1.2\Elements\Action.IconUrl.TwoActions`

## How Verified
(using default host config values)

Before PR:
image

After PR:
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5704)


dclaux:
I think the choice of 30px as a default is what is problematic here. The typical size of a button icon is 16px. I think you should change the default to 16px, which is what JS uses.

jwoo-msft:
> I think the choice of 30px as a default is what is problematic here. The typical size of a button icon is 16px. I think you should change the default to 16px, which is what JS uses.

@dclaux 
mobile platforms (iOS & Android) and probably UWP use the title height of the button for the image height when iconPlacement is LeftOfTitle. I don't have objections to change to use the iconsize for the icon even it's for LeftOfTitle. We probably have good reasons why we did this way. c.c. @jonmill @matthidinger @RebeccaAnne @paulcam206 @almaasrawi @golddove

RebeccaAnne:
My recollection was that it was intentional to use text height rather than icon size when the icon is left of the title, and in general I think that's a good approach. What would the above title button look like if we size the default down to 16? I suspect it would look a little too small in that case. For what it's worth the default of 30 is consistent with the documentation ([ActionsConfig](https://docs.microsoft.com/en-us/adaptive-cards/rendering-cards/host-config#schema-actionsconfig)).

That said, it **is** weird to have an iconSize property but not always respect it (especially given that that behavior is not documented). I like @golddove's idea of having it be optional and having the renderer default to text height for left of title and the documented default (30) for above title. That seems like the best of both worlds in terms of letting us do the reasonable default but also respecting any explicit value from the host.

In any case, if we stick with the approach in this PR we'll need to check the other platforms as well as I suspect everyone except JavaScript is doing it the way Android was. We should also make sure to update the docs with whatever we land on.

golddove:
> What would the above title button look like if we size the default down to 16? I suspect it would look a little too small in that case. For what it's worth the default of 30 is consistent with the documentation ([ActionsConfig](https://docs.microsoft.com/en-us/adaptive-cards/rendering-cards/host-config#schema-actionsconfig)).

16 looks fine on Android:

image

dclaux:
I only suggested using 16 in place of 30 where 30 is currently used as a default, not to change other logic such as using title height.

golddove:
Documenting outcome from standup: match current JS behavior (i.e. always respect iconSize, and default to 16 when unset)

source/shared/cpp/ObjectModel/HostConfig.h
```diff
@@ -292,7 +292,7 @@ namespace AdaptiveSharedNamespace
         unsigned int maxActions = 5;
         Spacing spacing = Spacing::Default;
         IconPlacement iconPlacement = IconPlacement::AboveTitle;
-        unsigned int iconSize = 30;
+        unsigned int iconSize = 16;
```
jwoo-msft, line 295: @golddove 
How would we know if the iconSize is not set? Shouldn't we change it to optional to tell if it has a value?

jwoo-msft (changes requested):

source/shared/cpp/ObjectModel/HostConfig.h
golddove: The plan of record was to always respect iconSize.

RebeccaAnne (approved):
If you haven't already could you log bugs to update the other renderers to use this behavior and reference this PR so we've got breadcrumbs back to the decision making? Otherwise looks good.

dclaux (approved):

golddove:
> If you haven't already could you log bugs to update the other renderers to use this behavior and reference this PR so we've got breadcrumbs back to the decision making? Otherwise looks good.

Logged UWP bug, and added comment to existing iOS bug,

jwoo-msft (approved):

Issue #5630, opened 2021-04-07
Table element spec and design review


RebeccaAnne:
## Try it for yourself
- This feature is implemented for JavaScript in this draft PR: https://github.com/microsoft/AdaptiveCards/pull/5705
- You can use the PR's [test site](https://adaptivecardsci.z5.web.core.windows.net/pr/5705) to try the feature for yourself. Note that it might be buggy and that designer support isn't complete.

## Table element for Adaptive Cards
This is a proposal for a simple Table element. The main goals of this element are:
- To allow the creation of table layouts in a straightforward way, as opposed to using a stack of ColumnSet elements which leads to a larger (potentially much larger) payload and potential headaches wrt column sizing
- To allow accessibility software to present table layouts as such, which is currently impossible

This is NOT a proposal for an advanced Table with features such as:
- Dynamic column sizing
- Dynamic column reordering
- Sorting by columns
- Etc.

### Table
| Name | Type | Required | Description |
| --- | --- | --- | --- |
| `columns` | `ColumnDefinition[]` | No | Defines the table's columns (number of columns, and column sizes). See `ColumnDefinition` below. |
| `rows` | `TableRow[]` | No | Defines the rows of the Table, each being a collection of cells. See `TableRow` below. Rows are not required, which allows empty Tables to be generated via templating without breaking the rendering of the whole card. |
| `showGridLines` | `Boolean` | No | Specifies if grid lines should be displayed around the table and its cells. Defaults to `true`. |
| `gridStyle` | string | No | When `showGridLines` is set to `true`, `gridStyle` controls the color of the grid. Its value must be a valid container style name (emphasis, accent, etc.). If `gridStyle` isn't specified, the style used is inherited from the Table's parent container.  This feature requires adding support for border colors in HostConfig (see below). |
| `horizontalCellContentAlignment` | `HorizontalAlignment` | No | Specifies how the content of each cell should be horizontally aligned by default. Defaults to `left`. This value can be overridden by setting `horizontalCellContentAlignment` on a column or row, or by setting the `horizontalAlignment` property on each individual element in a cell. |
| `verticalCellContentAlignment` | `VerticalAlignment` | No | Specifies how the content of each cell should be vertically aligned by default. Defaults to `top`. This value can be overridden by setting `verticalCellContentAlignment` on a column, row or cell. |
[...]

PR #5705, opened 2021-04-21
[JS] Table element

dclaux:
## Related Issue
- Implements https://github.com/microsoft/AdaptiveCards/issues/5729
- Also introduces a new "adaptivecards-extras" package

## Description
- Adds support for the Table element (spec: https://github.com/microsoft/AdaptiveCards/issues/5630) to the JS SDK.
- Introduces the "adaptivecards-extras" package
  - This package is designed to include additional AC features for the JS SDK
  - These features will be either demos, documentation samples, experiments or preview features destined to be eventually integrated in the main SDK
  - None of the features in this package are fully tested nor supported. They might change or even be removed without notice
  - With this PR, the package only contains an implementation of an **Extras.ProgressBar** custom element, which we have so far used as a sample for creating custom elements. 
  - Refer to the README.md file for more details
- Introduces the "adaptivecards-extras-designer" package
  - This package implements designer support for features exported by the "adaptivecards-extras" package
  - Refer to the package's README.md for more information

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5705)


source/nodejs/adaptivecards-extras-designer/src/adaptivecards-extras-designer.ts
```diff
@@ -0,0 +1,9 @@
+import { CardDesignerSurface } from "adaptivecards-designer";
```
paulcam206, line 1: missing license/copyright comment (should be in every source file)

source/nodejs/adaptivecards-extras/src/progress-bar.ts
```diff
@@ -0,0 +1,117 @@
...
+            let color = Adaptive.stringToCssColor(this.hostConfig.containerStyles.emphasis.foregroundColors.accent.default);
+
+            if (color) {
+                this._leftBarElement.style.backgroundColor = color;
+            }
+
+            this._rightBarElement = document.createElement("div");
+            this._rightBarElement.style.height = "6px";
+
+            color = Adaptive.stringToCssColor(this.hostConfig.containerStyles.emphasis.backgroundColor);
```
paulcam206, line 80: in general, it's a bad habit to reuse variables

source/nodejs/adaptivecards/src/table.ts
```diff
@@ -0,0 +1,586 @@
...
+import { CardElement, StylableCardElementContainer, SerializationContext, Container, ContainerStyleProperty } from "./card-elements";
+import { HorizontalAlignment, VerticalAlignment, SizeUnit, ValidationEvent, TypeErrorType, Spacing } from "./enums";
+import { TextStyleDefinition } from "./host-config";
+import { BaseSerializationContext, BoolProperty, CustomProperty, EnumProperty, property, PropertyBag,
+    PropertyDefinition, SerializableObject, SerializableObjectCollectionProperty, Versions } from "./serialization";
+import { SizeAndUnit, PaddingDefinition } from "./shared";
+import { Strings } from "./strings";
+import { stringToCssColor } from "./utils";
+
+export class ColumnDefinition extends SerializableObject {
```
paulcam206, line 12: Feel like this should be `TableColumnDefinition` to prevent confusion with `ColumnSet` `Column`s

paulcam206 (changes requested):

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -121,10 +125,12 @@ export class CardElementPeerRegistry extends DesignerPeerRegistry<CardElementTyp
...
     }
 
-    createPeerInstance(designerSurface: CardDesignerSurface, parent: DesignerPeers.DesignerPeer, cardElement: Adaptive.CardElement): DesignerPeers.CardElementPeer {
+    createPeerInstance(designerSurface: CardDesignerSurface, parent: DesignerPeers.DesignerPeer, cardElement: Adaptive.CardElement, initializeCardElement?: boolean): DesignerPeers.CardElementPeer {
         var registrationInfo = this.findTypeRegistration((<any>cardElement).constructor);
 
-        var peer = registrationInfo ? new registrationInfo.peerType(parent, designerSurface, registrationInfo, cardElement) : new DesignerPeers.CardElementPeer(parent, designerSurface, this.defaultRegistration, cardElement);
+        var peer = registrationInfo ?
+            new registrationInfo.peerType(parent, designerSurface, registrationInfo, cardElement, initializeCardElement) :
+            new DesignerPeers.CardElementPeer(parent, designerSurface, this.defaultRegistration, cardElement, initializeCardElement);
```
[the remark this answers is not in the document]
dclaux: This code has been like that for a long time, and I happen to prefer the way it reads now. I'll leave as is.

source/nodejs/adaptivecards-extras-designer/src/adaptivecards-extras-designer.ts
dclaux: Thanks, will fix in all affected files.

source/nodejs/adaptivecards-extras/src/progress-bar.ts
dclaux: Why?

source/nodejs/adaptivecards/src/table.ts
dclaux: I don't really mind, but this is per spec.

dclaux: If you feel strongly about this one, please send a mail to the team to suggest this name change which would also have to be reflected in the spec that Risheek has a PR for here: https://github.com/microsoft/AdaptiveCards/pull/5806

source/nodejs/adaptivecards-extras/src/progress-bar.ts
paulcam206: * `color` is inadequate to describe what the variable is holding. in its first assignment, it's holding a foreground color, and in its second, a background color. this leads to cognitive load for the reader to keep in mind which color is being referred to at which point in the function
* adding new code in the vicinity of this means having to first figure out if `color`'s second life has any bearing on its first
* if this method were to be refactored, this reuse might need to be reconciled

there's really no benefit to reusing a local variable anyways

dclaux: I don't really agree with any of this, but there's also no harm in doing what you suggest. So I'll make the change if it can help closing this PR faster.

dclaux: Done

paulcam206: for what it's worth, reusing variables being bad isn't just a me thing:

https://javascript.info/variables
> Reuse or create?
> And the last note. There are some lazy programmers who, instead of declaring new variables, tend to reuse existing ones.
> As a result, their variables are like boxes into which people throw different things without changing their stickers. What’s inside the box now? Who knows? We need to come closer and check.
> Such programmers save a little bit on variable declaration but lose ten times more on debugging.
> An extra variable is good, not evil.
> Modern JavaScript minifiers and browsers optimize code well enough, so it won’t create performance issues. Using different variables for different values can even help the engine optimize your code.

it also shows up in [*How To Write Unmaintainable Code*](https://cs.fit.edu/~kgallagher/Schtick/How%20To%20Write%20Unmaintainable%20Code.html), and as language checks in various places (e.g. [PHPStorm](https://www.jetbrains.com/help/phpstorm/javascript-and-typescript-reuse-of-local-variable.html))

jonmill (approved):
Thanks for all the hard work here David, looks good!

Not blocking this time, but in the future could you please break up large changes into smaller packages to review? Large code drops are difficult to properly review, so it'll help get prompter and better feedback in the future. Thank you!

source/nodejs/adaptivecards-designer-app/package.json
```diff
@@ -15,9 +15,11 @@
 	"devDependencies": {
 		"@typescript-eslint/eslint-plugin": "^3.4.0",
 		"@typescript-eslint/parser": "^3.4.0",
-		"adaptivecards": "^2.9.0",
-		"adaptivecards-controls": "^0.8.0",
-		"adaptivecards-designer": "^2.2.0",
+		"adaptivecards": "^2.8.0",
```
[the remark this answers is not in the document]
dclaux: Correct. Typically, we should NOT up this dependency version every time we release a new adaptivecards package, unless that package introduces new APIs that are required by the designer. 2.8.0 is a good minimal version to stick to for now.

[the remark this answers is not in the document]
dclaux: And... let me contradict myself. The designer has dependencies on the breaking changes we are making to support text styles, so this dependency needs to be pointing to the latest version this time around. I'll update.

PR #5710, opened 2021-04-21
[Android] Dropdown item rendering refactored.

karthikgopal36:
## Description

Dropdown item rendering is refactored, so the dropdown items can behave like primary action elements even if the client uses CustomActionRenderer.
For inline cards, rootLevelActions should be false. fixed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5710)

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
@@ -31,7 +31,7 @@ import io.adaptivecards.renderer.actionhandler.ICardActionHandler;
 
 public abstract class BaseActionElementRenderer implements IBaseActionElementRenderer
 {
-    protected final static int BOTTOM_SHEET_TAG = 0xffffffff;
+    public final static int PARENT_DROPDOWN_TAG = 0xffffffff;
 
     protected static int getColor(String colorCode)
     {
@@ -241,9 +241,9 @@ public abstract class BaseActionElementRenderer implements IBaseActionElementRen
             v.setSelected(m_invisibleCard.getVisibility() != View.VISIBLE);
             // Reset all other buttons
             ViewGroup parentContainer;
-            if (v.getTag(BOTTOM_SHEET_TAG) != null)
+            if (v.getTag(PARENT_DROPDOWN_TAG) != null)
             {
-                parentContainer = (ViewGroup) v.getTag(BOTTOM_SHEET_TAG);
+                parentContainer = (ViewGroup) v.getTag(PARENT_DROPDOWN_TAG);
             }
             else
             {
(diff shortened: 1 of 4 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
```diff
@@ -31,7 +31,7 @@
 
 public abstract class BaseActionElementRenderer implements IBaseActionElementRenderer
 {
-    protected final static int BOTTOM_SHEET_TAG = 0xffffffff;
+    public final static int PARENT_DROPDOWN_TAG = 0xffffffff;
```
golddove, line 34:
```suggestion
    /**
     * Describe purpose / use of the constant
     */
    public final static int PARENT_DROPDOWN_TAG = 0xffffffff;
```

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/OverflowActionLayoutRenderer.java
```diff
@@ -66,17 +66,22 @@ public View renderActions(final RenderedAdaptiveCard renderedCard, final Context
...
+            if (overflowActionRenderer == null || (overflowActionView = overflowActionRenderer.onRenderOverflowAction(viewGroup, menuItemList, renderArgs.isRootLevelActions())) == null) {
+                overflowActionView = renderDefaultOverflowAction(context, viewGroup, hostConfig);
+            }
+            final LinearLayout contentLayout = new LinearLayout(context);
+            contentLayout.setLayoutParams(new LinearLayout.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT));
+            contentLayout.setOrientation(LinearLayout.VERTICAL);
+            setDrawables(contentLayout);
+            overflowActionView.setOnClickListener(new OverflowActionOnClickListener(contentLayout, overflowActionRenderer, menuItemList));
+        }
+        catch (Exception e)
```
golddove, line 81: Is there a specific exception we are expecting, or is this just to continue rendering?

golddove (approved):

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/OverflowActionLayoutRenderer.java
karthikgopal36: This is for to continue rendering. Since client also does the custom overflow rendering, we want to make sure  card rendering should not be broken incase of exception.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
karthikgopal36: done

PR #5734, opened 2021-04-26
[Shared/UWP] Add IsEnabled property to actions

### LINKED ISSUE #5721: [UWP] Action isEnabled

### LINKED ISSUE #5723: [Shared] Action isEnabled

RebeccaAnne:
# Related Issue

Fixes #5723 
Fixes #5721

# Description

Add isEnabled property to actions in the shared model and UWP and updated the UWP renderer code to respect it

# Sample Card

New sample card added samples\v1.5\Elements\Action.IsEnabled.json

# How Verified

1. Updated Everything Bagel shared model test to validate new property
2. Updated UWP object model test to validate new property
3. Validated IsEnabled property works as expected in the visualizer using the above card.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5734)


source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -617,9 +625,19 @@ namespace AdaptiveNamespace::ActionHelpers
         ComPtr<IAdaptiveHostConfig> hostConfig;
         THROW_IF_FAILED(renderContext->get_HostConfig(&hostConfig));
 
-        if (ActionHelpers::WarnForInlineShowCard(renderContext, action, L"Inline ShowCard not supported for SelectAction"))
+        // We don't wrap the element if it's a show card or if the action's not enabled;
+        bool dontWrap = ActionHelpers::WarnForInlineShowCard(renderContext, action, L"Inline ShowCard not supported for SelectAction");
```
paulcam206, line 629: could we make this `shouldWrap` -- we should avoid negatives in bools :D

source/uwp/Renderer/lib/AdaptiveExecuteAction.h
```diff
@@ -90,6 +90,15 @@ namespace AdaptiveNamespace
             return AdaptiveActionElementBase::put_Tooltip(tooltip);
         }
 
+        IFACEMETHODIMP put_IsEnabled(boolean isEnabled) override
```
paulcam206, line 93: definitely not something to do in this PR, but I think if we refactor our classes using cppwinrt, we should be able to derive classes in a way that makes it so we can get out of the business of manually calling the base (`LinkButton` does something similar to this).

guessing it might require breaking API surface changes, so maybe part of our next major rev? just a thought :)

paulcam206 (approved):

source/uwp/Renderer/lib/AdaptiveExecuteAction.h
RebeccaAnne: I agree, I'd love to get more cppwinrt in here. I feel like it shouldn't require breaking changes? But I don't know. Definitely something to work towards at some point.

PR #5736, opened 2021-04-27
Feature/heading styles

RebeccaAnne:
# Related Issue

Merges feature branch that will contain fixes for  #5632, #5633, #5634, and #5631. This branch is not currently ready to merge as not all fixes are in yet.

# Description

Feature branch for shared model platforms

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5736)


source/shared/cpp/ObjectModel/ParseUtil.h
```diff
@@ -148,15 +145,7 @@ namespace AdaptiveSharedNamespace
...
-                if (isRequired)
-                {
-                    throw AdaptiveCardParseException(ErrorStatusCode::RequiredPropertyMissing,
-                                                     "Property is required but was found empty: " + propertyName);
-                }
-                else
-                {
-                    return defaultEnumValue;
-                }
+                return std::nullopt;
```
jwoo-msft, line 148: @RebeccaAnne,
Returnnig nullopt for empty value seems to be inconsistent with returnning the default value from the line 157 when the given string value doesn't map to enum values. For example, for "", will return nullopt, but "Hello World" for horizontal alignment will return HorizontalAlignmentLeft. returnning nullopt makes more sense since we have to know if the user set the value or not. But then the renderers have to know what is the proper default values. We could have a util function for returnning the default values for cross platforms.

RebeccaAnne: You're first point is a good one, and I updated the enum magic code to not set the default from something like "Hello World". Since most of the callers of this are coming through ParseUtil::GetEnumValue those callers will set the default anyway, but for the ones coming directly through ParseUtil::GetOptionalEnumValue should now get nullopt in that case. 

It is correct that with this setup the renderers do need to know the defaults. I'm not opposed to setting something up in shared model to ensure we've got consistent defaults cross platform. Could you log an issue to track adding that?

golddove: +1 to this, would love to move the defaults for all optionals back into the object model

golddove (approved):

PR #5737, opened 2021-04-27
[Samples] Update scenario samples to use heading style

RebeccaAnne:
# Related Issue

[ADO 29462597](https://microsoft.visualstudio.com/OS/_workitems/edit/29462597)

# Description

Updates the scenario samples called out in the above ADO bug to use the newly introduced heading style. This will cause it to be reported correctly for accessibility.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5737)

### DIFF
--- a/samples/Templates/Scenarios/ActivityUpdate.template.json
+++ b/samples/Templates/Scenarios/ActivityUpdate.template.json
@@ -1,98 +1,99 @@
 {
-    "type": "AdaptiveCard",
-    "body": [
-        {
-            "type": "TextBlock",
-            "size": "Medium",
-            "weight": "Bolder",
-            "text": "${title}",
-            "wrap": true
-        },
-        {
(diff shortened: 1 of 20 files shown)


golddove (approved):
The first textblocks in ActivityUpdate, CalendarReminder, and ExpenseApproval (and their respective *WithLabels versions) look like they should be headings, too, right?

RebeccaAnne:
Good point. I just went for the ones called out in the ADO bug, but I'll update those as well.

---
In reply to: [646973700](https://github.com/microsoft/AdaptiveCards/pull/5737#issuecomment-646973700) [](http://example.com/codeflow?ancestors=646973700,646973700)

PR #5740, opened 2021-04-28
[Android] Add heading style support

### LINKED ISSUE #5633: [Android] Add heading styling support to renderer

golddove:
# Related Issue

Fixes #5633 

# Description

- SWIG run to ingest model changes
- To avoid bloating the SWIG interface with type mappings from `std::optional` to plain Java enums for each style property, the mappings were written as a macro. Existing optional primitive mappings were removed to use this new macro.
- Updated renderer to resolve style-related properties based on `style`
- `AdaptiveCardSchemaKey::Level` removed as it was replaced by `AdaptiveCardSchemaKey::HeadingLevel`

## Breaking API Changes

- `io.adaptivecards.objectmodel.TextBlock` and `io.adaptivecards.objectmodel.TextRun` now return `null` by default for style-related properties (i.e. `GetFontType`, `GetIsSubtle`, `GetTextSize`, `GetTextColor`, `GetTextWeight`) after parsing
    - `null` indicates the property is unset in the JSON payload. Non-null value is returned if and only if a valid value is explicitly set in the payload. This change was required because the defaults are no longer constant (they now depend on `TextBlock.style`).
- `TextBlockRenderer` methods that apply these properties have new signatures to accept `TextStyle` as a parameter. Any subclasses relying on these will need to update their calls to the new signatures.

# Sample Card

`v1.5\Test\TextBlock.StyleOverrides.json`

# How Verified

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5740)


source/android/adaptivecards/src/main/java/io/adaptivecards/objectmodel/AdaptiveCardSchemaKey.java
```diff
@@ -104,6 +106,7 @@
   Language,
   Large,
   Left,
+  Level,
```
RebeccaAnne, line 109: >Level [](http://example.com/codeflow?start=2&length=5)

This one's not in use anymore i don't think and i should have taken it out. Could you possibly pull this from the shared model while you're in here?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
```diff
@@ -60,36 +63,38 @@ public static TextBlockRenderer getInstance()
...
     }
 
-    /**
-     * Applies given TextStyle to the given TextView
-     * @param textView TextView to apply style to
-     * @param style TextStyle to apply
-     */
-    private static void applyTextStyle(TextView textView, final TextStyle style)
+    public static void applyAccessibilityHeading(@NonNull final TextView textView,
+                                                    @Nullable final TextStyle textStyle)
```
RebeccaAnne, line 80: >    [](http://example.com/codeflow?start=49&length=3)

[nit] alignment

RebeccaAnne (approved):
:shipit:

source/android/adaptivecards/src/main/java/io/adaptivecards/objectmodel/AdaptiveCardSchemaKey.java
golddove: Done

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java
golddove: Fixed

golddove:
Merging without iOS pass, as that PR will come in later.

PR #5751, opened 2021-04-29
[SharedModel] Basic cmake support

paulcam206:
# Description

This change introduces `cmake` build support for the Shared Model. This allows us to independently build the Shared Model in a cross-platform way. As part of this, there were a number of code issues found that are fixed here (primarily from clang, which seems stricter than MSVC).

## Tested configurations
* cmake/VS (`Visual Studio 16 2019` / `Windows SDK version 10.0.19041.0` / `MSVC 19.28.29914.0` - Windows)
* cmake/gcc (`GNU 10.3.0` - WSL)
* cmake/clang (`Clang 9.0.1` - Windows and WSL)
* cmake/emscripten (`emcc 2.0.18` - Windows and WSL)

## Benefits
* Easy access to real clang tooling (e.g. `clang-tidy`)
* Different compilers catch different types of issues
* Necessary prerequisite to broader emscripten investigation

## How do I build this thing?
* Windows + VS
  1. Install prerequisites
      * Install VS 2019
      * `winget install cmake`
  1. Run `Developer Command Prompt for VS 2019` from start menu
  2. `cd <ac root>\source\shared\cpp\ObjectModel`
  3. `cmake .`
  4. `cmake --build .`
* Windows + clang
  1. Install prerequisites
      * `winget install cmake`
      * `winget install llvm`
  1. `cd <ac root>\source\shared\cpp\ObjectModel`
  2. `cmake .`
  3. `cmake --build .`
* Windows + emscripten
  1. Install prerequisites
      * `winget install python`
      * `winget install cmake`
      * `winget install git`
  1. Install `emscripten`
      1. `git clone https://github.com/emscripten-core/emsdk.git <location to place emscripten SDK>`
      2. `cd <emsdk directory>`
      3. `git pull`
      4. `emsdk install latest`
      5. `emsdk activate latest`
  1. `cd <ac root>\source\shared\cpp\ObjectModel`
  2. `<emsdk directory>\emsdk_env.bat` <- might not be needed depending on your environment
  2. `emcmake cmake .`
  3. `cmake --build .`
* WSL (Ubuntu) + gcc
  1. Install prerequisites: `sudo apt update && sudo apt install cmake gcc`
  2. `cd <ac root>/source/shared/cpp/ObjectModel`
  3. `cmake .`
  4. `cmake --build .`
* WSL (Ubuntu) + clang
  1. Install prerequisites: `sudo apt update && sudo apt install cmake clang`
      * You might need to change your compilers:
          *  Environment vars: 
              * `export CC=/usr/bin/clang`
              * `export CXX=/usr/bin/clang++`
          * Choose `clang` as your default C/C++ compiler:
              * `sudo update-alternatives --config cc` (choose `clang`)
              * `sudo update-alternatives --config c++` (choose `clang`)
  2. `cd <ac root>/source/shared/cpp/ObjectModel`
  3. `cmake .`
  4. `cmake --build .`
* WSL (Ubuntu) + emscripten
  1. Install prerequisites: `sudo apt update && sudo apt install cmake python3 git`
  1. Install `emscripten`
      1. `git clone https://github.com/emscripten-core/emsdk.git <location to place emscripten SDK>`
      2. `cd <emsdk directory>`
      3. `git pull`
      4. `emsdk install latest`
      5. `emsdk activate latest`
  1. `cd <ac root>/source/shared/cpp/ObjectModel`
  2. `source <emsdk directory>/emsdk_env.sh` <- might not be needed depending on your environment
  2. `emcmake cmake .`
  3. `cmake --build .`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5751)


paulcam206:
ooh -- build errors, that's exciting :)

paulcam206:
UWP break is interesting -- complaining because existing `max()` macro def. Can force it not to be defined, but we're using `min()` macro def in XamlHelpers.cpp. Will fix tomorrow.

source/shared/cpp/ObjectModel/HostConfig.cpp
```diff
@@ -631,16 +631,16 @@ unsigned int HostConfig::GetFontSize(FontType fontType, TextSize size) const
     // desired font size
     auto result = GetFontType(fontType).fontSizes.GetFontSize(size);
 
-    // UINT_MAX used to check if value was defined
-    if (result == UINT_MAX)
+    // std::numeric_limits<unsigned int>::max() used to check if value was defined
```
jonmill, line 634: Do we care about / should we care about the differences of `unsigned int` and `uint32_t` and the like?

source/shared/cpp/ObjectModel/ParseUtil.h
```diff
@@ -99,27 +102,29 @@ namespace AdaptiveSharedNamespace
 
         template<typename T>
         std::vector<std::shared_ptr<T>> GetElementCollectionOfSingleType(
-            ParseContext& context,
+            ParseContext & context,
```
jonmill, line 105: NIT: difference in formatting the dereference operator here from elsewhere in the file...is this on purpose?

jonmill (approved):
A couple questions but LGTM otherwise!

source/shared/cpp/ObjectModel/ParseUtil.h
paulcam206: hmm... we usually lean on `clang-format` to do this for us... maybe it didn't apply here? I'll fix up :)

source/shared/cpp/ObjectModel/HostConfig.cpp
paulcam206: I don't think we really care in this *specific* case since we're using this as a sentinel value (and presumably we'll hit platform font size limits before we approach the upper limit of `unsigned int` regardless of how many bits it has). additionally, this is strictly for HostConfig, so it's not controllable outside of the host.

source/shared/cpp/ObjectModel/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+cmake_minimum_required(VERSION 3.16)
```
jwoo-msft, line 1: Are we going to include CMake build process as part of the shared model CI? Is CMake-built shared model binary compatible with the shared model unit tests?

jwoo-msft (approved):

source/shared/cpp/ObjectModel/CMakeLists.txt
paulcam206: I think we will eventually want to stand up a pipeline here, but not as part of this checkin.

The generated binary would theoretically be binary compatible for linking, but we'd have to modify the unit tests. Getting that story sorted out is future work, though :)

source/shared/cpp/ObjectModel/ParseUtil.h
paulcam206: fixed in 2a01ac6

paulcam206:
@jwoo-msft - could you fix up the iOS project for me? `InternalID` was moved into its own header file, `InternalID.h`, which needs to be referenced by the iOS project :)

Thanks!

paulcam206:
@jwoo-msft - gentle ping :)

jwoo-msft:
> @jwoo-msft - gentle ping :)
@paulcam206
sorry!

jwoo-msft:
@paulcam206,
Done

PR #5757, opened 2021-04-30
[Shared] Fix link parsing for URLs with `)` chars

### LINKED ISSUE #5553: [Teams][Rendering] [Broken rendering of markup links on iOS / Android]
# Platform

- [x] Android
- [x] iOS

# Author or host

Are you an author (like sending something to Outlook), or a host that is rendering your own cards? If host, please specify the corresponding host (app) name.

Host - Teams Mobile

# Version of SDK

What version are you using? Ex: NuGet 1.0.2, or latest main, etc...

2.4.6

# Details

Explain your issue. If you just have a question, please post [on StackOverflow instead](https://stackoverflow.com/questions/tagged/adaptive-cards).

With the given JSON payload with markup links, desktop renders it as expected, while on iOS / Android, there's an extra ".aspx)" being rendered.

[
[markupLinkBugJson.txt](https://github.com/microsoft/AdaptiveCards/files/6159660/markupLinkBugJson.txt)
](url)

Desktop render:

image

Mobile render:

image

Note that this repros in the standalone visualizer app as shown above in addition to Teams Mobile clients.

paulcam206:
# Related Issue

Fixes #5553

# Description

Fix up handling of URLs with `)` chars in them -- they should work properly if escaped. While I was in the vicinity, I sped up the ObjectModel build by enabling multiprocessor build support.

# Sample card

Broken card from #5553:
 ```json
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "TextBlock",
   "text": "[News Time! (01/2021)](https://contoso.sharepoint.com/SitePages/News-Time-(01-2021).aspx)",
   "weight": "Bolder",
   "size": "Medium"
        }]}
```

Same card, but fixed (note escaped `)`):
 ```json
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
    {
      "type": "TextBlock",
   "text": "[News Time! (01/2021)](https://contoso.sharepoint.com/SitePages/News-Time-(01-2021\\).aspx)",
   "weight": "Bolder",
   "size": "Medium"
        }]}
```
# How Verified

* new unit test
* manual testing via UWP visualizer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5757)


paulcam206:
I should note that I noticed some more broken markdown parsing behavior that goes beyond the scope of this issue. I've filed https://github.com/microsoft/AdaptiveCards/issues/5758 to track.

jwoo-msft:
> I should note that I noticed some more broken markdown parsing behavior that goes beyond the scope of this issue. I've filed #5758 to track.

Would it be better matching the parenthesis inside link text block by tracking the number parenthesis instead of supporting escape char?

paulcam206:
> > I should note that I noticed some more broken markdown parsing behavior that goes beyond the scope of this issue. I've filed #5758 to track.
> 
> Would it be better matching the parenthesis inside link text block by tracking the number parenthesis instead of supporting escape char?

Well -- that's what [commonmark](https://commonmark.org/) does:

image

becomes
image
(note that the `[foobar]` link points to the supplied url correctly)

However, our parsing is much more closely aligned with older-style markdown parsing. If we want to move towards commonmark's implementation, we should look to ingest their code, but that's a much larger change:
* We'd need to ingest their code
* AFAIK they don't support disabling some of the features we'd want disabled, so we'd either need to try to PR changes upstream (possible, but I don't know how receptive they'd be) or manage a fork of their codebase (possible, but I'm not sure we have bandwidth for this at the moment)
* I don't have a good sense for how much of a breaking change this would be to existing card consumers -- we'd want to enumerate the differences and, if they're breaking, only ship them as part of a larger update.

Since this was a specific ask from Teams, I went with the point fix for this one scenario in the style of markdown that we currently support.

source/shared/cpp/ObjectModel/MarkDownBlockParser.h
```diff
@@ -97,6 +97,8 @@ namespace AdaptiveSharedNamespace
         // Callback function that handles the Emphasis State
         static EmphasisState MatchEmphasis(EmphasisParser&, std::stringstream&, std::string&);
 
+        static bool IsEmphasisToken(const int token);
```
paulcam206, line 100: shouldn't need `const` here

jonmill: It isn't necessary but isn't it considered const-correct to have it when `token` should not be modified?

paulcam206: current guidance is not to use `const` for simple params passed by value -- obviously that's not How I Was Raised, because I still `const` everything instinctively :)

paulcam206: fixed in 4dd0da8

jwoo-msft:
> > > I should note that I noticed some more broken markdown parsing behavior that goes beyond the scope of this issue. I've filed #5758 to track.
> > 
> > 
> > Would it be better matching the parenthesis inside link text block by tracking the number parenthesis instead of supporting escape char?
> 
> Well -- that's what [commonmark](https://commonmark.org/) does:
> 
> image
> 
> becomes
> image
> (note that the `[foobar]` link points to the supplied url correctly)
> 
> However, our parsing is much more closely aligned with older-style markdown parsing. If we want to move towards commonmark's implementation, we should look to ingest their code, but that's a much larger change:
> 
> * We'd need to ingest their code
> * AFAIK they don't support disabling some of the features we'd want disabled, so we'd either need to try to PR changes upstream (possible, but I don't know how receptive they'd be) or manage a fork of their codebase (possible, but I'm not sure we have bandwidth for this at the moment)
> * I don't have a good sense for how much of a breaking change this would be to existing card consumers -- we'd want to enumerate the differences and, if they're breaking, only ship them as part of a larger update.
> 
> Since this was a specific ask from Teams, I went with the point fix for this one scenario in the style of markdown that we currently support.

@paulcam206,
sorry about the late response... for the text inside the square brackets, matching the inner square brackets does work, so I was thinking we could use the same logic. I haven't really looked into it, so I could be wrong.

jonmill (approved):
LGTM!

PR #5760, opened 2021-05-02
Toggle input alignment and Handle isEnable property for action type

IbrahimSulai:
This PR consist of the below changes:
1. Updated the alignment of toggle input as same as native platform.
2. Handled isEnabled property for action type, select action, and inline action - [https://github.com/microsoft/AdaptiveCards/issues/4670](https://github.com/microsoft/AdaptiveCards/issues/4670)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5760)

### DIFF
--- a/source/community/reactnative/src/components/inputs/input.js
+++ b/source/community/reactnative/src/components/inputs/input.js
@@ -206,9 +206,10 @@ export class Input extends React.Component {
 			return null;
 		}
 		else {
+			let opacityStyle = { opacity: inlineAction.isEnabled == undefined ? 1.0 : inlineAction.isEnabled ? 1.0 : 0.4 };
 			return (
 				<View>
-					<ElementWrapper configManager={this.props.configManager}  json={payload} style={wrapperStyle} isError={this.props.isError} isFirst={this.props.isFirst}>
+					<ElementWrapper configManager={this.props.configManager} json={payload} style={wrapperStyle} isError={this.props.isError} isFirst={this.props.isFirst}>
 						<View style={styles.elementWrapper}>
 							<InputLabel configManager={this.props.configManager} isRequired={this.isRequired} label={label} />
 							<TextInput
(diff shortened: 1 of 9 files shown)


source/community/reactnative/src/components/inputs/input.js
```diff
@@ -206,9 +206,10 @@ export class Input extends React.Component {
 			return null;
 		}
 		else {
+			let opacityStyle = { opacity: inlineAction.isEnabled == undefined ? 1.0 : inlineAction.isEnabled ? 1.0 : 0.4 };
```
vivekvijayakrishnan, line 209: `undefined` check is not required as it's handled in **` BaseActionModel`**. can we remove this check?

source/community/reactnative/src/components/inputs/input.js
```diff
@@ -236,11 +237,14 @@ export class Input extends React.Component {
 								value={this.props.value}
 							/>
 						</View>
-						<TouchableOpacity onPress={() => { this.onClickHandle(onExecuteAction, Constants.InlineAction) }}>
+						<TouchableOpacity
+							disabled={inlineAction.isEnabled == undefined ? false : !inlineAction.isEnabled} //isEnabled defaults to true
+							opacity={inlineAction.isEnabled == undefined ? 1.0 : inlineAction.isEnabled ? 1.0 : 0.5}
```
vivekvijayakrishnan, line 242: We can remove the `undefined` check here as well.

vivekvijayakrishnan (changes requested):
Please check the comments

source/community/reactnative/src/components/inputs/input.js
IbrahimSulai: inlineAction inside the payload is not inherited from the BaseActionModel. Only ActionModel is inherited from the BaseActionModel. So we need to handle the null cases in the places where we consumed.

source/community/reactnative/src/components/inputs/input.js
IbrahimSulai: inlineAction inside the payload is not inherited from the BaseActionModel. Only ActionModel is inherited from the BaseActionModel. So we need to handle the null cases in the places where we consumed.

vivekvijayakrishnan (approved):
LGTM

Issue #5579, opened 2021-03-24
[UWP][Accessibility] Container contents are not read as the accessibility text for select actions


RebeccaAnne:
# Platform

What platform is your issue or question related to? (Delete other platforms).

- UWP

# Details

Using the below card in UWP, when focus lands on the select action, narrator reads only "Button" for the button. It should read all the content of the select action. (Adding `title` does result in the title being read as expected). In JavaScript the following is announced: "This is text block one. This is image one. This is text block two. This is image two. Button"

![image](https://user-images.githubusercontent.com/9091833/112391977-f5753300-8cb5-11eb-95e3-fad71433e431.png)

``` json
{
  "type": "AdaptiveCard",
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "version": "1.3",
  "body": [
    {
      "type": "ColumnSet",
      "columns": [
        {
          "type": "Column",
          "width": "stretch",
          "items": [
            {
              "type": "TextBlock",
              "text": "This is text block 1",
              "wrap": true
            }
          ]
        },
        {
          "type": "Column",
          "width": "stretch",
          "items": [
            {
              "type": "Image",
              "url": "https://adaptivecards.io/content/cats/1.png",
              "altText": "This is image 1"
            }
          ]
        },
        {
          "type": "Column",
          "width": "stretch",
          "items": [
            {
              "type": "TextBlock",
              "text": "This is text block 2",
              "wrap": true
            }
          ]
        },
        {
          "type": "Column",
          "width": "stretch",
          "items": [
            {
              "type": "Image",
              "url": "https://adaptivecards.io/content/cats/2.png",
              "altText": "This is image 2"
            }
          ]
        }
      ],
      "selectAction": {
        "type": "Action.Submit",
        "title": "I'm a button"
      }
    }
  ]
}

```

paulcam206:
Addressing this issue is not currently on our roadmap, so closing for now. We may return to this in the future. Feel free to continue the conversation in this issue.

PR #5763, opened 2021-05-04
[UWP] Add rendering support for tooltip property

### LINKED ISSUE #5470: [UWP] Renderer support for tooltip property

RebeccaAnne:
# Related Issue

Fixes #5470

# Description

Implemented UWP support for tooltips. Added logic to handle correctly setting the title and tooltip properties to the UI tooltip and automation name and description. Refactored automation name/description setting and tooltip creation for easier code sharing.

This PR does not fix related issue #5579. This means that container contents of a container are still not read in the case where the title is not present.

# Sample Card

Added new card at samples\v1.5\Test\TooltipTestCard.json

# How Verified

Using the above test card:
- Validated tooltips display correctly
- Validated Name and FullDescription properties were set properly in accessibility insights
- Validated via Narrator that title and tooltip properties were narrated as expected.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5763)


RebeccaAnne:
Thanks @jwoo-msft!

paulcam206 (approved):

jwoo-msft (approved):

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -43,6 +43,43 @@ namespace AdaptiveNamespace::ActionHelpers
...
+
+            ComPtr<IToolTipServiceStatics> toolTipService;
+            THROW_IF_FAILED(GetActivationFactory(HStringReference(RuntimeClass_Windows_UI_Xaml_Controls_ToolTipService).Get(),
+                                                 &toolTipService));
+
+            THROW_IF_FAILED(toolTipService->SetToolTip(tooltipTarget, toolTip.Get()));
+        }
+    }
+
+    void SetAutomationNameAndDescription(IDependencyObject* dependencyObject, HSTRING name, HSTRING description)
```
jonmill, line 68: Should we verify that `dependencyObject != nullptr` here too since I'm assuming that `SetName` and the like will dereference it?

jonmill (approved):
Just one question but LGTM otherwise

source/uwp/Renderer/lib/ActionHelpers.cpp
RebeccaAnne: So, the tooltip check in the above was because if there isn't a tooltip set in the card we don't want to create an empty tooltip in the UI. In general, we're not doing null parameter checking on required parameters in internal methods in this code. 

One thing that would have helped make this clearer though is that i'm missing SAL annotations on the parameters. I'll add those so it's more obvious what's going on.

samples/v1.5/Test/TooltipTestCard.json
```diff
@@ -0,0 +1,225 @@
+{
```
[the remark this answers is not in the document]
RebeccaAnne: Screenshot's not going to help too much here as the interesting behavior is in the tooltips and the accessibility text, neither of which will be captured by that. Maybe I can update the card to better describe the expected behavior for each case.

PR #5799, opened 2021-05-06
 [Android][Accessibility] Media control accessibility

### LINKED ISSUE #5150: [Android][Accessibility] Accessible video controls
# Platform

- [ ] Android

# Version of SDK

latest main

# Details

Source issue: https://microsoft.visualstudio.com/OS/_workitems/edit/28383710

Video player controls (play/pause) should be accessible with screen reader / keyboard navigation on Android.

golddove:
# Related Issue

Fixes #5150 (ADO 28383710)

# Description

Replaces third-party video UI with native `MediaController` widget, for default accessibility. The native control also disables auto-hide behavior when it detect a screen reader, which seems like a good experience.

# Sample Card

`Scenario/ProductVIdeo`

# How Verified

Manual verification with Talkback. Progress and controls popup appear when video begins. Play, pause, rewind, forward buttons are now announced.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5799)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/FullscreenVideoLayout.java
```diff
@@ -265,149 +151,93 @@ public boolean onError(MediaPlayer mp, int what, int extra)
...
+            showControlsForever();
         }
+    }
 
-        m_playButton.setBackground(icon);
+    public void hideControls()
+    {
+        m_mediaControlView.hide();
+        // return focus to player after hiding control view
+        requestFocus();
```
almedina-ms, line 204: is the hide and show methods still used or was the functionality of those functions replaced by this method (and the one below)?

golddove: These three methods (`hideControls`, `showControls`, `showControlsForever`) are used instead of direct calls to show/hide.

The only exception is when releasing the player (in `releaseObjects` and `onDetachFromWindow`), then `.hide` is called directly.

paulcam206 (approved):

almedina-ms (approved):

jonmill:
@golddove - is this still a relevant PR? If so, what do we need to merge?

PR #5807, opened 2021-05-10
[Shared] Remove ADAPTIVE_CARDS_WINDOWS macro (plus fallout)

paulcam206:
# Description

Remove `ADAPTIVE_CARDS_WINDOWS` macro as it is unneeded. Includes follow-on changes to remove namespacing macros. Followed up with a `clang-format` of affected files.

# How Verified

* build

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5807)


source/shared/cpp/ObjectModel/pch.h
```diff
@@ -50,7 +42,7 @@
 #include "Enums.h"
 #include "json/json.h"
 
-namespace AdaptiveSharedNamespace
+namespace AdaptiveCards
 {
-    constexpr const char * const c_sharedModelVersion = "1.2";
+    constexpr const char* const c_sharedModelVersion = "1.2";
```
RebeccaAnne, line 47: >1.2 [](http://example.com/codeflow?start=56&length=3)

What is this? Can we get rid of it?

RebeccaAnne (approved):

source/shared/cpp/ObjectModel/pch.h
paulcam206: looks like this should be updated:

https://github.com/microsoft/AdaptiveCards/blob/217105e5cafeeb86ddd347c5fde0c67787c559e6/source/shared/cpp/ObjectModel/FeatureRegistration.cpp#L11-L12

paulcam206: updated in 49e4d75

PR #5809, opened 2021-05-11
[Android] Icon downloaded and displayed for overflow Action

karthikgopal36:
To display icon on a dropdown item, we usually fetch it from rendered action element and this works only when the dropdown item added to window. Otherwise, icon will not be displayed.
In this PR, we downloaded icon for dropdown item separately instead of fetching it from rendered action element, so the icon can be fetched and displayed on any condition.

Also, rootLevelActions flag added to the `onDisplayOverflowActionMenu()` callback to distinguish between `Action & ActionSet`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5809)

### DIFF
--- a/source/nodejs/adaptivecards-designer/package-lock.json
+++ b/source/nodejs/adaptivecards-designer/package-lock.json
@@ -3129,9 +3129,9 @@
 			}
 		},
 		"hosted-git-info": {
-			"version": "2.8.8",
-			"resolved": "https://registry.npmjs.org/hosted-git-info/-/hosted-git-info-2.8.8.tgz",
-			"integrity": "sha512-f/wzC2QaWBs7t9IYqB4T3sR1xviIViXJRJTWBlx2Gf3g0Xi5vI7Yy4koXQ1c9OYDGHN9sBy1DQ2AB8fqZBWhUg==",
+			"version": "2.8.9",
+			"resolved": "https://registry.npmjs.org/hosted-git-info/-/hosted-git-info-2.8.9.tgz",
+			"integrity": "sha512-mxIDAb9Lsm6DoOJ7xH+5+X4y1LU/4Hi50L9C5sIswK3JzULS4bwk1FvjdBgvYR4bzT4tuUQiC15FE2f5HbLvYw==",
 			"dev": true
 		},
 		"hpack.js": {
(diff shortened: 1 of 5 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/IOverflowActionRenderer.java
```diff
@@ -34,11 +34,12 @@ default View onRenderOverflowAction(@NonNull ViewGroup viewGroup, @NonNull List<
     /**
      * This implementation is invoked when Overflow action view ("...") is pressed and rendered secondary view elements will be shown.
      *
-     * @param menuItemList list of view of rendered secondary action elements.
-     * @param view           Overflow action view.
+     * @param menuItemList       list of view of rendered secondary action elements.
+     * @param view               Overflow action view.
+     * @param isRootLevelActions indicates action is part of root level actions or action set elements in body.
```
golddove, line 39:
```suggestion
     * @param isRootLevelActions true if action is defined at root level, false if part of ActionSet
```

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/action/DropdownElementRenderer.java
```diff
@@ -82,14 +86,24 @@ public Button render(RenderedAdaptiveCard renderedCard, Context context, Fragmen
...
+            ActionElementRendererIconImageLoaderAsync imageLoader = new ActionElementRendererIconImageLoaderAsync(
+                renderedCard,
+                dropDownItem,
+                hostConfig.GetImageBaseUrl(),
+                iconPlacement,
+                hostConfig.GetActions().getIconSize(),
+                hostConfig.GetSpacing().getDefaultSpacing(),
+                context
+            );
+            imageLoader.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR, iconUrl);
```
golddove, line 106: I believe this is identical to code in ActionElementRenderer. Consider moving it into a helper within ActionElementRenderer and calling it, instead of duplicating.

golddove (approved):
Some suggestions

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/action/DropdownElementRenderer.java
karthikgopal36: moved to helper class.

PR #5813, opened 2021-05-12
[Schema] Add missing `id` to Actions

### LINKED ISSUE #5703: [Schema] Action.id docs
# Platform

- [ ] .NET HTML
- [ ] .NET WPF
- [ ] Android
- [ ] iOS
- [x] JavaScript
- [ ] UWP

# Author or host

Both author and host

Our own platform (Twist)

# Version of SDK

2.9.0

# Details

Not sure if this is a bug, missing documentation, or another of these "special nodejs only features" that I keep finding. In our integrations, we are utilising the `id` field on an [`Action.Submit`](https://adaptivecards.io/explorer/Action.Submit.html) but looking at the spec, there is no `id` field. However, if you go to the designer and add a new Action.Submit element, you get the id field as something you can fill in. Same applies to `Action.Execute`.
image

So is this another nodejs only "unsupported" feature?

paulcam206:
# Related Issue

Fixes #5703

# Description

Small schema change to add `id` property to Actions. `id` is already implemented everywhere, we just forgot to document it.

# How Verified

* local build and verification

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5813)


jonmill (approved):
LGTM!

PR #5849, opened 2021-05-19
Feature/is enabled file input color input

TemoeBOUISSOU:
# Description

- [JS, Dotnet] Add support for isEnabled property on Choice in ChoiceSetInput
- [JS, Dotnet] Add support for inlineAction (ToggleVisibility) inside Choice in ChoiceSetInput (multiSelect and unique selection)
- [JS, Dotnet] Add support for Input.File
- [JS, Dotnet] Add support for Input.Color
- [JS] Updated submit action result on adaptivecards-designer-app

Please note that i used base64 string to send the file, i'm pretty sure i'm wrong on that, so please, may i have an explanation on how you see how it should be done. I have updated the dotnet package because i need it to be accessible on a BotFramework chatbot app. Also, i don't know if i did everything that is expected for pull-request like this one, please notify me if anything is wrong and how i should do it to satisfy requirements.

# Sample Card
1. Sample card for unique selection expanded or compact :
```
{
  "type": "AdaptiveCard",
  "version": "1.3",
  "body": [
    {
      "id": "ChoiceSet",
      "type": "Input.ChoiceSet",
      "choices": [
        {
          "title": "Choice1",
          "value": "Choice1",
          "inlineAction": {
            "targetElements": [
              {
                "elementId": "FileInput",
                "isVisible": true
              },
              {
                "elementId": "ColorInput",
                "isVisible": false
              }
            ]
          }
        },
        {
          "title": "Choice2",
          "value": "Choice2",
          "inlineAction": {
            "targetElements": [
              {
                "elementId": "ColorInput",
                "isVisible": true
              },
              {
                "elementId": "FileInput",
                "isVisible": false
              }
            ]
          }
        }
      ],
      "value": "",
      "isMultiSelect": false,
      "style": "expanded"
    },
    {
      "id": "FileInput",
      "type": "Input.File",
      "isVisible": false
    },
    {
      "id": "ColorInput",
      "type": "Input.Color",
      "isVisible": false
    }
  ],
  "actions": [
    {
      "type": "Action.Submit",
      "title": "Submit"
    }
  ]
}
```
Rendering :

unique-choice1
unique-choice2

2. Sample card for multiselect :
```
{
  "type": "AdaptiveCard",
  "version": "1.3",
  "body": [
    {
      "id": "ChoiceSet",
      "type": "Input.ChoiceSet",
      "choices": [
        {
          "title": "Choice1",
          "value": "Choice1",
          "inlineAction": {
            "targetElements": [
              {
                "elementId": "FileInput",
                "isVisible": true
              }
            ]
          }
        },
        {
          "title": "Choice2",
          "value": "Choice2",
          "inlineAction": {
            "targetElements": [
              {
                "elementId": "ColorInput",
                "isVisible": true
              }
            ]
          }
        }
      ],
      "value": "",
      "isMultiSelect": true,
      "style": "expanded"
    },
    {
      "id": "FileInput",
      "type": "Input.File",
      "isVisible": false
    },
    {
      "id": "ColorInput",
      "type": "Input.Color",
      "isVisible": false
    }
  ],
  "actions": [
    {
      "type": "Action.Submit",
      "title": "Submit"
    }
  ]
}
```
Rendering :
multi-select-one-selection
multi-select-all

Sample card with isEnabled on Choice :
```
{
  "type": "AdaptiveCard",
  "version": "1.3",
  "body": [
    {
      "id": "ChoiceSet",
      "type": "Input.ChoiceSet",
      "choices": [
        {
          "title": "Choice1",
          "value": "Choice1",
          "isEnabled": false,
          "inlineAction": {
            "targetElements": [
              {
                "elementId": "FileInput",
                "isVisible": true
              }
            ]
          }
        },
        {
          "title": "Choice2",
          "value": "Choice2",
          "inlineAction": {
            "targetElements": [
              {
                "elementId": "ColorInput",
                "isVisible": true
              }
            ]
          }
        }
      ],
      "value": "Choice1",
      "isMultiSelect": true,
      "style": "expanded"
    },
    {
      "id": "FileInput",
      "type": "Input.File",
      "isVisible": true
    },
    {
      "id": "ColorInput",
      "type": "Input.Color",
      "isVisible": false
    }
  ],
  "actions": [
    {
      "type": "Action.Submit",
      "title": "Submit"
    }
  ]
}
```

Rendering :
isEnabled

# How Verified

I have tested it in using adaptivecards-designer-app for nodejs package, AdaptiveCards.Sample.WPFVisualizer for the dotnet package and also tried with local npm package in WebChat and BotFramework Emulator local execution.


jonmill:
@TemoeBOUISSOU - thanks so much for the PR! In order to take new features, we need to understand how the features will work on every platform (Desktop, Mobile, etc) as well as understand the schema changes that drive the features. Can we engage further to spec out how each feature will work on every platform and go through our process to get these to move forward? Also, where are you looking to have these features rendered (ex: a custom app? Teams? Outlook?)?

I see that our contribution guidelines are *not* well documented and are not very clear, so I apologize for that. I'll work on clarifying those to make sure this is clarified going forward.

TemoeBOUISSOU:
> @TemoeBOUISSOU - thanks so much for the PR! In order to take new features, we need to understand how the features will work on every platform (Desktop, Mobile, etc) as well as understand the schema changes that drive the features. Can we engage further to spec out how each feature will work on every platform and go through our process to get these to move forward? Also, where are you looking to have these features rendered (ex: a custom app? Teams? Outlook?)?
> 
> I see that our contribution guidelines are _not_ well documented and are not very clear, so I apologize for that. I'll work on clarifying those to make sure this is clarified going forward.

Just to clarify, for now, i have updated only Dotnet (Libraries + WPF visualizer) and nodeJs (adaptiveCards + designer but only submit action result). These were parts that i was needing to update to make these behaviours working on WebChat and BotFramework-Emulator. For the needs, i need it to be available on BotFramework-Emulator in order to have a development environment ready for these behaviours, and also, i need it on WebChat because i am using BotFramework SDK (and so adaptivecards dotnet library) on a Dotnet server to expose a bot building cards with these behaviours.
I hope that i was clear enough on that part, do not hesitate if you have any other question. I was not planning on updating every platform as i was not aware on the dev process, but i can manage to make it working for every platform if needed. I didn't look at all projects inside the repository for now, so from here and what i have explained here, can you confirm what projects are missing in order to make the PR fulfill my needs and requirements.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3231,7 +3455,27 @@ export class ChoiceSetInput extends Input {
                 }
             }
 
-            input.onchange = () => { this.valueChanged(); }
+            input.onchange = () => {
+                this.valueChanged();
+                if (choice.inlineAction instanceof ToggleVisibilityAction) {
```
dclaux, line 3460: I'm not getting this. The `onchange` handler should only do this:
```typescript
input.onchange = () => {
    this.valueChanged();

    if (choice.inlineAction) {
        choice.inlineAction.execute();
    }
}
```

You've already said that this can basically only be a ToggleVisibilityAction in the definition of the inlineAction property, and there is no reason to dynamically change the list of target elements.

`ToggleVisibilityAction` works the following way:
- If a target element is defined as just a string, then the action will hide the element if it is visible and show it if it is hidden
- If a target element's `isVisible` property is explicitly defined, then the action will set the visibility of the element accordingly

In other words, to actually "toggle" the visibility of an element, the associated `targetElement` must not have its `isVisible` property set.

If I apply this principle to the test payload you provided, a card author would have to do the following:

If the ChoiceSet is in multiselect mode:
```
{
  "type": "AdaptiveCard",
  "version": "1.3",
  "body": [
    {
      "id": "ChoiceSet",
      "type": "Input.ChoiceSet",
      "choices": [
        {
          "title": "Choice1",
          "value": "Choice1",
          "inlineAction": {
            "targetElements": [
              {
                "elementId": "FileInput"
              }
            ]
          }
        },
        {
          "title": "Choice2",
          "value": "Choice2",
          "inlineAction": {
            "targetElements": [
              {
                "elementId": "ColorInput"
              }
            ]
          }
        }
      ],
      "value": "",
      "isMultiSelect": true,
      "style": "expanded"
    }
  ]
}
```

If the ChoiceSet is NOT in multiselect mode:
```
{
  "type": "AdaptiveCard",
  "version": "1.3",
  "body": [
    {
      "id": "ChoiceSet",
      "type": "Input.ChoiceSet",
      "choices": [
        {
          "title": "Choice1",
          "value": "Choice1",
          "inlineAction": {
            "targetElements": [
              {
                "elementId": "FileInput",
		"isVisible": true
              },
              {
                "elementId": "ColorInput",
		"isVisible": false
              }
            ]
          }
        },
        {
          "title": "Choice2",
          "value": "Choice2",
          "inlineAction": {
            "targetElements": [
              {
                "elementId": "FileInput",
		"isVisible": false
              },
              {
                "elementId": "ColorInput",
		"isVisible": true
              }
            ]
          }
        }
      ],
      "value": "",
      "isMultiSelect": true,
      "style": "expanded"
    }
  ]
}
```

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2953,6 +2953,208 @@ export class TextInput extends Input {
...
+      }
+
+      input.oninput = () => {
+          this.valueChanged();
+          let fileName = (<HTMLInputElement>this.renderedInputControlElement).value;
+          let file = (<HTMLInputElement>this.renderedInputControlElement).files?.item(0) as File;
+          let fileReader = new FileReader();
+          fileReader.onload = () => {
+              let content = fileReader.result as string;
+              this.defaultValue = new FileResult(fileName, content);
```
dclaux, line 3028: For me this is a big problem with this implementation - the file is entirely read, converted to base64 and stored in memory. Then it is passed as a bsed64 encoded string to the backend when an Action.Submit or Action.Execute is executed. I don't think that scales beyond very small files.

IMO the right model here is more complicated. We need to have a way to let the host application get access to the byte stream of the selected file at the time the action is executed to allow it to handle the transfer of the bytes in an optimal way, which will most likely involve in many cases sending the file in chunks in order to never have to read it all into memory. The problem obviously is that making all of this work requires explicit client support, which currently doesn't exist.

This implementation definitely has the benefit of working with WebChat/Bots "as is" but I don't believe it is the right solution.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2953,6 +2953,208 @@ export class TextInput extends Input {
...
+  }
+}
+
+export class ColorInput extends Input {
+  //#region Schema
+
+  static readonly valueProperty = new StringProperty(Versions.v1_0, "value");
+  static readonly maxLengthProperty = new NumProperty(Versions.v1_0, "maxLength");
+  static readonly placeholderProperty = new StringProperty(Versions.v1_0, "placeholder");
+  static readonly styleProperty = new EnumProperty(Versions.v1_0, "style", Enums.InputTextStyle, Enums.InputTextStyle.Color);
```
dclaux, line 3080: `Enums.InputTextStyle` is used to specify the style of the otherwise generic Input.Text.

If you're going to add new File and Color styles to the enum, these will be available on TextInput itself, but it won't really work. So you have two choices:
- Either you do add the enum values but then:
  - You have to make TextInput honor them (and thus you have to update TextInput itself to implement the logic you currently have in your FileInput and ColorInput classes)
  - And thus you remove you FileInput and ColorInput classes
- Or:
  - You don't add the enum values
  - You don't declare a style property on either FileInput or ColorInput because it's no necessary (the classes themselves represent specialized inputs, no need to have a style property that repeats that)

To me there's no doubt the second option is the right one.

source/nodejs/adaptivecards/src/enums.ts
```diff
@@ -123,7 +123,9 @@ export enum InputTextStyle {
     Text,
     Tel,
     Url,
-    Email
```
dclaux, line 126: As per my comment above, this isn't desirable, those values shouldn't be added.

dclaux (changes requested):
Thank you so much for your contribution. This is not a full code review nor an indication that we are ready to merge your changes after you make the suggested updates - please see @jonmill's comments. This is just in order to suggest improvements to your current implementation.

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -332,7 +332,15 @@ export class CardDesignerSurface {
 
         if (this.isPreviewMode) {
             cardToRender.onExecuteAction = (action: Adaptive.Action) => {
+              if (action instanceof Adaptive.SubmitAction) {
```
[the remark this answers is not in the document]
TemoeBOUISSOU: Maybe i used it wrong, but i don't see the data payload from submit action, that is why i worked on it like that here.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3098,24 +3300,45 @@ export class Choice extends SerializableObject {
...
+    inlineAction?: Action;
+
     //#endregion
 
     protected getSchemaKey(): string {
         return "Choice";
     }
 
-    constructor(title?: string, value?: string) {
+    protected isDesignMode(): boolean {
```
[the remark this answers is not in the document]
TemoeBOUISSOU: Appeared to be mandatory in order to be able for me to test it in designer-app. So is still here, maybe i have another way to make it work without it ? I have "is not function" error without it.

source/nodejs/adaptivecards/src/card-elements.ts
TemoeBOUISSOU: I did some little changes on your suggestion, but thank you for that, i think i got lost trying to find a solution here !

source/nodejs/adaptivecards/src/card-elements.ts
TemoeBOUISSOU: Totally agree with you, i had no choice there for a first version of that feature, i saw in the upload process of botframework-emulator that files are stored as attachments. I don't really have all the spec on that part and decided to make a simple start. Still have nothing more to tell on how to change that part for now.

source/nodejs/adaptivecards/src/card-elements.ts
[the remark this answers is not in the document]
dclaux: `isDesignMode` is a method of CardElement and is only ever called on a CardElement (or descendant class) instance. So it will never be called here.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3305,9 +3280,8 @@ export class Choice extends SerializableObject {
         Versions.v1_0,
         "inlineAction",
         [
-            "Action.ShowCard",
-            "Action.Submit",
-            "Action.OpenUrl"
+            "*",
+            "-Action.ToggleVisibility"
```
dclaux, line 3284: Sorry if I was unclear - this won't just work. It's the right syntax, but it is not currently supported. It would have to be implemented in `SerializationContext.internalParseCardObject`, which might be a bit outside the scope of this PR. Feel free to revert to what you had before for now, but do include Action.Execute in the list.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3457,23 +3431,10 @@ export class ChoiceSetInput extends Input {
...
-                            targetElement = input.checked;
-                        }
-
-                        newTargetElements[elementId] = targetElement;
-                    }
 
-                    toggleVisibilityAction.targetElements = newTargetElements;
-                    toggleVisibilityAction.setParent(this);
-                    toggleVisibilityAction.execute();
+                if (choice.inlineAction) {
```
dclaux, line 3435: You removed the check that `inlineAction` is of type `ToggleVisibilityAction` which is fine, that said if `inlineAction` is set in code it can be set to another type of action given that `inlineAction` is defined as being of type `Action`. Maybe change the type of the field to `ToggleVisibilityAction`, given that's all it's supposed to be?

dclaux (changes requested):
Thank you for making these changes. There are a couple more things to look at.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Also, you do have to check for `input.checked` here but only when the input in in single select expanded mode. Otherwise what's going to happen is the `inlineAction` of both the newly selected radio button and that of the one that isn't selected anymore are going to be triggered. Given that the `inlineAction` as defined in the card is meant to be executed only when the radio button is checked, that's going to yield the wrong result.

dclaux:
@TemoeBOUISSOU quick update - we're still discussing this and will get back to you next week.

TemoeBOUISSOU:
Hello, i was in holidays i couldn't work on it lately, i'm sorry for that ! Thank you for your answers, i tried to fix things like you explained but i'm having trouble on the `isDesignMode` method from `Choice` class. 
I have found where the problem is but don't know how i should manage that problem. I would have to change something that is applied generically to all `Action` :
isDesignMode error

debug_value

card-elements ts - nodejs - Visual Studio Code

That is why i let the `isDesignMode` implementation on the `Choice` class for now otherwise i wouldn't be able to test my code because of the error.

dclaux:
@TemoeBOUISSOU ah, I see. The problem actually goes well beyond `isDesignMode()`. `ActionProperty.parse` currently assumes that the object (`sender`) on which the action is being parsed is a `CardElement` instance, which is no the case here (it's a `Choice`). It is necessary for an `Action` instance to have a parent of type `CardElement` for things to work properly. `Choice` is not a `CardElement`, so the `let parent = <CardElement>sender;` cast in `ActionProperty.parse` actually hides a bigger bug where the parsed action will have a parent of type `Choice`.

Your initial implementation worked because:
- You artificially added the `isDesignMode()` method on `Choice`
- Your handler for `Choice.inlineAction` duplicated the `ToggleVisibility` logic and was able to get to the target elements via the ChoiceSet's parent

But that's a workaround, we need an actual solution. That solution likely involves:
- Updating the `Choice` class so it has an `owner` property of type `ChoiceSet`
- Making sure that all choices added to a `ChoiceSet` (either during parsing or in code via the API) have their `owner` property set properly to that `ChoiceSet`. That probably means changing the actual type of `ChoiceSet.choices` from `Choice[]` to a class that extends `Array<Choice>` and handles setting `owner` while maintaining backward compatibility with the current model in order to not break existing applications
- Not using the `ActionProperty` class and instead creating a new `ChoiceInlineActionProperty` whose `parse` method would look like this:

```
parse(sender: SerializableObject, source: PropertyBag, context: SerializationContext): Action | undefined {
    let parent = (<Choice>sender).owner;

    return context.parseAction(
        parent,
        source[this.name],
        this.forbiddenActionTypes,
        parent.isDesignMode());
}
```

That last bit isn't as generic as I would like, but it's clean. There might be a better way.

TemoeBOUISSOU:
> @TemoeBOUISSOU ah, I see. The problem actually goes well beyond `isDesignMode()`. `ActionProperty.parse` currently assumes that the object (`sender`) on which the action is being parsed is a `CardElement` instance, which is no the case here (it's a `Choice`). It is necessary for an `Action` instance to have a parent of type `CardElement` for things to work properly. `Choice` is not a `CardElement`, so the `let parent = <CardElement>sender;` cast in `ActionProperty.parse` actually hides a bigger bug where the parsed action will have a parent of type `Choice`.
> 
> Your initial implementation worked because:
> 
> * You artificially added the `isDesignMode()` method on `Choice`
> * Your handler for `Choice.inlineAction` duplicated the `ToggleVisibility` logic and was able to get to the target elements via the ChoiceSet's parent
> 
> But that's a workaround, we need an actual solution. That solution likely involves:
> 
> * Updating the `Choice` class so it has an `owner` property of type `ChoiceSet`
> * Making sure that all choices added to a `ChoiceSet` (either during parsing or in code via the API) have their `owner` property set properly to that `ChoiceSet`
>   
>   * That probably means changing the actual type of `ChoiceSet.choices` from `Choice[]` to a custom collection class that handles setting `owner` whenever a choice is added
>   * Problem is that will likely be a breaking change
> * Not using the `ActionProperty` class and instead creating a new `ChoiceInlineActionProperty` whose `parse` method would look like this:
> 
> ```
> parse(sender: SerializableObject, source: PropertyBag, context: SerializationContext): Action | undefined {
>     let parent = (<Choice>sender).owner;
> 
>     return context.parseAction(
>         parent,
>         source[this.name],
>         this.forbiddenActionTypes,
>         parent.isDesignMode());
> }
> ```
> 
> That last bit isn't as generic as I would like, but it's clean.
> 
> This is all off-the-top-of-my-head and not thought through though. There might be a better way.

Can you please elaborate on how i do the custom collection ? I tried with `SerializableObjectCollectionProperty` but i can't manage to get the right parent in the parsing. Also i tried with `CustomProperty<PropertyBag>`, i managed to get it to be parsed (but it's not calling `ChoiceInlineActionProperty` and i don't know how to), but the parent is missing a lot of properties, like i'm having a serialization problem. Can you please help me @dclaux ? That's the only thing i'm missing right now to make it work as you said.

PR #5867, opened 2021-05-24
[iOS] Fixed #5559

### LINKED ISSUE #5559: [iOS][Sample App] [Focusing after input validation failure is not correct]
# Platform

- [ ] iOS

# Details
Sample's app's content offset calculation logic needs to be re-visited after recent sample app changes.

jwoo-msft:
# Related Issue

Fixed #5559 

# Description

Sample app refreshes the row using the visible row property of a UIViewTable when input validation returns the result.
The visible row values are unreliable since the value gets cleared when the table is reloaded. Changed to begin and endUpdates calls.

# Sample Card

# How Verified

Manually verified.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5867)


paulcam206 (approved):
:shipit:

PR #5946, opened 2021-06-01
[iOS] disable selection action is implemented

### LINKED ISSUE #5718: [iOS] Action isEnabled

jwoo-msft:
# Related Issue

Fixed #5718 

# Description

Disabled actions if `isEnabled` set false with one exception with RichTextBlock.
There is no way to disable the link section of the text, so prevented adding the selection action to the text if not enabled.

# How Verified

Tested it with the new sample card.
Verified with old cards.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5946)


paulcam206 (approved):
:shipit:

PR #5960, opened 2021-06-02
Users/clintwar/handletypeinfo

### LINKED ISSUE #2148: Bot Builder v4 IgnoreEmptyItemsConverter causes an issue when ReadJson
# Platform
Bot Builder v4.0 (dotnet core)

# Author or host
web api (Bot Framework)

# Version of SDK
1.1

# Issue
When deserialize the DialogState which contains the AdaptiveCard, it fails with error "Error reading JArray from JsonReader at 

var array = JArray.Load(reader);. 

Error: Current JsonReader item is not an array: StartObject. Path 'DialogState.DialogStack.$values[0].State.dialogs.DialogStack.$values[0].State.options.Prompt.attachments.$values[0].content.action"

This caused by [JsonConverter(typeof(IgnoreEmptyItemsConverter<AdaptiveAction>))] attribute on AdaptiveCard which will be called while deserialization but it's not actually an array yet, as $value holds an array not action (nor body)

```json
"actions": {
    "$type": "System.Collections.Generic.List`1[[AdaptiveCards.ActionBase, AdaptiveCards]], System.Private.CoreLib",
    "$values": [
    {
        "$type": "AdaptiveCards.SubmitAction, AdaptiveCards",
        "type": "Action.Submit",
        "data": "Test",
        "title": "Test"
    }
    ]
}
```
We should try/catch JArray.Load

clintwar:
fixes #2148 

# Description

If $type information is included in serialized content the reader will be at start of an object. Instead of loading into JArray load into JToken and test to see if it is JObject. If it is set token to $values token. if there isn't one it will handle correctly and leave for instances when it is null.

# How Verified

Added new test Test_TypeHandling
Since $type is included and AdditionalProperties is decorated with JsonExtensionData it gets stored there. Test removes this form expected elements and compares to ensure that the structure of the card is the same.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5960)


paulcam206 (approved):

clintwar:
The build pipeline seems to keep failing on the Android-CI check. Is there anything that can be done to correct this?

jonmill:
@RebeccaAnne - can you take a quick look and see if this looks good to you too? I think this PR hit our Android CI issue 2 weeks ago. if it can be merged, the Android CI should be unblocked

clintwar:
@jonmill  and @RebeccaAnne It seems that the checks continue to fail and that is what is holding up merging of this PR. Looks like now it is IOS instead of Android.

golddove:
@clintwar checks are all passed!

jonmill:
@golddove - Thank you! Can you also give this a final review so we have 2 folks with eyes on it?

source/dotnet/Test/AdaptiveCards.Test/SerializationTests.cs
```diff
@@ -256,6 +256,95 @@ public void TestStyleNullDeserialization()
...
+            // make card into JObject with types included
+            JObject cardObject = JObject.FromObject(card, new Newtonsoft.Json.JsonSerializer()
+            {
+                TypeNameHandling = Newtonsoft.Json.TypeNameHandling.All
+            });
+
+            // now bring it back
+            AdaptiveCard card2 = cardObject.ToObject<AdaptiveCard>();
+
+            // card2 will now have AdditionalProperties because $type is not known and it seems $type is not ignored by Newtonsoft JsonExtensionData
```
golddove, line 301: @clintwar should this be considered a bug we should look into?

golddove (approved):

source/dotnet/Test/AdaptiveCards.Test/SerializationTests.cs
clintwar: @golddove not sure I would call it a bug, since AdaptiveTypeElement has AdditionalProperties with JsonExtensionData attribute anything not defined gets thrown into it. $type is included by Newtonsoft when you serialize but I do not see any good means to ignore since there is not a good way to differentiate things included by the serialization vs things introduced in the json directly.

PR #5964, opened 2021-06-03
Fix issue with dotnet template null handling

### LINKED ISSUE #5963: [.NET][Templating] Null Substitution not working
# Target Application

Rendering in nodejs and/or Outlook, created using dotnet (AdaptiveCards.Templating 1.2.1, AdaptiveCards 2.7.1)

# Problem Description

Trying to render cards with null substitution, but not able to do so:

`string cardJson = template.Expand(data, (name) => "something");`

# Expected Outcome

The card should have the substitution text rendered (in this case "something")

# Actual Outcome

The tag is displayed like `${NullTagThatShouldBeReplaced}`

# Repro Steps

```
            AdaptiveCardTemplate template = new(text);
            var data = new
            {
                NullTagThatShouldBeReplaced= null
            };

            string cardJson = template.Expand(data, (name)=> "something");

            AdaptiveCardParseResult parseResult = AdaptiveCard.FromJson(cardJson);
            AdaptiveCard adaptiveCard = parseResult.Card;

            return adaptiveCard.ToJson();
```

# Schema Version

1.0

Greg-Hitchon:
# Related Issue

Fixes #5963

# Description

It seems the null substitution options were being overwritten within the `AdaptiveCardsTemplateVisitor` class.

# How Verified

Added two unit tests for the cases with/without null substitution.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5964)


source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
```diff
@@ -611,10 +612,13 @@ public static string Expand(string unboundString, IMemory data, bool isTemplated
                 return unboundString;
             }
 
-            var options = new Options
```
jwoo-msft, line 614: can you remove https://github.com/microsoft/AdaptiveCards/pull/5964/files#diff-75a381746eb6256f934ad5e0d008c5d87268fd74cd97162e6cd385f69c5ddd7fL614-L617 instead?

options are already initialized here
https://github.com/microsoft/AdaptiveCards/pull/5964/files#diff-75a381746eb6256f934ad5e0d008c5d87268fd74cd97162e6cd385f69c5ddd7fL124-L127

jwoo-msft (changes requested):

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
Greg-Hitchon: `Expand` is `static` though, which I think is causing the issue.

jwoo-msft: https://github.com/microsoft/AdaptiveCards/pull/5964/files/f7d9938a7333bbe5adf8ce1e88d13f2277534569#diff-75a381746eb6256f934ad5e0d008c5d87268fd74cd97162e6cd385f69c5ddd7fR615-R621. These lines seem to be redundant. I guess it won't hurt.

jwoo-msft (approved):

anna-dingler:
@paulcam206 Is it possible to get these changes in 1.2.1 as well?

https://github.com/microsoft/AdaptiveCards/issues/7210

PR #5969, opened 2021-06-07
[Shared] Only validate non-empty locales

paulcam206:
# Description

`std::locale::locale` was showing up as the top hot path during some perf measuring I was doing. The intent of the code in question is to validate the language identifier supplied in the `AdaptiveCard` `lang` property, which accepts 2-letter language codes according to [our docs](https://adaptivecards.io/explorer/AdaptiveCard.html). Since the `lang` property isn't super common, we should avoid instantiating `std::locale` unless a language was actually supplied.

# How Verified

* local build/unit tests

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5969)

### DIFF
--- a/source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
+++ b/source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
@@ -147,7 +147,11 @@ void AdaptiveCard::_ValidateLanguage(const std::string& language, std::vector<st
 {
     try
     {
-        if (language.empty() || language.length() == 2 || language.length() == 3)
+        if (language.empty())
+        {
+            // no need to validate locale
+        }
+        else if (language.length() == 2)
         {
             auto locale = std::locale(language.c_str());
         }


source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
```diff
@@ -147,7 +147,11 @@ void AdaptiveCard::_ValidateLanguage(const std::string& language, std::vector<st
 {
     try
     {
-        if (language.empty() || language.length() == 2 || language.length() == 3)
+        if (language.empty())
```
almedina-ms, line 150: from the c++ documentation, leaving the locale empty retrieves the user preferred locale, if we can initialize once the user preferred locale we should only deal with this problem once

https://en.cppreference.com/w/cpp/locale/locale#:~:text=An%20object%20of%20class%20std%3A%3Alocale%20is%20an%20immutable,locale%20object%20is%20associated%20with%20each%20std%3A%3Abasic_regex%20object.

PR #5993, opened 2021-06-10
[iOS] updated intrinsic content size for UICollectionView

jwoo-msft:
# Related Issue
Image collection view returns zero intrinsic content size when the super view's width is zero. 

# Description
iOS AdaptiveCards width constraint is no longer required, and the intrinsic content size can be zero because of that. This PR returns the intrinsic content size as if all images are laid in a single row if the width is not yet determined. When the width is determined, a new intrinsic content size is calculated and ask for the layout engine to update the layout.

# Sample Card
https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.0/Elements/ImageSet.json
https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.0/Scenarios/ImageGallery.json
https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.2/Scenarios/Agenda.json

# How Verified

1. run existing sample cards

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5993)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRImageSetUICollectionView.mm
```diff
@@ -105,26 +104,37 @@ - (CGSize)intrinsicContentSize
...
-    int numbersOfItemsInRow = frameWidth / imageWidthWithSpacing;
+    NSUInteger numbersOfItemsInRow = (frameWidth) ? (frameWidth) / imageWidthWithSpacing : cellCounts;
     // if addtional image can be fit by removing spacing, do so
     if (numbersOfItemsInRow * imageWidthWithSpacing + imageSize.width <= frameWidth) {
         numbersOfItemsInRow++;
     }
 
-    int numbersOfRows = ceil(((float)cellCounts) / numbersOfItemsInRow);
-    return CGSizeMake(self.frame.size.width, (numbersOfRows) * (imageSize.height) + (numbersOfRows - 1) * lineSpacing);
+    NSUInteger numbersOfRows = ceil(((CGFloat)cellCounts) / numbersOfItemsInRow);
```
paulcam206, line 126: is there a risk that `numbersOfItemsInRow` could be `0`?

paulcam206 (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRImageSetUICollectionView.mm
jwoo-msft: thanks made the `numberOfItemsInRow` to be at least 1.

PR #5999, opened 2021-06-11
[Android] Update project file to setup UI Testing

### LINKED ISSUE #5916: [Android] Update project to support UI Tests

almedina-ms:
# Related Issue

Fixes #5916

# Description

Updates the android project to set up the environment for the development of the ui test app and tests
* Declare a new project application
* Update the build.gradle file to include the espresso library
* Create mock test that always succeed

# Sample Card

No sample card

# How Verified

Local build

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/5999)


source/android/uitestapp/src/AdaptiveCardObjectModel.i
```diff
@@ -0,0 +1,217 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
```
paulcam206, line 1: Is there no way to consume the other `.i` file here? it'd be a bummer if we have to keep them both up to date...

source/android/uitestapp/src/androidTest/java/io.adaptivecards.uitestapp.testing/UiTests.java
```diff
@@ -0,0 +1,27 @@
+package io.adaptivecards.uitestapp.testing;
```
paulcam206, line 1: missing copyright banner

source/android/uitestapp/src/main/res/SimpleAdaptiveCards.json
```diff
@@ -0,0 +1,13 @@
+{
+	"@type": "AdaptiveCard",
```
paulcam206, line 2: `@type`?

source/android/uitestapp/src/AdaptiveCardObjectModel.i
golddove: +1, the object model should be included by depending on the `adaptivecards` module, right?

golddove (commented):
If I understand correctly, assets are duplicated from the sample app. I believe there's a way to link assets by path, to avoid having the same asset in two locations. It would be nice to not have to maintain two copies :)

source/android/uitestapp/src/AdaptiveCardObjectModel.i
golddove: +1, the object model should be included by depending on the adaptivecards module, right?

golddove (approved):

golddove (commented):
Looks good

golddove (approved):

PR #6000, opened 2021-06-13
[JS] visibility not correctly applied to images inside an ImageSet (#3251)

### LINKED ISSUE #3251: [JS][1.2] isVisible set to false on image in the imageSet doesn't work
# Platform

* JavaScript

# Version of SDK

1.2

# Issue

isVisible set to false doesn't work.

```html
<!-- cdn -->
<script src="https://unpkg.com/adaptivecards/dist/adaptivecards.min.js"></script>
```
```javascript
// card
{
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
      {
          "type": "ImageSet",
          "images": [
              {
                  "type": "Image",
                  "url": "http://adaptivecards.io/content/cats/1.png",
                  "size": "Medium",
                  "isVisible": false
              },
              {
                  "type": "Image",
                  "url": "http://adaptivecards.io/content/cats/1.png",
                  "size": "Medium"
              },
              {
                  "type": "Image",
                  "url": "http://adaptivecards.io/content/cats/1.png",
                  "size": "Medium"
              },
              {
                  "type": "Image",
                  "url": "http://adaptivecards.io/content/cats/1.png",
                  "size": "Medium"
              },
              {
                  "type": "Image",
                  "url": "http://adaptivecards.io/content/cats/1.png",
                  "size": "Medium"
              },
              {
                  "type": "Image",
                  "url": "http://adaptivecards.io/content/cats/1.png",
                  "size": "Medium"
              }
          ]
      }
  ]
}
```

render:
Capture2

ChristoperHowell:
# Related Issue
Fixes #3251 

# Description
It has been identified that the visibility property of images held within an ImageSet was not being properly applied when rendering the card. 

# How Verified
1) Built locally
2) Launched the designer and created a card with an ImageSet with three images
3) Tested various configurations of these images with the visibility flag checked and unchecked
4) Added a toggle visibility action and tested again with various configurations of the image visibility flag

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -2161,7 +2161,11 @@ export class ImageSet extends CardElementContainer {
                 let renderedImage = image.render();
 
                 if (renderedImage) {
-                    renderedImage.style.display = "inline-flex";
+
+                    if(image.isVisible){
+                        renderedImage.style.display = "inline-flex";
+                    }
+                    
                     renderedImage.style.margin = "0px";
                     renderedImage.style.marginRight = "10px";
 


ChristoperHowell:
So I have identified where the problem is occurring as you can see in my commit which does fix the issue. The core problem is that after the render method on the element is called, which is when the display property is set, an updated display property is then being added to the element without considering if the visibility is true or false. 

I'm open to discussion on if this is the best fix for this issue?

PR #6022, opened 2021-06-20
5973/update teams v1.4

ChristoperHowell:
# Related Issue

5973

# Description

MS Teams now supports up to version 1.4 of adaptive cards. However, the designer did not reflect this and has been updated so that version 1.4 is now the default card version when selecting MS Teams host containers.

# How Verified

1. Built locally
2. Open the designer and selected light/dark teams container
3. Created a card an added the new ActionExecute button to the card
4. Copied the card JSON and confirmed it was version 1.4


ChristoperHowell:
The original ticket that requested this change made referenced to ensuring that version 1.4 is also supported in the Outlook container. On the main branch this is already the supported version in the Outlook container, so I simply updated the MS Teams container.

jonmill:
User no longer exists

PR #6023, opened 2021-06-21
[UWP] Implementation of filtered choice sets

### LINKED ISSUE #5988: [UWP] Input.ChoiceSet static filtering

RebeccaAnne:
# Related Issue

Fixes #5988

# Description

Implemented Input.Choice set using XAML's AutoSuggestBox. Also refactored the ChoiceSetInputValue by splitting it into three subclasess (CompactChoiceSetInputValue, ExpandedChoiceSetInputValue, and FilteredChoiceSetInputValue). The three types shared very little code and this refactor reduced the switching in the class to handle each type differently.

# Sample Card

[Input.ChoiceSet.Filtered.json](https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.5/Elements/Input.ChoiceSet.Filtered.json)

# How Verified

Confirmed expected behavior using above sample card.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6023)


source/uwp/Renderer/lib/InputValue.cpp
```diff
@@ -537,112 +541,201 @@ try
...
+    IterateOverVector<AdaptiveChoiceInput, IAdaptiveChoiceInput>(choices.Get(), [&selectedChoice, text](IAdaptiveChoiceInput* choice) {
+        HString titleHString;
+        RETURN_IF_FAILED(choice->get_Title(titleHString.GetAddressOf()));
 
-        boolean isFocused;
-        RETURN_IF_FAILED(choiceAsControl->Focus(FocusState_Programmatic, &isFocused));
-    }
+        std::string title = HStringToUTF8(titleHString.Get());
+
+        if (!ParseUtil::ToLowercase(text).compare(ParseUtil::ToLowercase(title)))
```
almedina-ms, line 730: I don't think we talked about case sensitiveness during the spec review, I believe making the query case insensitive will provide a better user experience, but we may have to raise this detail to achieve parity in all platforms

source/uwp/Renderer/lib/InputValue.cpp
```diff
@@ -537,112 +541,201 @@ try
...
 
-        ComPtr<IVector<UIElement*>> panelChildren;
-        RETURN_IF_FAILED(panel->get_Children(panelChildren.ReleaseAndGetAddressOf()));
+    return S_OK;
+}
+
+HRESULT FilteredChoiceSetInputValue::GetSelectedChoice(IAdaptiveChoiceInput** adaptiveChoiceInput)
+{
+    HString textHString;
+    RETURN_IF_FAILED(m_autoSuggetBox->get_Text(textHString.GetAddressOf()));
```
almedina-ms, line 717: typo: m_autoSuggestBox

almedina-ms (approved):

source/uwp/Renderer/lib/InputValue.cpp
RebeccaAnne: Good catch, fixed!

source/uwp/Renderer/lib/InputValue.cpp
RebeccaAnne: I sent a mail to raise awareness of this.

PR #6045, opened 2021-07-01
[iOS] updated container style's color scheme for accessibility

jwoo-msft:
# Related Issue

Fixed accessibility issue, https://microsoft.visualstudio.com/OS/_queries/edit/29253854/?triage=true.

# Description

Updated the container style's color scheme values for better contrast.

# Sample Card

https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.0/Elements/Container.Style.json

# How Verified
verified all text present in the sample card using CCA.
Screen Shot 2021-07-01 at 2 08 25 AM

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6045)


almedina-ms (approved):

source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
```diff
@@ -50,23 +50,23 @@
 					"subtle": "#6f6f6f"
```
almedina-ms, line 50: why is this sample in the ADCIOSVisualizer project? Is this the same file as the one located in the HostConfigs folder in the samples? do both samples have the same colors?

jwoo-msft: @almedina-ms, iOS visualizer has its own host configs.

PR #6060, opened 2021-07-07
Accessibility fix for inline-actions on inputs

frags51:
The inline-action buttons had no role or name assigned previously. This PR adds the name, role and disabled state.
Also concatenates the Label and Asterisk in a single narration.

Also concatenate the fact set label and value into a single announcement.
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6060)


frags51:
@vivekvijayakrishnan / @IbrahimSulai Pls suggest if I am missing similar additions to any other place also - this was the only place I was able to find but I may be missing something.

IbrahimSulai (approved):
LGTM

IbrahimSulai (approved):
LGTM

PR #6079, opened 2021-07-13
Replace bespoke antlr json parser with JSON.NET fixing bugs and reducing library size by 10x

### LINKED ISSUE #6078: [Templating] AdaptiveCard.Templating should not be using custom json antlr parser
# Target Platform
- dotnet

# Problem Description

If you do a 
```json 
{  
      "$when" : "x > 3"
}
```
the templating engine will output incorrect JSON 
```json 
{  
       "x > 3"
}
```

Looking at the code I was surprised to see an custom antlr generated parser for JSON.  This despite the fact that  the dotnet libraries already have a dependency on JSON.NET.  The current implementation of the templating library is over 200k.  AdaptiveExpressions already has a parser to handle interpolated strings, as well as the ability to correctly return native types (via = syntax) and interpolated syntax ( ${path}  syntax ).

tomlm:
Fixes #6078 

This PR replaces the 5000 lines of bespoke antlr code with a a simple class of ~hundred lines of code which uses the highly tested and performant JSON.NET DOM to manipulate the json data structures.

All unit tests pass.  

NOTE: Some of the unit tests were validating incorrect behavior (aka the old code if you had "text":"${counter}" and counter was a number would emit invalid JSON of "text": 1, instead of "text":"1".   With AdaptiveExpressions if you want the string you do "${counter}" if you want the underlying type you should use "=counter".  Same with DateTime data structures.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6079)


jonmill:
@RebeccaAnne - FYI

jonmill:
@paulcam206 - can you take a look please

paulcam206:
taking a look -- @jwoo-msft and @almedina-ms should take a look too :)

paulcam206 (approved):
just a couple minor nits -- looks good!

source/dotnet/Test/AdaptiveCards.Templating.Test/TestTransform.cs
```diff
@@ -11231,7 +11240,7 @@ public void TestIntInExpression()
                     ""placeholder"": ""Enter Estimated Mileage"",
                     ""min"": 1,
                     ""max"": 10,
-                    ""value"": ""${car_type_1.average_mileage}"" 
+                    ""value"": ""=car_type_1.average_mileage""
```
jwoo-msft, line 11243: Can this create a backward compatibility issue? '=' is not formally introduced to the users. The users have been able to do the integer substitution without the new operator.

RebeccaAnne:
@tomlm Please reopen if you have a chance to follow up with Joseph's questions. Thanks!

PR #6084, opened 2021-07-14
Handled the bleed property

IbrahimSulai:
Handled the bleed property. On the web, it's not having conditions for the first and last elements. so removed this condition from the react-native.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6084)


pragadeeshk (approved):
LGTM

PR #6150, opened 2021-07-31
[UWP] Add support for action overflow

### LINKED ISSUE #5494: [UWP] Object model support for ActionSet overflow

### LINKED ISSUE #5495: [UWP] Renderer support for ActionSet overflow

RebeccaAnne:
# Related Issue

Fixes #5494 
Fixes #5495

# Description

Add support for action overflow to UWP. The majority of the code in this PR is in aid of supporting the spec'd show card behavior where a show card in an overflow menu needs to switch places with the last primary button when invoked. This required more logic and bookkeeping around show card actions.

As part of that bookkeeping, more information needed to be registered with the render context via AddInlineShowCard. Because of that, that function started to feel more implementation specific than felt right for the public API surface. With this change, hosts who are custom rendering Action**Sets** (not individual actions) will need to manage their own inline show cards. This seems like a reasonable requirement for a host custom rendering the entire action set. 

# Sample Card

Moved the existing sample card to the Test directory and added some more cases to it. Created a new sample card in elements to be the sample for the webpage. With this change the two sample cards for overflow are:

\samples\v1.5\Elements\Action.Mode.json
samples\v1.5\Test\ActionModeTestCard.json

# How Verified

- Updated and verified behavior in the two sample cards discussed above
- Updates to the object model unit tests to verify new properties
- Renderer test to validate OverflowMaxActions behavior

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6150)


almedina-ms:
> /* Header file automatically generated from AdaptiveCards.ObjectModel.Uwp.idl */

Should this file be added? It looks like it's a product of some compilation

---
Refers to: source/uwp/AdaptiveCardsObjectModel/AdaptiveCards.ObjectModel.Uwp.h:1 in cd79747. [](commit_id = cd797472b9c6f1508495df11ba4c9e738049ffab, deletion_comment = False)

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -48,6 +47,10 @@ namespace AdaptiveCards
                 Windows.UI.Xaml.ResourceDictionary OverrideStyles;
                 AdaptiveFeatureRegistration FeatureRegistration;
 
+                Boolean OverflowMaxActions;
```
almedina-ms, line 50: Is the overflow of actions only supposed to happen on the card level ActionSet (the counterpart to the card body)? Will other ActionSet elements inherit the same overflow text and accessibility texts?

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -319,6 +347,7 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
     HRESULT BuildAction(_In_ IAdaptiveActionElement* adaptiveActionElement,
                         _In_ IAdaptiveRenderContext* renderContext,
                         _In_ IAdaptiveRenderArgs* renderArgs,
+                        bool overflowAction,
```
almedina-ms, line 350: nit: rename to isOverflowAction, from the context of the method it looks like it's referring to the adaptiveActionElement being an overflowAction or not, also, I see the usage in line 914 sets adaptiveActionElement as nullptr, is there any case where the adaptiveActionElement is null and the overflowAction flag has false as value or the adaptiveActionElement is not null but the overflowAction flag is set tot true?

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -855,6 +907,260 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
...
+                HStringReference(RuntimeClass_Windows_UI_Xaml_Controls_ColumnDefinition));
+            RETURN_IF_FAILED(columnDefinitions->Append(columnDefinition.Get()));
+
+            if (buttonVisible)
+            {
+                RETURN_IF_FAILED(columnDefinition->put_Width({1.0, GridUnitType::GridUnitType_Star}));
+            }
+            else
+            {
+                RETURN_IF_FAILED(columnDefinition->put_Width({0, GridUnitType::GridUnitType_Auto}));
```
almedina-ms, line 1141: >GridUnitType_Auto [](http://example.com/codeflow?start=79&length=17)

is Auto the same as Pixel?

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -955,139 +1271,198 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
...
+            // allows it to be moved to the overflow menu to make space for a show card action.
+
+            if (lastPrimaryAction != nullptr)
+            {
+                RETURN_IF_FAILED(AddOverflowFlyoutItem(lastPrimaryAction.Get(), overflowButton.Get(), renderContext, &lastActionOverflowItem));
+                RETURN_IF_FAILED(lastActionOverflowItem->put_Visibility(Visibility_Collapsed));
+            }
+
+            // Create non-visible primary buttons for all overflow show card actions so they can be moved to the action bar when invoked
+            for (auto overflowShowCard : showCardOverflowButtons)
```
almedina-ms, line 1392: >auto [](http://example.com/codeflow?start=17&length=4)

const auto&

almedina-ms (approved):

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
RebeccaAnne: Overflow happens on all action sets and they will all use this same set of properties.

source/uwp/Renderer/lib/ActionHelpers.cpp
RebeccaAnne: Auto uses the size of the content to size the column. So, in this case, because the only content is a collapsed button, the column size also collapses and ends up with zero width.

RebeccaAnne:
> /* Header file automatically generated from AdaptiveCards.ObjectModel.Uwp.idl */

It should not. Unfortunately it was accidently added to this branch before it was added to .gitignore in another change list. Now i can't get git to notice I'm deleting it to remove it from this change list even if i take it out of the .gitignore. I might have to delete it in a separate change to un-confuse git.

---
In reply to: [892048249](https://github.com/microsoft/AdaptiveCards/pull/6150#issuecomment-892048249) [](http://example.com/codeflow?ancestors=892048249,892048249)

---
Refers to: source/uwp/AdaptiveCardsObjectModel/AdaptiveCards.ObjectModel.Uwp.h:1 in cd79747. [](commit_id = cd797472b9c6f1508495df11ba4c9e738049ffab, deletion_comment = False)

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -955,139 +1271,198 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
...
+                }
+
+                // Add a flyout item to the overflow menu
+                RETURN_IF_FAILED(AddOverflowFlyoutItem(action.Get(), overflowButton.Get(), renderContext, &actionControl));
+
+                // Show card actions in the overflow menu move to the action bar when selected, so we need to keep track
+                // of these so we can make non-visible buttons for them later. They need to go after all the actions we're creating as primary.
+                if (actionType == ActionType_ShowCard)
+                {
+                    showCardOverflowButtons.emplace_back(std::make_pair(action, actionControl));
```
paulcam206, line 1363: just a note -- IIRC you can replace the `std::make_pair(...)` here with `{action, actionControl}`. not sure if we have a standard convention on that or not though, so use what you prefer :)

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -955,139 +1271,198 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
...
+                PeekInnards<AdaptiveCards::Rendering::Uwp::AdaptiveRenderContext>(renderContext);
+
+            if (adaptiveActionSet != nullptr)
+            {
+                RETURN_IF_FAILED(contextImpl->AddOverflowButton(adaptiveActionSet, overflowButtonAsUIElement.Get()));
+            }
+            else
+            {
+                RETURN_IF_FAILED(contextImpl->AddOverflowButton(adaptiveCard, overflowButtonAsUIElement.Get()));
+            }
```
paulcam206, line 1463: maybe

```suggestion
            RETURN_IF_FAILED(contextImpl->AddOverflowButton((adaptiveActionSet) ? adaptiveActionSet : adaptiveCard,
                                                            overflowButtonAsUIElement.Get()));
```

source/uwp/Renderer/lib/AdaptiveHostConfig.cpp
```diff
@@ -292,6 +296,38 @@ namespace AdaptiveCards::Rendering::Uwp
...
+
+    HRESULT AdaptiveHostConfig::get_OverflowButtonAccessibilityText(_Outptr_ HSTRING* overflowButtonAccessibilityText)
+    {
+        return m_overflowButtonAccessibilityText.CopyTo(overflowButtonAccessibilityText);
+    }
+
+    HRESULT AdaptiveHostConfig::put_OverflowButtonAccessibilityText(_Out_ HSTRING overflowButtonAccessibilityText)
+    {
+        return m_overflowButtonAccessibilityText.Set(overflowButtonAccessibilityText);
+    }
```
paulcam206, line 329:
```suggestion
    HRESULT AdaptiveHostConfig::put_OverflowButtonText(_In_ HSTRING overflowButtonText)
    {
        return m_overflowButtonText.Set(overflowButtonText);
    }

    HRESULT AdaptiveHostConfig::get_OverflowButtonAccessibilityText(_Outptr_ HSTRING* overflowButtonAccessibilityText)
    {
        return m_overflowButtonAccessibilityText.CopyTo(overflowButtonAccessibilityText);
    }

    HRESULT AdaptiveHostConfig::put_OverflowButtonAccessibilityText(_In_ HSTRING overflowButtonAccessibilityText)
    {
        return m_overflowButtonAccessibilityText.Set(overflowButtonAccessibilityText);
    }
```

source/uwp/Renderer/lib/ActionHelpers.h
```diff
@@ -7,6 +7,7 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
     HRESULT BuildAction(_In_ ABI::AdaptiveCards::ObjectModel::Uwp::IAdaptiveActionElement* adaptiveActionElement,
                         _In_ ABI::AdaptiveCards::Rendering::Uwp::IAdaptiveRenderContext* renderContext,
                         _In_ ABI::AdaptiveCards::Rendering::Uwp::IAdaptiveRenderArgs* renderArgs,
+                        bool overflowAction,
```
paulcam206, line 10: `isOverflowAction` to be consistent with `HandleActionStyling()`?

source/uwp/Renderer/lib/RenderedAdaptiveCard.cpp
```diff
@@ -113,32 +114,151 @@ namespace AdaptiveCards::Rendering::Uwp
...
+                }
+
+                // Walk the buttons in the button bar. We're looking for the last visible action button (so we can hide
+                // it now that it's visible on the overflow menu). This will be the second to last visible button in the
+                // panel (the last being the overflow button itself)
+                UINT currentButtonIndex = 0;
+                UINT lastVisibleButtonIndex = 0;
+                UINT penultimateVisibleButtonIndex = 0;
+                ComPtr<IUIElement> lastVisibleButton;
+                ComPtr<IUIElement> penultimateVisibleButton;
```
paulcam206, line 165: love the choice in name 😎

source/uwp/Renderer/lib/RenderedAdaptiveCard.cpp
```diff
@@ -366,44 +503,62 @@ namespace AdaptiveCards::Rendering::Uwp
...
         return S_OK;
     }
     CATCH_RETURN;
 
     HRESULT RenderedAdaptiveCard::AddInlineShowCard(ABI::AdaptiveCards::ObjectModel::Uwp::IAdaptiveCard* adaptiveCard,
                                                     ABI::AdaptiveCards::ObjectModel::Uwp::IAdaptiveShowCardAction* showCardAction,
-                                                    ABI::Windows::UI::Xaml::IUIElement* showCardFrameworkElement,
+                                                    _In_ ABI::Windows::UI::Xaml::IUIElement* actionButtonUIElement,
+                                                    _In_ ABI::Windows::UI::Xaml::IUIElement* actionOverflowUIElement,
+                                                    ABI::Windows::UI::Xaml::IUIElement* showCardUIElement,
```
paulcam206, line 527: `_In_`?

paulcam206 (changes requested):
largely looks good. this was a ton a work!

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -855,6 +907,260 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
...
+            if (adaptiveActionSet)
+            {
+                RETURN_IF_FAILED(contextImpl->AddInlineShowCard(
+                    adaptiveActionSet, actionAsShowCardAction.Get(), actionButton, actionOverflowItem, uiShowCard.Get(), buttonIndex, renderArgs));
+            }
+            else
+            {
+                RETURN_IF_FAILED(contextImpl->AddInlineShowCard(
+                    adaptiveCard, actionAsShowCardAction.Get(), actionButton, actionOverflowItem, uiShowCard.Get(), buttonIndex, renderArgs));
+            }
```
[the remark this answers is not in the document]
RebeccaAnne: It seems like the compiler can't deal with this since the switch is between two different overloads:
> ':': no conversion from 'ABI::AdaptiveCards::ObjectModel::Uwp::IAdaptiveCard *' to 'ABI::AdaptiveCards::ObjectModel::Uwp::IAdaptiveActionSet *'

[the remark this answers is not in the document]
paulcam206: hmm... -- that's unfortunate :(

source/uwp/Renderer/lib/RenderedAdaptiveCard.cpp
RebeccaAnne: 🙂

source/uwp/Renderer/lib/ActionHelpers.cpp
RebeccaAnne: That doesn't immediately seem to compile, so leaving as is.

source/uwp/Renderer/lib/ActionHelpers.cpp
RebeccaAnne: Same issue here, the compiler can't figure this out.

paulcam206 (approved):

PR #6151, opened 2021-08-02
Rich text block schema as per Adaptive Card designer

frags51:
As per adaptive card designer, rich text block does not expect "paragraphs" to be the first JSON key. However, RN sdk is expecting it to be.
This PR ensures that both cases - with or without paragraph work, by synthetically injecting a paragraph as the parent of all inlines in case a paragraph is not present

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6151)

### DIFF
--- a/source/community/reactnative/src/components/elements/rich-text-block.js
+++ b/source/community/reactnative/src/components/elements/rich-text-block.js
@@ -125,6 +125,10 @@ export class RichTextBlock extends React.Component {
     }
 
     render() {
+        if (!this.payload.paragraphs) {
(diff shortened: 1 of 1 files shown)


frags51:
P.S - The test is failing even on removing my change (tried on local machine)

IbrahimSulai:
@frags51 - Few checks failing in this PR. So, I have include this change in the [PR](https://github.com/microsoft/AdaptiveCards/pull/6163).

Could you please close this PR?

PR #6168, opened 2021-08-05
[iOS] Commit version to podspec

### LINKED ISSUE #6065: Align release versioning with learnings from other teams
Investigate versioning practices by other library teams before/after release, and apply their learnings to our infrastructure

golddove:
# Related Issue

Fixes #6065 

# Description

Move ios release to yml file, and disable runtime version specification. Version is in podspec, and should be update there.

# Sample Card

N/A

# How Verified

Manual verification

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6168)


source/ios/tools/AdaptiveCards.podspec
```diff
@@ -2,7 +2,7 @@
 Pod::Spec.new do |spec|
   spec.name             = 'AdaptiveCards'
 
-  spec.version          = '1.0.3'
+  spec.version          = '2.8.0-beta.3'
```
jwoo-msft, line 5: so how do we update spec.version?

I don't see where it's being updated.

jwoo-msft (changes requested):

source/ios/tools/AdaptiveCards.podspec
golddove: This line should be updated manually ahead of release, following practice by Azure SDKs

jwoo-msft (approved):

PR #6214, opened 2021-08-17
[iOS] Fixed #5283

### LINKED ISSUE #5283: [iOS][Accessibility] Font size should adapt system settings.
**What platform is your feature request related to? (Delete other platforms).**
- iOS

**Additional context**
If we changed font size in system settings, I think font size of adaptive card should be adjusted automatically like the green part in screenshots.

imageimage
imageimage

### LINKED ISSUE #6207: [iOS][SampleApp][Testing] SampleApp on testing mode generate extra memory usage
# Target Platform

- iOS

# Target Application

ADCIOS Visualizer (AdaptiveCards repo)

# Problem Description

During UI testing the function reloadRowsAtIndexPaths: withRowAnimation: is called in order to update the Element Tree so that elements in a hidden card (now shown) can be interacted with. The previously mentioned function in turn tries to call the dequeueReusableCellWithIdentifier which is a function created with the idea of recycling elements that are out of the screen boundaries to free memory when not in use; the issue found was that the recycled cell may keep a bigger size than the card requires which is not a desired behaviour.

# Screenshot

Please provide a screenshot of the problem (if possible)

# Card JSON

Please provide the JSON of the Card that is being rendered

# Sample Code

Please provide any relevant code samples to help reproduce the issue

# Version of SDK

Introduced by #6094

jwoo-msft:
# Related Issue

Fixed #5283, Fixed #6207

# Description

* Changed font to use dynamic type.
* If HTML is used for font, then font attribute is checked in the whole range of the text, and font is updated to use dynamic type.
* Updated sample app to demonstrate the feature
  * Added ability to handle accessibility status change event
  * Changed to use automatic dimension in the UITableView to handle the accessibility status change event.
  * Refactored & cleaned up the codes
* Bumped up minimum iOS version from 10 to 11 

# Sample Card

N/A

# How Verified

1. Existing relevant unit/regression tests are run.

https://user-images.githubusercontent.com/4112696/129647927-09c10573-2c6b-4364-a561-9777168bc672.mov

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6214)


source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ACRChatWindow.mm
```diff
@@ -63,7 +63,7 @@ - (void)insertCard:(NSString *)card
     numberOfCards += 1;
 }
 
-- (void)insertView:(UIView *)view
+- (void)insertView:(ACRView *)view
```
almedina-ms, line 66: this change is ok, but I would like to know why changing to a more specific type of view

source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ACRChatWindow.mm
```diff
@@ -146,8 +144,61 @@ - (void)renderCards:(NSString *)card
...
+
+- (void)updateLayoutConstraints
+{
+    _contentViewConstratins = @[
+        [self.adaptiveCardView.centerXAnchor constraintEqualToAnchor:self.contentView.centerXAnchor],
+        [self.adaptiveCardView.centerYAnchor constraintEqualToAnchor:self.contentView.centerYAnchor],
+        [self.contentView.heightAnchor constraintEqualToAnchor:self.adaptiveCardView.heightAnchor],
+        [self.contentView.widthAnchor constraintEqualToAnchor:self.adaptiveCardView.widthAnchor]
+    ];
+    [NSLayoutConstraint activateConstraints:_contentViewConstratins];
```
almedina-ms, line 201: thank you for taking on this task, you may want to add "Fixes #6207" to the description

source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ACRChatWindow.mm
```diff
@@ -146,8 +144,61 @@ - (void)renderCards:(NSString *)card
                            widthConstraint:adaptiveCardsWidth
                                   delegate:self.adaptiveCardsDelegates];
     }
+    return renderResult.view;
+}
 
-    [adaptiveCardsViews addObject:renderResult.view];
+- (void)renderCards:(NSString *)card
```
almedina-ms, line 150: could this method or the method in line 129 be renamed?  while some rendering is being performed inside of it, it seems like it's adding the rendered card to the cards list and the plural in card would suggest as it is rendering multiple cards; a possible update could be having the method in line 129 changed to: renderCardFromJson and this method changed to renderCard (or it could follow the pattern from lines 59 & 66 where the methods start with the word 'insert')

jwoo-msft: you are correct. But this is an existing function & public. Its function hasn't changed, so I rather keep it, for now, to keep it simple and on point to addressing the issue of supporting dynamic font type.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRFactSetRenderer.mm
```diff
@@ -65,6 +65,12 @@ + (ACRUILabel *)buildLabel:(NSString *)text
             content = [[NSMutableAttributedString alloc] initWithData:htmlData options:options documentAttributes:nil error:nil];
             // Drop newline char
             [content deleteCharactersInRange:NSMakeRange([content length] - 1, 1)];
+            [content enumerateAttribute:NSFontAttributeName
```
almedina-ms, line 68: it looks like this block of code is the same across all 3 files, is there some way to create a static function that does the same?

almedina-ms (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRFactSetRenderer.mm
jwoo-msft: okay

source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ACRChatWindow.mm
```diff
@@ -146,8 +144,61 @@ - (void)renderCards:(NSString *)card
...
+        for (NSInteger i = 0; i < adaptiveCardsViews.count; i++) {
+            adaptiveCardsViews[i] = [NSNull null];
+        }
+    }
+}
+@end
+
+// configure the content cell
+@implementation ACRChatWindowCell {
+    NSArray<NSLayoutConstraint *> *_contentViewConstratins;
```
paulcam206, line 169: typo

```suggestion
    NSArray<NSLayoutConstraint *> *_contentViewConstraints;
```

source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
```diff
@@ -11,7 +11,7 @@
 	"imageBaseUrl": "https://pbs.twimg.com/profile_images/3647943215/",
 	"fontTypes": {
 		"default": {
-			"fontFamily": "'Segoe UI', 'sans-serif', 'Oxygen', 'Ubuntu', 'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue'",
+			"fontFamily": "San Francisco, American Typewriter, Segoe UI, Oxygen, Ubuntu, Cantarell, Fira Sans, Droid Sans, Helvetica Neue",
```
paulcam206, line 14: we should probably document that dynamic font size changes for hosts requires use of Apple's new fonts...

paulcam206 (approved):

source/ios/AdaptiveCards/ADCIOSVisualizer/resources/sample.json
jwoo-msft: https://developer.apple.com/documentation/uikit/uifont/scaling_fonts_automatically?language=objc
`Although custom fonts are supported in Dynamic Type, the preferred font is designed to look good at any size (for more information, see Typography). Also, using the preferred font ensures consistency within the system and with other apps.`

PR #6310, opened 2021-09-03
[Android] Fix Markdown ordered lists to honour start

### LINKED ISSUE #6238: [Rendering] Numbered item in TextBlock always displaying number 1
### Target Platforms

Android

### SDK Version

2.8.0 beta

### Application Name

Microsoft Teams

### Problem Description

Numbered items in the `TextBlock`  is always displaying number 1 instead of other numbers.

### Screenshots

Screenshot_1629758184

### Card JSON

```JSON
{

  "type": "AdaptiveCard",

  "version": "1.3",

  "body": [

	{

	  "type": "TextBlock",

	  "text": "1. First item in the list;"

	},

	{

	  "type": "TextBlock",

	  "text": "2. second item in the list;"

	},

	{

	  "type": "TextBlock",

	  "text": "3. Third item in the list;"

	}

  ]

}
```

### Sample Code Language

n/a

### Sample Code

n/a

almedina-ms:
# Related Issue

Fixes  #6238

# Description

Context: In the Android rendered all texts to be rendered are preprocessed to convert the markdown components to HTML

In the android renderer, any text containing "<number>. <text>" is parsed as the start of an ordered markdown list, this approach works most of the time but due to some limitations the ordered lists would always start at 1 as we would miss context on the html tag attributes, for this reason this PR overrides the ContentHandler class behaviour inside of the currently used XMLReader to retrieve the attributes for any given html element and thus providing a correct starting number for any ordered list.

# Sample Card

v1.5/Test/TextBlock.Markdown.NumberStart.json

# How Verified

1. A unit test was added to verify the rendered TextBlocks include the right number

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6310)


samples/v1.5/Test/TextBlock.Markdown.NumberStart.json
```diff
@@ -0,0 +1,19 @@
...
+	  "type": "TextBlock",
+	  "text": "1. First item in the list;"
+	},
+	{
+	  "type": "TextBlock",
+	  "text": "2. Second item in the list;"
+	},
+	{
+	  "type": "TextBlock",
+	  "text": "3. Third item in the list;"
```
paulcam206, line 16: we should have a test to make sure the second and third items in a markdown list are incremented correctly:

```json
{
    "type": "TextBlock",
    "text": "10. the tenth thing\r10. the list is still going!\r10. should be 12!"
}
```

paulcam206 (changes requested):

paulcam206 (approved):

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/RendererUtil.java
```diff
@@ -148,13 +163,121 @@ public void handleTag(boolean opening, String tag, Editable output, XMLReader xm
...
+        // ContentHandler override methods
+        @Override
+        public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException
+        {
+            boolean isHandled = handleTag(true, localName, text, attributes);
+            tagStatusQueue.addLast(isHandled);
+
+            if (!isHandled)
+            {
+                defaultContentHandler.startElement(uri, localName, qName, attributes);
```
jwoo-msft, line 215: I'm a little concerned here. what happens when the tag is not handled inside the `startElement`? Would it create an infinite loop? since it will call the method with the same arguments?

jwoo-msft (changes requested):

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/RendererUtil.java
almedina-ms: we're handling a small subset of the html tags, if it's not handled instead of handling it ourselves we're passing the tag to the original/default content handler that we have been using all this time hence no infinite loop should ever happen

jwoo-msft (approved):

Issue #6313, opened 2021-09-08
[Designer] Needs to be usable at higher zoom levels


paulcam206:
### Problem Description

Original bug: #6263

WCAG 2.1 requirements [state](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) that content needs to be able to be read/interacted with using only vertical scrolling at 1280x768 400% zoom. 

The original bug tracks getting the designer to the point where everything is at least accessible, if not convenient. This bug tracks giving the designer a more thoughtful redesign to be useful and usable at high zoom levels.

### Card JSON

```JSON
{}
```

### Screenshot

width 1280 @ 400% zoom:
![image](https://user-images.githubusercontent.com/16614499/132572823-592880ff-6afb-46de-b5db-eedbead335c7.png)

JeanRoca:
PR addressing issue has been merged

PR #6320, opened 2021-09-14
[Designer] Make everything accessible at high zoom levels

### LINKED ISSUE #6263: [Designer]: At 400% zoom, All content present on "Pick a sample as a starting point" window are not visible.
### Problem Description

[30645314](https://microsoft.visualstudio.com/OS/_workitems/edit/30645314)

At 400% zoom, All content present on "Pick a sample as a starting point" window are not visible.
Note: Same issue is repro for 'Pic2card Dialog for Image Upload' window

### Card JSON

```JSON
No card
```

### Screenshot

_No response_

paulcam206:
# Related Issue

Fixes #6263

# Description

A bevy of fixes for the designer on small screens. Pushes a bunch of runtime-defined CSS styles into classes so that they can be overridden by media queries at render time. Below 500px width, we shift to a vertical scroll model which, while not ideal from a usability perspective, at least allows the user to access all of the content, which was not possible before. Having a more usable experience is something that will require design resources which we'll have to engage at a later time (see #6313).

# Screenshots

### Main
(note: no scrollbars)
image

### New Card Dialog
(large margins)
image

### Main
image

### New Card Dialog
image

# How Verified

Manual testing

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6320)


dclaux (commented):
Looks good overall, however:
- In "mini" mode, dragging a new element onto the card makes the card basically disappear. I believe this happens because the associated item in the treeview is forced into view, which works with the normal layout but not when all panes are stacked on top of each other
- You apparently have the threshold set to 501px horizontally, but the designer starts being unusable well before that. Would it make sense to change the threshold to something a little higher?

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/designer.ejs
```diff
@@ -7,9 +7,11 @@
...
 
 <style type="text/css">
-	.content {
-		height: calc(100vh - 97px) !important;
-	}
+    @media all and (min-width: 501px) {
+        .content {
+		    height: calc(100vh - 97px) !important;
+	    }
+    }
```
RebeccaAnne, line 14: [nit] Tabs and spaces

RebeccaAnne (approved):

paulcam206:
> Looks good overall, however:
> 
> * In "mini" mode, dragging a new element onto the card makes the card basically disappear. I believe this happens because the associated item in the treeview is forced into view, which works with the normal layout but not when all panes are stacked on top of each other

the scroll-in-to-view fix was pretty easy, but the drag/drop overlay fix was a little gnarly (there was some math that didn't account for scroll position (because there wasn't a need to before)). 

> * You apparently have the threshold set to 501px horizontally, but the designer starts being unusable well before that. Would it make sense to change the threshold to something a little higher?

I set it to 651px since that's about as small as you can go and still sort of use everything...

jwoo-msft (approved):

PR #6322, opened 2021-09-14
Merge Jonwis' object model conversion into UWP

### LINKED ISSUE #6309: Fold Jonwis cppwinrt conversion back into UWP
Jon has created a version of our object model code that has been converted to cppwinrt. This work tracks folding that work back into the existing UWP object model.

RebeccaAnne:
# Related Issue

Fixes #6309

# Description

Jonwis converted our Object Model to C++/WinRT. This PR merges that change back into the UWP object model, and adds testing and bug fixes.

# How Verified

- Ran existing UWP tests
- Updated UWP object model tests to be more robust and ensure they tested full round trip through the object model code (Construct object model in code, serialize to json, deserialize back to object model from json, serialize to json)
- Smoke testing on the scenario cards
    - Found an issue in fallback this way that wasn't caught by the tests and added a new fallback test to cover it.


source/uwp/AdaptiveCardsObjectModel/AdaptiveCardsObjectModel.vcxproj
```diff
@@ -43,13 +43,15 @@
     <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
     <AppContainerApplication>true</AppContainerApplication>
     <ApplicationType>Windows Store</ApplicationType>
-    <WindowsTargetPlatformVersion>10.0.18362.0</WindowsTargetPlatformVersion>
-    <WindowsTargetPlatformMinVersion>10.0.15063.0</WindowsTargetPlatformMinVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
```
jonwis, line 46: Is this change OK? I was being lazy and updated these to make VS happy; not sure this is what you want.

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveColumn.cpp
```diff
@@ -30,7 +31,10 @@ namespace winrt::AdaptiveCards::ObjectModel::Uwp::implementation
             BackgroundImage = winrt::make<implementation::AdaptiveBackgroundImage>(backgroundImage);
         }
 
-        Rtl = sharedColumn->GetRtl();
+        if (sharedColumn->GetRtl().has_value())
+        {
+            Rtl = winrt::box_value(sharedColumn->GetRtl().value()).as<Windows::Foundation::IReference<bool>>();
```
jonwis, line 36: Note: coming soon, the `wrl::winrt::property<std::optional<bool>>` will do this conversion for you so you can just say `Rtl = sharedColumn->GetRtl()` in one line.

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveNumberInput.cpp
```diff
@@ -9,9 +9,21 @@ namespace winrt::AdaptiveCards::ObjectModel::Uwp::implementation
 {
     AdaptiveNumberInput::AdaptiveNumberInput(const std::shared_ptr<::AdaptiveCards::NumberInput>& sharedNumberInput)
     {
-        Min = sharedNumberInput->GetMin();
-        Max = sharedNumberInput->GetMax();
-        Value = sharedNumberInput->GetValue();
+        if (sharedNumberInput->GetMin().has_value())
```
jonwis, line 12: Can these be extracted to a function? Like:

```c++
template<typename T> winrt::Windows::Foundation::IReference<T> box_or_default(std::optional<T> const& t)
{
    if (t.has_value())
    {
        return winrt::box_value(t.value()).as<Windows::Foundation::IReference<T>>();
    }
    else
    {
        return nullptr;
    }
}
```

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveTextElement.h
```diff
@@ -17,13 +17,38 @@ namespace winrt::AdaptiveCards::ObjectModel::Uwp::implementation
...
         {
-            IsSubtle = sharedModel->GetIsSubtle();
-            FontType = opt_cast<Uwp::FontType>(sharedModel->GetFontType());
-            Size = opt_cast<Uwp::TextSize>(sharedModel->GetTextSize());
-            Weight = opt_cast<Uwp::TextWeight>(sharedModel->GetTextWeight());
-            Color = opt_cast<Uwp::ForegroundColor>(sharedModel->GetTextColor());
             Text = UTF8ToHString(sharedModel->GetText());
             Language = UTF8ToHString(sharedModel->GetLanguage());
+
+            if (sharedModel->GetIsSubtle().has_value())
```
jonwis, line 23: Huh, really curious about what I messed up with the `operator=` on `property_opt` here. Was it a compile error or runtime?

jonwis (approved):

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveTextElement.h
RebeccaAnne: It was a runtime error, the values were just not being set. At the time i hadn't realized that property_opt was your creation so i didn't look too hard at the code there and just updated the caller side. I went back and did what i should have don't in the first place and fixed the property_opt set method. Let me know if my change looks right to you.

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveColumn.cpp
RebeccaAnne: This goes away with the property_opt fix.

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveNumberInput.cpp
RebeccaAnne: This goes away with the property_opt fix

source/NuGet.config
```diff
@@ -4,6 +4,7 @@
     <!-- remove any machine-wide sources with <clear/> -->
     <clear />
     <add key="AdaptiveCards-Public" value="https://pkgs.dev.azure.com/ms/AdaptiveCards/_packaging/AdaptiveCards-Public/nuget/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
jonwis, line 7: Maybe the cppwinrt 210913.7 should get pushed into the AC feed? I'm nervous about adding the additional source here, but having it definitely made life easier.

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveTextElement.h
jonwis: Huh! Thank you, I'd thought `std::optional<T>::operator bool()` ([see here](https://en.cppreference.com/w/cpp/utility/optional/operator_bool)) would have done this.  I'll have to dig a little more for my own edification. Your change looks great.

source/NuGet.config
RebeccaAnne: Agreed, I'm following up on my end on what our process is supposed to be here for picking up the new version.

source/uwp/AdaptiveCardsObjectModel/AdaptiveCardsObjectModel.vcxproj
RebeccaAnne: Visual Studio seems happy enough without it on my end, so reverting this change.

source/NuGet.config
RebeccaAnne: Ok, I've pushed the updated packages and this is now working against the AC feed.

jonwis (approved):

paulcam206 (approved):
this looks so nice! appreciate all of the hard work it took to get this done :)

jwoo-msft (approved):

PR #6377, opened 2021-09-19
[Shared] Add `password` `Input.Text` support

### LINKED ISSUE #6340: Input.Text masking/password Shared Model support

paulcam206:
# Related Issue

Fixes #6340

# Description

Add support for the `password` style on `Input.Text`. As written, the shared model will lie in `TextInput::GetIsMultiline` if `"style": "password"`, but we'll still serialize it as `true`. Added tests to verify the new style as well as serialization behavior. Also added generated SWIG wrappers.

# How Verified

* local build, step-through debugging, and new/existing unit tests

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6377)


source/shared/cpp/ObjectModel/TextInput.cpp
```diff
@@ -78,7 +78,8 @@ void TextInput::SetValue(const std::string& value)
 
 bool TextInput::GetIsMultiline() const
 {
-    return m_isMultiline;
+    // isMultiline is ignored if the TextInputStyle is password
+    return m_isMultiline && m_style != TextInputStyle::Password;
```
RebeccaAnne, line 82: I'm not sure i like the idea that GetIsMultiline will return a different value than what is serialized if you serialize the card. 

I see the charm in doing this here so the renderers don't have to think about it (and don't get a say), but if we're going to let the shared model take over this logic maybe we just clear isMultiline in this case so that the serialization will come out the same as the object model?

paulcam206: that's a fair point -- I'll fix it up.

jwoo-msft (approved):

source/shared/cpp/ObjectModel/TextInput.cpp
paulcam206: actually, given that we don't have precedent to do this kind of thing in the shared model, I think I'll pull it for now...

RebeccaAnne (approved):

paulcam206:
verified builds/runs clean on .NET

PR #6395, opened 2021-09-22
Update custom.props for release

### LINKED ISSUE #6351: Update XAML package version for schema v1.5

RebeccaAnne:
# Related Issue
Fixes #6351 

# Description

Update custom.props to set the UWP package version. Due to significant breaking changes in the UWP renderer (Object model namespace change, several properties changing type from `foo` to `IReference<foo>` (see #6354)), I have incremented the major version. Please holler if that doesn't seem right.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6395)


paulcam206 (approved):

jwoo-msft (approved):

custom.props
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup Label="Version">
-    <VersionMajor>2</VersionMajor>
-    <VersionMinor>7</VersionMinor>
+    <VersionMajor>3</VersionMajor>
```
almedina-ms, line 4: >VersionMajor>3</VersionMaj [](http://example.com/codeflow?start=5&length=26)

so far we have been releasing 2.8.0-beta versions for android and iOS, are we sure we should be moving to version 3.0.0?

almedina-ms (changes requested):
:clock1:

custom.props
paulcam206: per standup, we're cool with this on UWP

almedina-ms (approved):

PR #6407, opened 2021-09-24
[JS] Add Password support to Input.Text

### LINKED ISSUE #6337: JavaScript

almedina-ms:
# Related Issue
 
Fixes #6337

# Description

Adds Password to the Input.Text style list

# Sample Card

-

# How Verified

-

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6407)


jwoo-msft (commented):
we don't need to set the input field to password?

almedina-ms:
if you take a look at card-elements.ts  line 2846, the inputTextStyle is set to the `type` property, so it's setting type="password" automatically

---
In reply to: [763547867](https://github.com/microsoft/AdaptiveCards/pull/6407#issuecomment-763547867) [](http://example.com/codeflow?ancestors=763547867,763547867)

source/nodejs/adaptivecards/src/enums.ts
```diff
@@ -123,7 +123,8 @@ export enum InputTextStyle {
     Text,
     Tel,
     Url,
-    Email
+    Email,
+    Password
```
paulcam206, line 127: It's pretty cool that the type maps directly to the HTML type !

I think we need some schema change on the `TextInput` class in order to version this properly, though... here's an example of a versioned property value in `ChoiceSetInput`:

https://github.com/microsoft/AdaptiveCards/blob/bf65cddc824b64b95a0acd635504b314ae3a94ad/source/nodejs/adaptivecards/src/card-elements.ts#L3128-L3136

this info will help the designer give accurate warnings and allow the renderer to know whether to honor the `password` value when rendering in lower versions...

paulcam206 (changes requested):

paulcam206 (approved):

paulcam206:
looks good -- have you tested in the designer to see if it does the Right Thing? it should just magically work :)

almedina-ms:
> looks good -- have you tested in the designer to see if it does the Right Thing? it should just magically work :)

passowrdJS

jwoo-msft (approved):

Issue #6393, opened 2021-09-22
[Shared] `TableRow` and `TableCell` elements parse outside of `Table`


paulcam206:
### Target Platforms

iOS, Android, UWP, Other

### SDK Version

Schema 1.5

### Application Name

cl.exe

### Problem Description

The shared model allows `TableRow` and `TableCell` to be valid elements outside of the `Table` context leading to unfortunate behavior in dependent platforms.

* Android does not render the content
* iOS does render content
* UWP crashes

### Card JSON

```JSON
{
  "type": "AdaptiveCard",
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "version": "1.5",
  "body":   [
    {
      "type": "TableRow",
      "cells": [
        {
          "type": "TableCell",
          "items": [
            {
              "type": "TextBlock",
              "text": "."
            }
          ]
        }
      ]
    }
  ]
}
```

paulcam206:
have a fix done, just need to get it cleaned up

PR #6412, opened 2021-09-26
[UWP] Fix risky cast in `GenerateElementProjection()`

### LINKED ISSUE #6394: [UWP] `GenerateElementProjection()` performs risky cast in default case
https://github.com/microsoft/AdaptiveCards/blob/61677f2d4f9fe52de9551cb6238fe415e8b7d688/source/uwp/AdaptiveCardsObjectModel/lib/ObjectModelUtil.cpp#L313-L317

In the above, if we encounter an element that isn't covered in a `case`, but *is* a known element (*shouldn't* happen, but see #6393), we can end up incorrectly casting a known element to `AdaptiveCards::UnknownElement`, leading to undefined behavior later on. We should probably split the `default` case out on its own and throw an exception (and do a quick examination to see if this pattern is repeated elsewhere).

paulcam206:
# Related Issue

Fixes #6394

# Description

See #6393 and #6394 for more details, but basically, in this instance we were casting an object implemented with formal element class (`TableRowElement` or `TableCellElement`) to an object of type `UnknownElement`. This is risky because we're making assumptions about what the shared model is doing. It's possible that we would end up in this scenario if the shared model implemented support for an element type that the UWP renderer had yet to implement.

# Sample Card

See [`StandaloneTableRow.json`](https://github.com/microsoft/AdaptiveCards/pull/6408/files#diff-55134bf1ed8dd065b7368a93c45f3c21ce7c2c2fccda7db45f099471b89cb20c).

# How Verified

* step through debugging with UWP visualizer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6412)


jwoo-msft (approved):

source/uwp/AdaptiveCardsObjectModel/lib/ObjectModelUtil.cpp
```diff
@@ -307,13 +307,17 @@ winrt::AdaptiveCards::ObjectModel::Uwp::IAdaptiveCardElement GenerateElementProj
...
     case CardElementType::Unknown:
-    default:
         return winrt::make<winrt::AdaptiveCards::ObjectModel::Uwp::implementation::AdaptiveUnsupportedElement>(
             std::AdaptivePointerCast<AdaptiveCards::UnknownElement>(baseElement));
+    default:
+        // It should not be possible for an element represented by a CardElementType to not have a matching case here.
+        // Either a new element has been introduced in the shared model without full support here in UWP, or there's a
+        // bug in the shared model (see https://github.com/microsoft/AdaptiveCards/issues/6393).
+
+        throw winrt::hresult_error();
```
licanhua, line 320: suggest a meaning for hr code
```
throw winrt::hresult_error(E_INVALIDARG, L"CardElementType is not supported");
```

almedina-ms (approved):

licanhua (approved):

PR #6457, opened 2021-09-30
[Designer] Fix the card structure order problem

### LINKED ISSUE #6375: [Designer] Items in `Card Structure` pane are out of order
### Problem Description

At some point, the designer version in `main` had a regression introduced -- the order of items in the `Card Structure` pane is incorrect. I came across this while working on the high zoom level issue, and couldn't figure out how I broke it.

Turns out the bug is there without my change (see [CI](http://adaptivecards-ci.azurewebsites.net/designer/)), so we need to track down the problem.

### Card JSON

```JSON
{}
```

## On page load

note the position of the two actions:
image

## After dragging a textblock to the bottom of the card (above the actions)
image

licanhua:
Fix #6375
In the code, action is insert to the begin the tree, so for any cards with action, the order are messed up.
After the code change, the action is moved to the end.

Before 
image

After
image

I did the manual testing, and will like if I can write automation for this

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6457)


licanhua:
Click this [link](https://adaptivecardsci.z5.web.core.windows.net/pr/6457/designer) to verify the change

almedina-ms (approved):
:shipit:

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -1682,8 +1682,10 @@ export class CardElementPeer extends DesignerPeer {
             }
         }
 
+        let itemCount: number = 0;
```
paulcam206, line 1685: since we're immediately assigning this to `0`, we don't need an explicit type declaration

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -1682,8 +1682,10 @@ export class CardElementPeer extends DesignerPeer {
...
         if (this.cardElement instanceof Adaptive.CardElementContainer) {
-            for (var i = 0; i < this.cardElement.getItemCount(); i++) {
+            itemCount = this.cardElement.getItemCount();
+            for (var i = 0; i < itemCount; i++) {
```
paulcam206, line 1688: I know this isn't your issue, but while we're here, we should switch `var` to `let` (`var` scopes variables to the containing function)

paulcam206 (approved):
a couple of minor nits - good find!

source/nodejs/adaptivecards-designer/src/designer-peers.ts
licanhua: @paulcam206 good call. there are 43 places with `for (var`. Should I fix them all in this PR or another PR

licanhua: I will create a separate PR to address the var problem

PR #6527, opened 2021-10-07
[Sample] Add sample for Multiline Input.Text with Password style

almedina-ms:
Adds a test json as other samples don't contain this defined behaviour

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6527)


jwoo-msft:
I'm looking at the iOS CI failure, I have seen it twice at the same place. It complains about the parse failure. the failure is not related to this PR.

jwoo-msft:
> I'm looking at the iOS CI failure, I have seen it twice at the same place. It complains about the parse failure. the failure is not related to this PR.

the duplicate id is causing the CI failure.

PR #6556, opened 2021-10-08
[iOS] Updated Podspec file as bitcode is not enabled for fluentUI pod

jwoo-msft:
# Description
When tested on actual device, XCode wasn't able to create bitcode for arm64 due to fluent-ui pod did not enable it by default.
This PR fixes that.

[reference](https://stackoverflow.com/questions/34666247/bitcode-bundle-could-not-be-generated-because-ios-charts)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6556)


jonmill:
Known Android issues - merging as they are unrelated

PR #6145, opened 2021-07-28
React.js Adaptive Card Component

marakim:
A React.js wrapper for Adaptive Cards

Also:
* Update to Typescript 4
* Update to Webpack 5

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6145)

PR #6577, opened 2021-10-11
[JS] Add hostConfig to adaptivecards-react project

### LINKED ISSUE #6504: [Feature Request] allows host config to pass from adaptivecards-react to adaptivecards
### Problem Statement

We don't have examples in react project to allow host config to be passed. Maybe need code change on adaptivecards-react project too.

### Proposed solution

n/a

_No response_

licanhua:
In #6145, adaptivecards-react is added, and which implemented HostConfigContext, but user has to implement its own Provider to customize the hostConfig.

In this PR, it allows react to pass hostConfig from props.
hostConfig takes high priority than customized host config Provider if both hostConfig and provider are used. 

Fix #6504

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6577)

### DIFF
--- a/source/nodejs/adaptivecards-designer/README.md
+++ b/source/nodejs/adaptivecards-designer/README.md
@@ -47,9 +47,6 @@ To load the designer component you have 2 options:
 <!-- OPTIONAL: markdown-it isn't required but enables out-of-the-box markdown support -->
 <script src="https://unpkg.com/markdown-it@8.4.0/dist/markdown-it.min.js"></script>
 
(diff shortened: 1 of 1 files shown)


marakim:
This change shouldn't be necessary as the host config is meant to be made available via the React context provider `HostConfigContext`.  I could see it being convenient to be able to pass in a host config directly (perhaps as a component level override?). @paulcam206 what do you think?

PR #6664, opened 2021-10-29
[Styling] Update swiper style

almedina-ms:
Clean up the style to have arrows having a different icon and setting the arrows and bullets outside of the card view

## Visual Verification

image

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6664)


source/nodejs/adaptivecards/src/adaptivecards.css
```diff
@@ -281,14 +281,54 @@ a.ac-anchor:visited:active {
     -ms-flex-align: center;
     -webkit-align-items: center;
     align-items: center;
+    padding-left: calc(var(--swiper-navigation-size)/ 44 * 27) !important;
```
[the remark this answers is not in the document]
paulcam206: the problem here is that padding is defined in the element's `style` property -- CSS rules apply in order of [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity), and an explicit `style` tag always takes top precedence (unless you override with `!important`).

the JS renderer emits the `style` tag with styles calculated from the schema and host config. in this case, it's the `CarouselPage` itself emitting the padding value (see `CardElementContainer.applyPadding()`). I'm guessing we won't expect card authors to be able to affect padding for v1, so maybe you can just have an empty `applyPadding()` implementation at the `CarouselPage` level? that should make it so the `padding` style isn't emitted for `CarouselPage`s.

source/nodejs/adaptivecards/src/adaptivecards.css
```diff
@@ -281,14 +281,54 @@ a.ac-anchor:visited:active {
...
+
+.swiper-button-prev,
+.swiper-rtl.swiper-button-next {
+    left: 0px !important;
+} 
+
+.ac-carousel-pagination.swiper-pagination.swiper-pagination-bullets.swiper-pagination-horizontal {
+    bottom: 0px !important;
+
 }
```
paulcam206, line 334: do these actually need `!important`?

paulcam206 (changes requested):

source/nodejs/adaptivecards/src/adaptivecards.css
almedina-ms: if I don't add the !important then the properties are not overriden, not sure why

source/nodejs/adaptivecards/src/adaptivecards.css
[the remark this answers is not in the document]
almedina-ms: so, is it ok to set inline style properties and expect consumers to override them with the !important css option?

source/nodejs/adaptivecards/src/adaptivecards.css
jwoo-msft: Swiper's css sets some of those style properties, and we need to add !important to be able to override those attributes. we could do away with not using its css, but it would be way too many attributes to track all of them down.

almedina-ms: I removed all important comments

paulcam206 (approved):
looks good - we should revisit high contrast post-v0

PR #6667, opened 2021-11-01
[JS] Carousel schema changes

paulcam206:
# Description

Per design discussions, I've reworked the JS Carousel implementation to fit the new schema. This work is not yet fully complete... I just wanted to get these changes on everybody's radar (and get some early feedback if folks are inclined). The following should be fixed before checking in:

- [x] Complete "forbidden" items support (ripped out and only partially replaced while moving everything over)
- [x] Requirement for Designer change
- [ ] Tests:
  - [ ] Make sure `Carousel` not allowed outside of the one acceptable context as a direct child object of `body`
  - [x] Make sure `CarouselPage` objects are allowed to skip defining `type` (but validate that `type` matches if it's specified) -- should get this for free
  - [x] Make sure `CarouselPage` not allowed to exist outside of `Carousel`
  - [ ] Serialization
  - [ ] Object model

## Old Schema (snippet)

```json
{
    "display": {
        "type": "carousel"
    }
    "body": [
         { "type": "CarouselPage" }
     ]
}
```

## New Schema (snippet)

```json
{
    "body": {
        "type": "Carousel",
        "pages": [
            { "type": "CarouselPage" }
        ]
    }
}
```

# Sample Card

Samples in `samples/v1.6` have been updated to use the new schema.

# How Verified

* local build, dev tools

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6667)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -7524,58 +7326,60 @@ class InlineAdaptiveCard extends AdaptiveCard {
...
         allowFallback: boolean,
         createInstanceCallback: (typeName: string | undefined) => T | undefined,
         logParseEvent: (typeName: string | undefined, errorType: Enums.TypeErrorType) => void): T | undefined {
         let result: T | undefined = undefined;
 
         if (source && typeof source === "object") {
-            let typeName = Utils.parseString(source["type"]);
+            const typeName = Utils.parseString(source["type"]);
+
+            // if (typeName && this._forbiddenElementTypes.indexOf(typeName) >= 0) {
```
jwoo-msft, line 7346: nit: commetns

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5948,6 +5825,10 @@ export class Container extends ContainerBase {
                 }
             }
         }
+        else if (isSingletonContainer(this) && this.singletonChild) {
```
jwoo-msft, line 5828: `this.singletonChild` is already checked by `isSingletonContainer()`.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -7124,8 +6943,18 @@ export class Authentication extends SerializableObject {
     tokenExchangeResource?: TokenExchangeResource;
 }
 
+export interface ISingletonContainer {
```
jwoo-msft, line 6946: nit: can you consider adding comments or re-naming Singleton? In the context of programming, the first thing that comes to mind is the singleton design pattern; not the singleton set.

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -0,0 +1,264 @@
...
+        }
+        else {
+            return undefined;
+        }
+    }
+
+    protected internalParse(source: any, context: SerializationContext) {
+        super.internalParse(source, context);
+
+        this._pages = [];
```
jwoo-msft, line 128: nit: `timer` doesn't seem to be parsed.

samples/v1.6/Elements/Carousel.json
```diff
@@ -2,51 +2,51 @@
 	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
 	"type": "AdaptiveCard",
 	"version": "1.6",
-	"display": {
```
jwoo-msft, line 5: `display` property at LN 5676 needs to be removed from AdatpiveCards.

jwoo-msft (commented):
LGTM with just a few minor things

licanhua:
why don't we use list directly instead of object in this proposal
```
{
    "body": [{
        "type": "Carousel",
        "pages": [
            { "type": "CarouselPage" }
        ]
    }]
}
```
Here is the information I get:
1. only one carousel is supported, at least at first
2. If one carousel is supported, using list will confuse customer because customer is able to put anything in the list. so object is proposed here
3. fallback for the carousel. If two carousels are in body, and fallback will show only two. If only one is supported, it cause visual difference.

Why do we only support one carousel?

dclaux:
@licanhua I think the confusion you're referring to will exist anyway. The existing AC model is ` body[ ... ]`, chances are card authors will not get why they can't just put a `Carousel` in there. The bottom line is they have to understand that a card is either a single Carousel and it needs to be specified with `body { ... }` or a card can be what it is today and use the `body [ ... ]` notation. They will only understand that by either:
- Reading documentation
- Or being guided by the designer which will abstract away the syntax anyway

So bottom line, if they have to learn and understand something, learning that a card with a `Carousel` it its body cannot contain any other element is just as easy as learning to use the `body { ... }` syntax IMO.

That being said, I'd say there's value in supporting the `body: { ... }` syntax generically, as a shortcut to creating a container (be it an AdaptiveCard, a Container, a Column, a TableCell etc.) with just one element in it. It would just be a very minor change in the way Container parses its elements:

```typescript
    protected internalParse(source: any, context: SerializationContext) {
        super.internalParse(source, context);

        this.clear();
        this.setShouldFallback(false);

        let jsonItems = source[this.getItemsCollectionPropertyName()];

        if (typeof jsonItems === "object") {
            jsonItems = [ source[this.getItemsCollectionPropertyName()] ];
        }

        if (Array.isArray(jsonItems)) {
            for (let item of jsonItems) {
                let element = context.parseElement(this, item, !this.isDesignMode());

                if (element) {
                    this.insertItemAt(element, -1, true);
                }
            }
        }
    }
```

The code simply creates an one object array on the fly if the payload contains an object, nothing else. The parsing logic remains the unchanged otherwise, no need to introduce the concept of a "singleton child". And in terms of the registry, instead of introducing a new "singleton registry", I think it should be possible to associate a constraint with each registration in the elements registry:
```typescript
GlobalRegistry.defaultElements.register("Container", Container, true /* singletonOnly */);
```

`Container`'s parsing logic can then invoke `SerializationContext.parseElement` passing it a flag indicating whether it accepts singletons (basically, if it's empty) which would then pass that along to `CardObjectRegistry.createInstance`.

source/nodejs/adaptivecards/src/carousel.ts
paulcam206: handled here:

https://github.com/microsoft/AdaptiveCards/pull/6667/files/6b34a781fd14f54c944e1df2b15136ce1ce07f4b#diff-c83b27dc841a20f8725b8b4aeb7f518336e1c5b77a9ae989824bb6908adc6204R64-R66

samples/v1.6/Elements/Carousel.json
```diff
@@ -2,51 +2,51 @@
...
-	  		"url": "https://adaptivecards.io/content/cats/2.png",
-	  		"size": "medium"
-	  	}
+	"body": {
+		"type": "Carousel",
+		"timer": 5000,
+		"pages": [
+			{
+				"type": "CarouselPage",
+				"id": "firstCarouselPage",
```
licanhua, line 11: AdaptiveCards has some properties like `verticalContentAlignment`, `rtl`. Should we add it to page

samples/v1.6/Tests/Carousel.ForbiddenActions.json
```diff
@@ -2,194 +2,201 @@
...
-										"text": "What do you think?"
-									}
-									],
-									"actions": [
-									{
-										"type": "Action.Submit",
-										"title": "Neat!"
+	"body": {
+		"type": "Carousel",
+		"pages": [
```
licanhua, line 7: I prefer CarouselItems, CarouselSet or CarouselPages.

samples/v1.6/Elements/Carousel.json
```diff
@@ -2,51 +2,51 @@
...
-		"id": "theSecondCarouselPage",
-		"items": [
-	  	{
-	  		"type": "Image",
-	  		"url": "https://adaptivecards.io/content/cats/2.png",
-	  		"size": "medium"
-	  	}
+	"body": {
+		"type": "Carousel",
+		"timer": 5000,
```
licanhua, line 7: add height which is  mandatory at the beginning

jwoo-msft:
> @licanhua I think the confusion you're referring to will exist anyway. The existing AC model is ` body[ ... ]`, chances are card authors will not get why they can't just put a `Carousel` in there. The bottom line is they have to understand that a card is either a single Carousel and it needs to be specified with `body { ... }` or a card can be what it is today and use the `body [ ... ]` notation. They will only understand that by either:
> 
> * Reading documentation
> * Or being guided by the designer which will abstract away the syntax anyway
> 
> So bottom line, if they have to learn and understand something, learning that a card with a `Carousel` it its body cannot contain any other element is just as easy as learning to use the `body { ... }` syntax IMO.
> 
> That being said, I'd say there's value in supporting the `body: { ... }` syntax generically, as a shortcut to creating a container (be it an AdaptiveCard, a Container, a Column, a TableCell etc.) with just one element in it. It would just be a very minor change in the way Container parses its elements:
> 
> ```ts
>     protected internalParse(source: any, context: SerializationContext) {
>         super.internalParse(source, context);
> 
>         this.clear();
>         this.setShouldFallback(false);
> 
>         let jsonItems = source[this.getItemsCollectionPropertyName()];
> 
>         if (typeof jsonItems === "object") {
>             jsonItems = [ source[this.getItemsCollectionPropertyName()] ];
>         }
> 
>         if (Array.isArray(jsonItems)) {
>             for (let item of jsonItems) {
>                 let element = context.parseElement(this, item, !this.isDesignMode());
> 
>                 if (element) {
>                     this.insertItemAt(element, -1, true);
>                 }
>             }
>         }
>     }
> ```
> 
> The code simply creates a one object array on the fly if the payload contains an object and not an array, nothing else. The parsing logic remains unchanged otherwise, no need to introduce the concept of a "singleton child", and no change is needed in the rendering logic. And in terms of the registry, instead of introducing a new "singleton registry", I think it should be possible to associate a constraint with each registration in the elements registry:
> 
> ```ts
> GlobalRegistry.defaultElements.register("Container", Container, true /* singletonOnly */);
> ```
> 
> `Container`'s parsing logic can then invoke `SerializationContext.parseElement` passing it a flag indicating whether it accepts singletons (basically, if it's empty) which would then pass that along to `CardObjectRegistry.createInstance`.

TBH, I was uneasy about the body becoming an object for Carousel, but the generalization makes sense to me and makes me feel good about this schema change.

source/nodejs/adaptivecards/tsconfig.jest.json
```diff
@@ -1,8 +1,5 @@
 {
 	"extends": "./tsconfig.json",
```
paulcam206, line 2: probably don't even need this file any more... should probably delete it...

samples/v1.6/Tests/Carousel.ForbiddenActions.json
paulcam206: we can discuss `items` vs. `pages`, but all of our existing container-like elements have their collections in a property with a single word all-lowercase name (e.g. `items` or `columns`)

source/nodejs/adaptivecards/src/card-elements.ts
paulcam206: not quite -- `isSingletonContainer()` checks to see if `singletonChild` is either a valid `CardElement` OR is `undefined`

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -7642,16 +7436,45 @@ export class SerializationContext extends BaseSerializationContext {
...
 
         let forbiddenTypes: string[] = [];
-        if (parent?.isInCarouselCard())
-        {
-            forbiddenTypes = Array.from(GlobalRegistry.forbiddenCarouselElements.values());
-        }
+        // if (parent?.isInCarouselCard())
+        // {
+        //     forbiddenTypes = Array.from(GlobalRegistry.forbiddenCarouselElements.values());
+        // }
```
almedina-ms, line 7477: why is this commented?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -7683,16 +7506,10 @@ export class SerializationContext extends BaseSerializationContext {
         forbiddenActionTypes: string[],
         allowFallback: boolean): Action | undefined {
 
-        let forbiddenTypes: string[] = [];
-        if (parent.isInCarouselCard())
-        {
-            forbiddenTypes = Array.from(GlobalRegistry.forbiddenCarouselActions.values()).concat(forbiddenActionTypes);
-        }
```
almedina-ms, line 7690: what is taking care of the forbidden items inside a carousel?

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        new Swiper(swiperContainer, swiperOptions);
+    }
+
+    private createCarouselPageInstance(source: any, context: SerializationContext): CarouselPage | undefined {
+        return context.parseCardObject<CarouselPage>(
+            this,
+            source,
+            CarouselPage.unsupportedElements,
```
almedina-ms, line 281: >CarouselPage.unsupportedElements [](http://example.com/codeflow?start=12&length=32)

is this array propagated? for example, in ColumnSets, where the parsing of Columns is done differently (the same issue may happen on Tables)

paulcam206: I'm currently reworking this stuff (it was only half-implemented here) -- I'll unmark this PR as draft once it's good to go :)

source/nodejs/adaptivecards/src/card-elements.ts
paulcam206: as noted in the PR description, it's not implemented here yet...

source/nodejs/adaptivecards/src/card-elements.ts
paulcam206: as noted in the PR description, it's not implemented here yet...

paulcam206:
I still need to do a top-down review, but I believe all the pieces are in place now. I still have some tests to write, and there's one thing I want to address around naming for getting items out of `Carousel`'s `pages` property -- there's currently two sets of functions: `getItem[At|Count]` and `getPage[At|Count]`. Needs to be fixed before commit.

paulcam206:
Build failures are because I haven't pushed updated `package-lock.json`s yet -- will do once I have a clean from-scratch build.

dclaux:
I'll insist a bit here.

I don't believe the whole "singleton" stuff is the way to go. The whole thing can be implemented in a straightforward way that fits the existing model without requiring special parsing or rendering logic, as I described above.

dclaux:
That `singletonChild` doesn't fit in any of the built-in behaviors. `updateLayout`, `getElementById`, `getActionById`, `validateProperties` don't work because they don't consider that "singleton child" (and I'm probably forgetting other things). So of course, you could sprinkle `if` statements everywhere and make it work, but that's not the right way to go. A "singleton container" is a "singleton container" because it has only one child. The goal is simply to not allow multiple children when the first child of a container is a `Carousel` (or any other element that identifies itself as only be usable as a singleton). And with syntactic sugar on top of that which allows `body` to be specified as an object instead of an array you have everything this feature requires without introducing an unnecessary new concept.

paulcam206:
> That `singletonChild` doesn't fit in any of the built-in behaviors. `updateLayout`, `getElementById`, `getActionById`, `validateProperties` don't work because they don't consider that "singleton child" (and I'm probably forgetting other things). So of course, you could sprinkle `if` statements everywhere and make it work, but that's not the right way to go. A "singleton container" is a "singleton container" because it has only one child. The goal is simply to not allow multiple children when the first child of a container is a `Carousel` (or any other element that identifies itself as only be usable as a singleton). And with syntactic sugar on top of that which allows `body` to be specified as an object instead of an array you have everything this feature requires without introducing an unnecessary new concept.

Thanks for the insight David -- I'll take a look :)

dclaux:
I like the latest changes, thanks for taking a stab at my suggestion!

paulcam206:
> I like the latest changes, thanks for taking a stab at my suggestion!

thanks for the help!

source/nodejs/adaptivecards/src/card-elements.ts
paulcam206: happy to entertain other naming ideas -- I just picked the first thing that came to mind, and haven't been able to come up with something better :)

samples/v1.6/Elements/Carousel.json
paulcam206: we should definitely have samples that show this off. I believe that @JeanRoca is on the hook to come up with our official samples, which we can add later...

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -0,0 +1,345 @@
...
+export class Carousel extends ContainerBase {
+    //#region Schema
+
+    static readonly timerProperty = new NumProperty(Versions.v1_6, "timer", undefined);
+    @property(Carousel.timerProperty)
+    get timer(): number | undefined {
+        let timer = this.getValue(Carousel.timerProperty);
+
+        if (timer && timer < this.hostConfig.carousel.minAutoplayDelay) {
+            console.warn(Strings.errors.tooLittleTimeDelay);
```
paulcam206, line 61: @jwoo-msft - `console.warn` here isn't really the right thing to do. we should be logging these via `logEvent`. can you file an issue to track?

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -0,0 +1,345 @@
...
+            }
+        }
+
+        swiperContainer.appendChild(swiperWrapper as HTMLElement);
+
+        cardLevelContainer.appendChild(swiperContainer);
+
+        this.initializeSwiper(swiperContainer, nextElementDiv, prevElementDiv, pagination);
+
+        cardLevelContainer.onfocus = () => {
```
paulcam206, line 256: @almedina-ms - instead of `onfocus`/`onmouseenter`/etc., we should be preferring [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener). can you file an issue to track?

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -0,0 +1,345 @@
...
+            ShowCardAction.JsonTypeName,
+            "Media",
+            "ActionSet",
+            "Input.Text",
+            "Input.Date",
+            "Input.Time",
+            "Input.Number",
+            "Input.ChoiceSet",
+            "Input.Toggle",
+            ...super.forbiddenChildElements()
```
jwoo-msft, line 95: nit: is this needed, we are parsing from top to bottom and adding a forbidden item list to the children.

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -0,0 +1,345 @@
...
+    }
+}
+
+export class Carousel extends ContainerBase {
+    //#region Schema
+
+    static readonly timerProperty = new NumProperty(Versions.v1_6, "timer", undefined);
+    @property(Carousel.timerProperty)
+    get timer(): number | undefined {
+        let timer = this.getValue(Carousel.timerProperty);
```
jwoo-msft, line 58: simple question, will `getValue()` return undefined if the timer value was not present?

jwoo-msft (approved):
LG

source/nodejs/adaptivecards/src/carousel.ts
jwoo-msft: https://github.com/microsoft/AdaptiveCards/issues/6699

source/nodejs/adaptivecards/src/carousel.ts
jwoo-msft: https://github.com/microsoft/AdaptiveCards/issues/6701

almedina-ms: Issue opened

https://github.com/microsoft/AdaptiveCards/issues/6700

almedina-ms (approved):

source/nodejs/adaptivecards-site/package.json
```diff
@@ -5,6 +5,7 @@
 	"scripts": {
 		"clean": "hexo clean",
 		"build": "hexo generate",
+		"release-deps": "npx lerna run --scope adaptivecards-site --include-dependencies release --stream",
```
almedina-ms, line 8: >"release-deps": "npx lerna run --scope adaptivecards-site --include-dependencies release --stream", [](http://example.com/codeflow?start=2&length=99)

will this require pipeline changes?

paulcam206: no - this is just a little helper script... you can `npm run release-deps` in `source/nodejs/adaptivecards-site`, and it causes `lerna` to build only what's needed for the site. I just got tired of typing that command :)

source/nodejs/adaptivecards/src/carousel.ts
paulcam206: I'll take a look and remove in a subsequent PR :)

licanhua (approved):

source/nodejs/adaptivecards/src/carousel.ts
paulcam206: that's right :)

PR #6677, opened 2021-11-02
fixed a11y tab index issue not being able to use space on pagination bullets

jwoo-msft:
# Description
tabbing through pagination works, but pressing `space` or `enter` doesn't allow moving to the carousel page as it would if the bullet was clicked. Added support for those key-down events.

* swiper adds the bullets on its own.
* bullets are not created at the time we init the swiper module
* if bullets are empty, a11y key down events are not added to swiper.
* Made onfocus event on carousel card to init the swiper module for a11y.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6677)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5868,9 +5869,17 @@ export class Container extends ContainerBase {
...
+        cardLevelContainer.appendChild(swiperContainer);
+
         this.initializeSwiper(swiperContainer, nextElementDiv, prevElementDiv, pagination, displayProperties);
 
-        cardLevelContainer.appendChild(swiperContainer);
+        cardLevelContainer.onfocus = () => {
+            if (!this._isSwiperInitialized) {
+                this._isSwiperInitialized = true;
+                this._swiper?.destroy();
+                this.initializeSwiper(swiperContainer, nextElementDiv, prevElementDiv, pagination, displayProperties);
```
almedina-ms, line 5880: why is this reinitialization needed?

jwoo-msft: swiper's a11y init expects bullets, but the bullet is not generated; i believe it's because adaptivecard is not in the document. so on focus on the carousel, we check if the swiper is initialized. this way, if not we re-init, then the bullets will be there and allow a11y config to occur.

almedina-ms (approved):

PR #6695, opened 2021-11-05
Auto generate content for VNext Preview feature

### LINKED ISSUE #5619: [Website] We need a vNext moniker to opt-into preview features
# Platform

- [ ] Website
- [ ] Designer
- [ ] Infrastructure
- [ ] Documentation

# Description

As we ship website builds out of `main` we need to clearly indicate to users on Schema Explorer and the Designer which features are vNext and "not final"

licanhua:
Fix #5619

As we ship website builds out of main we need to clearly indicate to users on Schema Explorer and the Designer which features are vNext preview feature and "not final"

1. if you want to mark some version as preview in schema explorer, you just need to change style like 
```
.ac-schema-version-1-6::after
  content: "Preview";
  color: White;
  background-color: blue;
  padding: 0px 4px 4px 4px;
  margin: 1px;
  font-size: 120%;
  border-radius: 4px;

```
2. maxSupportedVersion is introduced. latest <= maxSupportedVersion. if any class or property is over latest version number, then it's assumed the preview version in designer.  

Because /explorer heavily depends on markdown, image instead of css is used to mark the 'preview'. 

Below is screenshot which assume 1.5 is the preview feature.
- Designer warning when preview property is used but target version is less than preview version
image

- Designer target version
image

Below is screenshot which assume 1.1 is the preview feature.
- document on /explorer site
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6695)


source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -83,6 +83,11 @@ export class Version {
         return 0;
     }
 
+    isVNextFeature(): boolean
+    {
+        return isVersionLessOrEqual(Versions.latest, this);
+    }
```
paulcam206, line 89:
```suggestion
    get isPreview(): boolean
    {
        return isVersionLessOrEqual(Versions.latest, this);
    }
```

paulcam206 (changes requested):

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -109,7 +114,12 @@ export class Versions {
     static readonly v1_3 = new Version(1, 3);
     static readonly v1_4 = new Version(1, 4);
     static readonly v1_5 = new Version(1, 5);
+    static readonly v1_6 = new Version(1, 6);
     static readonly latest = Versions.v1_5;
+    // version over latest is considered "vNext feature"
+    // if latest or maxSupportedVersion is changed,
+    // don't forget to update getPreviewMarkDown in ac-typed-schema\src\markdown\generate-markdown.ts
```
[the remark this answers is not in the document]
licanhua: what do you want to catch? generate-markdown.ts has no knowledge about latest and maxSupportedVersion.
It only catches things like maxSupportedVersion changed to 1.7, but not the latest is changed.

source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
```diff
@@ -13,6 +13,14 @@ import * as console from "console";
 import * as mdTable from "markdown-table";
 import * as style from "./style";
 
+const getPreviewMarkDown = (version: string) => {
+	if (version === '1.6') {
+		return '![Preview Feature](preview.png "This is a vNext preview feature.")';
+	} else {
+		return '';
+	}
```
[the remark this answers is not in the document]
licanhua: I tried, it makes the table in markup doesn't work as expect

image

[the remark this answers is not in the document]
licanhua: keep image implementation.

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -479,6 +489,16 @@ export class ValueSetProperty extends PropertyDefinition {
...
 
+                    if (targetVersion.isVNextFeature()) {
+                        context.logParseEvent(
+                            sender,
+                            Enums.ValidationEvent.VNextFetureUsed,
+                            Strings.errors.propertyIsVNextFeature(
+                                sourceValue,
+                                this.name,
+                                targetVersion.toString()));
+                    }
```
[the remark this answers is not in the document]
licanhua: another change is in SerializableObject which is not PropertyDefinition. but I made it as a common function

source/nodejs/adaptivecards/src/serialization.ts
licanhua: To avoid conflict with IsPreviewMode, I rename it to isPreviewFeature

source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
```diff
@@ -47,7 +51,7 @@ const __ = markdownConfig.i18n.__;
 export { __ };
 
 export function createPropertiesSummary(classDefinition: SchemaClass, knownTypes, autoLink, includeVersion, elementVersion) {
-	var md = '';
+	var md = "";
```
licanhua, line 54:
```suggestion
	var md = '';
```

paulcam206 (changes requested):
code looks good -- have one style suggestion and a request for us to be able to put labels on properties details as well 👍

source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
```diff
@@ -94,7 +98,7 @@ export function createPropertiesSummary(classDefinition: SchemaClass, knownTypes
 			var summary = getPropertySummary(property, knownTypes, autoLink, elementVersion);
 
 			var formattedProperty: any = {
-				Property: style.propertyNameSummary(name),
+				Property: style.propertyNameSummary(name) + getPreviewMarkDown(summary.version),
```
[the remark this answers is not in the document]
licanhua: image

paulcam206 (approved):

source/nodejs/adaptivecards-designer/src/shared.ts
```diff
@@ -14,6 +14,6 @@ export class GlobalSettings {
         Adaptive.Versions.v1_3,
         Adaptive.Versions.v1_4,
         Adaptive.Versions.v1_5,
-        Adaptive.Versions.v1_6
+        Adaptive.Versions.v1_6,
```
almedina-ms, line 17: >Adaptive.Versions.v1_6, [](http://example.com/codeflow?start=8&length=23)

nit: extra change

source/nodejs/ac-typed-schema/src/markdown/generate-markdown.ts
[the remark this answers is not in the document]
jwoo-msft: `Property: style.propertyNameSummary(name) + getPreviewMarkDown(summary.version)`
nit:Do we have a separator token around '+'?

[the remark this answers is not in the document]
jwoo-msft: `Property: style.propertyNameSummary(name) + getPreviewMarkDown(summary.version)`
nit:Do we have a separator token around '+'?

almedina-ms (approved):
:shipit:

source/nodejs/adaptivecards-site/themes/adaptivecards/scripts/helper-properties-details.js
```diff
@@ -20,7 +20,12 @@ hexo.extend.helper.register('properties_details', function (locals, properties,
 		// Get the markdown for the property and turn it into HTML
 		html += '<div class="w3-container">';
 
-		html += marked(typedschema.markdown.createPropertyDetails(property, 3, null, false, true, elementVersion), { headerPrefix: "dedupe-header" });
+		// mark header with class name like ac-schema-version-1.5
+		html +=  `<div class="ac-schema-version-${elementVersion?.replace(/\./, '-')}" style="display: flex;">`
```
jwoo-msft, line 24: nit: why we don't refactor this as we do in createPropertyDetailsHeader as etc?

source/nodejs/adaptivecards/src/enums.ts
```diff
@@ -177,7 +177,8 @@ export enum ValidationEvent {
     UnsupportedProperty,
     RequiredInputsShouldHaveLabel,
     RequiredInputsShouldHaveErrorMessage,
-    Other
+    Other,
+    PreviewFeatureUsed,
```
jwoo-msft, line 181: nit: PreviewFeatureInUse?

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -111,6 +116,22 @@ export class Versions {
     static readonly v1_5 = new Version(1, 5);
     static readonly v1_6 = new Version(1, 6);
     static readonly latest = Versions.v1_6;
+    // version over latest is considered "vNext feature"
+    // if latest or maxSupportedVersion is changed,
+    // don't forget to update styles .ac-schema-version-1-?::after to add/remove Preview tag from adaptivecards-site\themes\adaptivecards\source\css\style.css
+    static readonly maxSupportedVersion = Versions.v1_6;
```
jwoo-msft, line 122: nit: why can't we `do static readonly maxSupportedVersion = latest` since the latest and maxSupportedVersion is same atm?

jwoo-msft (approved):

source/nodejs/adaptivecards-designer/src/shared.ts
licanhua: trailing commas are recommended in my old project. It has some benefit, and there is tslint rule for this.
I expect we didn't setup the tslint correctly in our project.
https://stackoverflow.com/questions/52218260/tslint-missing-trailing-comma-trailing-comma-on-the-last-line

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

source/nodejs/adaptivecards-site/themes/adaptivecards/scripts/helper-properties-details.js
licanhua: The problem is `marked` function, `div` is converted to `%3Cdiv%3E`, so it's not a `<div>` anymore.
markdown has special handle for `<span>` but it has side effect that it impacted layout. It's only be expected to use in table of markdown.

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -111,6 +116,22 @@ export class Versions {
     static readonly v1_5 = new Version(1, 5);
     static readonly v1_6 = new Version(1, 6);
     static readonly latest = Versions.v1_6;
```
licanhua, line 118:
```suggestion
    static readonly latest = Versions.v1_5;
```

source/nodejs/adaptivecards/src/serialization.ts
licanhua: It's a merge bug here. Likely Carousel changed this value, latest should be v1_5 here.

source/nodejs/adaptivecards/src/enums.ts
licanhua: good name

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -109,7 +114,10 @@ export class Versions {
     static readonly v1_3 = new Version(1, 3);
     static readonly v1_4 = new Version(1, 4);
     static readonly v1_5 = new Version(1, 5);
-    static readonly latest = Versions.v1_5;
+    static readonly v1_6 = new Version(1, 6);
+    static readonly latest = Versions.v1_1;
```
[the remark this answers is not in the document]
dclaux: Ah - never mind, I see your comment below.

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -840,7 +865,7 @@ export type PropertyBag = { [propertyName: string]: any };
 
 export abstract class SerializableObject {
     static onRegisterCustomProperties?: (sender: SerializableObject, schema: SerializableObjectSchema) => void;
-    static defaultMaxVersion: Version = Versions.latest;
+    static defaultMaxVersion: Version = Versions.maxSupportedVersion;
```
dclaux, line 868: Why is this new `maxSupportedVersion` even necessary? `Version.latest` is meant to mean exactly that: it's the latest (max) supported version. The two are now redundant unless I am missing something.

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -111,6 +116,22 @@ export class Versions {
     static readonly v1_5 = new Version(1, 5);
     static readonly v1_6 = new Version(1, 6);
     static readonly latest = Versions.v1_6;
+    // version over latest is considered "vNext feature"
+    // if latest or maxSupportedVersion is changed,
+    // don't forget to update styles .ac-schema-version-1-?::after to add/remove Preview tag from adaptivecards-site\themes\adaptivecards\source\css\style.css
+    static readonly maxSupportedVersion = Versions.v1_6;
+}
+
+const LogEventIfVersionIsPreview = (sender: SerializableObject, name: string, value: any, context: BaseSerializationContext, version: Version) => {
```
dclaux, line 125: Why is this declared as a const equal to a lambda instead of simply a function like everywhere else in the codebase?

Also, why is it not a method on `BaseSerializationContext`?

source/nodejs/adaptivecards/src/serialization.ts
dclaux, line 122: Why is this necessary when `Versions.latest` already means the same?

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -83,6 +83,11 @@ export class Version {
         return 0;
     }
 
+    isPreviewFeature(): boolean
```
dclaux, line 86: A "version" isn't a "feature", this public method name is odd. It should probably simply be `isPreview` or `isPreviewVersion`.

dclaux (changes requested):

source/nodejs/adaptivecards/src/serialization.ts
licanhua: `latest` is a little confusion here. actually it means 'stable'. maybe I need to rename it to stableVersion
 the version above `stable` but <= maxSupportedVresion will be the preview features. 
If it's above maxSupportedVersion, it raises `The specified card version (${version}) is not supported`, and that's why maxSupportedVersion is introduced.

source/nodejs/adaptivecards/src/serialization.ts
licanhua: Look like isPreviewVersion is a better name than isPreviewFeature. I avoid isPreview because we have IsPreviewMode or something, and it's easy to cause confusion

source/nodejs/adaptivecards/src/serialization.ts
licanhua: I can't remember where I read(possible react), arrow function can avoid 'this' binding problem in a class. And that's why I'm used to code with lambda.

I agree with you it's better to move it to BaseSerializationContext

source/nodejs/adaptivecards/src/serialization.ts
dclaux: We can't rename `Versions.latest` as it's a shipped API and changing it would risk breaking application code when it's not really necessary.

For the rest, please help me understand. Currently you have this:

```typescript
export class Versions {
    static readonly v1_0 = new Version(1, 0);
    static readonly v1_1 = new Version(1, 1);
    static readonly v1_2 = new Version(1, 2);
    static readonly v1_3 = new Version(1, 3);
    static readonly v1_4 = new Version(1, 4);
    static readonly v1_5 = new Version(1, 5);
    static readonly v1_6 = new Version(1, 6);
    static readonly latest = Versions.v1_5;
    static readonly maxSupportedVersion = Versions.v1_6;
}
```

You're saying that if `Versions.latest < version <= Versions.maxSupportedVersion` then `version` is a preview version. How is that different from just saying `version` is a preview if `version > Versions.latest`? When would you ever hit a situation where there is another declared version that is > maxSupportedVersion but isn't a preview version?

I'm fairly certain `maxSupportedVersion` isn't necessary.

Furthermore, if you want to be granular in terms of which version is a preview version, you are probably better off making that an explicit flag in the version instance itself:

```typescript
export class Versions {
    static readonly v1_0 = new Version(1, 0);
    static readonly v1_1 = new Version(1, 1);
    static readonly v1_2 = new Version(1, 2);
    static readonly v1_3 = new Version(1, 3);
    static readonly v1_4 = new Version(1, 4);
    static readonly v1_5 = new Version(1, 5);
    static readonly v1_6 = new Version(1, 6, undefined /* label */, true /* isPreview */);
    static readonly latest = Versions.v1_5;
}
```

The above would make v1_6 and v1_6 only a preview version.

source/nodejs/adaptivecards/src/serialization.ts
dclaux: That is only when you set an event handler within the method of a class. Your `const` is global, there's not `this` to resolve in that context.

The below will not work:

```typescript
class MyClass {
    private _privateFields: string;

    someMethod() {
        let button = document.createElement("button");

        button.onclick = function() {
            this._privateField = "Private field set";
        }
    }
}
```

But this will:
```typescript
class MyClass {
    private _privateFields: string;

    someMethod() {
        let button = document.createElement("button");

        button.onclick = () => {
            this._privateField = "Private field set";
        }
    }
}
```

source/nodejs/adaptivecards/src/serialization.ts
licanhua: I was thinking to add isPreview parameter to Version as you mentioned. but I finally dropped it.

Here is the logic:
if (version <= latest) {
   Ok
} else if (version <= maxVersion) {
   raise "it's preview"
} else {
   raise "The specified card version (${version}) is not supported"
}

If I implements 'isPreview' in version, the logic 
```
if version in [v1_0, v1_1, v1_2.... v1_7] {
      if (version.isPreview() {}
     else {
    }
} else {
}
```

So 
1. I need to implement new logic to convert "1.7' to a valid Version first or implement a new comparator
2. make v1_6, v1_7 somewhere too to do the comparison. Then if I add a new version, I will need to make change on multiple place.

In these solution, I only need to change maxSupportedVersion, and change are grouped together.

dclaux: Sorry, I still don't get it. The goal seems to be to warn the user in the designer that they're using a property value that's only supported in a version that happens to be a preview version. So all you need to know is that a version is a preview, so a flag on the Version class would work just fine.

Besides, the Version class already has a `label` property that is designed to and has been used for this type of thing. Essentially, in your case, you'd want to define v1_6 as follows:
```typescript
export class Versions {
    static readonly v1_6 = new Version(1, 6, "1.6 (Preview)");
}
```

That's enough for the version picker in the designer to show "1.6 (Preview)" and thus let the user know they're authoring their card against a preview version.

Now the designer already has the property warning you seem to want:
image

But, it so happens that it is emitted with this code:
image

Because it uses `toString()`, the label isn't displayed. If you just changed that to use `targetVersion.label` instead, you'd get exactly what you're after without needing to introduce anything new.

dclaux: In the end, using `Version.label` will yield this result:
image

Which is basically what you want based on the screenshot in the description of this PR.

licanhua: I see code like this:

```
    static parse(versionString: string, context: BaseSerializationContext): Version | undefined {
        if (!versionString) {
            return undefined;
        }

        let result = new Version();
        result._versionString = versionString;
```

How do you check if the parsed version is preview or not?

dclaux: Why would I need to if all I need is a designer warning?

But if you _do_ need it, then an `isPreview` property on `Version` is the way to go.

dclaux:
```typescript
export class Version {
    ...
    private _isPreview: boolean;
    ...
    constructor(major: number = 1, minor: number = 1, label?: string, isPreview: boolean = false) {
        ...
        this._isPreview = isPreview;
    }

    get isPreview(): boolean {
        return this._isPreview;
    }
}

```

dclaux: Oh, and of course you'd need to:
- Update the `Version.parse` logic...
- ...and thus introduce a version format that makes it possible to identify the version string as a preview

But I really don't see why you'd need all that. Using the label should be all you need to achieve the result you want, unless I am missing one of your requirements.

source/nodejs/adaptivecards/src/serialization.ts
licanhua: add declaredVersions which can be shared by both designer and serialization

source/nodejs/adaptivecards/src/serialization.ts
licanhua: not need it anymore

source/nodejs/adaptivecards/src/serialization.ts
licanhua: removed

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -846,7 +867,8 @@ export type PropertyBag = { [propertyName: string]: any };
 
 export abstract class SerializableObject {
     static onRegisterCustomProperties?: (sender: SerializableObject, schema: SerializableObjectSchema) => void;
-    static defaultMaxVersion: Version = Versions.declaredVersions[Versions.declaredVersions.length-1];
+
+    static defaultMaxVersion: Version = Versions.getMaxiumDeclaredVersion();
```
dclaux, line 871: Should be
```typescript
static defaultMaxVersion: Version = Versions.latest;
```

dclaux (commented):
Please just revert the value of defaultMaxVersion back to Versions.latest

Looks good otherwise!

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -840,6 +862,7 @@ export type PropertyBag = { [propertyName: string]: any };
 
 export abstract class SerializableObject {
     static onRegisterCustomProperties?: (sender: SerializableObject, schema: SerializableObjectSchema) => void;
+
```
licanhua, line 865:
```suggestion
```

dclaux:
Looks good!

PR #6707, opened 2021-11-09
UI Tests for carousel

almedina-ms:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6707)


source/nodejs/tests/ui-tests/src/ui-tests.test.ts
```diff
@@ -50,4 +51,85 @@ describe("Mock function", function() {
...
         Assert.strictEqual(firstInputId, activeElementId);
     }));
+
+    test("Test actions are rendered and active below carousel", (async() => {
+        await testUtils.goToTestCase("v1.6/Carousel.HostConfig");
+
+        await testUtils.clickOnActionWithTitle("See more");
+
+        // wait half a second to wait for page refresh
+        await testUtils.delay(500);
```
jwoo-msft, line 61: nit: can you make it a globally configurable property, so we can change the value in the future in one place? one test case for half a second is fine, but I'm worried in the future where we have hundreds of such tests.

jwoo-msft (approved):
looks great

samples/v1.6/Elements/Carousel.ScenarioCards.Timer.json
```diff
@@ -0,0 +1,467 @@
...
+				"type": "Action.OpenUrl",
+				"title": "Click for more information about the first carousel page!",
+				"url": "https://adaptivecards.io/"
+			},
+				"items": [
+				{
+						"type": "ColumnSet",
+						"columns": [
+							{
+								"type": "Column",
```
paulcam206, line 22: this doesn't seem quite right...

source/nodejs/adaptivecards-ui-testapp/src/rendering-utils.ts
```diff
@@ -63,25 +63,33 @@ export function renderCard(cardJson: any, callbackFunction: Function): void {
...
+
+            inputs.forEach((input) => {
+                const inputId: string = input.id;
+                const inputValue: string = input.value;
 
-            if (!isFirstInput) {
-                inputsAsJson = inputsAsJson.concat(",");
-            }
+                if (!isFirstInput) {
+                    inputsAsJson = inputsAsJson.concat(",");
```
paulcam206, line 80: instead of building up a JSON string by hand, you should be able to do something like this:

```typescript
const inputs = card.getAllInputs(true);
const inputsObject = {};
inputs.forEach((input) => {
    inputsObject[input.id] = input.value; // beware: last writer wins
});

// later, when you need the string
const inputsAsJsonString = JSON.stringify(inputsObject);
```

paulcam206 (changes requested):

source/nodejs/adaptivecards-ui-testapp/src/rendering-utils.ts
paulcam206: note: you don't have to do this -- just thought it might be helpful :)

paulcam206 (approved):

PR #6712, opened 2021-11-10
[JS][Carousel] RTL update & other small improvements

jwoo-msft:
Updated for RTL / Updated Swiper init / Fixed voiceover to announce current slide position / fixed accessibility focus 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6712)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -7153,6 +7165,10 @@ export class AdaptiveCard extends ContainerWithActions {
         return "AdaptiveCard";
     }
 
+    hasCarousel(): boolean {
+        return (this.getItemCount() === 1 && this.getItemAt(0).getJsonTypeName() === "Carousel");
+    }
```
paulcam206, line 7170: I'm not a big fan of this method -- what happens if we add another singleton-type element?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -7103,6 +7108,13 @@ export class AdaptiveCard extends ContainerWithActions {
             renderedElement.style.removeProperty("minHeight");
         }
 
+        // if rtl is set to dir at the parent view of carousel, carousel's view will get disappeared
+        if (this.hasCarousel()){
+            if (this.rtl && renderedElement) {
+                renderedElement.dir = "ltl";
```
paulcam206, line 7114: shouldn't this be:

```suggestion
                renderedElement.dir = "rtl";
```

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -7103,6 +7108,13 @@ export class AdaptiveCard extends ContainerWithActions {
             renderedElement.style.removeProperty("minHeight");
         }
 
+        // if rtl is set to dir at the parent view of carousel, carousel's view will get disappeared
```
paulcam206, line 7111: is this an issue in `swiper`? I'm interested to find out what's happening...

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -270,23 +271,25 @@ export class Carousel extends Container {
...
-        cardLevelContainer.addEventListener("focusin", (event) => {
-            if (!this._isSwiperInitialized) {
-                this._isSwiperInitialized = true;
-                this._swiper?.destroy();
-                this.initializeCarouselControl(swiperContainer, nextElementDiv, prevElementDiv, pagination);
-            }
-        });
+        cardLevelContainer.addEventListener("keydown", (event) => {
+            // we don't need to check which key was pressed, we only need to reinit swiper once, then remove this event listener
+           let activeIndex = this._swiper?.activeIndex
```
paulcam206, line 282: missing trailing `;`

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -270,23 +271,25 @@ export class Carousel extends Container {
...
 
-        cardLevelContainer.addEventListener("focusin", (event) => {
-            if (!this._isSwiperInitialized) {
-                this._isSwiperInitialized = true;
-                this._swiper?.destroy();
-                this.initializeCarouselControl(swiperContainer, nextElementDiv, prevElementDiv, pagination);
-            }
-        });
+        cardLevelContainer.addEventListener("keydown", (event) => {
+            // we don't need to check which key was pressed, we only need to reinit swiper once, then remove this event listener
```
paulcam206, line 281: what issue is this working around? what's the scenario?

paulcam206 (changes requested):

source/nodejs/adaptivecards/src/card-elements.ts
jwoo-msft: I believe so 
Screen Shot 2021-11-10 at 11 02 38 AM
note that the position of actions is correctly shown with rtl set.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -6660,6 +6660,11 @@ export abstract class ContainerWithActions extends Container {
 
         if (element) {
             let renderedActions = this._actionCollection.render(this.hostConfig.actions.actionsOrientation, false);
+            // we can't set rtl at the card level if the card has carousel, setting the rtl explicitly at the action container takes 
+            // care of card level rtl for actions
+            if (this.rtl !== undefined && renderedActions) {
```
licanhua, line 6665: Can we move this to AdativeCards, likely AdaptiveCards has knowledge about carousel and rtl

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: I understand that somehow having `rtl` set to "ltr" on the Swiper's parent element causes Swiper to not work correctly, and so that's why you're overriding here? Seems quite weird.

If you do need to override, then a better way to do it in a generic way is to introduce a new `forceRtlOnParentContainer(): boolean` method on `CardElement` that returns `false` by default, and override it in `Carousel` to return true.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Agree with @paulcam206. Related to that, `get currentCarouselPageId()` in `SubmitActionBase` isn't used anywhere and should be removed. Per my below suggestion, this `hasCarousel` method shouldn't be necessary anyway.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Why would this be necessary? The `renderedActions` element already inherits the `rtl` property of its parent container.

source/nodejs/adaptivecards/src/card-elements.ts
jwoo-msft: my comments weren't published, so the comments are out of order. @dclaux
AdaptiveCards 
           |
     Carousel
           |
  CarouselPage
if we set rtl at AdaptiveCard level, the carousel will disappear, so I'm trying to override the rtl of AdaptiveCards, not the rtl of Carousel.

source/nodejs/adaptivecards/src/card-elements.ts
jwoo-msft: @dclaux,
my above comments should have been pushed a few hours ago, so it's out of order.

`currentCarousePageId()` is used in action handler in card-designer-surface.ts to get the pageId when submit action is triggered.

jwoo-msft:
closing this PR to cherry-pick onto rebased feature/carousel branch and address review comments.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: @jwoo-msft  that's precisely what I'm saying.

Your code in AdaptiveCard.internalRender should read

```typescript
protected internalRender(): HTMLElement | undefined {
    ...
    for (let i = 0; i < this.getItemCount(); i++) {
        if (this.getItemAt(i).forceRtlOnParentContainer() && this.rtl && renderedElement) {
            renderedElement.dir = "rtl";
            break;
        }
    }
    ...
}
```

With `forceRtlOnParentContainer` defined in CardElement:
```typescript
export class CardElement extends CardObject {
    ...
    forceRtlOnParentContainer(): boolean {
        return false;
    }
    ...
}
```

And on Carousel:
```typescript
export class Carousel extends Container {
    ...
    forceRtlOnParentContainer(): boolean {
        return true;
    }
    ...
}
```

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Ok, `currentCarouselPageId()` is used exclusively for display purposes in the designer - it shouldn't be a public API available in the renderer SDK. Just move the logic directly into card-designer-surface (as below) and remove `currentCarouselPageId` from `SubmitActionBase`:

```typescript
                let carouselPageId: string | undefined = undefined;

                const root = action.getRootObject() as AdaptiveCard;

                if (root) {
                    for (let i = 0; i < root.getItemCount(); i++) {
                        let element = root.getItemAt(i);

                        if (element instanceof Carousel) {
                            carouselPageId = element.currentPageId;
                        }
                    }
                }

                if (carouselPageId) {
                    message += `\ncarousel page id: "${carouselPageId}"`;
                }

```

source/nodejs/adaptivecards/src/card-elements.ts
jwoo-msft: i found the solution to this.

source/nodejs/adaptivecards/src/card-elements.ts
jwoo-msft: I appreciate the detailed explanation, @dclaux!
i found the reason that the swiper was misbehaving with the root level, rtl. https://github.com/microsoft/AdaptiveCards/pull/6723

source/nodejs/adaptivecards/src/card-elements.ts
jwoo-msft: @dclaux, thanks for the suggestion, i will update it.

jwoo-msft: @dclaux addressed it in this [PR](https://github.com/microsoft/AdaptiveCards/pull/6728)
for some reasons, i can't add you to PR.

source/nodejs/adaptivecards/src/card-elements.ts
jwoo-msft: this is addressed in #6723

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: I think it's because someone in the team revoke some of my permissions.

PR #6720, opened 2021-11-11
Jwoo/feature/carousel 4

jwoo-msft:
# Description
* re-initializing swiper; swiper needs to be re-initialized because it won't configure pagination bullets for a11y.
* Fixed the issue where it resets the active page and swiper wrapper's attributes. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6720)

### DIFF
--- a/source/nodejs/adaptivecards/src/carousel.ts
+++ b/source/nodejs/adaptivecards/src/carousel.ts
@@ -270,23 +270,25 @@ export class Carousel extends Container {
 
         swiperContainer.appendChild(swiperWrapper as HTMLElement);
 
+        swiperContainer.tabIndex = 0;
+
         containerForAdorners.appendChild(swiperContainer);
 
         this.initializeCarouselControl(swiperContainer, nextElementDiv, prevElementDiv, pagination);
 
-        cardLevelContainer.addEventListener("focusin", (event) => {
-            if (!this._isSwiperInitialized) {
-                this._isSwiperInitialized = true;
-                this._swiper?.destroy();
-                this.initializeCarouselControl(swiperContainer, nextElementDiv, prevElementDiv, pagination);
-            }
-        });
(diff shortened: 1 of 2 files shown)


source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -270,23 +270,25 @@ export class Carousel extends Container {
...
         this.initializeCarouselControl(swiperContainer, nextElementDiv, prevElementDiv, pagination);
 
-        cardLevelContainer.addEventListener("focusin", (event) => {
-            if (!this._isSwiperInitialized) {
-                this._isSwiperInitialized = true;
-                this._swiper?.destroy();
-                this.initializeCarouselControl(swiperContainer, nextElementDiv, prevElementDiv, pagination);
-            }
-        });
+        cardLevelContainer.addEventListener("keydown", (event) => {
```
almedina-ms, line 279: why is the event changed for keydown? the issue can be hit if the controller is clicked on

almedina-ms (changes requested):
See comment

source/nodejs/adaptivecards/src/carousel.ts
jwoo-msft: this is for an a11y config for pagination bullet since pages are not available at the point of execution, thus, pressing the spacebar on the bullet won't do anything. also key down events are narrower than focus in events.

almedina-ms: yeah, but if focusin handles clicking and keyboard events why handling keyboard only now?

paulcam206 (approved):

almedina-ms (approved):

PR #6723, opened 2021-11-11
Updated RTL support for Carousel

jwoo-msft:
Added RTL support to Carousel.
Screen Shot 2021-11-11 at 2 12 43 PM
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6723)

### DIFF
--- a/source/nodejs/adaptivecards/src/carousel.ts
+++ b/source/nodejs/adaptivecards/src/carousel.ts
@@ -205,7 +205,6 @@ export class Carousel extends Container {
         // `isRtl()` will set the correct value of rtl by reading the value from the parents
         this.rtl = this.isRtl();
         this.applyRTL(swiperContainer);
-        //swiperContainer.dir = 'rtl';
         cardLevelContainer.appendChild(swiperContainer);
 
         const swiperWrapper: HTMLElement = document.createElement("div");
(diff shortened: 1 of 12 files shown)


source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -199,13 +200,12 @@ export class Carousel extends Container {
         }
 
         const cardLevelContainer: HTMLElement = document.createElement("div");
-
-        const containerForAdorners: HTMLElement = document.createElement("div");
```
almedina-ms, line 203: this container should not be removed, this allows the arrows and pagination to be rendered on the sides and below the carousel correctly

source/nodejs/tests/ui-tests/src/ui-tests.test.ts
```diff
@@ -14,6 +14,7 @@ describe("Mock function", function() {
     const delayForInputRetrieval: number = 500;
     const delayForCarouselArrows: number = 1000;
     const delayForCarouselTimer: number = 5500;
+    const timeOutValue: number = 9000;
```
almedina-ms, line 17: can we rename this variable to specify what this timeout is actually for?

almedina-ms (changes requested):

source/nodejs/adaptivecards/src/carousel.ts
jwoo-msft: @almedina-ms, this is causing quite an issue. with this, rtl won't work correctly.

almedina-ms (approved):

PR #6732, opened 2021-11-12
[JS] Change how we produce/consume CSS

paulcam206:
# Related Issue

Not sure we have an actual bug tracking this. Details below.

Related #6710 -- this is a partial solution (basically bootstrapping `adaptivecards` for this work and doing the initial `.scss` authoring)

# Description

The carousel element **requires** CSS to be present in order to function, so we need to make sure we're producing our CSS in a way that is structured for consumption. Our prior approach of just packing up an optional flat `.css` isn't sufficient for this case.  Consumers might already be using their own styles for AdaptiveCards, in which case it would be bad form to force them to take in all of our styles just so carousel works. At the same time, consumers might be new and just want an "easy button" way to stand up a host with our defaults, so we need to provide that as well. 

The fix is to move over to using [Sass](https://sass-lang.com/) to define our CSS styling. Basically, we declare things in the SCSS language, and then process them to produce CSS. This makes it easy for us to generate a single flat `.css` with all of our defaults (including carousel styles) as well as a targeted carousel-only CSS file. As a bonus, we can also publish our `.scss` files for customers to pick and choose what they want if they want a custom build.

There's additional fallout addressed here around making sure that the carousel CSS is consumed (and in some cases, republished) properly in all of our packages.

## Detailed description

in PR file list order:
* `adaptivecards-designer-app`
  * `dist/*`: these files shouldn't have been committed in the first place, so removing (noticed after adding a `clean` npm task)
  * `package.json`: added clean task (for hygiene's sake)
  * `webpack.config.js`:
    * differentiate between dev and prod builds
    * place `adaptivecards-designer.css` in the correct location
* `adaptivecards-designer`
  * `package.json`: add inner loop task to rebuild the designer and its dependencies
  * `webpack.config.js`:
    * strip out prior hacky fix (bb0c9d07f5686bb063cc937010c2a2c415cde381)
    * add carousel styles to `adaptivecards-designer.css`
    * remove an odd double-copy we were doing (see [here](https://github.com/microsoft/AdaptiveCards/blob/77f8a693a238957aaf57f4fb86ed6db5d9a04bb6/source/nodejs/adaptivecards-designer/webpack.config.js#L82-L83) and [here](https://github.com/microsoft/AdaptiveCards/blob/77f8a693a238957aaf57f4fb86ed6db5d9a04bb6/source/nodejs/adaptivecards-designer/webpack.config.js#L92-L95) - looks to have been around for quite some time)
* `adaptivecards-react`
  * `package.json`: add clean step to `release`
  * `webpack.config.js`:
    * formatting fixes
    * remove unnecessary `.json` resolve extension
    * remove unnecessary CSS processing rule
    * copy `adaptivecards` CSS files to the package
* `adaptivecards-site`
  * `.../designer.ejs`: formatting fixes. had previously introduced a change here, but that didn't end up being necessary
  * `.../generator-adaptiveassets.js`: formatting fixes and make sure `adaptivecards-designer.css` is available as a non-hashed asset (the designer has an [existing dependency](https://github.com/microsoft/AdaptiveCards/blob/77f8a693a238957aaf57f4fb86ed6db5d9a04bb6/source/nodejs/adaptivecards-designer/src/card-designer.ts#L1101-L1107) on being able to load it by name)
* `adaptivecards-ui-testapp`
  * `index.html`: no need to manually bring in script or CSS -- that's webpack's job
  * `package.json`:
    * mark package as private so we [can't accidentally publish it](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#private). this is also noticed by `lerna`.
    * we don't need to produce a published package of this project, so there's no need to build loose javascript files. remove `dts` task and reference to `lib` dir
    * add `clean` task to `release`
    * remove `cpy-cli` dependency since it's not used
 * `src/adaptivecards-ui-testapp.css`: remove carousel styles
 * `src/adaptivecards-ui-testapp.ts`:
   * no need to export things since we're not a library and the main page doesn't have its own script references any more
   * import the CSS we need for the app
   * add script back in that used to live in `index.html`
 * `src/rendering-utils.ts`:
   * formatting fixes
   * use object model to set properties on `itemLink` variable
   * move from using `XMLHttpRequest` to the now-preferred `fetch()`
 * `webpack.config.js`:
   * remove hacky fix
   * configure CSS loaders for dev/prod builds
   * since we're ingesting CSS through the `import` directives in `adaptivecards-ui-testapp.ts`, they'll get included via the CSS loaders, so we can remove all of the CSS processing we were doing here before
   * I broke samples publishing when we moved to `webpack@^5` (7bccd1056fdc43eb871b367531c37603c43b1c39), so fix that
* `adaptivecards`
  * `.gitignore`: CSS should be generated for this project, so have git ignore any CSS files that might be added
  * `package.json`:
    * add `generate-css` task to process our `.scss` files into `.css` - we may want to have a more scalable solution in the future, but this works for now
    * add `build-source` task that performs the TypeScript build as well as generate CSS, and have this run before `release` and `build` tasks
  * `src/adaptivecards.css`: no longer need this, so remove it
  * `scss/`:
    * `adaptivecards-base.scss`:
      * renamed from an old checked-in `base.scss` that wasn't actually being utilized
      * updated to include styles that were only added to `adaptivecards.css`
    * `adaptivecards-carousel.scss`:
      * `swiper` styles previously defined in `adaptivecards.css` go here
      * references `swiper` `scss` files for the modules we're using
      * removed some hard overrides of `swiper` styles that are configurable via [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) (e.g. [`--swiper-pagination-color`](https://github.com/nolimits4web/swiper/blob/2824b0cf35398c97b7027b11bc6a28153bfa4b5c/src/modules/pagination/pagination.scss#L3-L15) used [here](https://github.com/microsoft/AdaptiveCards/pull/6732/files#diff-2a0ac0fc4fd40f8a5200042344f9ae135d2cfa28485b6ed81ffb096f429e7b7aR94-R101))
      * factor some things (e.g. `.swiper-button-[next|prev].ac-carousel-[right|left]` out to avoid duplicating markup
    * `adaptivecards-default.scss`:
      * consumes `-base` and `-carousel` and produces what we used to call `adaptivecards.css`
    * `adaptivecards-vars.scss`: basically just exists for us to share variables between `-default` and `-carousel`
  * `webpack.config.js`:
    * rip out all of the old CSS processing stuff
    * copy `.scss` and generated `.css` files to appropriate locations
* `nodejs/package.json`: add `sass` dependency

# How Verified

* local builds

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6732)


source/nodejs/adaptivecards-site/themes/adaptivecards/scripts/generator-adaptiveassets.js
```diff
@@ -64,10 +62,22 @@ var customAssets = [
 		noHash: true
 	},
 	{
+        // designer module (hashing not working for CSS files; the designer expects certain filenames)
+        path: "node_modules/adaptivecards-designer/dist/adaptivecards-designer.css",
+		dest: function (p) { return p; },
+		noHash: true
```
paulcam206, line 68: this indenting is what happens when you copy/paste 🤦‍♂️

almedina-ms (approved):

source/nodejs/adaptivecards-designer/webpack.config.js
```diff
@@ -42,7 +39,7 @@ module.exports = (env, argv) => {
 				exclude: /(node_modules|__tests__)/
 			},
 			{
-				test: /\.css$/,
+				test: /\.css$/i,
```
jwoo-msft, line 42: nit: looks like a typo, but i'm certain it's not.

source/nodejs/adaptivecards-designer/webpack.config.js
```diff
@@ -97,13 +94,13 @@ module.exports = (env, argv) => {
 					from: 'src/adaptivecards-designer.css',
 					to: './[name][ext]'
 				},
+                {
```
jwoo-msft, line 97: nit: spacing

jwoo-msft (approved):

source/nodejs/adaptivecards-designer/webpack.config.js
paulcam206: this just makes the regex [case-insensitive](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#advanced_searching_with_flags), though it's not really necessary here. I'll back it out

source/nodejs/adaptivecards-designer-app/webpack.config.js
```diff
@@ -36,19 +36,26 @@ module.exports = (env, argv) => {
...
 			]
 		},
 		plugins: [
 			new CopyWebpackPlugin({
-				patterns: [{
-					from: 'node_modules/adaptivecards-designer/src/containers/*',
-					to: 'containers/[name][ext]'
-				}]
+				patterns: [
+                    {
```
paulcam206, line 48: more indenting woes

source/nodejs/adaptivecards-designer/webpack.config.js
```diff
@@ -84,7 +80,11 @@ module.exports = (env, argv) => {
 				bundles: [
 					{
 						dest: 'dist/adaptivecards-designer.css',
-						src: ['./node_modules/adaptivecards-controls/dist/adaptivecards-controls.css', './src/adaptivecards-designer.css']
+						src: [
```
paulcam206, line 83: more indentation nonsense

jwoo-msft (approved):

PR #6743, opened 2021-11-16
prettier evaluation

licanhua:
This PR is for evaluation purpose.
I converted the adaptivecards folder with the default prettier setting.
```
.\node_modules\.bin\prettier --write "adaptivecards/**/*.{js,jsx,ts,tsx,json,css,scss,md}"
```

Prettier is just a code formatter, and you will see more from [why prettier](https://prettier.io/docs/en/why-prettier.html)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6743)


paulcam206:
@licanhua - digging in to this a bit deeper, I think you're right that we should go the `prettier` route (avoiding having to run the linter to get formatting is pretty compelling given that `eslint` is a bit on the slow side).

However, this is going to have to be coordinated with #6757, which I'm currently working on... given that I'm already updating `eslint` stuff, I think it makes sense for `prettier` changes to go in at the same time... are you okay with me taking the `prettier` integration task?

licanhua:
> @licanhua - digging in to this a bit deeper, I think you're right that we should go the `prettier` route (avoiding having to run the linter to get formatting is pretty compelling given that `eslint` is a bit on the slow side).
> 
> However, this is going to have to be coordinated with #6757, which I'm currently working on... given that I'm already updating `eslint` stuff, I think it makes sense for `prettier` changes to go in at the same time... are you okay with me taking the `prettier` integration task?

Feel free to take it

PR #6760, opened 2021-11-19
[Android] Updates for Android UI testing

almedina-ms:
# Related Issue

# Description

The current pipeline uses an API 27 emulator, sadly there's no way to reproduce the same environment (agent OS where ci tests run and emulator) so bumping the API version to the next one so we have a way to reproduce UI testing in development machines.

* Adds a general timeout for of 1 minute for every test to avoid them hanging builds

* Also macos 10.14 as agent is being deprecated so moving to next version

# How Verified

Running the pipeline

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6760)


paulcam206 (approved):

beervoley (approved):

almedina-ms:
While this PR has been approved it doesn't completely solve the tests failing ocassionally so I'll close it in the meantime while we find a better solution

PR #6790, opened 2021-11-29
Remove pic2card and experimental source

### LINKED ISSUE #6789: Remove unneeded pic2card source

paulcam206:
# Related Issue

Fixes #6789 

# Description

Simple removal of pic2card and experimental source since it should not live in our codebase.

# How Verified

* local build and test of site

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6790)


licanhua:
Can you keep source/pic2card/README.md, and state that that project is deprecated and source code is in https://github.com/microsoft/AdaptiveCards/releases/tag/21.09?

licanhua:
workflow, bug template and adaptivecards/readme need to update too

image
image

source/pic2card/ui/README.md
```diff
@@ -1,29 +0,0 @@
-# ann-app
-
```
licanhua, line 2: keep this file and mark it as deprecated.
update /readme.md
update bug template, workflow

image
image

source/pic2card/README.md
```diff
@@ -1,152 +0,0 @@
-# Pic2Card
-![pic2card-build](https://github.com/Microsoft/AdaptiveCards/workflows/pic2card-build/badge.svg)
```
licanhua, line 2: keep this file and update all links related to Pic2Cards on /

licanhua (changes requested):

almedina-ms:
With the removal of this project and the vs extension which were provided by the community, is there some plan in place to move these tools into a different repository?

paulcam206:
> With the removal of this project and the vs extension which were provided by the community, is there some plan in place to move these tools into a different repository?

The VS extension was already shipping out of a different repository, so there's really no action to take. I think we would want to partner with folks to spotlight community work (and be available to provide guidance or support if they'd like), but it's not really appropriate to host external tools here...

source/pic2card/ui/README.md
paulcam206: I've updated `/README.md`, but I don't want to leave breadcrumbs around the repo of projects that are no longer hosted here (see, e.g. the old NodeJS visualizer codebase).

almedina-ms (approved):

licanhua (approved):

PR #6831, opened 2021-12-08
Deleting unofficial VsCode extension

jonmill:
# Related Issue

N/A

# Description

Removes the unofficial VsCode extension from the repo

# Sample Card

N/A

# How Verified

N/A

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6831)


paulcam206 (approved):

almedina-ms (approved):

source/vscode/README.md
```diff
@@ -1,42 +0,0 @@
-## Features
```
licanhua, line 1: Can you keep this file, and add something like `vscode is removed, and please goto link to find the source code`

jonmill: That's documented in the extension page itself: https://marketplace.visualstudio.com/items?itemName=madewithcardsio.adaptivecardsstudiobeta

I don't think we should be responsible for keeping a readme up-to-date for that as we don't own it and that repo could change

licanhua: You are right. look like the old extension is deleted, so we should not have link to this

licanhua (approved):

Issue #6746, opened 2021-11-16
[JS] react project `Cannot find module: 'swiper'. Make sure this package is installed.`


licanhua:
```
c:\repo\my-app>npm run build

> my-app@0.1.0 build
> react-scripts build

Creating an optimized production build...
Failed to compile.

.\node_modules\adaptivecards\lib\carousel.js
Cannot find module: 'swiper'. Make sure this package is installed.

You can install this package by running: npm install swiper.

```

Repro steps:
```
npx create-react-app my-app --template typescript --use-npm
npm i adaptivecards-react@1.1.0-alpha.0 --save
```

then modify app.tsx
```tsx
import React from 'react';
import logo from './logo.svg';
import './App.css';
import { AdaptiveCard } from "adaptivecards-react";
var card = {
  "type": "AdaptiveCard",
  "version": "1.0",
  "body": [
      {
          "type": "TextBlock",
          "text": "Here is a ninja cat"
      },
      {
          "type": "TextBlock",
          "text": "Here is another ninja cat"
      },
      {
          "type": "Image",
          "url": "http://adaptivecards.io/content/cats/1.png"
      }
  ]
};

var hostConfig = {
  fontFamily: "Segoe UI, Helvetica Neue, sans-serif"
};


function App() {
  return (
    <div className="App">
      <header className="App-header">
        <img src={logo} className="App-logo" alt="logo" />
      </header>
      <AdaptiveCard payload={card} hostConfig={hostConfig} />

    </div>
  );
}

export default App;

```

paulcam206:
This is an issue with how the `adaptivecards` package declares its dependency on `swiper`. It should probably be a `peerDependency`:

https://github.com/microsoft/AdaptiveCards/blob/d0beb67cc1d3affe7847bc6ffc21e40dd7a5d0f2/source/nodejs/adaptivecards/package.json#L47

alternatively, `adaptivecards-react` could import `adaptivecards` through the `dist` folder, which has `swiper` bundled in to it

PR #6833, opened 2021-12-09
[JS] Create a JS pipeline

### LINKED ISSUE #6656: Add and Enable JS test on pipeline

### LINKED ISSUE #6749: [JS] setup new pipeline to consume npm packages

licanhua:
Fix #6656. which enabled JS test
Fix #6749. It's used to catch bugs like #6746.
Currently the js-ci pipeline failed because of #6843. It's expected. 
note: all .ts files' change are not related to my PR and it's automatically formatted by huskey hook when I merged the branch with main

Because lerna workspace, webpack and npm pack, the actual package we released is a little different what we see in source/nodejs/adaptivecards*/. So even the build passed on our dev environment, we still don't have 100% confidence on our package.
This PR used `npm pack` to build the package we `npm publish` to npmjs website, and then the two apps consumes these packages.

Here is the logic.
1. make a copy for the two test project: adaptivecards-react-testapp and adaptivecards-ui-testapp
2. build the js projects as usual
3. run `npm pack` to make packages for adaptivecards, adaptivecards-controls, adaptivecards-designer, adaptivecards-templating and adaptivecards-react
4. install the .tgz package on the copied testapp, and rebuilt them.

Example of carousel failure(#6746) detected by this pipeline:
image
https://microsoft.visualstudio.com/AdaptiveCards/_build/results?buildId=42543322&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=35f53584-79fc-539f-463c-6471d0663a3a

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6833)


.pipelines/js-ci.yml
```diff
@@ -0,0 +1,104 @@
...
+  branches:
+    include:
+    - main
+  paths:
+    include:
+    - source/nodejs
+
+pool:
+  name: Azure Pipelines
+  vmImage: ubuntu-20.04
```
almedina-ms, line 13: >ubuntu-20.04 [](http://example.com/codeflow?start=11&length=12)

is there some way to keep the pipeline running in windows hosts?

licanhua: why do you ask?
I guess I only need to change the  vmImage to windows.
-bash should be supported in Windows too.

.pipelines/js-ci.yml
```diff
@@ -0,0 +1,104 @@
...
+    displayName: '[${{ target_package }}] npm pack'
+    inputs:
+      command: custom
+      customCommand: pack
+      workingDir: source/nodejs/${{ target_package }}
+  
+  - task: CopyFiles@2
+    inputs:
+      sourceFolder: source/nodejs/${{ target_package }}
+      contents: '*.tgz'
```
paulcam206, line 79: is this going to repeatedly copy files as each `target_package` is built?

should it maybe be:

```suggestion
      contents: '${{target_package}}.tgz' 
```

(here and below)

.pipelines/js-ci.yml
```diff
@@ -0,0 +1,104 @@
...
+  workingDirectory: source/adaptivecards-ui-testapp
+  displayName: 'Install the built package and build the adaptivecards-ui-testapp'
+
+- bash: |
+    npm i *.tgz
+    cat package.json
+    npm i
+    npm run build
+  workingDirectory: source/adaptivecards-react-testapp
+  displayName: 'Install the built package and build the adaptivecards-react-testapp'
```
paulcam206, line 104: should this be genericized like you did with `target_packages` above? might be easier when we add new apps to the list...

source/nodejs/adaptivecards-ui-testapp/package.json
```diff
@@ -36,6 +36,18 @@
...
+		"html-webpack-plugin": "^5.4.0",
+		"mini-css-extract-plugin": "^2.4.3",
+		"typescript": "^4.4.4",
+		"webpack": "^5.60.0",
+		"webpack-cli": "^4.9.1",
+		"webpack-concat-files-plugin": "^0.5.2",
+		"webpack-dev-server": "^4.3.1",
+		"ts-loader": "^9.2.6",
+		"css-loader": "^6.4.0",
+		"style-loader": "^3.3.1"
```
paulcam206, line 51: these dependencies should already be available in `source/nodejs`... rather than duplicating the dependencies here, you should just `npm i` in the parent directory before building.

removal of these dependencies was intentional to prevent us from managing version updates across a bunch of diverging packages...

paulcam206 (changes requested):

source/nodejs/adaptivecards-ui-testapp/package.json
licanhua: To make the test app consumes the built package, this project is copied to source/ and run outside of lerna. so I need to dup them

.pipelines/js-ci.yml
licanhua: It doesn't work because the package name is something like `ac-typed-schema-0.7.0-alpha.0.tgz`

.pipelines/js-ci.yml
licanhua: I believe so. will make the change

.pipelines/js-ci.yml
almedina-ms: tests use the edge driver, we may have to update the driver used if no edge is installed in ubuntu vm

.pipelines/js-ci.yml
paulcam206: `${{target_package}}*.tgz`?

source/nodejs/adaptivecards-ui-testapp/package.json
paulcam206: you shouldn't have to copy this outside of `source/nodejs`... you should be able to do something like:

1. Build and pack the target packages
1. Save packed `tgz` files outside of source (a temp directory should suffice)
2. `git clean -dxf`
3. `npm i` in `source/nodejs`
4. `npm i /path/to/saved/packs/*.tgz && npm i && npm run build` in `source/nodejs/adaptivecards-ui-testapp`

Managing each of our individual projects' npm dependencies is a pain, so I'd really like to avoid duplicating them to support this one CI task.

paulcam206 (changes requested):

source/nodejs/adaptivecards-ui-testapp/package.json
licanhua: It will not catch the bug like missing `swiper`. It only happens when command is run outside of lerna workspace

paulcam206: `lerna` only comes into play when you `npx lerna bootstrap`. when you `git clean -dxf`, it will delete any cross-linking that `lerna` does:

image

and then after doing an `npm i swiper@latest` in `source/nodejs/adaptivecards-ui-testapp`:

image

almedina-ms (approved):

.pipelines/js-ci.yml
```diff
@@ -0,0 +1,84 @@
...
+
+- bash: |
+   npm i -g npm@latest
+   npm ci 
+   npx lerna bootstrap --ci
+   npx lerna run release
+  workingDirectory: source/nodejs
+  displayName: 'Bash - lerna bootstrap'
+
+- bash: |
```
almedina-ms, line 50: the tests at source/nodejs/tests/ui-tests are not being run

licanhua: build-and-test all project in /tests/

paulcam206 (approved):

PR #6838, opened 2021-12-10
[UWP]: Don't use AdaptiveImage to render BackgroundImage

### LINKED ISSUE #2712: [UWP] Don't re-use custom AdaptiveImage renderer for background images

beervoley:
# Related Issue
Closes #2712.

# Description
We opt out of using `AdaptiveImageRenderer` in the case of rendering a background image. 
This will allow to override `AdaptiveImageRenderer` without overriding background image rendering.

# Sample Card
Old implementation on the left, new implementation on the right:

image

# How Verified
Manually.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6838)

### DIFF
--- a/source/uwp/Renderer/lib/XamlHelpers.cpp
+++ b/source/uwp/Renderer/lib/XamlHelpers.cpp
@@ -359,44 +359,53 @@ namespace AdaptiveCards::Rendering::Uwp::XamlHelpers
         return S_OK;
     }
 
+    void CreateBackgroundImage(_In_ IAdaptiveRenderContext* renderContext, _In_ HSTRING url, _Outptr_ IImage** backgroundImage)
+    {
+        ComPtr<IImage> image = XamlHelpers::CreateABIClass<IImage>(HStringReference(RuntimeClass_Windows_UI_Xaml_Controls_Image));
+
+        ComPtr<IDependencyObject> imageAsDependencyObject;
+        THROW_IF_FAILED(image.As(&imageAsDependencyObject));
+
+        ComPtr<IAdaptiveHostConfig> hostConfig;
+        THROW_IF_FAILED(renderContext->get_HostConfig(&hostConfig));
+
+        ComPtr<IUriRuntimeClass> imageUrl;
+        GetUrlFromString(hostConfig.Get(), url, imageUrl.GetAddressOf());
+
+        ComPtr<IBitmapImage> bitmapImage =
+            XamlHelpers::CreateABIClass<IBitmapImage>(HStringReference(RuntimeClass_Windows_UI_Xaml_Media_Imaging_BitmapImage));
+        THROW_IF_FAILED(bitmapImage->put_UriSource(imageUrl.Get()));
+
+        ComPtr<IImageSource> bitmapImageSource;
+        THROW_IF_FAILED(bitmapImage.As(&bitmapImageSource));
+
+        ComPtr<IUIElement> imageAsUIElement;
+        THROW_IF_FAILED(image.As(&imageAsUIElement));
+        THROW_IF_FAILED(image->put_Source(bitmapImageSource.Get()));
+        THROW_IF_FAILED(image.CopyTo(backgroundImage));
+    }
+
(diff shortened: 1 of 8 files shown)


source/uwp/Renderer/lib/XamlHelpers.cpp
```diff
@@ -359,44 +359,48 @@ namespace AdaptiveCards::Rendering::Uwp::XamlHelpers
...
 
-        ComPtr<IAdaptiveCardElement> adaptiveCardElement;
-        THROW_IF_FAILED(adaptiveImage.As(&adaptiveCardElement));
+        ComPtr<IAdaptiveHostConfig> hostConfig;
+        THROW_IF_FAILED(renderContext->get_HostConfig(&hostConfig));
 
-        ComPtr<IAdaptiveElementRendererRegistration> elementRenderers;
-        THROW_IF_FAILED(renderContext->get_ElementRenderers(&elementRenderers));
+        ComPtr<IUriRuntimeClass> imageUrl;
+        GetUrlFromString(hostConfig.Get(), url, imageUrl.GetAddressOf());
```
beervoley, line 373: @paulcam206 should I do this in **applyBackgroundToRoot** and pass an actual working url here?

source/uwp/Renderer/lib/XamlHelpers.cpp
```diff
@@ -359,44 +359,48 @@ namespace AdaptiveCards::Rendering::Uwp::XamlHelpers
...
         ComPtr<IImage> xamlImage;
-        THROW_IF_FAILED(background.As(&xamlImage));
+        CreateBackgroundImage(renderContext, url.Get(), &xamlImage);
+        if (!xamlImage)
+        {
+            return;
+        }
 
-        ABI::AdaptiveCards::ObjectModel::Uwp::BackgroundImageFillMode fillMode;
-        THROW_IF_FAILED(backgroundImage->get_FillMode(&fillMode));
```
paulcam206, line 399: this is just simple dead code, yes? have you verified that [`fillMode`](https://adaptivecards.io/explorer/BackgroundImage.html) still works?

paulcam206 (approved):

source/uwp/Renderer/lib/XamlHelpers.cpp
```diff
@@ -359,44 +359,48 @@ namespace AdaptiveCards::Rendering::Uwp::XamlHelpers
         return S_OK;
     }
 
-    void ApplyBackgroundToRoot(_In_ IPanel* rootPanel,
-                               _In_ IAdaptiveBackgroundImage* backgroundImage,
-                               _In_ IAdaptiveRenderContext* renderContext,
-                               _In_ IAdaptiveRenderArgs* renderArgs)
+    void CreateBackgroundImage(_In_ IAdaptiveRenderContext* renderContext, _In_ HSTRING url, _Outptr_ IImage** backgroundImage)
```
licanhua, line 362: nit: HSTRING const& url

licanhua (approved):

source/uwp/Renderer/lib/XamlHelpers.cpp
beervoley: @licanhua
image
there's no need for const& :)

source/uwp/Renderer/lib/XamlHelpers.cpp
beervoley: **fillMode** is being applied in **TileControl**, checking now..

PR #6849, opened 2021-12-14
Added App extension to podspec

jwoo-msft:
# Description

* SDK users currently can't choose not including the FluentUI dependency
* This change allows it.
* Based on Janus'PR created AdpativeCardsCore subspec that doesn't take dependency on FluentUI
* Added extension subspec that takes FluentUI
* Added CPP macro to enable extension via proprocessor macro.
* the pattern is well documented in [this thread](https://github.com/CocoaPods/CocoaPods/issues/5373#issuecomment-220486539) and [a blog post](https://miqu.me/blog/2016/11/28/app-extensions-xcode-and-cocoapods-omg/)
* Added the macro to sample project as it doesn't consume pod.
# How Verified

Tested it with Sample app with the extension.
Janus tested without extension.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6849)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards.xcodeproj/project.pbxproj
```diff
@@ -2279,6 +2279,10 @@
 					"DEBUG=1",
 					"$(inherited)",
 				);
+				"GCC_PREPROCESSOR_DEFINITIONS[arch=*]" = (
```
beervoley, line 2282: if your intention is to define it for all the flavors, what is the need for **[arch=*]**?
Can we simply:
`GCC_PREPROCESSOR_DEFINITIONS = (
                "DEBUG=1",
                "USEFLUENTUI=1",
                "$(inherited)",
            );`

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards.xcodeproj/project.pbxproj
```diff
@@ -2336,6 +2340,8 @@
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				GCC_C_LANGUAGE_STANDARD = gnu99;
 				GCC_NO_COMMON_BLOCKS = YES;
+				GCC_PREPROCESSOR_DEFINITIONS = "USEFLUENTUI=1";
+				"GCC_PREPROCESSOR_DEFINITIONS[arch=*]" = "";
```
beervoley, line 2344: won't these 2 lines simply counteract each other?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards.xcodeproj/project.pbxproj
jwoo-msft: good catch

source/ios/tools/AdaptiveCards.podspec
```diff
@@ -1,20 +1,25 @@
-  
-Pod::Spec.new do |spec|
+od::Spec.new do |spec|
   spec.name             = 'AdaptiveCards'
 
-  spec.version          = '2.8.2'
-  
+  spec.version          = '2.8.3'
```
beervoley, line 4: I see that you've already tagged 2.8.3 - https://github.com/microsoft/AdaptiveCards/releases/tag/2.8.3
but to a different commit. Was that intentional?

source/ios/tools/AdaptiveCards.podspec
```diff
@@ -1,20 +1,25 @@
-  
-Pod::Spec.new do |spec|
+od::Spec.new do |spec|
```
beervoley, line 1: **P**od here, missing **P**

source/ios/tools/AdaptiveCards.podspec
```diff
@@ -27,21 +32,23 @@ Pod::Spec.new do |spec|
   end
 
   spec.subspec 'AdaptiveCardsPrivate' do | sspec |
-    sspec.source_files = 'source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/PrivateHeaders/*.h'
+    sspec.source_files = 'source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/PrivateHeaders/**/*.{h,m,mm}'
     sspec.header_mappings_dir = 'source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/PrivateHeaders/'
     sspec.private_header_files = 'source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/PrivateHeaders/*.h'
   end
 
+  spec.subspec 'UIProviders' do | sspec |
```
licanhua, line 40: should the name be FlucentUIAppExtension other than UIProviders?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRBaseTarget.mm
```diff
@@ -6,8 +6,10 @@
 //
 
 #import "ACRBaseTarget.h"
-#import <FluentUI/FluentUI-Swift.h>
 #import <UIKit/UIKit.h>
+#if defined(ADAPTIVECARDS_APP_EXTENSION)
```
licanhua, line 10: As a user which consume the AdaptiveCards, what's the impact to them? Is there any document for that: For example, if you want the tooltip, please do this setting on your project.

Should it be a major change from user's aspect?

janusbal: @licanhua   There should be no impact for the consumers. This flag will always exist unless the consumer explicitly exclude the new subspec (UIProviders)

jwoo-msft: by default, users will get tooltip

if they want not to include tooltip, they can choose individual subspecs.


pod 'AdaptiveCards' ~> 'x.x.x' 
get everything.

pick their own, they can do:
pod 'AdaptiveCards/FolderWeNeed1' ~> 'x.x.x'

pod 'AdaptiveCards/FolderWeNeed2' ~> 'x.x.x'

pod 'AdaptiveCards/FolderWeNeed3' ~> 'x.x.x'

source/ios/tools/AdaptiveCards.podspec
jwoo-msft: thanks

source/ios/tools/AdaptiveCards.podspec
jwoo-msft: we will have more third-party libraries.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards.xcodeproj/project.pbxproj
jwoo-msft: it's xcode gui doing.

source/ios/tools/AdaptiveCards.podspec
jwoo-msft: yes, the hash is for testing.

source/ios/tools/AdaptiveCards.podspec
jwoo-msft: it's weird. i noticed it happened again with another commit; it could be due to tmux i'm usig.

licanhua (approved):

beervoley (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRBaseTarget.mm
dclaux, line 10: `ADAPTIVECARDS_APP_EXTENSION` sounds weird as a flag name. Why not `ADAPTIVECARDS_USE_FLUENT_TOOLTIPS` since that's what it enables?

source/ios/tools/AdaptiveCards.podspec
```diff
@@ -1,20 +1,25 @@
...
+
   spec.summary          = 'Adaptive Cards are a new way for developers to exchange card content in a common and consistent way'
   
-  spec.source       = { :git => 'https://github.com/microsoft/AdaptiveCards.git', :tag => 'ios-v2.8.2' }
+  spec.source       = { :git => 'https://github.com/microsoft/AdaptiveCards.git', :tag => 'ios-v2.8.3' }
 
-  spec.source_files           = 'source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/*.{h,m,mm}'
+  spec.default_subspecs = 'AdaptiveCardsCore', 'AdaptiveCardsPrivate', 'ObjectModel', 'UIProviders'
+
+  spec.subspec 'AdaptiveCardsCore' do | sspec |
```
almedina-ms, line 18: >sspec [](http://example.com/codeflow?start=40&length=5)

what does sspec stand for, is it short for subspec?

almedina-ms (approved):
:shipit:

source/ios/tools/AdaptiveCards.podspec
jwoo-msft: correct

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRBaseTarget.mm
jwoo-msft: I had a similar name initially. After seeing what others do, I changed to this name. the suffix APP_EXTENSION, used throughout frameworks for this pattern. I didn't include Fluent nor Tooltips in the name because the flag is meant for the subspec, `UIProvider`.

dclaux: @jwoo-msft that doesn't make much sense to me. What about the next feature/capability that a host should be able to disable independently of the FluentUI tooltip? What name will you give that next flag?

beervoley: > @jwoo-msft that doesn't make much sense to me. What about the next feature/capability that a host should be able to disable independently of the FluentUI tooltip? What name will you give that next flag?

that's true...

beervoley (changes requested):
`ADAPTIVECARDS_APP_EXTENSION` may not be the best name because in this case it only affects  `FLUENT_TOOLTIPS` and if we add more features that host should be able to disable `independently`, it's going to be a mess :)

jwoo-msft:
> Member

jwoo-msft:
> Member

No, it's not. We are going to add an extension library on top of this and will remove the flag when we add more controls.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRBaseTarget.mm
jwoo-msft: @dclaux, sure, I will change the name. the use case you mentioned is a good one, but it's not a realistic one since we won't be using the preprocessor macro for this purpose anymore once we come up with a general solution for third-party controls.

beervoley (approved):
I will approve, but the naming still doesn't make much sense to me :)

beervoley (approved):

jwoo-msft:
> I will approve, but the naming still doesn't make much sense to me :)

I changed the name.

beervoley:
> > I will approve, but the naming still doesn't make much sense to me :)
> 
> I changed the name.

thanks Joseph :)

PR #6878, opened 2022-01-10
[iOS] Fixed Observer Issue

jwoo-msft:
# Related Issue

We have an occasional test failure with IconsInSomeActions.json card. I'm pretty certain that this PR fixes the issue. If not, this PR will fix one of the possible failure paths. 

The unit tests do not utilize resource resolvers. There is a small chance where image is loaded very fast, and activate the observer removal path. Because actions have been increasing the subscribers counts without checking if resolver interface is used, it will try to remove the observer, and will cause this exception. 
This won't impact sdk users since the users use resource resolvers unlike in unit test scenarios.

Removed dead code that's not used.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6878)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.mm
```diff
@@ -509,37 +509,29 @@ - (void)loadImage:(std::string const &)urlStr
...
-                std::vector<char> decodedDataUri = AdaptiveCards::AdaptiveBase64Util::Decode(dataUri);
-                NSData *decodedBase64 = [NSData dataWithBytes:decodedDataUri.data() length:decodedDataUri.size()];
-                img = [UIImage imageWithData:decodedBase64];
-            } else {
-                img = [UIImage imageWithData:[NSData dataWithContentsOfURL:url]];
-            }
-            return img;
-        };
-    }
+    ImageLoadBlock imageloadblock = ^(NSURL *url) {
```
licanhua, line 512: You skipped the resolver. will it impact the user if they https://docs.microsoft.com/en-us/adaptive-cards/sdk/rendering-cards/ios/extensibility#adcresolver

jwoo-msft: this is a dead code. I removed it. user of resolver won't call this method.

paulcam206 (approved):

PR #6815, opened 2021-12-07
[JS] Modernize and automate code quality/formatting

paulcam206:
# Related Issue

Fixes #6657
Fixes #6757 

# Description

The commits on this PR are structured such that each is self-contained:

* e6d74951455deaafd2919bcf0603e9ed24203253
  * Updated configuration for `eslint`
  * Install/configure `prettier`
  * `npm` scripts updated for `lint`
* d683d1cae0ce46563fe4ee2f83b9c031e0c983a5
  * Install/configure `lint-staged`, a tool that can run sets of tools against changes staged in `git`
* 1a54be7c0f25380d2cd16a10e0724b8f02096539
  * Install/configure `husky`, a client-side `git` hook manager
  * Wire `husky` up to our root `prepare` lifecycle script
* 5b55c77a156b88242ff0f8630e6410725fea8772
  * Run `eslint --fix` on the `adaptivecards` package - no manual edits
* 46bcce1bea81bf77b628b0a097d38f5ece90b721
  * Manual fixes for `eslint` issues. **This is probably the most important thing to review!**
* 0455e3b88595f4d94b57c53f3c31cdb7a2b47cdc
  * Run `prettier` on stuff in `adaptivecards` - no manual edits
* subsequent commits address PR feedback

With these changes, after this branch is merged, doing an `npm install` in `source/nodejs` will automatically install `husky`-managed `git` hooks. After a file under `source/nodejs` is modified and staged, attempting to commit it will cause `eslint` and `prettier` to run. Automatically-fixable issues detected by `eslint` will be fixed and any formatting changes that need to be made will also be done automatically. Any errors from `eslint` that *aren't* auto-fixable will yield an error and refuse the commit. This hook can be bypassed using the `--no-verify` `git` commandline argument.

I should note that I didn't apply any autofixing or formatting outside of the `adaptivecards` package for now, as each individual package will need its own dedicated time.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/6815)

PR #7051, opened 2022-02-17
Add stub device emulation dropdown.

thure:
# Related Issue

A Pull Request should close a **single** issue; multiple issues can be closed when the issues are **small and related**, but that should be an exception not the rule. Please keep Pull Requests small and targeted; large 'code drops' with dozens of files will be closed and asked to split into reviewable pieces. Reviews that need to be large due to dependencies will be
reviewed on a case-by-case basis.

Please use one of the well-known [github fixes keywords](https://help.github.com/en/articles/closing-issues-using-keywords) to reference
the issue fixed with this PR (eg Fixes #<github issue number>). If an issue doesn't yet exist please create one to aid
in issue tracking.

page above.

# Description

For all Pull Requests, please describe how the issue was fixed or how the feature was implemented from a summary level. This information will be used to help provide context to the reviewers of the pull request and should be additive to the issue being closed.

# Sample Card

# How Verified

2. Existing relevant unit/regression tests that you ran
3. Manual scenario verification if any; ***Do include .gif's or screenshots of the testing you performed here if you think that it

### DIFF
--- a/source/nodejs/adaptivecards-designer-app/src/app.ts
+++ b/source/nodejs/adaptivecards-designer-app/src/app.ts
@@ -61,7 +61,7 @@ window.onload = function() {
     }
     */
 
-	let designer = new ACDesigner.CardDesigner(ACDesigner.defaultMicrosoftHosts);
+	let designer = new ACDesigner.CardDesigner(ACDesigner.defaultMicrosoftHosts, ACDesigner.defaultMicrosoftDeviceEmulations);
 	designer.sampleCatalogueUrl = window.location.origin + "/sample-catalogue.json";
 	designer.attachTo(document.getElementById("designerRootHost"));
 
--- a/source/nodejs/adaptivecards-designer/src/adaptivecards-designer.ts
+++ b/source/nodejs/adaptivecards-designer/src/adaptivecards-designer.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import * as containers from "./containers";
+import {DeviceEmulation} from "./device-emulation";
 
 export * from "./adaptivecards-designer-standalone";
 export * from "./containers/index";
@@ -16,4 +17,9 @@ export const defaultMicrosoftHosts: containers.HostContainer[] = [
 	new containers.BotFrameworkContainer("Bot Framework Other Channels (Image render)", "containers/bf-image-container.css"),
 	new containers.CortanaClassicContainer("Cortana Skills (Classic)", "containers/cortana-classic-container.css"),
 	...containers.BerlinContainer.allContainers(),
+]
+
+export const defaultMicrosoftDeviceEmulations: DeviceEmulation[] = [
+	{ name: 'Unconstrained', styleSheet: 'unconstrained.css' },
+	{ name: 'Mobile, typical', styleSheet: 'mobile.css'},
 ]
--- a/source/nodejs/adaptivecards-designer/src/card-designer.ts
+++ b/source/nodejs/adaptivecards-designer/src/card-designer.ts
@@ -26,6 +26,7 @@ import * as Shared from "./shared";
 import { TreeView } from "./tree-view";
 import { SampleCatalogue } from "./catalogue";
 import { HelpDialog } from "./help-dialog";
+import {DeviceEmulation} from "./device-emulation";
 
 export class CardDesigner extends Designer.DesignContext {
     private static internalProcessMarkdown(text: string, result: Adaptive.IMarkdownProcessingResult) {
@@ -48,6 +49,7 @@ export class CardDesigner extends Designer.DesignContext {
     private _cardEditor: monaco.editor.IStandaloneCodeEditor;
     private _sampleDataEditor: monaco.editor.IStandaloneCodeEditor;
     private _hostContainers: Array<HostContainer>;
+    private _deviceEmulations: Array<DeviceEmulation>;
     private _isMonacoEditorLoaded: boolean = false;
     private _designerSurface: Designer.CardDesignerSurface;
     private _designerHostElement: HTMLElement;
@@ -55,6 +57,7 @@ export class CardDesigner extends Designer.DesignContext {
     private _draggedElement: HTMLElement;
     private _currentMousePosition: IPoint;
     private _hostContainer: HostContainer;
+    private _deviceEmulation: DeviceEmulation;
     private _undoStack: Array<object> = [];
     private _undoStackIndex: number = -1;
     private _startDragPayload: object;
@@ -610,6 +613,7 @@ export class CardDesigner extends Designer.DesignContext {
     private _fullScreenHandler = new FullScreenHandler();
     private _fullScreenButton: ToolbarButton;
     private _hostContainerChoicePicker: ToolbarChoicePicker;
+    private _deviceEmulationChoicePicker: ToolbarChoicePicker;
     private _versionChoicePicker: ToolbarChoicePicker;
     private _undoButton: ToolbarButton;
     private _redoButton: ToolbarButton;
@@ -751,6 +755,27 @@ export class CardDesigner extends Designer.DesignContext {
             };
 
             this.toolbar.addElement(this._hostContainerChoicePicker);
+        }
+
+        if (this._deviceEmulations && this._deviceEmulations.length > 0) {
+            this._deviceEmulationChoicePicker = new ToolbarChoicePicker(CardDesigner.ToolbarCommands.DeviceEmulationPicker);
+            this._deviceEmulationChoicePicker.separator = true;
+            this._deviceEmulationChoicePicker.label = "Emulate device:"
+
+            for (let i = 0; i < this._deviceEmulations.length; i++) {
+                this._deviceEmulationChoicePicker.choices.push(
+                    {
+                        name: this._deviceEmulations[i].name,
+                        value: i.toString(),
+                    }
+                );
+            }
+
+            this._deviceEmulationChoicePicker.onChanged = (sender) => {
+                this.deviceEmulation = this._deviceEmulations[parseInt(this._deviceEmulationChoicePicker.value)];
+            };
+
+            this.toolbar.addElement(this._deviceEmulationChoicePicker);
         }
 
         this._undoButton = new ToolbarButton(
@@ -995,7 +1020,7 @@ export class CardDesigner extends Designer.DesignContext {
 
     lockDataStructure: boolean = false;
 
-    constructor(hostContainers: Array<HostContainer> = null) {
+    constructor(hostContainers: Array<HostContainer> = null, deviceEmulations: Array<DeviceEmulation> = null) {
         super();
 
         Adaptive.GlobalSettings.enableFullJsonRoundTrip = true;
@@ -1007,6 +1032,7 @@ export class CardDesigner extends Designer.DesignContext {
         }
 
         this._hostContainers = hostContainers ? hostContainers : [];
+        this._deviceEmulations = deviceEmulations ? deviceEmulations : [];
 
         this.prepareToolbar();
     }
@@ -1398,6 +1424,16 @@ export class CardDesigner extends Designer.DesignContext {
         }
     }
 
+    get deviceEmulation(): DeviceEmulation {
+        return this._deviceEmulation;
+    }
+
+    set deviceEmulation(value: DeviceEmulation) {
+        if (this._deviceEmulation !== value) {
+            this._deviceEmulation = value;
+        }
+    }
+
     get canUndo(): boolean {
         return this._undoStackIndex >= 1;
     }
@@ -1460,6 +1496,7 @@ export class CardDesigner extends Designer.DesignContext {
 export module CardDesigner {
     export class ToolbarCommands {
         static readonly HostAppPicker = "__hostAppPicker";
+        static readonly DeviceEmulationPicker = "__deviceEmulationPicker";
         static readonly VersionPicker = "__versionPicker";
         static readonly Undo = "__undoButton";
         static readonly Redo = "__redoButton";
--- a/source/nodejs/adaptivecards-designer/src/device-emulation.ts
+++ b/source/nodejs/adaptivecards-designer/src/device-emulation.ts
@@ -1,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export abstract class DeviceEmulation {
+  readonly name: string;
+  readonly styleSheet: string;
+
+  constructor(name: string, styleSheet: string) {
+    this.name = name;
+    this.styleSheet = styleSheet;
+  }
+
+}


source/nodejs/adaptivecards-designer/src/adaptivecards-designer.ts
```diff
@@ -17,3 +18,8 @@ export const defaultMicrosoftHosts: containers.HostContainer[] = [
 	new containers.CortanaClassicContainer("Cortana Skills (Classic)", "containers/cortana-classic-container.css"),
 	...containers.BerlinContainer.allContainers(),
 ]
+
+export const defaultMicrosoftDeviceEmulations: DeviceEmulation[] = [
+	{ name: 'Unconstrained', styleSheet: 'unconstrained.css' },
```
dclaux: I'm not sure I understand the reason for a device-emulation-specific CSS when the device-emulation itself is disconnected from the HostContainer.

The HostContainer implementation is the entity responsible for emulating a device's look and feel, and it has its own CSS. What's needed on top of that is emulating the screen width of a device, but that shouldn't change the device's look and feel.

When would this `styleSheet` be applied, and what does it control?

source/nodejs/adaptivecards-designer/src/adaptivecards-designer.ts
```diff
@@ -20,6 +20,7 @@ export const defaultMicrosoftHosts: containers.HostContainer[] = [
 ]
 
 export const defaultMicrosoftDeviceEmulations: DeviceEmulation[] = [
-	{ name: 'Unconstrained', styleSheet: 'unconstrained.css' },
-	{ name: 'Mobile, typical', styleSheet: 'mobile.css'},
+	{ name: 'Unconstrained', maxWidth: null },
```
dclaux, line 23: Ah, this answers my previous comment, which is now moot :)

source/nodejs/adaptivecards-designer/src/adaptivecards-designer.ts
dclaux: Moot per subsequent commits.

source/nodejs/adaptivecards-designer/src/device-emulation.ts
```diff
@@ -3,11 +3,11 @@
 
 export abstract class DeviceEmulation {
   readonly name: string;
-  readonly styleSheet: string;
+  readonly maxWidth: string | null;
```
dclaux, line 6: I know this is contentious and people tend to disagree, but... I suggest using `undefined` instead of `null` pretty much everywhere (there are still parts of the designer codebase that use null that I've wanted to change for a while, but the renderer SDK has entirely moved to `undefined`). The main reasons are:
- Uninitialized variables are `undefined`, not `null`. Using `null` thus leads to potentially having to do multiple checks to determine what to do with a variable
- `typeof null == object` whereas `typeof undefined == undefined`. So if you have say `let s: string | null = null`, even though `s` is explicitly typed as a `string` `typeof s` will return `object`. Not cool in my book.

dclaux, line 6: Any reason why `maxWidth` is a `string` and not a `number`? Only a width expressed in pixels is meaningful in the context of this feature.

dclaux (changes requested):

source/nodejs/adaptivecards-designer/src/device-emulation.ts
thure: I have my own thoughts on the matter, but I'll gladly defer to however the maintainers of this project want to handle non-values, so I'll refactor this to use `undefined`.

thure: I didn't want to proscribe the use of CSS units other than `px`; I figure developers can choose what they like, in case there is some more useful value to them than a CSS pixel value.

Is there a reason this should be limited to just `px` values?

dclaux: It's just that in real life a screen width is expressed in pixels, and the idea with this feature is that the tool matches real life screens.

dclaux: Thanks - and if you want to share your thoughts with me offline, I'd love to know.

thure: I would agree if it weren't for [CSS pixels having not much bearing on hardware pixels](https://juiceboxinteractive.com/blog/a-pixel-is-not-a-pixel-designing-for-a-new-generation-of-mobile-devices/#:~:text=A%20hardware%20pixel%20is%20an%20individual%20dot%20of,a%20concept%20known%20as%20the%20device%20pixel%20ratio.).

As an example, I have an iPhone 13 Pro, and on Apple's [tech specs page](https://www.apple.com/iphone-13-pro/specs/) they say the screen is "2532‑by‑1170-pixel resolution at 460 ppi".

If I visit whatismyscreenresolution.net, however, it reports my iPhone's CSS pixel dimensions as 390x844.

I propose developers should put the whole CSS string as this value as a reminder that the width of the card is being determined by CSS and not some other graphics package, and that its `px` comes with its caveats and also that other units are available (developers could, if they like, use `in`, which happens to always be 96 times larger than `px` and thus potentially less ambiguous).

thure: All that said, since you have a better handle on the norms in this project, I'll gladly defer to your choice, just wanted to contextualize my choices here.

dclaux: The width of the card isn't being defined by CSS, especially not on Teams mobile which this feature is meant to help with. On mobile, cards are rendered natively (as is the whole app) and CSS isn't even involved.

Resolutions as reported by design systems and programming APIs are usually brought back to 96dpi so that developers don't have to deal with dpi conversions unless they want/need to.

To me, the questions are:
- What are the values that make sense to the end user in that drop down
- How do we translate these values into an actual width in the designer

To the first question, I'd say the values that make most sense are likely "iPhone 6", "iPhone 13", "iPad mini", "iPad Pro 10'", "Galaxy S10", etc more than actual pixel dimensions.

To the second question, "iPhone 13" as you mention above translates into 390 pixels (at 96 dpi), never to other units. That's why I think only pixels make sense: no other unit will actually be used to specify those dimensions.

All that said, it doesn't fundamentally matter that much, and if you feel strongly (as you apparently do) about width expressed as a string, I suppose we can live with it.

dclaux (approved):

thure:
I should point out, as a general note, that the code quality tooling introduced in #6815 is no longer running.

When setting up for this project, the `prepare` script did run, and Husky reported successful installation, however the pre-commit hook did not run. It looks as if it hasn't run in some time, since running Prettier manually overwrites much of the code, and I needed to temporarily disable ESLint in my IDE since extant linting issues were crowding out other issues.

I think maybe #6657 and #6757 should be reopened.

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -1398,6 +1424,26 @@ export class CardDesigner extends Designer.DesignContext {
...
+                this._designerHostElement.style.setProperty('margin-left', 'auto');
+                this._designerHostElement.style.setProperty('margin-right', 'auto');
+            }else{
+                this._designerHostElement.style.removeProperty('max-width');
+                this._designerHostElement.style.removeProperty('margin-left');
+                this._designerHostElement.style.removeProperty('margin-right');
+            }
+            this._designerSurface.updateLayout(true);
+        }
+    }
```
paulcam206, line 1445:
```suggestion
    set deviceEmulation(value: DeviceEmulation) {
        if (this._deviceEmulation !== value) {
            this._deviceEmulation = value;
            if (value.maxWidth) {
                this._designerHostElement.style.setProperty('max-width', value.maxWidth);
                this._designerHostElement.style.setProperty('margin-left', 'auto');
                this._designerHostElement.style.setProperty('margin-right', 'auto');
            } else {
                this._designerHostElement.style.removeProperty('max-width');
                this._designerHostElement.style.removeProperty('margin-left');
                this._designerHostElement.style.removeProperty('margin-right');
            }
            this._designerSurface.updateLayout(true);
        }
    }
```

source/nodejs/adaptivecards-designer/src/adaptivecards-designer.css
```diff
@@ -36,7 +36,8 @@
 }
 
 .acd-designer-host {
-    margin: 20px 40px 20px 20px;
+    margin: 20px 0;
+    padding: 0 40px 0 20px;
```
paulcam206, line 40: How does this look when the entire Designer has a narrower width? 

image

paulcam206 (changes requested):

source/nodejs/adaptivecards-designer/src/adaptivecards-designer.css
thure: This is how it looks in Edge, from a comfortable width down to when the viewport begins scaling (this behavior appears to be automatic since there is no viewport `<meta>`):

pr

thure: You can emulate the change this PR introduces by duplicating the host app selector element as seen in production. Elements in the toolbar continue to wrap as before:

pr

The toolbar pattern used in the designer conforms to WCAG SC 1.4.10 even with this PR, but the design lacks any appeal. IMO the designer should be redesigned responsively (i.e. mobile-first), but that would seem well out-of-scope for this PR.

paulcam206: agree that the designer could definitely be more mobile-friendly... I retrofitted this reflow behavior to address accessibility concerns, but didn't have enough time to do a proper redesign. thanks for the gif! :)

paulcam206 (approved):

source/nodejs/adaptivecards-designer-app/src/app.ts
```diff
@@ -61,9 +61,13 @@ window.onload = function() {
     }
     */
 
-	let designer = new ACDesigner.CardDesigner(ACDesigner.defaultMicrosoftHosts);
-	designer.sampleCatalogueUrl = window.location.origin + "/sample-catalogue.json";
-	designer.attachTo(document.getElementById("designerRootHost"));
+  let designer = new ACDesigner.CardDesigner(
```
licanhua, line 64: I can't see the change on https://adaptivecardsci.z5.web.core.windows.net/pr/7051/designer

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -1398,6 +1424,26 @@ export class CardDesigner extends Designer.DesignContext {
         }
     }
 
+    get deviceEmulation(): DeviceEmulation {
+        return this._deviceEmulation;
+    }
+
+    set deviceEmulation(value: DeviceEmulation) {
+        if (this._deviceEmulation !== value) {
```
licanhua, line 1432: We should only show the picker when it's in preview mode.
but I didn't see you have code to check this condition

licanhua (changes requested):

source/nodejs/adaptivecards-designer/src/card-designer.ts
dclaux: @licanhua I don't agree with that. The picker is useful as the user designs the card, so they can make quick adjustments after switching from one form factor to another. That is a desirable capability for Teams developers who this feature is designed for.

@thure I assume the selected device setting also applies to the preview when in preview mode?

source/nodejs/adaptivecards-designer-app/src/app.ts
thure: Strange, I can't either. I wasn't aware there was CI for this in this project. I can attempt to look at the action that ran to see if there was an issue.

source/nodejs/adaptivecards-designer/src/card-designer.ts
thure: I might be missing some context, is the preview part of the designer ever _not_ in preview mode?

Certainly the idea is to offer this control exactly as often as it is relevant, and based on what I know about the experience here, that is all the time.

licanhua: Do you know how often the user will use templating? If templating is used, this will provide false boundary conditions which doesn't mean too much for developers.

thure: > If templating is used, this will provide false boundary conditions

Can you explain this point a bit more? What boundary conditions are you referring to, and how would this provide false ones?

dclaux: @thure you toggle between design and preview modes by clicking the "Preview" button.

source/nodejs/adaptivecards-designer-app/src/app.ts
thure: It looks like [an action to build and deploy this PR succeeded](https://microsoft.visualstudio.com/AdaptiveCards/_build/results?buildId=45740634&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=08d41a1b-6013-5b01-b962-661a2fb402c0&l=1694), but since I'm unfamiliar with this project's infrastructure details I can't know for sure why you wouldn't see this PR's changes.

source/nodejs/adaptivecards-designer/src/card-designer.ts
dclaux: @licanhua it makes little sense to:
- Let a developer author a card using one specific width
- Then switch to preview mode
- Then switch to a different device emulation just to see that something's off
- Exit preview just to be returned to the previous width they were using to design the card
  - The developer needs to be able to tweak the card while seeing it using the emulation in which it looks off

Besides, Preview mode is basically only needed IF templating is used. If templating is not used and the device emulation selector is only available in preview mode, then the developer may never even see it.

The device emulation selector really has to be available in design mode.

licanhua: For example, if weather card is selected, the "$" part is not the actual rendering until preview mode is clicked

image

dclaux: As I say above, the selector being available in design mode is crucial when not using templating.

licanhua: I leave this to @kat-y and @JeanRoca

dclaux:
@thure I believe you will also need to update the website code in order to add the default list of device emulations to the designer. That is likely why the selector doesn't appear in the [test site](https://adaptivecardsci.z5.web.core.windows.net/pr/7051).

source/nodejs/adaptivecards-designer-app/src/app.ts
licanhua: you may need to update adaptivecards\source\nodejs\adaptivecards-site\themes\adaptivecards\layout\designer.ejs too.
Your change is made to adaptivecards-designer-app, but the site doesn't use it

thure: I think I found the code that initializes the designer for the page in question and pushed the appropriate changes just now.

thure: The device emulation picker now appears at https://adaptivecardsci.z5.web.core.windows.net/pr/7051/designer

source/nodejs/adaptivecards-designer/src/card-designer.ts
thure: Since you are deferring this, I will re-request your review. Please feel free to update your review as needed.

dclaux:
@thure also, and I'm not sure it's related to your changes (but I thought I'd mention just in case), the Outlook and Viva apps don't work as they should anymore. For example, Outlook looks like this now:
image

When it should look like this:
image

licanhua (approved):

dclaux:
@licanhua I'm seeing issues on the test site. I don't think we should merge the code in that state?

source/nodejs/adaptivecards-designer/src/card-designer.ts
licanhua: From code aspect, it looks good. You can merge it if Kat or J.P. think templating is good

thure: Understood, thanks! ✨

thure:
@dclaux Got it, investigating now. Will implement the solutions we discussed offline.

source/nodejs/adaptivecards-designer/src/card-designer.ts
JeanRoca: > Do you know how often the user will use templating? If templating is used, this will provide false boundary conditions which doesn't mean too much for developers.

Hey @licanhua could you explain what you mean by false boundary conditions? I am not sure I understand how the views would impact a developer using templating. 

@thure is this feature available for all host apps or only bot framework / teams?

thure:
> @thure I'm finding that the device emulation doesn't work well for host applications that impose their own width, namely WebChat, Teams, Federated search and Berlin.

@dclaux I've implemented the changes we discussed offline. Device emulation is now only enabled for hosts that don't currently specify a specific width for the card host.

source/nodejs/adaptivecards-designer/src/adaptivecards-designer.ts
```diff
@@ -1,20 +1,29 @@
...
-	new containers.DarkCortanaContainer("Cortana Skills - Dark", "containers/cortana-container-dark.css"),
-	new containers.LightCortanaContainer("Cortana Skills - Light", "containers/cortana-container-light.css"),
-	new containers.BotFrameworkContainer("Bot Framework Other Channels (Image render)", "containers/bf-image-container.css"),
+	new containers.DarkCortanaContainer("Cortana Skills - Dark", "containers/cortana-container-dark.css", true),
+	new containers.LightCortanaContainer("Cortana Skills - Light", "containers/cortana-container-light.css", true),
+	new containers.BotFrameworkContainer("Bot Framework Other Channels (Image render)", "containers/bf-image-container.css", true),
 	new containers.CortanaClassicContainer("Cortana Skills (Classic)", "containers/cortana-classic-container.css"),
-	new containers.FederatedSearchContainer("Federated Search","containers/federated-search-container.css"),
+	new containers.FederatedSearchContainer("Federated Search","containers/federated-search-container.css", true),
 	...containers.BerlinContainer.allContainers(),
```
thure, line 20: @JeanRoca You can see here which hosts are opted into device emulation (`true` as the third argument opts into device emulation). I assigned them based on whether their CSS rules specify a width for the card container.

Please let me know if any of this should change.

dclaux:
@thure  thanks - I do have a nit though. I think it should be the responsibility of the `HostContainer` implementation to say whether or not it supports device emulation, as opposed to being a parameter passed to the constructor. The reason for that is you could very well pass `true` to the Viva container constructor, the device emulation selector would be enabled but it wouldn't work properly.

I think you should introduce this:
```typescript
class HostContainer {
    ...
    get allowDeviceEmulation(): boolean {
        return false;
    }
    ....
}
```

And override in the HostContainer implementations that do actually support device emulation, e.g:

```typescript
class OutlookContainer extends HostContainer {
    ...
    get allowDeviceEmulation(): boolean {
        return true;
    }
    ....
}
```

thure:
@dclaux I've made the changes you suggested.

dclaux:
thure thank you very much :)

source/nodejs/adaptivecards-designer/src/card-designer.ts
thure: @JeanRoca I've just adjusted the implementation to only apply to hosts that don't specify a card host width in their CSS.

licanhua (approved):

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -1101,7 +1140,7 @@ export class CardDesigner extends Designer.DesignContext {
         let styleSheetLinkElement = document.createElement("link");
         styleSheetLinkElement.id = "__ac-designer";
         styleSheetLinkElement.rel = "stylesheet";
-		styleSheetLinkElement.type = "text/css";
+		    styleSheetLinkElement.type = "text/css";
```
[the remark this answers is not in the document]
thure: Took me a while to clock this one since it looks right in my IDE, looks like somehow this line had tabs. Now fixed.

PR #7065, opened 2022-02-24
removed padding added for background image

### LINKED ISSUE #7052: [Rendering] Padding Added When Background Image is present
### Target Platforms

NodeJS

### SDK Version

latest

### Application Name

N/A

### Problem Description

When background image is added to a collection type view, the view gets padding.

### Screenshots

_No response_

### Card JSON

```JSON
N/A
```

### Sample Code Language

_No response_

### Sample Code

_No response_

jwoo-msft:
# Related Issue
removed padding added to background image.
added global setting to turn on and off of it.
enabled this behavior in berlin container.
Fixed #7052

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -5906,13 +5906,17 @@ export abstract class StylableCardElementContainer extends CardElementContainer 
     }
 
     protected getHasBackground(): boolean {
+        return this.getHasPadding(true);
+    }
+
+    protected getHasPadding(shouldCountBackgroundImage : boolean): boolean {
         let currentElement: CardElement | undefined = this.parent;
 
         while (currentElement) {
             const currentElementHasBackgroundImage =
                 currentElement instanceof Container
-                    ? currentElement.backgroundImage.isValid()
-                    : false;
+                    ? (currentElement.backgroundImage.isValid() && shouldCountBackgroundImage) 
+		    : false;
 
             if (currentElement instanceof StylableCardElementContainer) {
                 if (
@@ -5931,7 +5935,11 @@ export abstract class StylableCardElementContainer extends CardElementContainer 
     }
 
     protected getDefaultPadding(): PaddingDefinition {
-        return this.getHasBackground() || this.getHasBorder()
+        let hasPadding = GlobalSettings.addPaddingToContainerWithBackgroundImage ? 
+            this.getHasBackground() : this.getHasPadding(false);
+
+        return hasPadding  || 
+               this.getHasBorder()
             ? new PaddingDefinition(
                   Enums.Spacing.Padding,
                   Enums.Spacing.Padding,
--- a/source/nodejs/adaptivecards/src/shared.ts
+++ b/source/nodejs/adaptivecards/src/shared.ts
@@ -34,6 +34,7 @@ export class GlobalSettings {
     static enableFallback: boolean = true;
     static useWebkitLineClamp: boolean = true;
     static allowMoreThanMaxActionsInOverflowMenu: boolean = false;
+    static addPaddingToContainerWithBackgroundImage: boolean = true;
 
     static readonly applets: AppletsSettings = {
         logEnabled: true,
--- a/source/nodejs/adaptivecards-designer/src/containers/berlin/berlin-container.ts
+++ b/source/nodejs/adaptivecards-designer/src/containers/berlin/berlin-container.ts
@@ -27,6 +27,11 @@ export class BerlinContainer extends HostContainer {
 		this._colorTheme = theme;
 	}
 
+	public initialize(): void {
+		super.initialize();
+		Adaptive.GlobalSettings.addPaddingToContainerWithBackgroundImage = false; 
+	}
+
 	public renderTo(hostElement: HTMLElement) {
 		this.cardHost.classList.add(`berlin-${this._containerSize.toLowerCase()}-card`);
 		const outerFrame = document.createElement("div");


source/nodejs/adaptivecards/src/shared.ts
```diff
@@ -34,6 +34,7 @@ export class GlobalSettings {
     static enableFallback: boolean = true;
     static useWebkitLineClamp: boolean = true;
     static allowMoreThanMaxActionsInOverflowMenu: boolean = false;
+    static addPaddingToContainerWithBackgroundImage: boolean = true;
```
licanhua, line 37: @paulcam206 who is the owner to approve this kind of API review?
How about the name `extendBackgroundImageToContainerPaddingArea`

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5906,13 +5906,17 @@ export abstract class StylableCardElementContainer extends CardElementContainer
     }
 
     protected getHasBackground(): boolean {
+        return this.getHasPadding(true);
```
almedina-ms, line 5909: should this always be true or should it depend upon GlobalSettings.addPaddingToContainerWithBackgroundImage value?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5906,13 +5906,17 @@ export abstract class StylableCardElementContainer extends CardElementContainer
...
+    }
+
+    protected getHasPadding(shouldCountBackgroundImage : boolean): boolean {
```
licanhua, line 5912: name here is very confusion.
Possible add a function getHasBackgroundImage which is checking if background image is set or not.
then getHasBackground 
```
condition && getHasBackgroundImage()
```

Possible the `getHasPadding`
```
return shouldCountBackgroundImage && getHasBackgroundImage
```

jwoo-msft: @licanhua; I'm not quite sure if I'm following you. getHasPadding return true if this container has padding and return false if it doesn't.

source/nodejs/adaptivecards/src/shared.ts
beervoley: @jwoo-msft I would assume @dclaux is the right person?:)

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5906,13 +5906,17 @@ export abstract class StylableCardElementContainer extends CardElementContainer
...
         let currentElement: CardElement | undefined = this.parent;
 
         while (currentElement) {
             const currentElementHasBackgroundImage =
                 currentElement instanceof Container
-                    ? currentElement.backgroundImage.isValid()
-                    : false;
+                    ? (currentElement.backgroundImage.isValid() && shouldCountBackgroundImage)
```
dclaux, line 5918: Test should more likely be
```typescript
!shouldCountBackgroundImage || currentElement.backgroundImage.isValid()
```

No?

source/nodejs/adaptivecards/src/card-elements.ts
dclaux, line 5909: This can't be. We can't determine that something has a background because it has padding when we determine that it has padding because it has a background

dclaux: To be clear, the logic has so far been that something (typically a container, but the JS SDK could technically support that on any element) has padding because it either has a background (via a container style or a background image). The base implementation in CardElement cannot return `true`. And I don't see any reason why the base implementation should be changed from its current implementation.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux, line 5912: No space before the colon

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Plus the implementation of `StylableCardElementContainer.getDefaultPadding` is wonky:

```typescript
protected getDefaultPadding(): PaddingDefinition {
        let hasPadding = GlobalSettings.addPaddingToContainerWithBackgroundImage ? 
            this.getHasBackground() : this.getHasPadding(false);
        ...
```

It should be
```typescript
let hasPadding = this.getHasPadding(GlobalSettings.addPaddingToContainerWithBackgroundImage);
```

source/nodejs/adaptivecards/src/card-elements.ts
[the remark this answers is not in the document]
dclaux: @almedina-ms it's really the container that gets padded, not the background image

source/nodejs/adaptivecards/src/shared.ts
dclaux: It depends what the feature actually is. `addPaddingToContainersWithBackgroundImage` pretty explicitly means that padding is added to containers that have a background image, which is how things work today. So it seems like the right name for a flag that allows to not add padding to containers that have a background image.

`extendBackgroundToContainerPaddingArea` seems to mean that, when set to `false`, background images will simply not cover the padding of their container, but not that the container isn't going to get padding.

dclaux: Why is this feature needed in the first place? Is there a screenshot of the desired experience that cannot be accomplished without this?

This flag basically makes an AC authored for a host that has it set to false not render properly in a host that has it set to true (and the other way around). That isn't a good interop story, it doesn't seem like a good solution at first glance.

dclaux (changes requested):

source/nodejs/adaptivecards/src/shared.ts
jwoo-msft: @dclaux; this is needed because we have a customer who needs to have custom styling of button. the customer wants to have a button with green background and corner radius. they are trying to achieve this by setting background image in the column. with padding added, their button becomes bigger and the corner radius disappears. this change is designed to support only one specific host at the moment.

source/nodejs/adaptivecards/src/card-elements.ts
jwoo-msft: Thanks David.
You're correct.

source/nodejs/adaptivecards/src/shared.ts
dclaux: @jwoo-msft Wow, that is just not the right way to enable the customer. It's a major hack.

There are likely multiple right ways to support this. Here are a couple of them:
- Enhance `Container`
  - Allow custom paddings in containers. This feature is already supported in JS, it just requires adding a `padding` property on `Container`. You can disable the capability by default via a global setting and only allow specific hosts to support it.
  - Plus adding a `borderRadius` property to `Container`, which is somethingv that we've discussed multiple times in the past
- Introduce a dedicated `Button` element

_Please_ reconsider this.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5906,16 +5906,16 @@ export abstract class StylableCardElementContainer extends CardElementContainer
     }
 
     protected getHasBackground(): boolean {
-        return this.getHasPadding(true);
+        return this.getHasPadding(GlobalSettings.addPaddingToContainerWithBackgroundImage);
```
dclaux, line 5909: No - `CardElement.getHasBackground` MUST return false, always. This isn't the place where I suggested to make the change. Here is my original comment:

Plus the implementation of `StylableCardElementContainer.getDefaultPadding` is wonky:

```typescript
protected getDefaultPadding(): PaddingDefinition {
    let hasPadding = GlobalSettings.addPaddingToContainerWithBackgroundImage ? 
        this.getHasBackground() : this.getHasPadding(false);
    ...
```

It should be
```typescript
protected getDefaultPadding(): PaddingDefinition {
    let hasPadding = this.getHasPadding(GlobalSettings.addPaddingToContainerWithBackgroundImage);
    ...
```

dclaux (changes requested):

source/nodejs/adaptivecards/src/card-elements.ts
jwoo-msft: @dclaux; before the change getHasBackground() of `StyleableCardElementContainer` doesn't consider background image of current element since it start with current element's parent. we could change that to current, but I don't know about the side effects if I do. the reason that it has been working is because this.getHasBackground calls the Container's method which checks if the container's background is valid, only if fails it calls getHasBackground of `StylableCardElementContainer`.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Sorry, I misread the change here. You're making the change in `StylableCardElementContainer.getHasBackground`, and not `CardElement.getHasBackground`. While that's better, it's still not good. Padding is determined by checking if a container has a background. So we cannot determine if a container has a background by checking if it has padding.

dclaux (changes requested):

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -398,7 +398,7 @@ export abstract class CardElement extends CardObject {
         return new PaddingDefinition();
     }
 
-    protected getHasBackground(): boolean {
+    protected getHasBackground(ignoreBackgroundImage: boolean): boolean {
```
dclaux, line 401: Pluralize `ignoreBackgroundImages`

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5905,18 +5905,17 @@ export abstract class StylableCardElementContainer extends CardElementContainer
...
+    protected getHasBackground(ignoreBackgroundImage: boolean = false): boolean {
         let currentElement: CardElement | undefined = this.parent;
 
         while (currentElement) {
-            const currentElementHasBackgroundImage =
-                currentElement instanceof Container
-                    ? (!shouldCountBackgroundImage || currentElement.backgroundImage.isValid()) 
-		    : false;
+            let currentElementHasBackgroundImage: boolean = false;
+            if (ignoreBackgroundImage) {
```
dclaux, line 5913: Empty line before `if`

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -6313,8 +6308,16 @@ export class Container extends ContainerBase {
         }
     }
 
-    protected getHasBackground(): boolean {
-        return this.backgroundImage.isValid() || super.getHasBackground();
+    getEffectivePadding(): PaddingDefinition {
```
dclaux, line 6311: Please move to lower in the class so it lives with the other public methods

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -7706,7 +7709,7 @@ export class AdaptiveCard extends ContainerWithActions {
         return renderedElement;
     }
 
-    protected getHasBackground(): boolean {
+    protected getHasBackground(ignoreBackgroundImage: false): boolean {
```
dclaux, line 7712: Declare default value

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -6313,8 +6308,16 @@ export class Container extends ContainerBase {
         }
     }
 
-    protected getHasBackground(): boolean {
-        return this.backgroundImage.isValid() || super.getHasBackground();
+    getEffectivePadding(): PaddingDefinition {
+        if (GlobalSettings.removePaddingFromContainersWithBackgroundImage && !this.getHasBackground(true)) {
+            return new PaddingDefinition();
+        }
+        return super.getEffectivePadding();
```
dclaux, line 6315: Empty line before `return`

dclaux (changes requested):

dclaux:
Thanks for making the changes! Just a few nits and you're good to go

source/nodejs/adaptivecards/src/card-elements.ts
jwoo-msft: @dclaux Thanks David; your offline suggestion really helped.

dclaux (approved):

almedina-ms (approved):

paulcam206 (approved):

Issue #7012, opened 2022-02-08
[Programmatic access - Adaptive Cards - ExpenseReport.JSON] The name property of focusable element must not be NULL


jwoo-msft:
https://microsoft.visualstudio.com/OS/_workitems/edit/37140182

carlos-zamora:
Closing. This is a duplicate of https://github.com/microsoft/AdaptiveCards/issues/6274.

PR #7066, opened 2022-02-24
Provide name and tooltip for collapsed/expanded arrows under 'Amount' in the 'Expense approval card'

### LINKED ISSUE #6274: [Accessibility] ExpenseReport.Json: Screen Reader does not announce the name and state of the collapse down arrow icon buttons present under 'Amount' under 'Expense approval card'.
### Target Platforms

UWP

### SDK Version

Application Version: 2.7.2108.2001

### Application Name

Visualizer

### Problem Description

[35357132](https://microsoft.visualstudio.com/OS/_workitems/edit/35357132)

Screen Reader does not announce the name and state of the collapse down arrow icon buttons present under 'Amount' under 'Expense approval card'. It just announces "Button".

### Screenshots

_No response_

### Card JSON

[ExpenseReport.json](https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.2/Scenarios/ExpenseReport.json)

### Sample Code Language

_No response_

### Sample Code

_No response_

PankajBhojwani:
# Related Issue
Fixes #6274, #7012

# Description

The issues were regarding an edge case where we use a `Column` as a toggle button, i.e. the column contains 2 items, one visible and one invisible, and clicking the column swaps the visibilities between the items. In this case, the column itself does not have a title or a tooltip, and we need to reach into the visible item to get the `AltText`. 

The fix has 2 parts - 

1. When we first try to set the column's title/tooltip, we reach into the only visible item and use its `AltText`
2. Whenever the column is clicked, we invoke a callback function that updates the column's title/tooltip based on which item is now visible

# Sample Card

Expense report

# How Verified

Manual verification, can no longer reproduce the bugs listed above

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7066)

### DIFF
--- a/source/uwp/Renderer/lib/ActionHelpers.cpp
+++ b/source/uwp/Renderer/lib/ActionHelpers.cpp
@@ -541,10 +541,31 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
                     tooltip = title;
                 }
             }
-            else
+            else if (!tooltip.empty())
             {
                 // If we don't have a title, use the tooltip as the name
                 name = tooltip;
+            }
+            else
+            {
+                // There is an edge case here where the column is being used as a toggle button,
+                // i.e. it contains 2 items, one visible and one invisibile, and they swap visibilities
+                // upon the column being clicked. In this case, we need the AltText of the visible item
+                // since the column itself does not have one
+                if (const auto col = adaptiveCardElement.try_as<winrt::AdaptiveColumn>())
+                {
+                    for (auto item : col.Items())
+                    {
+                        if (item.IsVisible())
+                        {
+                            if (const auto img = item.try_as<winrt::AdaptiveImage>())
+                            {
+                                name = img.AltText();
+                                tooltip = img.AltText();
+                            }
+                        }
+                    }
+                }
             }
 
             // Disable the select action button if necessary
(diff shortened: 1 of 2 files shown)


source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -541,11 +541,32 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
...
+                        if (item.IsVisible())
+                        {
+                            if (const auto img = item.try_as<winrt::AdaptiveImage>())
+                            {
+                                name = img.AltText();
+                                tooltip = img.AltText();
+                            }
+                        }
+                    }
+                }
```
PankajBhojwani, line 568: I am unsure about this - can this be generalized? This code assumes the specific case where the `adaptiveCardElement` is an `AdaptiveColumn` and that its items are of the type `AdaptiveImage`. However I could imagine a scenario where we have the same problem but with a different type of Container and different type of Item.

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -563,22 +584,56 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
...
+                                }
+                                else
+                                {
+                                    img.IsVisible(true);
+                                    SetAutomationNameAndDescription(button, img.AltText(), description);
+                                    SetTooltip(img.AltText(), button);
+                                }
+                            }
+                        }
+                    }
```
PankajBhojwani, line 612: Similar concern (mentioned above) regarding the generalization of this code

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -563,22 +584,56 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
...
+                    {
+                        for (auto item : col.Items())
+                        {
+                            if (const auto img = item.try_as<winrt::AdaptiveImage>())
+                            {
+                                // Swap the visibilities
+                                if (img.IsVisible())
+                                {
+                                    img.IsVisible(false);
+                                }
```
PankajBhojwani, line 603: The `SendActionEvent` gets called before we get into this lambda, so the XAML tree has already been updated with the correct visibilities, however there doesn't seem to be an easy way to get to the XAML visibility from this AdaptiveItem, so we do this manual update here to keep it in sync with the XAML tree. If there is a better way to do this please let me know!

source/uwp/Renderer/lib/ActionHelpers.cpp
jwoo-msft: [ToggleVisibilty](https://adaptivecards.io/explorer/Action.ToggleVisibility.html) has property `targetElements` array.

It lists the elements whose visibility should be toggled when triggered. You could use the list and element's id in the list to figure out which will be visible and use it to grab the tool tip.

PankajBhojwani:
Update: we will be deprecating the `ToggleVisibility` action from our samples, including this one, in favour of implementing an Expander control at some point.

PR #7088, opened 2022-03-02
[Website] Apply fixed table layout to schema tables

### LINKED ISSUE #7014: [Schema Explorer- TextBlock]: On resizing the page to 200% link "https://aka.ms/ACTextFeatures" is getting cropped.
https://microsoft.visualstudio.com/OS/_workitems/edit/37791949

carlos-zamora:
# Related Issue

Closes #7014 

# Description

Applies the `table-layout:fixed` styling to the schema tables on the website. This makes the table fully visible, which forces the "aka.ms/ACTextFeatures" link visible when the page is very zoomed in.

# How Verified

Followed the repro steps:
1. Navigate to 'Schema Explorer' link through tab key under the header controls and activate it.
2. Navigate to the 'TextBlock' in the left side through tab key and activate it.
3. Navigate to the 'https://aka.ms/ACTextFeatures' link through tab key.
4. Verify that on resizing the page to 200% link "https://aka.ms/ACTextFeatures" is not getting cropped

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7088)


paulcam206 (changes requested):
I think maybe we should push back on this bug...

On the live site, I'm seeing the link just fine, but with a horizontal scroll:

image

According to [these docs](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) we *should* be okay with requiring a horizontal scroll:

> Complex data tables have a two-dimensional relationship between the headings and data cells. This relationship is essential to convey the content. This Success Criterion therefore does not apply to whole data tables. However, cells within data tables should fit within the size requirement unless the cell contains types of content that also requires two-dimensional layout for usage or meaning.

FWIW, I'm seeing this with your change at high zoom levels:

image

carlos-zamora:
I've reached out to the accessibility team and am pushing back on it. I'm keeping this open for now until that gets resolved.

carlos-zamora:
We didn't hear back from the accessibility team regarding our concerns.

PR #7096, opened 2022-03-04
[JS] Adding style property to IAction

### LINKED ISSUE #7078: [Authoring] IAction doesn't have the style property
### Target Application

Microsoft Teams, SharePoint

### Application Operating System

Other

### Schema Version

1.2

### Problem Description

The IAction interface in the nodejs SDK (source/nodejs/adaptivecards/src/schema.ts) doesn't have the style property. When typing our cards properly, the TypeScript compiler (tsc) throws an error for this when we set the style of an action.

### Expected Outcome

Should be able to set the style (ActionStyle.Positive, ActionStyle.Destructive, etc.) to an IAction.

### Actual Outcome

Setting the style to an IAction throws an error.

### Card JSON

```TypeScript
import { IAdaptiveCard } from "adaptivecards"

const card: IAdaptiveCard = {
    schema: "http=//adaptivecards.io/schemas/adaptive-card.json",
    type: "AdaptiveCard",
    version: "1.2",
    body: [],
    actions: [
        {
            type: "Action.Submit",
            title: "Some text",
            style: "Positive", // This will give an error
            id "someId",
            data {
                id: "someId",
            }
        }
    ],
}

```

### Repro Steps

_No response_

JCMartell:
# Related Issue

Fixes #7078 

# Description

Adding the style property to the IAction interface.

# How Verified

No unit tests were added because we're simply adding an optional property on an interface.
Ran the adaptivecards tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7096)


source/nodejs/adaptivecards/src/schema.ts
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+import { ActionStyle } from "./enums";
```
licanhua, line 3: We should not import anything from enums. instead, I guess
```
export ActionStyle = "default" | ...
```

licanhua: Please refer to HorizontalAlignment in this file

paulcam206 (approved):
looks good (once other feedback is addressed)

source/nodejs/adaptivecards/src/schema.ts
JCMartell: Is there a reason why we shouldn't import from enums? Doing your suggested change makes it redundant. Let's say a new ActionStyle gets added in the future, we would need to add it both to the enum and here.

licanhua: @dclaux may know why we designed this. I suggest we just follows the rule.

dclaux: Those schema interfaces should be completely decoupled from the rest of the SDK. The idea is that they can exist without the rest of the code. For that matter, they should never have been included in the main SDK in the first place.

Now reality is what it is, and they are in the SDK and could import other files. So I don't know how strongly I feel about this anymore.

dclaux: More on this - the reason they shouldn't have been included in the SDK is that they are mostly useful server-side, but the rest of the SDK is dependent on the DOM.

JCMartell: @dclaux I see, thanks for clarifying. I'll make the suggested change in case your team plans on decoupling this so it's as simple as possible. However, this is error prone. For instance, the schema ContainerStyle is missing some of its possible values defined in the enum. Just an FYI.

dclaux: > However, this is error prone

I certainly don't disagree with that. And as I had written and somehow decided to remove in my previous comment, the interfaces are not up to date, even beyond the problem you're fixing in this PR. I've been wanting to remove them all for a while, but the team doesn't like the idea.

Out of curiosity, how do you use them? And what is they didn't exist?

licanhua (approved):

source/nodejs/adaptivecards/src/schema.ts
JCMartell: @dclaux We're generating our UI with Typescript classes instead of plain JSON. It's easier to pass in variables and it's more compact that way. To do so, we had to create a class for each element in the schema (TextBlock, Container, etc.) and we have a generator method that looks like this:
```Typescript
export function createTemplate(
  body: BaseElement[],
  actions: (ISubmitAction | IOpenUrlAction | IShowCardAction)[] = []
): ISPFxAdaptiveCard {
  return {
    schema: "http=//adaptivecards.io/schemas/adaptive-card.json",
    type: "AdaptiveCard",
    version: "1.2",
    body,
    actions,
  };
}
```
The BaseElement class is an abstract class that all is inherited by every element and it contains the common properties. Here's a sample of what it looks like:

```Typescript
return createTemplate([
    new Container([
        new TextBlock("Hello World").setFontSize(FontSize.Medium).setFontWeight(FontWeight.Bolder);
    ]).setStyle(ContainerStyle.Emphasis);
]);
```

dclaux: @JCMartell thanks for sharing. Indeed the interfaces have not really been maintained.

Any reason why you wouldn't simply use the built-in SDK classes? As in:
```typescript

let textBlock = new TextBlock();
textBlock.text = "Hello World";
textBlock.fontSize = FontSize.Medium;
textBlock.fontWeight = FontWeight.Bolder;

let container = new Container();
container.style = ContainerStyles.Emphasis;
container.addItem(textBlock);

let card = new AdaptiveCard();
card.addItem(container); // Not sure you actually need a Container here though, the card itself is a container
```

It is true that the syntax isn't necessarily the one you're after, but you're essentially redoing a lot of what already exists.

JCMartell: @dclaux It's true, we're redoing a lot of what exists. However, we're using a dependency that extends AdaptiveCards, so there are different elements that have more/different props than the one defined here. At first, they were exporting the extended elements, but they changed that recently for some reason, so we had to do it ourselves.

Issue #7103, opened 2022-03-09
[JS] Only enable action buttons if the values of their associated inputs have changed


dclaux:
### Problem Statement

Currently, action buttons are always enabled or always disabled (via the `isEnabled` property).

Let's take the below scenario as a generic example:
![image](https://user-images.githubusercontent.com/1334689/157349272-ad0da907-e601-4274-853a-08162ad108b3.png)

In the above card, the "Save" button is always enabled, making it look like there's always something to save, which isn't true. Unless the user actually enters a value in either of the two inputs, there is nothing to save.

This existing model provides for a less than optimal user experience. A much better model would make it possible for the card author to specify that the "Save" button is to be disabled unless a value has been entered in one or more of its associated inputs. Essentially, the experience should be the following.

Before the user enters any value:
![image](https://user-images.githubusercontent.com/1334689/157349576-8fab49e5-8b0a-4a66-bab7-259b546d0b58.png)

After the user has modified at least one input:
![image](https://user-images.githubusercontent.com/1334689/157349619-281d0b5f-a63c-4e01-9400-649b7a41aed8.png)

### Proposed solution

We could simply make all this happen by default, but it is likely that the current behavior will need to be preserved for some existing scenarios. So we should make this an opt-in option for card authors.

We should introduce a new `disabledUnlessAssociatedInputsChange` Boolean property on Action.Submit and Action.Execute (which are the only two input-dependent action types). When `disabledUnlessAssociatedInputsChange` is set to `false` (the default), the current behavior (the action is always enabled) applies; if set to `true`, then the action is disabled if its `isEnabled` property is set to `false` OR if none of its associated inputs have changed (i.e. their values are the same as when the card was initially disaplyed).

As the end user changes the values of inputs and clicks an action button associated with those inputs, the "dirty state" of those inputs should be reset (as in, their new values are now considered current) and thus the action button should again become disabled. It is likely though that host applications might want to reset the inputs' dirty state on their own term, and thus we should add a global setting controlling whether or not the SDK automatically performs that reset.

### Alternatives or Workarounds

There are no alternatives or workaround. Some scenarios have opted for pushing inputs and associated "Save" button into  ShowCard just so that there wouldn't be a misleading, always enabled "Save" button on the main card.

paulcam206:
I like this approach and only have one piece of input to consider (I didn't look at how difficult it would be to implement):

```diff
[...]

PR #7104, opened 2022-03-09
[JS] Only enable action buttons when their associated inputs have changed

dclaux:
# Related Issue
Implements https://github.com/microsoft/AdaptiveCards/issues/7103

# Description

This change adds support for disabling action buttons unless their associated inputs have changed
- Adds a `CardElement.getAllActions` method that makes it easy to retrieve a list of all actions in the card
- Adds a `CardElement.updateActionsEnabledState` that refreshes the visual state of all actions in the card
- Adds the `Action.disabledUnlessAssociatedInputsChange: boolean` property on `Action`
  - Add support for that property in the designer
- Changes the logic for a button to be disabled if its `isEnabled` property is set to false OR if none of its associated inputs have changed
- Adds a `SubmitActionBase.updateEnabledState` method that refreshes the visual state of an action button according to the current values of its associated inputs
- Adds a `SubmitActionBase.resetReferencedInputsDirtyState` method that allows a host app to indicate that an action's associated inputs are not "dirty" anymore, i.e. their values are current (and the action should therefore be disabled)
- Adds the `GlobalSettings.resetInputsDirtyStateAfterActionExecution: boolean` flag that allows a host application to disable the automatic reset of inputs' dirty state after an action button has been clicked

# Sample Card

```json
{
    "type": "AdaptiveCard",
    "version": "1.6",
    "body": [
        {
            "type": "Input.Text",
            "id": "abc",
            "label": "Enter some text:"
        },
        {
            "type": "Input.ChoiceSet",
            "choices": [
                {
                    "title": "Choice 1",
                    "value": "Choice 1"
                },
                {
                    "title": "Choice 2",
                    "value": "Choice 2"
                }
            ],
            "id": "def",
            "label": "Make a choice:"
        }
    ],
    "actions": [
        {
            "type": "Action.Execute",
            "title": "Save",
            "disabledUnlessAssociatedInputsChange": true
        }
    ]
}
```

**Initial render:**
image

**After the user has updated at least one input:**
image

# How verified

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7104)


BILLYGameR (changes requested):
source/nodejs/adaptivecards-designer/src/designer-peers.ts

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -1544,14 +1544,16 @@ export abstract class BaseSubmitActionPeer<TAction extends Adaptive.SubmitAction
             { targetVersion: Adaptive.Versions.v1_3, name: "Automatic", value: "auto" },
             { targetVersion: Adaptive.Versions.v1_3, name: "None", value: "none" }
         ]);
+    static readonly disabledUnlessAssociatedInputsChangeProperty = new BooleanPropertyEditor(Adaptive.Versions.v1_0, "disabledUnlessAssociatedInputsChange", "Only enable when associated inputs change");
```
licanhua, line 1547: Actions is not udpated after I flipped the settings on ElementProperty
image

licanhua, line 1547: `Only enable when associated inputs change` is too long

dclaux: @licanhua What other name do you suggest?

licanhua: remove it? and associatedinputs != none means disabledUnlessAssociatedInputsChange

dclaux: Good catch. The checkbox shouldn't appear in the property inspector because you haven't selected version 1.6. This is a designer issue, I'll fix. In the meantime, you can test by setting the target to 1.6.

dclaux: That's completely different. `associatedInputs` controls which inputs are associated with the action. The new property controls if the action should be disabled if none of its associated inputs have changed.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -774,7 +798,7 @@ export class ActionProperty extends PropertyDefinition {
     }
 
     toJSON(
-        _sender: SerializableObject,
```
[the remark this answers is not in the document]
dclaux: Unless we don't want to remove the "_" prefixes (which Paul told me were added because of the `noUnusedParameters` compiler directive, which doesn't seem to be applied anyway) then can you please avoid me the trouble or reverting those changes and opening another PR? Those name changes simply revert to how things used to be and have zero impact on how the code works. They're easily ignorable while reviewing.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -579,6 +581,14 @@ export abstract class CardElement extends CardObject {
         this.applyPadding();
     }
 
+    updateActionsEnabledState() {
+        let allActions = this.getRootElement().getAllActions();
```
licanhua, line 585: nit: const

licanhua (approved):
Code looks good to me. I prefer `const` over `let` in some places, but it's not a blocker for this PR

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -670,10 +680,24 @@ export abstract class CardElement extends CardObject {
         return undefined;
     }
 
-    getAllInputs(_processActions: boolean = true): Input[] {
+    getAllInputs(processActions: boolean = true): Input[] {
         return [];
     }
 
+    getAllActions(): Action[] {
+        let result: Action[] = [];
```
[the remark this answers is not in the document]
dclaux: Done

jwoo-msft (approved):

source/nodejs/adaptivecards-designer/src/designer-peers.ts
JeanRoca: I can only think of 

EnableAssociatedAction or DisableAssociatedAction

dclaux: @JeanRoca doesn't sound too good to me. This property lives on Action. This name would make it sound like the action has an associated action.

Issue #5655, opened 2021-04-12
Update element samples to consume new features (Text Heading, etc)


RebeccaAnne:

carlos-zamora:
> [33779637](https://microsoft.visualstudio.com/OS/_workitems/edit/33779637) is an iOS bug so it is out of scope.

Mind filing an issue to specifically track that one? That way we can close this one once everything else is fixed.

PankajBhojwani:
> Mind filing an issue to specifically track that one? That way we can close this one once everything else is fixed.

Good call, done!

PR #7106, opened 2022-03-09
Delete old samples that don't have latest features (causing accessibility issues)

PankajBhojwani:
# Related Issue

Resolves [35357399](https://microsoft.visualstudio.com/OS/_workitems/edit/35357399) and [35368881](https://microsoft.visualstudio.com/OS/_workitems/edit/35368881), tracked in #5655 

# Description

Update `Inputs.json` and `InputForm.json` to use new features like `label` and `errorMessage`, to close the issues listed in #5655 

Delete the samples `Inputs.json` and `InputForm.json`, because we already have the samples `InputFormWithLabels.json` and `InputsWithValidation.json`, which solve the issues mentioned 

# How Verified

Files no longer exist so neither do the issues with them


samples/v1.0/Scenarios/InputForm.json
```diff
@@ -31,35 +31,25 @@
...
 						},
-						{
-							"type": "TextBlock",
-							"text": "Your name",
-							"wrap": true
-						},
 						{
 							"type": "Input.Text",
 							"id": "myName",
+							"label": "Your name",
```
PankajBhojwani, line 37: Does the version number need to be updated? The `label` attribute seems to be working fine without updating the version number but I'm not sure if we should still update it for consistency?

samples/v1.0/Scenarios/InputForm.json
```diff
@@ -31,35 +31,25 @@
...
-						{
-							"type": "TextBlock",
-							"text": "Your email",
-							"wrap": true
-						},
 						{
 							"type": "Input.Text",
 							"id": "myEmail",
+							"label": "Your email",
+							"regex": "^(.+)@(.+)$",
```
PankajBhojwani, line 44: I've added a very simple regex here for the email validation (it is just a sample after all), but I don't think there is an easy regex for phone number or full name, so those won't have error messages.

carlos-zamora: **Regex for phone number**
- Try this: `/^\\(?([0-9]{3})\\)?[-.\\s]?([0-9]{3})[-.\\s]?([0-9]{4})$/` ([Source](https://regexpattern.com/phone-number/))
- I would rather we have something than nothing here (don't want them to send a similar bug back). It doesn't have to be perfect though. At the end of the day, this is a sample that the author can build off of and introduce their own regex if they want.

**Regex for name**
- I don't think this needs/deserves validation, so I'm fine with it.

carlos-zamora (changes requested):
> Does the `version` number at the top of the file need to be updated? The `label` attribute seems to be working fine without updating the version number but I'm not sure if we should still update it for consistency?

Just blocking over this comment. We definitely should update the version number because an author may copy-and-paste this card directly and expect it to work in 1.0. For reference, unsupported features are generally just dropped. So if you took this new card and put it in a v1.0 renderer, the label should just be dropped entirely (wow, so this card would look really bad haha).

A question I have though is, should we move this card out of the v1.0 folder then? Would that be bad for testing? Need insight from @paulcam206 or @licanhua here.

samples/v1.0/Scenarios/InputForm.json
PankajBhojwani: Yeah I personally wasn't a fan of putting a regex for the phone number that was region-specific but sure, its just a sample. Added it!

samples/v1.0/Scenarios/InputForm.json
paulcam206: the card version is... poorly named. it's effectively the minimum version required to display the card, so leaving it as-is is perfectly fine for this case 👍

paulcam206:
> > Does the `version` number at the top of the file need to be updated? The `label` attribute seems to be working fine without updating the version number but I'm not sure if we should still update it for consistency?
> 
> Just blocking over this comment. We definitely should update the version number because an author may copy-and-paste this card directly and expect it to work in 1.0. For reference, unsupported features are generally just dropped. So if you took this new card and put it in a v1.0 renderer, the label should just be dropped entirely (wow, so this card would look really bad haha).
> 
> A question I have though is, should we move this card out of the v1.0 folder then? Would that be bad for testing? Need insight from @paulcam206 Paul Campbell (DXP) FTE or @licanhua CanHua Li FTE here.

I think in this case it's probably fine to leave it (or update if you feel like pushing another commit) -- labels shipped long enough ago that at this point all major partners support them.

As for the various `vX.X` folders, we don't actually enforce anything based on the file path of the card, so it can stay in `v1.0` :)

paulcam206 (approved):

paulcam206 (changes requested):
actually, I think the right fix here is to just delete these two samples. double-check me on this, but I believe we have `InputsWithLabels.json` and `InputFormWithLabels.json` that are the labeled versions of these...

PankajBhojwani:
> actually, I think the right fix here is to just delete these two samples. double-check me on this, but I believe we have `InputsWithLabels.json` and `InputFormWithLabels.json` that are the labeled versions of these...

That's a good call! Deleted the files

carlos-zamora (approved):

paulcam206 (approved):

Issue #7111, opened 2022-03-11
[JS] Inline YouTube video playback


dclaux:
### Problem Statement

Currently, it is not possible to embed a YouTube video in an Adaptive Card. This feature would make that possible.

### Proposed solution

We extend the existing `Media` element with no schema change. The updated logic does the following:
- If a URL matching a YouTube URL pattern is found in the `sources` collection, the `Media` element is setup to embed an inline YouTube player instead of the standard HTML5 media player
  - The `mimeType` property isn't required for YouTube sources
  - Only one YouTube source is supported. If a YouTube video source is detected, all other sources are ignored
- The element renders the poster as it normally would. However, if the selected source is a YouTube video URL and there is no `poster` property is specified, the `Media` element automatically pulls the YouTube video's thumbnail
- As the `play` button is clicked, the `Media` element instantiates an embedded YouTube video player and automatically starts playback

### Alternatives or Workarounds

The only alternative currently is to use an image that represents a YouTube video thumbnail and to include an OpenURL action somewhere in the card to open the YouTube web site.

paulcam206:
Presumably to support this in non-browser renderers we'd either need to host a web control of some sort or allow card authors to specify other sources as fallback. How would `poster` work in that case? Perhaps it should be optional in the source entry? I'd expect that `poster` in the `Media` element would apply to all sources, with each source being able to override it (do we need a `"poster": "auto"` for HTML5?).

Perhaps something like this:

```json
{
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "type": "AdaptiveCard",
  "version": "1.6",
  "body": [
        {
            "type": "Media",
            "sources": [
                {
                    "url": "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
                },
                {
                    "mimeType": "video/mp4",
                    "url": "https://adaptivecardsblob.blob.core.windows.net/assets/AdaptiveCardsOverviewVideo.mp4",
                    "poster": "https://adaptivecardsblob.blob.core.windows.net/assets/AdaptiveCardsOverviewVideoPoster.png"
                }
            ]
        }
    ]
}
```

dclaux:
It's an interesting idea, although IMO it's overkill.

One of the reasons why the Media element as it currently stands hasn't been able to satisfy the needs developers is because there's not fallback to a YouTube video: a YouTube video only exists in YouTube. So while we could implement what you're suggesting, IMO that fallback model would basically never be exercised.

[...]

PR #7112, opened 2022-03-11
[JS] Add inline YouTube, Vimeo and Dailymotion video playback

dclaux:
# Related Issue

Implements https://github.com/microsoft/AdaptiveCards/issues/7111

# Description

This PR:
- Updates the behavior of the `Media` element to use an embedded player (hosted in an iFrame) for YouTube, Vimeo and Daily motion videos
- Updates the `Media` element so it attempts to automatically pull a video thumbnail from the appropriate provider to use as a poster, unless the `poster` property is set
- Introduces a new public `releaseDOMResources` API that, when called, discards any expensive DOM resource that might have been instantiated
  - In the current implementation, only the `Media` element implements this method in a meaningful way, namely by disposing of the iFrame that embeds the inline player. After the iFrame has been disposed of, the `Media` element reverts to showing the poster.
  - This ability to release resources is **very important** in Teams where iFrames need to be disposed of when they are scrolled off screen to save memory and processing power.
- Refactors the `Media` element source to introduce a new extensibility point, making it possible for host applications to easily add support for additional embedded video players (e.g. Twitch and more)
- Fixes a problem with `adaptivecards-designer-app` where the various host app CSS files were no longer copied into the `dist` folder

# Sample Card

```json
{
    "type": "AdaptiveCard",
    "version": "1.5",
    "body": [
        {
            "type": "TextBlock",
            "text": "Inline YouTube video player demo",
            "wrap": true,
            "size": "Large",
            "weight": "Bolder"
        },
        {
            "type": "Media",
            "sources": [
                {
                    "url": "https://www.youtube.com/watch?v=YsqcODOEO-M&ab_channel=DavidClaux"
                }
            ]
        }
    ]
}
```

Replace the `url` in the above payload with these additional examples to test Vimeo and Dailymotion:
- https://vimeo.com/5543457
- https://www.dailymotion.com/video/x6wxwq4

Initial render, with poster pulled from YouTube:
image

After the play button is clicked:
image

# How Verified

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7112)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2452,7 +2460,234 @@ export class MediaSource extends SerializableObject {
...
+    }
+
+    render(): HTMLElement {
+        let container = document.createElement("div");
+        container.style.position = "relative";
+        container.style.width = "100%";
+        container.style.height = "0";
+        container.style.paddingBottom = "56.25%";
+
+        let iFrame = document.createElement("iframe");
```
paulcam206, line 2603: I believe we need to [specify the `title` property for accessibility](https://dequeuniversity.com/tips/provide-iframe-titles) purposes....

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2452,7 +2460,234 @@ export class MediaSource extends SerializableObject {
...
+
+export interface ICustomMediaPlayer {
+    urlPatterns: RegExp[];
+    createMediaPlayer: (matches: RegExpExecArray) => CustomMediaPlayer
+}
+
 export class Media extends CardElement {
+    static customMediaPlayers: ICustomMediaPlayer[] = [
+        {
+            urlPatterns: [ /https?:\/\/www.youtube.com\/watch\?v=([\w\d-_]+)&?/ig, /https?:\/\/youtu.be\/([\w\d-_]+)/ig ],
```
paulcam206, line 2678: this regex will break if e.g. the youtube url includes a starting time index:

`https://www.youtube.com/watch?v=dQw4w9WgXcQ&t=66s`

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2452,7 +2460,234 @@ export class MediaSource extends SerializableObject {
     }
 }
 
+export abstract class MediaPlayer {
```
paulcam206, line 2463: we should update the schema to account for `MediaSources` changes

paulcam206 (changes requested):

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: In my tests, `title` is never read, which is why I didn't include it. It's easy enough to add it if you insist. We just need to make it localizable.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: The first is ready for that, but not the second. Good catch.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Will do

jwoo-msft (approved):

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Fixed anyway.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Updated the regular expressions for all three providers.
Added start time support for YouTube.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Done. I had to introduce schema 1.6 which wasn't there even though v1.6 and Carousel were already introduced. Someone will have to add the schema for Carousel.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2452,7 +2460,255 @@ export class MediaSource extends SerializableObject {
...
+        for (const source of this.owner.sources) {
+            const mimeComponents = source.mimeType ? source.mimeType.split("/") : [];
+
+            if (mimeComponents.length === 2) {
+                if (!this._selectedMediaType) {
+                    const index = HTML5MediaPlayer.supportedMediaTypes.indexOf(mimeComponents[0]);
+
+                    if (index >= 0) {
+                        this._selectedMediaType = HTML5MediaPlayer.supportedMediaTypes[index];
+                    }
```
licanhua, line 2505: The code is more readable if you add `else {continue;}`
it lets the dev knows that: hi, mediatype is not supported and not need to do selectedSource.push

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2452,7 +2460,255 @@ export class MediaSource extends SerializableObject {
...
+    }
+
+    get videoId(): string | undefined {
+        return this._videoId;
+    }
+}
+
+export class VimeoPlayer extends IFrameMediaMediaPlayer {
+    async fetchVideoDetails(): Promise<void> {
+        const oEmbedUrl = `https://vimeo.com/api/oembed.json?url=${this.getEmbedVideoUrl()}`;
```
licanhua, line 2631: Do you need to make the url in configuration? Likely 3rd party can't change it without code changing in our codebase. How about put it in customMediaPlayers

schemas/1.6.0/adaptive-card.json
```diff
@@ -0,0 +1,3298 @@
+{
+  "$schema": "http://json-schema.org/draft-06/schema#",
```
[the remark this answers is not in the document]
dclaux: Specific changes in this commit: https://github.com/microsoft/AdaptiveCards/pull/7112/commits/cc394cdecd226bbffb9ad4edeca68785e1f5575e

It updates the Media and MediaSource definitions.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2452,7 +2460,255 @@ export class MediaSource extends SerializableObject {
...
+    canPlay(): boolean {
+        return this._videoId !== undefined;
+    }
+
+    render(): HTMLElement {
+        let container = document.createElement("div");
+        container.style.position = "relative";
+        container.style.width = "100%";
+        container.style.height = "0";
+        container.style.paddingBottom = "56.25%";
```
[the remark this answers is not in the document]
dclaux: Great question - I have no idea. It's a magic number. It's "documented" all over the web. It's "the way" to do this in order to force the player to resize dynamically.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: `continue` forces the code to start the next iteration in the nearest loop statement. If I added it, the next `if (mimeComponents[0] === this._selectedMediaType` statement would be skipped, which isn't what needs to happen.

source/nodejs/adaptivecards/src/card-elements.ts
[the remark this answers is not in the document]
sujghosh: 16:9 aspect ratio translates to 56.25%

source/nodejs/adaptivecards/src/card-elements.ts
licanhua: I mean can we make `https://vimeo.com/api/oembed.json` in global definition or somewhere, then if vimeo changed their API endpoint, we don't need to make code change.

dclaux: I suppose we could always do something like that in the future, but I don't think Vimeo is likely to change their API as it would break a lot of client applications.

licanhua (approved):

schemas/src/elements/Media.json
```diff
@@ -14,7 +14,7 @@
 		},
 		"poster": {
 			"type": "uri-reference",
-			"description": "URL of an image to display before playing. Supports data URI in version 1.2+"
+			"description": "URL of an image to display before playing. Supports data URI in version 1.2+. If poster is omitted, the Media element will either use a default poster (controlled by the host application) ot will attempt to automatically pull the poster from the target video service when the source URL points to a video from a Web provider such as YouTube."
```
paulcam206, line 17: small typo:

```suggestion
			"description": "URL of an image to display before playing. Supports data URI in version 1.2+. If poster is omitted, the Media element will either use a default poster (controlled by the host application) or will attempt to automatically pull the poster from the target video service when the source URL points to a video from a Web provider such as YouTube."
```

schemas/src/elements/MediaSource.json
```diff
@@ -8,8 +8,8 @@
 	"properties": {
 		"mimeType": {
 			"type": "string",
-			"description": "Mime type of associated media (e.g. `\"video/mp4\"`).",
-			"required": true
+			"description": "Mime type of associated media (e.g. `\"video/mp4\"`). For YouTube and other Web video URLs, mimeType can be omitted.",
```
paulcam206, line 11: small formatting suggestion:

```suggestion
			"description": "Mime type of associated media (e.g. `\"video/mp4\"`). For YouTube and other Web video URLs, `mimeType` can be omitted.",
```

paulcam206 (approved):
approved, but there are a couple of small tweaks to the schema text that we should probably take

dclaux:
Thanks @paulcam206. I have made the changes you suggested.

Issue #7116, opened 2022-03-11
[Designer- Undo/Redo]: In tab mode focus lands to the 'Undo/Redo' when it is in disable state.


PankajBhojwani:
https://microsoft.visualstudio.com/OS/_workitems/edit/37768448

PR #7119, opened 2022-03-11
Tab focus can no longer land on disabled 'undo'/'redo' buttons

PankajBhojwani:
# Related Issue

#7116 

# Description

For toolbar buttons, we now set the tab index to be -1 if it is disabled or 0 if it is enabled (so it is no longer possible to tab to it when it is disabled, but it is still possible to tab to it when it is enabled).

# Sample Card

N/A

# How Verified

Manual verification

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7119)


beervoley (approved):
:shipit:

jwoo-msft (approved):

Issue #7146, opened 2022-03-16
[Designer- Help]: Summary/Caption is not defined for the table present under the 'Help' dialog.


PankajBhojwani:
https://microsoft.visualstudio.com/OS/_workitems/edit/37784262

PR #7149, opened 2022-03-16
Add a caption to the table present in the 'help' dialog

PankajBhojwani:
# Related Issue

#7146 

# Description

The text `These keyboard shortcuts only work when the keyboard focus is within either the card payload or sample data editors` is now a caption to the keyboard shortcut table.

# Sample Card

N/A

# How Verified

Manually verified that a caption now exists for the keyboard shortcut table

### DIFF
--- a/source/nodejs/adaptivecards-designer/src/help-dialog.ts
+++ b/source/nodejs/adaptivecards-designer/src/help-dialog.ts
@@ -34,12 +34,14 @@ export class HelpDialog extends Dialog {
         keyboardHeading.innerText = "Keyboard Shortcuts";
         this._renderedElement.appendChild(keyboardHeading);
 
-        let keyboardText = document.createElement("div");
-        keyboardText.innerText = "These keyboard shortcuts only work when the keyboard focus is within either the card payload or sample data editors";
-        this._renderedElement.appendChild(keyboardText);
-        this._renderedElement.appendChild(document.createElement("br"));
-
         let shortcutTable = document.createElement("table");
+
+        let tableCaption = shortcutTable.createCaption();
+        tableCaption.style.textAlign = "left";
+        tableCaption.innerHTML = "These keyboard shortcuts only work when the keyboard focus is within either the card payload or sample data editors"
+
+        shortcutTable.appendChild(document.createElement("br"));
+
         shortcutTable.appendChild(this._getShortcutRow("Shortcut", "Description", true));
         shortcutTable.appendChild(this._getShortcutRow("Alt-F1", "Show help dialog for editor"));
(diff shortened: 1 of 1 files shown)


source/nodejs/adaptivecards-designer/src/help-dialog.ts
```diff
@@ -34,12 +34,14 @@ export class HelpDialog extends Dialog {
...
-        let keyboardText = document.createElement("div");
-        keyboardText.innerText = "These keyboard shortcuts only work when the keyboard focus is within either the card payload or sample data editors";
-        this._renderedElement.appendChild(keyboardText);
-        this._renderedElement.appendChild(document.createElement("br"));
-
         let shortcutTable = document.createElement("table");
+
+        let tableCaption = shortcutTable.createCaption();
+        tableCaption.style.textAlign = "left";
+        tableCaption.innerHTML = "These keyboard shortcuts only work when the keyboard focus is within either the card payload or sample data editors"
```
paulcam206, line 41: protip: [avoid `innerHTML`](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations)

this _specific_ use is not problematic, but better to avoid introducing new uses of it...

```suggestion
        tableCaption.innerText = "These keyboard shortcuts only work when the keyboard focus is within either the card payload or sample data editors"
```

paulcam206 (changes requested):
on small issue

source/nodejs/adaptivecards-designer/src/help-dialog.ts
PankajBhojwani: Fixed! Thank you

paulcam206 (approved):

source/nodejs/adaptivecards-designer/src/help-dialog.ts
```diff
@@ -34,12 +34,14 @@ export class HelpDialog extends Dialog {
...
-        let keyboardText = document.createElement("div");
-        keyboardText.innerText = "These keyboard shortcuts only work when the keyboard focus is within either the card payload or sample data editors";
-        this._renderedElement.appendChild(keyboardText);
-        this._renderedElement.appendChild(document.createElement("br"));
-
         let shortcutTable = document.createElement("table");
+
+        let tableCaption = shortcutTable.createCaption();
+        tableCaption.style.textAlign = "left";
+        tableCaption.innerText = "These keyboard shortcuts only work when the keyboard focus is within either the card payload or sample data editors"
```
licanhua, line 41: nit: missing `;`

licanhua, line 41:
```suggestion
        tableCaption.innerText = "These keyboard shortcuts only work when the keyboard focus is within either the card payload or sample data editors";
```

licanhua (approved):

source/nodejs/adaptivecards-designer/src/help-dialog.ts
PankajBhojwani: Thanks!

PR #7172, opened 2022-03-29
Trigger onExecuteAction for Toggle Visibility Action

### LINKED ISSUE #6742: [Rendering] Event Handler for ToggleVisibilityAction
### Target Platforms

NodeJS

### SDK Version

1.4

### Application Name

Berlin

### Problem Description

From dev - I'm currently working on telemetry for Adaptive Card actions like ShowCard and ToggleVisibility.  One concern that was brought up to me was that OnElementVisibilityChanged is called for each target element listed within a ToggleVisibility action and with the current way our telemetry is setup, there would be 3 telemetry events logged for each target element. We'd like to avoid duplicate events and ideally want one telemetry event per ToggleVisibility action.
 
From what I can tell in card-elements.d.ts, there's no handler for ToggleVisibility actions so logging a telemetry event for the ToggleVisibility invocation itself would require some work on our end. Is there an easy way to do this that I missed or examples of how others have handled this in the past?

David's solution - ToggleVisibilityAction indeed doesn’t trigger the onExecuteAction event; I think it should, which would solve your problem. It’s a very simple change that I assume the team could take for you guys. It just requires adding super.execute(); at the beginning of the ToggleVisibilityAction.execute() implementation.

Paul asked to file a GitHub issue.

### Screenshots

_No response_

### Card JSON

```JSON
NA
```

### Sample Code Language

_No response_

### Sample Code

_No response_

eunia:
Fixes #6742 

Issue was fixed by adding super.execute() at the beginning of the ToggleVisibilityAction.execute() implementation, per David's recommendation on the issue. This will make the ToggleVisibilityAction trigger the onExecuteAction event.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7172)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5175,6 +5175,8 @@ export class ToggleVisibilityAction extends Action {
     }
 
     execute() {
+		super.execute();
```
jwoo-msft, line 5178: nit: spacing

jwoo-msft (changes requested):
~could you fix the spacing issue?~

jwoo-msft (approved):

eunia:
Sorry, @jwoo-msft for some reason did not get your updated comment and added the spacing fix. Should I add it again?

beervoley:
> Sorry, @jwoo-msft for some reason did not get your updated comment and added the spacing fix. Should I add it again?

fixed it for you:)

paulcam206 (approved):

PR #7175, opened 2022-03-30
[NodeJS] Change cursor to pointer when hover on action

### LINKED ISSUE #7161: [Rendering] Actions do not display correct mouse pointer unless specified by host application in CSS.
### Target Platforms

NodeJS

### SDK Version

1.5

### Application Name

All

### Problem Description

When a user places any Action on an image the mouse pointer should change to indicate it is a clickable image on hover. This does not work unless the host specifies this behavior explicitly through CSS and not just for images but for other elements like containers as well.

### Screenshots

_No response_

### Card JSON

```JSON
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "TextBlock",
            "size": "Medium",
            "weight": "Bolder",
            "text": "${title}"
        },
        {
            "type": "ColumnSet",
            "columns": [
                {
                    "type": "Column",
                    "items": [
                        {
                            "type": "Image",
                            "style": "Person",
                            "url": "${creator.profileImage}",
                            "size": "Small",
                            "selectAction": {
                                "type": "Action.OpenUrl",
                                "url": "www.google.com"
                            }
                        }
                    ],
                    "width": "auto"
                },
                {
                    "type": "Column",
                    "items": [
                        {
                            "type": "TextBlock",
                            "weight": "Bolder",
                            "text": "${creator.name}",
                            "wrap": true
                        },
                        {
                            "type": "TextBlock",
                            "spacing": "None",
                            "text": "Created {{DATE(${createdUtc},SHORT)}}",
                            "isSubtle": true,
                            "wrap": true
                        }
                    ],
                    "width": "stretch"
                }
            ]
        },
        {
            "type": "Container",
            "style": "emphasis",
            "selectAction": {
                "type": "Action.OpenUrl",
                "url": "www.google.com"
            }
        },
        {
            "type": "TextBlock",
            "text": "${description}",
            "wrap": true
        },
        {
            "type": "FactSet",
            "facts": [
                {
                    "$data": "${properties}",
                    "title": "${key}:",
                    "value": "${value}"
                }
            ]
        }
    ],
    "actions": [
        {
            "type": "Action.OpenUrl",
            "title": "View",
            "url": "${viewUrl}"
        }
    ],
    "version": "1.5"
}
```

### Sample Code Language

_No response_

### Sample Code

_No response_

anna-dingler:
# Related Issue

Fixes #7161

# Description

Issue was fixed by indicating in SCSS that the cursor should be a pointer when hovering over an action.

# How Verified

Verified manually in the AdaptiveCards UI Test App.

Before:

https://user-images.githubusercontent.com/98650930/160898548-542dd108-f4bf-4532-9a25-1213a824b2d5.mp4

After:

https://user-images.githubusercontent.com/98650930/160898576-dee8108d-2c5b-4426-ba9b-df33cad98def.mp4

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7175)


source/nodejs/adaptivecards/src/scss/adaptivecards-base.scss
```diff
@@ -13,6 +13,7 @@ $font-family: 'Segoe UI', sans-serif;
 
             &:hover {
                 background-color: $selectable-hover-background !important;
+				cursor: pointer;
```
beervoley, line 16: Is there a difference between `hover { cursor: pointer}` or simply `{cursor: pointer}` ?
I don't think so and it's probably a better to have `{cursor: pointer}` for this class in general although it yields the same result :)

beervoley, line 16: We then probably don't need `{cursor: pointer}` [here](https://github.com/microsoft/AdaptiveCards/blob/951131c964682b6c0d32ebb25528b10ff5ec186f/source/nodejs/adaptivecards/src/scss/adaptivecards-base.scss#L25)
image

since we'll have it on `.ac-selectable`

source/nodejs/adaptivecards/src/scss/adaptivecards-base.scss
```diff
@@ -165,6 +166,7 @@ $font-family: 'Segoe UI', sans-serif;
             background-color: $hover-background-color;
             color: $hover-color;
             border: $hover-border;
+			cursor: pointer;
```
beervoley, line 169: Same as previous comment: I assume we can move `{cursor: pointer}` out of the `hover` state.
nit: spaces

jwoo-msft (commented):
looks good except those spacing issues

jwoo-msft (approved):

source/nodejs/adaptivecards/src/scss/adaptivecards-base.scss
paulcam206: just to jump on here with the spacing :)

@anna-dingler - if possible you should configure your editor to apply `prettier` formatting. If you're using VSCode, you can use the `Prettier` extension. More info on setting that up [here](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode).

anna-dingler: Making that change now :)

I need to set my tabs to be spaces, and I'll look at the prettier formatting too

anna-dingler: @beervoley Don't we need it here because we only have it on `ac-container.ac-selectable` and `ac-columnSet.ac-selectable` above?

beervoley: oh, you're right - my mistake. We still need it :)
I didn't read the code correctly :)

beervoley: @paulcam206 may be we should extract one liner `.ac-selectable {cursor: pointer}`? since it's natural for selectable items to have this?:)

source/nodejs/adaptivecards/src/scss/adaptivecards-base.scss
```diff
@@ -149,7 +150,7 @@ $font-family: 'Segoe UI', sans-serif;
         text-overflow: ellipsis;
         text-align: center;
         vertical-align: middle;
-        cursor: default;
+        cursor: pointer;
```
licanhua, line 153: can we check with designer? As I know, in xaml, we used backgroundcolor/bordercolor to signal the hover other than pointer. Also I'm wondering if it introduces issue on xbox.

beervoley: image

licanhua: Add @paulcam206 the question here is that should we take the pointer change?

paulcam206: I think the change is fine as-is. 

I don't know think we want to apply a default hover color -- there's no generically-correct color to use. It looks like some hosts already define one in CSS (note that this particular style wouldn't work correctly if the rest of its theme were dark-mode oriented):

https://github.com/microsoft/AdaptiveCards/blob/3a35997792bd7f7b4d4deb155f21804803084e70/source/nodejs/adaptivecards-designer/src/containers/outlook/outlook-container.css#L60-L66

paulcam206 (approved):

source/nodejs/adaptivecards/src/scss/adaptivecards-base.scss
anna-dingler: @paulcam206 Did you have any thoughts on this?

Issue #4963, opened 2020-10-23
Media captions support for accessibility


productboard-ac:
# Description

Reference: https://captionmax.com/common-subtitle-prerecorded-closed-caption-file-types/

Accessibility bug: [28383785](https://microsoft.visualstudio.com/OS/_workitems/edit/28383785)

## Open Issues

- [ ] iOS Sidecar support?
- [ ] Android sidecar support?
- [ ] WPF sidecar support?
- [x] UWP sidecar support?
- [x] JS sidecar support?
- [ ] What about per-language captions?

## Requirements

1. If a video is encoded with closed captions they should be automatically supported on each platform (if the media mimeType is supported)
2. If the video file does not have captions, one or more "sidecar" caption formats can be provided, which will be passed along to the Renderer/Native Platform Media player as appropriate.

### Media

One new property on `Media`

| Property       | Type         | Required | Description |
| -------------- | ------------ | -------- | ----------------- |
| captions | `CaptionSource[]` | No | Array of "sidecar" close caption sources that will be passed to the underlying platform.

### CaptionSource

New type called `CaptionSource`

| Property       | Type         | Required | Description |
| -------------- | ------------ | -------- | ----------------- |
| mimeType | `string` | Yes | Mime type of associated closed captions (e.g. "scc").
| url | `uri` | Yes | URL to closed captions file.
| label | `label` | Yes | Label to display for this caption track.

## Examples

```json
{
	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
	"type": "AdaptiveCard",
	"version": "1.1",
	"fallbackText": "This card requires Media to be viewed. Ask your platform to update to Adaptive Cards v1.1 for this and more!",
	"body": [
		{
			"type": "Media",
			"poster": "https://adaptivecards.io/content/poster-video.png",
			"altText": "Adaptive Cards overview video",
			"sources": [
				{
					"mimeType": "video/mp4",
					"url": "https://adaptivecardsblob.blob.core.windows.net/assets/AdaptiveCardsOverviewVideo.mp4"
				}
			],
			"captionSources": [
				{
					"mimeType": "vtt",
					"label": "English (vtt)",
					"url": "https://gist.githubusercontent.com/PankajBhojwani/d11855a1e0c0b1f70a0e64f81a1fb696/raw/46fa93da5c3703d03843d59cedd8b5f10aa3747b/test.vtt"
				}
			]
		}
	],
	"actions": [
		{
[...]

PR #7105, opened 2022-03-09
Adaptive Card v1.6

sowrabh-msft:
## Changes made:
- Add support for top level property `metadata` and `metadata.webUrl`
- Add support for `expires` property in `refresh`

### Metadata and `metadata.webUrl`
Metadata is a top level property to host non-functionality related metadata properties that do not affect rendering or actionability. The first property we are introducing here is the `webUrl` which can be used as a fallback by hosts to redirect users to the URL and uniquely map cards to their source URL.

### `refresh.expires` property
`refresh.expires` is a way for card authors to indicate the freshness of content delivered. The canonical scenario is a flight tracking card. If the flight is a month away, the developer could set the expiration weekly, but as the flight time approaches the expiration window can be reduced. 

The client is able to use this information as it sees fit, up to and including automatically refreshing the content if the expiration time has passed or avoiding refresh while the content is still valid.

### Example payload
```json
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "metadata": { 
        "webUrl": "[https://](https://../)asana.com/ticket/5"
    },
    "refresh": {
        "expires": "2022-01-01T12:00:00Z",
        "action": { 
            "type": "Action.Execute",
            "verb": "..."   
         }
    },
    "body": { ... }
}
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7105)

PR #7178, opened 2022-03-31
Implement closed captions in JS and UWP

PankajBhojwani:
# Related Issue

#4963 

# Description

Card authors can now add a `closedCaptions` property to `Media`, specifying a list of caption sources that the media consumer can turn on. The `closedCaptions` property follows what is defined in #4963, with an added "language" property.

***NOTE***: The captions in the below sample card are not really the captions for the video (I'm not sure if they exist yet/I am not sure how to generate them) - they are just vtt/srt files that I wrote and put some random text in.

# Sample Card

```json
{
	"type": "AdaptiveCard",
	"version": "1.1",
	"fallbackText": "This card requires Media to be viewed. Ask your platform to update to Adaptive Cards v1.1 for this and more!",
	"body": [
		{
			"type": "Media",
			"poster": "https://adaptivecards.io/content/poster-video.png",
			"altText": "Adaptive Cards overview video",
			"sources": [
				{
					"mimeType": "video/mp4",
				}
			],
			"captionSources": [
				{
					"mimeType": "vtt",
					"label": "English (vtt)",
					"url": "https://gist.githubusercontent.com/PankajBhojwani/d11855a1e0c0b1f70a0e64f81a1fb696/raw/46fa93da5c3703d03843d59cedd8b5f10aa3747b/test.vtt"
				},
				{
					"mimeType": "srt",
					"label": "English (srt)",
					"url": "https://gist.githubusercontent.com/PankajBhojwani/fab75d6493dddd9f5ee964228ca35c71/raw/1706044f46c56216f6a66b8b9e3bb9ecdbf2c56a/zzz.srt"
				}
			]
		}
	],
	"actions": [
		{
			"type": "Action.OpenUrl",
			"title": "Learn more",
		}
	]
}
```

# How Verified

Used the sample card above in the nodejs designer and the uwp visualizer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7178)


source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -224,25 +276,17 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
-                        bool audioOnly = mediaElement.IsAudioOnly();
-                        auto posterSource = mediaElement.PosterSource();
-
-                        if (audioOnly && posterSource == nullptr)
-                        {
-                            // If this is audio only and there's no poster, set the height so
-                            // that the controls are visible.
-                            mediaElement.Height(c_audioHeight);
-                        }
-                        mediaElement.Play();
```
PankajBhojwani, line 245: The controls show up even if the media is audio only and posterSource is null, so I went ahead and removed this code

PankajBhojwani:
One outstanding issue:

Despite `srt` files having the functionality to configure font options, the XAML `MediaPlayerElement` refuses to display the text with the given font options - it only displays the text with its internal default font style. Since the default font style has a white foreground, the captions are invisible if the video has a white background. This is obviously an accessibility concern, but I believe this is on the XAML `MediaPlayerElement`, not us

source/uwp/Visualizer/Helpers/PayloadValidator.cs
```diff
@@ -68,7 +68,8 @@ public static async Task<List<ErrorViewModel>> ValidateAsync(string payload)
                         // We don't report this error to user, it's a coding error
                         if (Debugger.IsAttached)
                         {
-                            Debugger.Break();
+                            // Commenting this out right now because its annoying, remember to un-comment before merging
```
licanhua, line 71: Do you need this change? It's better to keep it as it is

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2476,6 +2476,49 @@ export class ImageSet extends CardElementContainer {
...
+
+    constructor(url?: string, mimeType?: string) {
+        super();
+
+        this.url = url;
+        this.mimeType = mimeType;
+    }
+
+    isValid(): boolean {
+        return this.mimeType && this.url ? true : false;
```
licanhua, line 2505: `return this.mimeType && this.url`

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2476,6 +2476,49 @@ export class ImageSet extends CardElementContainer {
     }
 }
 
+export class CaptionSource extends SerializableObject {
```
licanhua, line 2479: @paulcam206 and @dclaux 

CaptionSource and MediaSource are nearly the same except getSchemaKey and render. What's the strategy for this? Create a base class like mimeSource, and then make CaptionSource and MediaSource subclasses of it, or duplicate the code like in this file

source/uwp/Visualizer/Helpers/PayloadValidator.cs
PankajBhojwani: No I don't - I just have it there while testing, will revert before merging

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2598,6 +2647,9 @@ export class HTML5MediaPlayer extends MediaPlayer {
         );
         this._mediaElement.setAttribute("webkit-playsinline", "");
         this._mediaElement.setAttribute("playsinline", "");
+        // I had to put this here for testing because the caption file has a different domain as the video and html doesn't like that.
+        // We should probably remove this before merging
+        this._mediaElement.setAttribute("crossorigin", "");
```
licanhua, line 2652: You may need to remove this in code review

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2545,10 +2588,12 @@ export abstract class MediaPlayer {
 export class HTML5MediaPlayer extends MediaPlayer {
     private _selectedMediaType?: string;
     private _selectedSources: MediaSource[] = [];
+    private _captionSources: CaptionSource[] = [];
```
licanhua, line 2591: From high level, there are two implementations, and implementation 1 is what you did.
Personally, I think 2 is easier to understand and will have less change.
Option 1 is harder to maintain the mapping between source and caption

Implementation 1:
```
"sources": [
				{
					"mimeType": "video/mp4",
				}
			],
			"captionSources": [
				{
					"mimeType": "vtt",
					"url": "https://gist.githubusercontent.com/PankajBhojwani/d11855a1e0c0b1f70a0e64f81a1fb696/raw/46fa93da5c3703d03843d59cedd8b5f10aa3747b/test.vtt"
				},
				{
					"mimeType": "srt",
					"url": "https://gist.githubusercontent.com/PankajBhojwani/fab75d6493dddd9f5ee964228ca35c71/raw/1706044f46c56216f6a66b8b9e3bb9ecdbf2c56a/zzz.srt"
				}
			]
		}
```

Implementation 2:
```
"sources": [
				{
					"mimeType": "video/mp4",
					"captionMimeType": "vtt",
					"captionUrl": "https://gist.githubusercontent.com/PankajBhojwani/d11855a1e0c0b1f70a0e64f81a1fb696/raw/46fa93da5c3703d03843d59cedd8b5f10aa3747b/test.vtt"
				}
			]
		}
```

source/shared/cpp/ObjectModel/CaptionSource.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "CaptionSource.h"
+#include "ParseUtil.h"
+
+using namespace AdaptiveCards;
+
+CaptionSource::CaptionSource()
+{
+}
+
+Json::Value CaptionSource::SerializeToJsonValue() const
```
licanhua, line 13: I'm wondering if you need propagate change to iOS and Android

source/nodejs/adaptivecards/src/card-elements.ts
PankajBhojwani: If we end up going with this implementation, `captionUrl` will probably need to be `captionSources` and take in an array, because there could be multiple caption tracks for the same video

dclaux: I don't think we can do that. The tracks apply to all the media sources. #2 would imply that each source has its own tracks, which isn't the case.

PankajBhojwani: Also, the implementation I followed here is what was outlined in #4963

source/nodejs/adaptivecards/src/card-elements.ts
PankajBhojwani: I left it in so that reviewers can test it and see it working

source/nodejs/adaptivecards/src/card-elements.ts
PankajBhojwani: That does not work:

```
adaptivecards: [0] src/card-elements.ts(2505,9): error TS2322: Type 'string | undefined' is not assignable to type 'boolean'.
adaptivecards: [0]   Type 'undefined' is not assignable to type 'boolean'.
adaptivecards: [0] tsc exited with code 2
```

source/shared/cpp/ObjectModel/CaptionSource.cpp
PankajBhojwani: I believe we just need to do this for UWP and JS, @jonmill can correct me here if I'm wrong

source/nodejs/adaptivecards/src/card-elements.ts
licanhua: Look like 1 is the way to go. Possible rename CaptionSources to VideoTrack? Also it should include more properties like srclang, kind...

I found an example of multiple sources and tracks from [Adding_captions_and_subtitles_to_HTML5_video](
https://developer.mozilla.org/en-US/docs/Web/Guide/Audio_and_video_delivery/Adding_captions_and_subtitles_to_HTML5_video)

```
<video id="video" controls preload="metadata">
   <source src="video/sintel-short.mp4" type="video/mp4">
   <source src="video/sintel-short.webm" type="video/webm">
   <track label="English" kind="subtitles" srclang="en" src="captions/vtt/sintel-en.vtt" default>
   <track label="Deutsch" kind="subtitles" srclang="de" src="captions/vtt/sintel-de.vtt">
   <track label="Español" kind="subtitles" srclang="es" src="captions/vtt/sintel-es.vtt">
</video>

```

source/shared/cpp/ObjectModel/CaptionSource.h
```diff
@@ -0,0 +1,50 @@
...
+{
+class CaptionSource
+{
+public:
+    CaptionSource();
+    CaptionSource(const CaptionSource&) = default;
+    CaptionSource(CaptionSource&&) = default;
+    CaptionSource& operator=(const CaptionSource&) = default;
+    CaptionSource& operator=(CaptionSource&&) = default;
+    virtual ~CaptionSource() = default;
```
beervoley, line 20: Do we need to define any of those?:) they're still **default** and compiler will do it for us:)

source/shared/cpp/ObjectModel/CaptionSource.h
```diff
@@ -0,0 +1,50 @@
...
+
+class CaptionSourceParser
+{
+public:
+    CaptionSourceParser() = default;
+    CaptionSourceParser(const CaptionSourceParser&) = default;
+    CaptionSourceParser(CaptionSourceParser&&) = default;
+    CaptionSourceParser& operator=(const CaptionSourceParser&) = default;
+    CaptionSourceParser& operator=(CaptionSourceParser&&) = default;
+    virtual ~CaptionSourceParser() = default;
```
beervoley, line 45: Same as the comment above

source/shared/cpp/ObjectModel/CaptionSource.h
```diff
@@ -0,0 +1,50 @@
...
+#include "BaseCardElement.h"
+#include "ElementParserRegistration.h"
+#include "Util.h"
+
+namespace AdaptiveCards
+{
+class CaptionSource
+{
+public:
+    CaptionSource();
```
beervoley, line 15: Default constructor you've defined in `.cpp` doesn't do anything special, so may be `CaptionSource() = default` will suffice?

source/shared/cpp/ObjectModel/CaptionSource.h
```diff
@@ -0,0 +1,50 @@
...
+#pragma once
+
+#include "pch.h"
+#include "BaseCardElement.h"
+#include "ElementParserRegistration.h"
+#include "Util.h"
+
+namespace AdaptiveCards
+{
+class CaptionSource
```
beervoley, line 12: How is this class any different from `MediaSource` ?:) It's identical. In order to not copy the same code, may we can extract the base class and have two children `MediaSource` and `CaptionSource` to differentiate?:)

source/uwp/AdaptiveCardsObjectModel/idl/AdaptiveCards.ObjectModel.Uwp.idl
```diff
@@ -324,6 +324,14 @@ namespace AdaptiveCards.ObjectModel.Uwp
         Windows.Foundation.Collections.IVector<IAdaptiveInline> Inlines { get; };
     };
 
+    runtimeclass AdaptiveCaptionSource
```
beervoley, line 327: This is exactly the same as `AdaptiveMediaSource` - let's refactor into a single base class with two children?:)

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -181,22 +181,49 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
-                mediaElement.SetSource(randomAccessStream, mimeType);
+                auto mediaSrc = winrt::MediaSource::CreateFromStream(randomAccessStream, mimeType);
+                if (adaptiveMedia.CaptionSources().Size() > 0)
+                {
+                    for (const auto captionSource : adaptiveMedia.CaptionSources())
+                    {
+                        const auto timedTextURL = GetUrlFromString(renderContext.HostConfig(), captionSource.Url());
+                        const auto timedTextSrc = winrt::TimedTextSource::CreateFromUri(timedTextURL);
+                        timedTextSrc.Resolved(
+                            [captionSource](winrt::TimedTextSource sender, winrt::TimedTextSourceResolveResultEventArgs args)
```
beervoley, line 202:
```suggestion
                            [captionSource](winrt::TimedTextSource const& sender, winrt::TimedTextSourceResolveResultEventArgs const& args)
```

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -181,22 +181,49 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
+                            {
+                                if (!args.Error())
+                                {
+                                    args.Tracks().GetAt(0).Label(captionSource.MimeType());
+                                }
+                            });
+                        mediaSrc.ExternalTimedTextSources().Append(timedTextSrc);
+                    }
+                }
+                auto playbackItem = winrt::MediaPlaybackItem(mediaSrc);
```
beervoley, line 212:
```suggestion
                winrt::MediaPlaybackItem playbackItem{mediaSrc};
```

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -215,7 +242,30 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
+                playbackItem.TimedMetadataTracksChanged(
+                    [playbackItem](winrt::IInspectable const& /*sender*/, winrt::IInspectable const& /*args*/)
+                    {
+                        playbackItem.TimedMetadataTracks().SetPresentationMode(
+                            0, winrt::TimedMetadataTrackPresentationMode::PlatformPresented);
+                    });
+                mediaElement.Source(playbackItem);
```
beervoley, line 268: Could you please extract that code into a function to not repeat the code twice?:)

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -224,25 +274,17 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
                 auto getResourceStreamOperation = resourceResolver.GetResourceStreamAsync(args);
```
beervoley, line 274: @paulcam206 since we allow for custom resolver for `mediaSource` - should we allow for custom resolver for `captionSource` then as well?

source/uwp/Renderer/lib/MediaHelpers.cpp
beervoley: If that's the case, then we probably don't need `c_audioHeight` - could you please check for usages and delete it if it's the only place where it is used?

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -224,25 +274,17 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
+                    { return HandleMediaResourceResolverCompleted(operation, status, mediaElement, mimeType, adaptiveMedia, renderContext); });
             }
 
-            mediaElement.MediaOpened(
-                [](winrt::IInspectable const& sender, winrt::RoutedEventArgs const& /*args*/) -> void
+            mediaElement.MediaPlayer().MediaOpened(
+                [](winrt::IInspectable const& sender, winrt::IInspectable const& /*args*/) -> void
                 {
-                    if (const auto mediaElement = sender.try_as<winrt::MediaElement>())
+                    if (const auto mediaPlayer = sender.try_as<winrt::Windows::Media::Playback::MediaPlayer>())
```
beervoley, line 285:
```suggestion
                    if (const auto mediaPlayer = sender.try_as<winrt::MediaPlayer>())
```

source/uwp/Renderer/dll/CppWinRTIncludes.h
```diff
@@ -32,6 +34,8 @@ namespace winrt
 	using namespace ::winrt::Windows::Foundation;
```
beervoley, line 34: nit: could you please fix the spacing here?:)
@paulcam206 @licanhua I can't remember, have we added automatic formatting for `UWP Renderer` code?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2476,6 +2476,49 @@ export class ImageSet extends CardElementContainer {
...
+    }
+
+    render(): HTMLElement | undefined {
+        let result: HTMLTrackElement | undefined = undefined;
+
+        if (this.isValid()) {
+            result = document.createElement("track");
+            result.src = this.url!;
+            result.kind = "captions";
+            result.label = this.mimeType!;
```
beervoley, line 2515: As the specs say:
image

So label should reflect the **language** of the chosen track instead of its **format**: `vtt` or `srt` (for example).

source/nodejs/adaptivecards/src/card-elements.ts
beervoley: @dclaux @licanhua @PankajBhojwani on top of that, we should be able to figure out what **language** each `captionSource` represents. We can enforce the naming of the file url to include the language like `subtitles-en.vtt` for example or add another entry to the schema to indicate the language.

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -181,22 +181,49 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
+                    for (const auto captionSource : adaptiveMedia.CaptionSources())
+                    {
+                        const auto timedTextURL = GetUrlFromString(renderContext.HostConfig(), captionSource.Url());
+                        const auto timedTextSrc = winrt::TimedTextSource::CreateFromUri(timedTextURL);
+                        timedTextSrc.Resolved(
+                            [captionSource](winrt::TimedTextSource sender, winrt::TimedTextSourceResolveResultEventArgs args)
+                            {
+                                if (!args.Error())
+                                {
+                                    args.Tracks().GetAt(0).Label(captionSource.MimeType());
```
beervoley, line 206: `Label` corresponds to user friendly name of the track - in this case the language of this subtitles track, not the **format** of the file :)
We also need to validate the `mimeType` (extension) to match one of the ones we allow like we do it for `MediaSource`: [supported mimeTypes](https://github.com/microsoft/AdaptiveCards/blob/3a35997792bd7f7b4d4deb155f21804803084e70/source/uwp/Renderer/lib/MediaHelpers.cpp#L10) and [how it's being used](https://github.com/microsoft/AdaptiveCards/blob/main/source/uwp/Renderer/lib/MediaHelpers.cpp#L150-L180)

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -215,7 +242,30 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
+                if (adaptiveMedia.CaptionSources().Size() > 0)
+                {
+                    for (const auto captionSource : adaptiveMedia.CaptionSources())
+                    {
+                        const auto timedTextURL = GetUrlFromString(renderContext.HostConfig(), captionSource.Url());
+                        const auto timedTextSrc = winrt::TimedTextSource::CreateFromUri(timedTextURL);
+                        timedTextSrc.Resolved([captionSource](winrt::TimedTextSource sender,
+                                                              winrt::TimedTextSourceResolveResultEventArgs args) {
+                                if (!args.Error()){
+                                    args.Tracks().GetAt(0).Label(captionSource.MimeType());
```
beervoley, line 255: Same as previous comment about the label :)

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: Yes, it would be good to introduce a base `ContentSource` class and and have both `MediaSource` and the new `CaptionSource` classes extend it, only overriding `getSchemaKey` and `render`.

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: @beervoley the spec being referred to up to this PR doesn't mention anything like that. That spec dates back from 2020. Someone should probably revise it and get that revision reviewed before the feature is implemented?

source/shared/cpp/ObjectModel/CaptionSource.h
PankajBhojwani: Good call! Removed

source/shared/cpp/ObjectModel/CaptionSource.h
PankajBhojwani: Removed

source/nodejs/adaptivecards/src/card-elements.ts
beervoley: @dclaux I agree :) @paulcam206 may be it's a good idea for Anna to look at if it's only adding the language?:)

source/uwp/Renderer/dll/CppWinRTIncludes.h
beervoley: for some reason it looks fine now, my bad :)

source/shared/cpp/ObjectModel/CaptionSource.h
PankajBhojwani: Makes sense, done!

source/uwp/AdaptiveCardsObjectModel/idl/AdaptiveCards.ObjectModel.Uwp.idl
PankajBhojwani: Yep, done

source/nodejs/adaptivecards/src/card-elements.ts
PankajBhojwani: Makes sense! Done

source/nodejs/adaptivecards/src/card-elements.ts
PankajBhojwani: Added a language property, and label gets set to that now

source/uwp/Renderer/lib/MediaHelpers.cpp
PankajBhojwani: Added a language property, and label gets set to that now

source/nodejs/adaptivecards/src/card-elements.ts
PankajBhojwani: I added a language property, and `label` gets set to that now. `kind` is set to `captions` and according to the [`track` docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track), `srclang` only needs to be set if `kind` is `subtitles`, but since we are implementing `captions` here, not `subtitles`, we don't need to provide `srclang`. Resolving this thread

source/uwp/Renderer/lib/MediaHelpers.cpp
PankajBhojwani: `Label` is now set to the language property, and we validate the `mimeType` now

source/uwp/Renderer/lib/MediaHelpers.cpp
PankajBhojwani: yep we don't need it, removed!

dclaux (changes requested):

dclaux:
@PankajBhojwani what about designer support?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2505,10 +2505,10 @@ export abstract class ContentSource extends SerializableObject {
 export class CaptionSource extends ContentSource {
     //#region Schema
 
-    static readonly languageProperty = new StringProperty(Versions.v1_1, "lang");
+    static readonly labelProperty = new StringProperty(Versions.v1_1, "label");
```
dclaux, line 2508: Still needs to be mapped to v1_6, not v1_1

dclaux (changes requested):

dclaux (changes requested):

source/nodejs/adaptivecards/src/card-elements.ts
PankajBhojwani: Thanks for the catch! Fixed

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2774,12 +2821,20 @@ export class Media extends CardElement {
         "sources",
         MediaSource
     );
+    static readonly captionSourcesProperty = new SerializableObjectCollectionProperty(
+        Versions.v1_1,
```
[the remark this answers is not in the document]
PankajBhojwani: Fixed

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveCaptionSource.cpp
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#include "pch.h"
+#include "AdaptiveCaptionSource.h"
+#include "AdaptiveCaptionSource.g.cpp"
+
+namespace winrt::AdaptiveCards::ObjectModel::Uwp::implementation
+{
+    AdaptiveCaptionSource::AdaptiveCaptionSource(const std::shared_ptr<::AdaptiveCards::CaptionSource>& sharedCaptionSource)
```
beervoley, line 9:
```suggestion
    AdaptiveCaptionSource::AdaptiveCaptionSource(std::shared_ptr<::AdaptiveCards::CaptionSource> const& sharedCaptionSource)
```
to be consistent :)

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveContentSource.cpp
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#include "pch.h"
+#include "AdaptiveContentSource.h"
+
+namespace winrt::AdaptiveCards::ObjectModel::Uwp::implementation
+{
+    void AdaptiveContentSource::InitializeContentSource(const std::shared_ptr<::AdaptiveCards::ContentSource>& sharedContentSource)
```
beervoley, line 8:
```suggestion
    void AdaptiveContentSource::InitializeContentSource(std::shared_ptr<::AdaptiveCards::ContentSource> const& sharedContentSource)
```
to be consistent :)

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2614,6 +2653,14 @@ export class HTML5MediaPlayer extends MediaPlayer {
             Utils.appendChild(this._mediaElement, renderedSource);
         }
 
+        for (const captionSource of this.owner.captionSources) {
+            if (captionSource.mimeType == "vtt") {
```
beervoley, line 2657: In UWP you've included:
image
maybe `srt` should be supported in JS too?:)

PankajBhojwani: JS [does not support `srt`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track)!

dclaux: Then should we really support it in UWP?

beervoley: > JS [does not support `srt`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track)!

my apologies :) had no idea :)

beervoley: > Then should we really support it in UWP?

It's a good question :)
I'm pretty certain we can easily convert `srt` to `vtt` with a simple routine (so we can support both in JS) but not certain if it's worth doing :)

PankajBhojwani:
> @PankajBhojwani what about designer support?

@dclaux I assume you mean for the 'Element Properties' pane (like how you can add a `MediaSource` from that pane) - I looked into this and the designer currently does not have the functionality to support this because `CaptionSource` has 3 name/value pairs and the `NameValuePairPropertyEditor`, which renders the section for the `MediaSource` in the pane, is hardcoded to only allow 2 name/value pairs. Since @carlos-zamora and I are going to be writing up a document on the various aspects of the designer that need an overhaul anyway, I will add this to that document.

dclaux:
@PankajBhojwani the designer does have support for this. It has support for the Media Sources, so it can support the Caption Sources as well

PankajBhojwani:
> @PankajBhojwani the designer does have support for this. It has support for the Media Sources, so it can support the Caption Sources as well

It does not have support for 3 text boxes in a row right now (which is what `CaptionSource` would need).

dclaux:
@PankajBhojwani ah, my mistake, I forgot about the addition of the `label` property. Yet, adding designer support isn't really optional... It's not that hard to build a new property editor that would allow for the entry of all three properties.

dclaux:
To be clear, every single feature should have designer support. There is only one exception, namely RichTextBlock simply because it's a very complicated and expensive thing.

source/uwp/Renderer/lib/MediaHelpers.cpp
PankajBhojwani: Thanks for pointing that out! Added

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -194,17 +194,55 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
+                            timedTextSrc.Resolved(
+                                [captionSource](winrt::TimedTextSource const& /*sender*/,
+                                                winrt::TimedTextSourceResolveResultEventArgs const& args)
+                                {
+                                    if (!args.Error())
+                                    {
+                                        args.Tracks().GetAt(0).Label(captionSource.Label());
+                                    }
+                                });
+                            mediaSrc.ExternalTimedTextSources().Append(timedTextSrc);
```
beervoley, line 216: We repeat this code twice: here and in case resolver is not null.
Could you please create a variable to hold this lambda before this logic and just pass it into `.Resolved()` instead?
I don't think it's worth extracting this as a method but lambda in a var will probably be fine :)

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -179,24 +179,105 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
+                              winrt::MediaSource const& mediaSrc)
+    {
+        if (adaptiveMedia.CaptionSources().Size() > 0)
+        {
+            for (const auto captionSource : adaptiveMedia.CaptionSources())
+            {
+                const auto currentCaptionType = captionSource.MimeType();
+                for (uint32_t i = 0; i < std::size(supportedCaptionTypes); i++)
+                {
+                    if (currentCaptionType == supportedCaptionTypes[i])
```
beervoley, line 194:
```suggestion
                for (auto&& supportedCaptionType: supportedCaptionTypes)
                {
                    if (currentCaptionType == supportedCaptionType)
```

source/uwp/Renderer/lib/MediaHelpers.cpp
PankajBhojwani: Sounds good

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -179,24 +179,97 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
+                        }
+
+                        const auto timedTextSrcResolvedHelper = [captionSource](winrt::TimedTextSource const& /*sender*/,
+                                                                                winrt::TimedTextSourceResolveResultEventArgs const& args)
+                                {
+                                    if (!args.Error())
+                                    {
+                                        args.Tracks().GetAt(0).Label(captionSource.Label());
+                                    }
+                                };
```
beervoley, line 211: I'm not too sure if this will leak. I supposed not due to **timedTextSrcResolvedHelper** getting destroyed when both this function has finished and **timedTextSrc** is destroyed.
But to be safe:
```suggestion
                        const auto timedTextSrcResolvedHelper = [label = captionSource.Label()](winrt::TimedTextSource const& /*sender*/,
                                                                                winrt::TimedTextSourceResolveResultEventArgs const& args)
                                {
                                    if (!args.Error())
                                    {
                                        args.Tracks().GetAt(0).Label(label);
                                    }
                                };
```

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -179,24 +179,97 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
+                            const auto timedTextSrc = winrt::TimedTextSource::CreateFromUri(timedTextURL);
+                            timedTextSrc.Resolved(timedTextSrcResolvedHelper);
+                            mediaSrc.ExternalTimedTextSources().Append(timedTextSrc);
+                        }
+                        else
+                        {
+                            auto args = winrt::make<winrt::implementation::AdaptiveCardGetResourceStreamArgs>(timedTextURL);
+                            auto getResourceStreamOperation = resourceResolver.GetResourceStreamAsync(args);
+                            getResourceStreamOperation.Completed(
+                                [mediaSrc, captionSource, timedTextSrcResolvedHelper](winrt::IAsyncOperation<winrt::IRandomAccessStream> operation,
```
beervoley, line 223: probably no need to capture `captionSource` anymore :)

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2567,6 +2599,10 @@ export class HTML5MediaPlayer extends MediaPlayer {
                 }
             }
         }
+
+        for (const captionSource of this.owner.captionSources) {
+            this._captionSources.push(captionSource);
+        }
```
paulcam206, line 2613:
```suggestion
        this._captionSources.push(...captionSource);
```

source/uwp/Renderer/lib/MediaHelpers.cpp
PankajBhojwani: Yep, thanks for catching that

source/uwp/Renderer/lib/MediaHelpers.cpp
PankajBhojwani: makes sense

PankajBhojwani:
Created a new property editor so `CaptionSource` can be edited in the "Element Properties" pane. This property editor accepts a variable number of properties and `MediaSource` has also been updated to use this new property editor because I believe it was only using the previous one out of convenience - `url`/`mimeType` are not actually a key/value pair but the property editor it was using previously is meant for key/value pairs like in `FactSet`.

This way if we ever add more properties to `MediaSource`/`CaptionSource` we will get the additional text boxes automatically.

source/nodejs/adaptivecards/src/card-elements.ts
paulcam206: Each platform should declare support for every common sidecar caption format it understands.

If a card author has a variety of caption formats at their disposal, they should add them all as caption sources. 

Each platform renderer should select a caption source that it supports from the author-supplied options (and should be encouraged to select the richest format it can).

This also enables scenarios where someone is interested in an accessible media story, but isn't interested in going cross-platform (e.g. IT app dev has a .wmv and a .srt file on a Sharepoint and wants to author a card to show the video in their Windows-only UWP LOB app -- they shouldn't have to transcode the video, convert the captions file, and reupload to Sharepoint in order to render a card). We should, if possible, document which formats are supported on each platform, but I wouldn't block this PR on that.

paulcam206: Also, we should definitely avoid doing any file format conversions if at all possible -- just as a general rule.

source/nodejs/adaptivecards/src/card-elements.ts
PankajBhojwani: _what is this sorcery_

source/nodejs/adaptivecards/src/card-elements.ts
PankajBhojwani: Okay, so uwp will support both `vtt` and `srt`, JS will only support `vtt`. Resolving this thread

source/nodejs/adaptivecards/src/card-elements.ts
beervoley: > _what is this sorcery_

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

PankajBhojwani: haha thanks yeah I looked it up too, was mainly expressing my dumbfoundedness that this is a thing

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -962,6 +962,123 @@ class NameValuePairPropertyEditor extends PropertySheetEntry {
...
+            messageTextBlock.text = this.messageIfEmpty;
+
+            result.addItem(messageTextBlock);
+        }
+        else {
+            for (let i = 0; i < innerPropertiesList.length; i++) {
+                let columnSet = new Adaptive.ColumnSet();
+                columnSet.spacing = Adaptive.Spacing.Small;
+
+                for (let j = 0; j < this.innerPropertiesLength; j++) {
```
dclaux, line 1010: Problem here is that you're going to have three (or more) input on one single line, that doesn't scale given the restricted space available in the "Element Properties" pane. We need to have logic that switches to a better layout when there are more than two inputs, something like the below:
```json
{
    "type": "ColumnSet",
    "separator": true,
    "columns": [
        {
            "type": "Column",
            "width": "stretch",
            "items": [
                {
                    "type": "Input.Text",
                    "placeholder": "Label for first input"
                },
                {
                    "type": "Input.Text",
                    "placeholder": "Label for second input"
                },
                {
                    "type": "Input.Text",
                    "placeholder": "Label for thrid input"
                }
            ]
        },
        {
            "type": "Column",
            "width": "auto",
            "items": [
                {
                    "type": "ActionSet",
                    "actions": [
                        {
                            "type": "Action.OpenUrl",
                            "title": "X"
                        }
                    ]
                }
            ],
            "verticalContentAlignment": "Center"
        }
    ]
}
```

which renders like this:
image

dclaux (changes requested):

source/nodejs/adaptivecards/src/card-elements.ts
dclaux: I'm responsible for more occurrences of this type of `for` loop in the codebase... maybe at some point we could change them all to using the spread operator. For that matter I wonder if it has a performance benefit.

source/nodejs/adaptivecards-designer/src/designer-peers.ts
PankajBhojwani: I've changed it to always render vertically (screenshot of what it looks like below). That way we don't need to render differently based on the number of properties. 

sources

dclaux: I suppose that's fine. Thanks for adding designer support!

beervoley (approved):

beervoley:
@PankajBhojwani amazing job!!

source/nodejs/adaptivecards/src/card-elements.ts
beervoley: don't forget to remove test pieces :)

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -2424,16 +2544,24 @@ export class ImagePeer extends TypedCardElementPeer<Adaptive.Image> {
...
         "sources",
-        "url",
-        "mimeType",
-        (name: string, value: string) => { return new Adaptive.MediaSource(name, value); },
-        "URL",
-        "MIME type",
+        ["url", "mimeType"],
+        (innerProperties: string[]) => { return new Adaptive.MediaSource(innerProperties[0], innerProperties[1]); },
+        ["URL", "MIME type"],
+        2,
```
paulcam206, line 2553: I think this can be simplified - check out this (non-specific) [snippet I was toying with](https://www.typescriptlang.org/play?#code/C4TwDgpgBAsiAKBDAlgJygXigbwFBQKgDtEBbCALigGdhVkiBzAGn0IDdEAbAVwgH4qtek1wBfXLlCRYIAEKJGmHGwIBtEuSF0GSgLpU4SNOMlcIwKADMA9jaiH5i5XkJQARD1Rd3VV28JNSg8AVQAlABl3VSgxVjd3UmRyABVwCF8VAMCyYPcYAEkYAFEoNMhotwkxAG5JWxs1T293PQA6Tl5oLHcAC2BgMGoKAHoR6htyYF7ddzrcAHkAIwArCABjYDaAawgQagAKBoBKNttUYsR13oPdkEwAPih1myIJ8zauG0YDgAMAEmwdzEVEBDTUd3aQTEUAOnT4oOw4MhHW4fCg-H4HiIPC4PjEx1+x2OdSAA)

paulcam206 (changes requested):
overall, looking good!

source/shared/cpp/ObjectModel/CaptionSource.cpp
```diff
@@ -0,0 +1,32 @@
...
+{
+    m_label = value;
+}
+
+std::shared_ptr<CaptionSource> CaptionSourceParser::Deserialize(ParseContext&, const Json::Value& json)
+{
+    std::shared_ptr<CaptionSource> captionSource = std::make_shared<CaptionSource>();
+
+    captionSource->SetMimeType(ParseUtil::GetString(json, AdaptiveCardSchemaKey::MimeType, false));
+    captionSource->SetUrl(ParseUtil::GetString(json, AdaptiveCardSchemaKey::Url, false));
```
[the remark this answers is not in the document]
PankajBhojwani: So do we also want a `ContentSourceParser`? Or will it just be `ContentSource::Deserialize`?

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveMediaSource.h
```diff
@@ -2,21 +2,19 @@
...
+        : AdaptiveMediaSourceT<AdaptiveMediaSource, ITypePeek>, AdaptiveContentSource
     {
         AdaptiveMediaSource(std::shared_ptr<::AdaptiveCards::MediaSource> const& sharedMediaSource =
                                 std::make_shared<::AdaptiveCards::MediaSource>());
 
-        property<hstring> MimeType;
-        property<hstring> Url;
-
-        virtual std::shared_ptr<::AdaptiveCards::MediaSource> GetSharedModel();
+        virtual std::shared_ptr<::AdaptiveCards::ContentSource> GetSharedModel() override;
```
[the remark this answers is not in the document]
PankajBhojwani: Then it doesn't work as an `override` anymore, I guess we could just remove it from `ContentSource` since that will never be called anyway?

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveMediaSource.cpp
```diff
@@ -8,11 +8,10 @@ namespace winrt::AdaptiveCards::ObjectModel::Uwp::implementation
 {
     AdaptiveMediaSource::AdaptiveMediaSource(const std::shared_ptr<::AdaptiveCards::MediaSource>& sharedMediaSource)
     {
-        MimeType = UTF8ToHString(sharedMediaSource->GetMimeType());
-        Url = UTF8ToHString(sharedMediaSource->GetUrl());
+        InitializeContentSource(sharedMediaSource);
     }
 
-    std::shared_ptr<::AdaptiveCards::MediaSource> AdaptiveMediaSource::GetSharedModel()
+    std::shared_ptr<::AdaptiveCards::ContentSource> AdaptiveMediaSource::GetSharedModel()
```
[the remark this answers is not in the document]
PankajBhojwani: See comment below about this being `override`, resolving this and we can continue the thread there

source/uwp/AdaptiveCardsObjectModel/lib/AdaptiveCaptionSource.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+namespace winrt::AdaptiveCards::ObjectModel::Uwp::implementation
+{
+AdaptiveCaptionSource::AdaptiveCaptionSource(std::shared_ptr<::AdaptiveCards::CaptionSource> const& sharedCaptionSource)
+    {
+        Label = UTF8ToHString(sharedCaptionSource->GetLabel());
+        InitializeContentSource(sharedCaptionSource);
+    }
+
+    std::shared_ptr<::AdaptiveCards::ContentSource> AdaptiveCaptionSource::GetSharedModel()
```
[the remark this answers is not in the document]
PankajBhojwani: Same thing here as for `MediaSource` where this is an `override`, resolving this to continue the discussion there

source/uwp/Renderer/lib/MediaHelpers.cpp
```diff
@@ -179,24 +179,97 @@ namespace AdaptiveCards::Rendering::Uwp::MediaHelpers
...
+                                    {
+                                        // Get the random access stream
+                                        if (const auto randomAccessStream = operation.GetResults())
+                                        {
+                                            auto timedTextSrc = winrt::TimedTextSource::CreateFromStream(randomAccessStream);
+                                            timedTextSrc.Resolved(timedTextSrcResolvedHelper);
+                                            mediaSrc.ExternalTimedTextSources().Append(timedTextSrc);
+                                        }
+                                    }
+                                });
```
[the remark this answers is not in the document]
PankajBhojwani: I tried this, (calling `randomAccessStream.get()` since `randomAccessStream` right now is just an `IAsyncOperation`), and I get this compiler error:

`'promise_type': is not a member of 'std::experimental::coroutine_traits<void,const winrt::Windows::UI::Xaml::Controls::MediaElement &,const winrt::AdaptiveCards::ObjectModel::Uwp::AdaptiveMedia &,const winrt::AdaptiveCards::Rendering::Uwp::AdaptiveRenderContext &,const winrt::Windows::Media::Core::MediaSource &>'`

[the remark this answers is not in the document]
PankajBhojwani: I tried adding `#include <winrt/Windows.Foundation.h>` to `CppWinRTIncludes.h` (guidance from [this article](https://devblogs.microsoft.com/oldnewthing/20210809-00/?p=105539)) and that did not fix it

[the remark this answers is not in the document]
beervoley: @paulcam205 we can't use `co_await` since we're not inside a co routine. We can use a blocking `.get()`.
@PankajBhojwani I've just tried and `.get()` call compiles fine without any issue:)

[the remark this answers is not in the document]
PankajBhojwani: Ah yes removing the `co_await` and adding `get` works, thanks!

source/nodejs/adaptivecards-designer/src/designer-peers.ts
PankajBhojwani: That's a lot better, implemented! Thank you

source/shared/cpp/ObjectModel/CaptionSource.cpp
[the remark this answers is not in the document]
paulcam206: derp - yeah, forgot the pattern here :)

`CaptionSourceParser::Deserialize()` should be calling `ContentSource::Deserialize()` to handle inherited properties... 

see e.g.

https://github.com/microsoft/AdaptiveCards/blob/5e0f39fe414829e776a101b76ea2fa0045b5a7c1/source/shared/cpp/ObjectModel/TextInput.h#L11

https://github.com/microsoft/AdaptiveCards/blob/5e0f39fe414829e776a101b76ea2fa0045b5a7c1/source/shared/cpp/ObjectModel/TextInput.h#L56

https://github.com/microsoft/AdaptiveCards/blob/5e0f39fe414829e776a101b76ea2fa0045b5a7c1/source/shared/cpp/ObjectModel/TextInput.cpp#L129-L136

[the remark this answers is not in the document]
paulcam206: derp - yeah, forgot the pattern here :)

`CaptionSourceParser::Deserialize()` should be calling `ContentSourceParser::Deserialize<CaptionSource>()` to handle inherited properties... 

see e.g.

https://github.com/microsoft/AdaptiveCards/blob/5e0f39fe414829e776a101b76ea2fa0045b5a7c1/source/shared/cpp/ObjectModel/TextInput.h#L11

https://github.com/microsoft/AdaptiveCards/blob/5e0f39fe414829e776a101b76ea2fa0045b5a7c1/source/shared/cpp/ObjectModel/TextInput.h#L56

https://github.com/microsoft/AdaptiveCards/blob/5e0f39fe414829e776a101b76ea2fa0045b5a7c1/source/shared/cpp/ObjectModel/TextInput.cpp#L129-L136

https://github.com/microsoft/AdaptiveCards/blob/5e0f39fe414829e776a101b76ea2fa0045b5a7c1/source/shared/cpp/ObjectModel/BaseInputElement.h#L45-L54

(LMK if that doesn't make any sense) :)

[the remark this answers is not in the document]
PankajBhojwani: Thanks for the example! I followed that style (`BaseInputElement` doesn't actually have a `BaseInputElementParser`, so I didn't implement a `ContentSourceParser` either)

source/nodejs/adaptivecards/src/card-elements.ts
PankajBhojwani: Updated the comment, we will leave crossorigin enabled

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -2424,16 +2537,20 @@ export class ImagePeer extends TypedCardElementPeer<Adaptive.Image> {
...
-    static readonly sourcesProperty = new NameValuePairPropertyEditor(
+    static readonly sourcesProperty = new InnerStructPropertyEditor(
         Adaptive.Versions.v1_1,
         "sources",
-        "url",
-        "mimeType",
-        (name: string, value: string) => { return new Adaptive.MediaSource(name, value); },
-        "URL",
-        "MIME type",
+        {"url": {name: "URL"}, "mimeType": {name: "MIME Type"}},
```
licanhua, line 2543: why not {"url":  "URL", "mimeType": "MIME Type"},

licanhua, line 2543: For the mineType, I'm wondering if it's better to use combobox to avoid wrong input

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2504,6 +2500,48 @@ export class MediaSource extends SerializableObject {
...
+
+    static readonly labelProperty = new StringProperty(Versions.v1_6, "label");
+
+    @property(CaptionSource.labelProperty)
+    label?: string;
+
+    //#endregion
+
+    constructor(url?: string, mimeType?: string, label?:string) {
+        super();
```
licanhua, line 2516: super(url, mineType)

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -962,6 +962,119 @@ class NameValuePairPropertyEditor extends PropertySheetEntry {
...
+        context.target[this.collectionPropertyName] = [];
+
+        const collectionItems = innerPropertiesList.map(e => this.createCollectionItem(e));
+        context.target[this.collectionPropertyName].push(...collectionItems);
+
+        context.peer.changed(refreshPropertySheet);
+    }
+
+    render(context: PropertySheetContext): Adaptive.CardElement {
+        let result = new Adaptive.Container();
```
licanhua, line 985: possible to use const

source/nodejs/adaptivecards-designer/src/designer-peers.ts
PankajBhojwani: That does not work, it needs to be a dictionary where keys are `string`s and values are `NameAndValue`

PankajBhojwani: We did not use a combo box for `MediaSource` before, so to maintain consistency I think it is okay for `CaptionSource` to also use a text box. If we want a combo box for this, it can be added in a follow up PR I think

PankajBhojwani: Also, the `Adaptive` name space does not have a `ComboBox` at this time

source/nodejs/adaptivecards/src/card-elements.ts
PankajBhojwani: Thank you for the catch! Fixed

schemas/1.6.0/adaptive-card.json
```diff
@@ -303,6 +303,11 @@
           "description": "Specifies the background image of the card.",
           "version": "1.2"
         },
+        "metadata": {
+          "$ref": "#/definitions/Metadata",
+          "description": "Defines various metadata properties typically not used for rendering the card",
+          "version": "1.6"
+        },
```
PankajBhojwani, line 310: I am not sure how this ended up in this diff - this was added and merged in #7105

PankajBhojwani: @paulcam206 is this an issue?

paulcam206: I'll take a look :)

paulcam206 (changes requested):

schemas/1.6.0/adaptive-card.json
PankajBhojwani: Great, resolving this convo

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -962,6 +962,119 @@ class NameValuePairPropertyEditor extends PropertySheetEntry {
...
+            for (let i = 0; i < innerPropertiesList.length; i++) {
+                let columnSet = new Adaptive.ColumnSet();
+                columnSet.spacing = Adaptive.Spacing.Medium;
+                columnSet.separator = true;
+
+                let newColumn = new Adaptive.Column("stretch");
+                Object.keys(innerPropertiesList[i]).forEach(key => {
+                    let textInput = new Adaptive.TextInput();
+                    textInput.defaultValue = innerPropertiesList[i][key].value;
+                    textInput.placeholder = innerPropertiesList[i][key].name;
```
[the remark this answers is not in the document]
PankajBhojwani: I'd say it looks fine! Added the `label`

captionsources

paulcam206:
Just as a heads up -- the Android and iOS build failures are legit. 

Don't Panic, though, I'll push fixes here for you :)

PankajBhojwani:
> Just as a heads up -- the Android and iOS build failures are legit.
> 
> Don't Panic, though, I'll push fixes here for you :)

Thank you!

paulcam206 (approved):

beervoley:
@PankajBhojwani amazing job, thank you!!

PR #7199, opened 2022-04-07
[NodeJS Templating] Enable host parameters

anna-dingler:
# Description

Allow card author to supply `$host` data in addition to `$root` data. 

I added `$host` variable to `IEvaluationContext` and `TemplateObjectMemory` so that binding with `${$host.keyHere}` will search through the provided `$host` JSON.

# Sample Card

An example card can be found at: `source/nodejs/tests/unit-tests/template-test-resources/complex-template-host.json`

# How Verified

Added new unit tests to `source/nodejs/tests/unit-tests` for the templating library.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7199)


source/nodejs/tests/unit-tests/src/unit-tests.test.ts
```diff
@@ -14,3 +17,123 @@ describe("test Text property on TextBlock", () => {
...
+					"text": "Hello Adaptive Cards! You're using the Light theme!" 
+				} 
+			]
+		};
+
+		expect(card).toStrictEqual(expectedOutput);
+	});
+
+	it("ComplexTemplate", () => {
+		const templateObject = fs.readFileSync("template-test-resources/complex-template.json", "utf8");
```
licanhua, line 98: Nit: It's better to refactor the test cases and remove duplication.

For example
```
const testData = [{template: "complex-template.json", expect: "complex-template.output.json"},  ];

for (test of testData) {
    runTest(test);
}
```

or for input/output from files 
```
runTest(templateFile, outputFile....)
```

or even more generic:
```
runTest(loadFile(templateFile), loadFile(outputFile)....)
```

source/nodejs/tests/unit-tests/src/unit-tests.test.ts
```diff
@@ -14,3 +17,99 @@ describe("test Text property on TextBlock", () => {
         expect('{"type":"TextBlock","text":"Some text"}').toBe(JSON.stringify(textBlockJson));
     });
 });
+
+describe("Test Templating Library", () => {
+    it("BasicTemplate", () => {
```
paulcam206, line 22: love these tests - I think I'd like to see a few additional scenarios covered:

* what happens if I pass `[ "foo", {"bar": "baz"}]` as my `$host`?
* how about `"just some string"`

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -115,7 +118,12 @@ export interface IEvaluationContext {
      * The root data object the template will bind to. Expressions that refer to $root in the template payload
      * map to this field. Initially, $data also maps to $root.
      */
-    $root: any
+    $root: any;
+    /**
+     * The host data object the template will bind to. Expressions that refer to $host in the template payload
+     * map to this field.
+     */
```
paulcam206, line 125:
```suggestion
    /**
     * The host data object the template will bind to. Expressions that refer to $host in the template payload
     * map to this field -- this allows a host process to supply additional context to the template.
     */
```

source/nodejs/tests/unit-tests/src/unit-tests.test.ts
```diff
@@ -1,4 +1,7 @@
 import * as AdaptiveCards from "adaptivecards";
+import * as ACData from "adaptivecards-templating";
+
+var fs = require("fs");
```
paulcam206, line 4: not sure if the `unit-tests` project is set up correctly, but you may be able to do:

```suggestion
import { readFileSync } from "fs";
```

(and then refer to it as just `readFileSync` in `loadFile()` below)

if it doesn't work, feel free to leave as-is :)

paulcam206 (changes requested):
looks great! just found a few minor things 👍

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -115,7 +118,12 @@ export interface IEvaluationContext {
...
+    $host?: any;
```
paulcam206, line 126: Forgot to mention -- with this addition, we should probably bump the `adaptivecards-templating` `package.json` version by one :)

anna-dingler: Bump major or minor?

licanhua (approved):

source/nodejs/tests/unit-tests/src/unit-tests.test.ts
anna-dingler: To clarify, $host and $root should only accept valid JSON, correct?

Also, did we want to accept host data without root data - don't remember what we decided on the call :)

anna-dingler: @paulcam206 I added the following tests

- BasicTemplate, only provide host data
- Host data given as string
- Host data given as array

I also went ahead and made both $host and $root optional - I can change this back if we want

source/nodejs/adaptivecards-designer-app/package.json
```diff
@@ -20,7 +20,7 @@
 	"dependencies": {
 		"adaptivecards": "^3.0.0-beta.1",
 		"adaptivecards-designer": "^2.4.0-alpha.0",
-		"adaptivecards-templating": "^2.3.0-alpha.0",
```
licanhua, line 23: revert this change. we never publish 2.3.0-alpha. you can check it from https://www.npmjs.com/package/adaptivecards-templating

anna-dingler: thanks for pointing this out! just reverted the change

almedina-ms (approved):

paulcam206 (approved):

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -3,17 +3,19 @@
 import * as AEL from "adaptive-expressions";
 
 class EvaluationContext {
```
licanhua, line 5: Please update the doc too. You can create a bug and separate PR for the document

PR #7204, opened 2022-04-08
[UWP Nuget]: Add ObjectModel .nuspec

beervoley:
Changes added to pack `ObjectModel` UWP nuget as well as adding a dependency for the `Renderer` on `ObjectModel` were not propagated from https://github.com/microsoft/AdaptiveCards/pull/6576 to the main. This PR fixes it.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7204)


source/uwp/NuGet/AdaptiveCards.Rendering.Uwp.nuspec
```diff
@@ -5,13 +5,16 @@
...
-        <description>Used to render Adaptive Cards into UWP XAML or to work with the object model for C++ projects</description>
+        <description>Used to render Adaptive Cards into UWP XAML</description>
         <requireLicenseAcceptance>false</requireLicenseAcceptance>
         <tags>adaptivecards;adaptive-cards</tags>
         <projectUrl>https://adaptivecards.io</projectUrl>
         <iconUrl>https://adaptivecards.io/content/icons_blue/blue-48.png</iconUrl>
         <license type="file">EULA-Windows.txt</license>
         <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+        <dependencies>
+            <dependency id="AdaptiveCards.ObjectModel.Uwp" version="1.0.0" />
```
carlos-zamora, line 16: would we need to change the version here? Since the object model has probably changed since it was last published

beervoley: new object model was not published yet :)
once the new release will be out, we'll update the verison

almedina-ms (approved):

source/uwp/NuGet/AdaptiveCards.ObjectModel.Uwp.targets
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <Target Name="CheckNonUwpApp" Condition="'$(TargetPlatformIdentifier)' != 'UAP'" BeforeTargets="PrepareForBuild">
+        <Error Text="Only UWP apps/libraries are supported." />
+    </Target>
+    <ItemGroup Condition="'$(ProjectExt)' != '.vcxproj'">
+        <SDKReference Include="Microsoft.VCLibs, Version=14.0">
```
licanhua, line 7: I may be cautious about this change. instead, I may warn user with error if possible. 

The reason for it is that For C# project, it may depend on Microsoft.VCLibs.140.00.Debug or than Microsoft.VCLibs

paulcam206 (approved):

source/uwp/NuGet/AdaptiveCards.ObjectModel.Uwp.targets
paulcam206: good callout here -- we should keep an eye out for folks running in to issues, but probably don't have cycles to investigate the possibility at the moment...

Issue #5888, opened 2021-05-27
Create UWP rendering packages for WinUI2 and WinUI3


RebeccaAnne:

paulcam206:
should probably split this out into two issues covering WinUI2 and WinUI3

PR #7205, opened 2022-04-08
[UWP] Ingest WinUI 2.7

carlos-zamora:
# Related Issue

#5888 
- [X] WinUI 2
- [ ] WinUI 3

# Description

Adds a dependency on WinUI 2.7.1 so that the UWP renderer gets the updated visual styling. These changes include...
- updating the dependencies on the published nuget package
- bumping up the minimum version to 10.0.18362.0
- modifying the namespaces a bit to take advantage of `Microsoft::UI::Xaml` as much as possible

On the UWP visualizer side, the project now also takes a dependency on WinUI 2.7.1. This makes it so that it can just keep a project reference to the UWP renderer as opposed to taking a reference to the NuGet package (which allows for a better debugging experience when using it). As a bonus, the WinUI 2 visuals were applied.

# How Verified

Verification of...
- Ingestion of WinUI 2
   - [X] UWP Renderer builds
   - [X] UWP Renderer can output a NuGet package
   - [X] Visualizer can consume that NuGet package
   - [x] Visualizer displays updated controls on rendered adaptive card and the visualizer surface itself

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7205)


carlos-zamora:
Two things I've noticed that I need to fix on Monday:
1. Rounded corners
   - **Problem**: The new UI doesn't completely take advantage of rounded corners, which means that we're not inheriting the new styling.
   - **Fix**: I need to update the built-in styles to be `BasedOn` the `DefaultButtonStyle`. This way, we directly inherit things like rounded colors and (hopefully) background/foreground color.
2. Color contrast
   - **Problem**: The a11y bug is specifically filed on insufficient contrast between the button and background color. That's still not solved!
   - **Fix**: Aside from just using `BasedOn`, I'll have to be a bit more clever about how the styling fallback occurs. This may require a few code changes in `ActionHelpers.h`'s `HandleActionStyling()`.

carlos-zamora:
## Notes to reviewer

### Why WinUI 2.7.1 and not WinUI 2.8?
WinUI 2.8 takes a dependency on WebView2. Need to discuss with the team if that's something we want as well. In the meantime, WinUI 2.7 should definitely suffice.

### Where's WinUI 3?
That's coming soon. Most of the changes to the namespace are actually in preparation of that. I want to make sure I have WinUI 2 working well before moving on to WinUI 3. Depending on how big of a diff ingesting WinUI 3 is, I might need to split this up into two PRs with the following base branch structure: `main` <-- WinUI 2 <-- WinUI 3.

source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
```diff
@@ -44,7 +44,7 @@
     <AppContainerApplication>true</AppContainerApplication>
     <ApplicationType>Windows Store</ApplicationType>
     <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
-    <WindowsTargetPlatformMinVersion>10.0.15063.0</WindowsTargetPlatformMinVersion>
+    <WindowsTargetPlatformMinVersion>10.0.18362.0</WindowsTargetPlatformMinVersion>
```
licanhua, line 47: this doesn't sound right. Are you trying to set WindowsTargetPlatformVersion to 10.0.18362.0?

source/uwp/Renderer/dll/CppWinRTIncludes.h
```diff
@@ -10,22 +10,31 @@
...
-#include <winrt/Windows.UI.Xaml.Documents.h>
-#include <winrt/Windows.UI.Xaml.Media.Imaging.h>
 #include <winrt/Windows.Data.Json.h>
 #include <winrt/Windows.Foundation.h>
 #include <winrt/Windows.Foundation.Collections.h>
 #include <winrt/Windows.Data.Xml.Dom.h>
 #include <winrt/AdaptiveCards.ObjectModel.Uwp.h>
+
 #include <winrt/Windows.UI.Text.h>
+#include <winrt/Windows.UI.Xaml.Automation.h>
```
licanhua, line 20: Is there reason you moved #include <winrt/Windows.UI.Xaml.Automation.h> and others from line 12 to line 20?

source/uwp/NuGet/AdaptiveCards.Rendering.Uwp.nuspec
```diff
@@ -12,6 +12,10 @@
         <iconUrl>https://adaptivecards.io/content/icons_blue/blue-48.png</iconUrl>
         <license type="file">EULA-Windows.txt</license>
         <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+        <dependencies>
```
licanhua, line 15: There should be a major version change on Version somewhere. Possible version 3.1.0

licanhua: https://www.nuget.org/packages/AdaptiveCards.Rendering.Uwp/

source/uwp/Renderer/lib/AdaptiveCardRendererComponent.cpp
```diff
@@ -107,6 +107,7 @@ namespace winrt::AdaptiveCards::Rendering::Uwp::implementation
                     ::AdaptiveCards::Rendering::Uwp::XamlBuilder::BuildXamlTreeFromAdaptiveCard(adaptiveCard,
                                                                                                 *renderContext,
                                                                                                 m_xamlBuilder.get());
+                xamlTreeRoot.Resources().MergedDictionaries().Append(winrt::XamlControlsResources());
```
licanhua, line 110: we should not do this in the library. instead, we should make app to do it.

Can you connect with react-native-windows team to see how do they achieved this?
https://github.com/microsoft/react-native-windows

source/uwp/Visualizer/AdaptiveCardVisualizer.csproj
```diff
@@ -11,8 +11,8 @@
     <AssemblyName>XamlCardVisualizer</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.17763.0</TargetPlatformVersion>
-    <TargetPlatformMinVersion>10.0.16299.0</TargetPlatformMinVersion>
+    <TargetPlatformVersion>10.0.18362.0</TargetPlatformVersion>
+    <TargetPlatformMinVersion>10.0.18362.0</TargetPlatformMinVersion>
```
licanhua, line 15: Don't change the minVersion unless it's required

source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
carlos-zamora: We need to bump the minimum version to ingest WinUI. See here: https://github.com/Microsoft/microsoft-ui-xaml#version-support

That said, it looks like I should correct the change to be this:
```
<WindowsTargetPlatformVersion>10.0.18362.0</WindowsTargetPlatformVersion>
<WindowsTargetPlatformMinVersion>10.0.15063.0</WindowsTargetPlatformMinVersion>
```

source/uwp/Renderer/dll/CppWinRTIncludes.h
carlos-zamora: Yes! It's just not relevant (yet) haha. All the changes in this file are intended to make my life easier when ingesting WinUI 3. Depending on how big of a change that is, that may get lumped into this PR or a separate one that targets this one (and takes advantage of these changes.)

source/uwp/NuGet/AdaptiveCards.Rendering.Uwp.nuspec
carlos-zamora: Makes sense. @beervoley any idea where that is?

source/uwp/Visualizer/AdaptiveCardVisualizer.csproj
carlos-zamora: Same as this comment above: https://github.com/microsoft/AdaptiveCards/pull/7205#discussion_r848703347

I assume changing the version on the visualizer is more ok though right? Because it's an internal tool, really.

source/uwp/Renderer/dll/CppWinRTIncludes.h
```diff
@@ -10,22 +10,31 @@
...
+#include <winrt/Windows.UI.Xaml.Documents.h>
+#include <winrt/Windows.UI.Xaml.Input.h>
```
asklar, line 22: you will definitely want to put all these includes in a PCH if you aren't using it yet, otherwise your compile times will be bad

source/uwp/Renderer/dll/CppWinRTIncludes.h
```diff
@@ -34,18 +43,21 @@ namespace winrt
...
-    using namespace ::winrt::Windows::UI;
-	using namespace ::winrt::Windows::UI::Xaml;
-    using namespace ::winrt::Windows::UI::Xaml::Controls;
-    using namespace ::winrt::Windows::UI::Xaml::Controls::Primitives;
-    using namespace ::winrt::Windows::UI::Xaml::Automation;
-	using namespace ::winrt::Windows::UI::Xaml::Automation::Peers;
+    using namespace ::winrt::AdaptiveCards::ObjectModel::Uwp;
+
+    // WinUI 2
+    using namespace ::winrt::Microsoft::UI;
```
asklar, line 49: in general avoid using namespace in header files as it pollutes. Instead create aliases (E.g. `namespace xaml = winrt::Windows::UI::Xaml` etc.)

source/uwp/Renderer/dll/CppWinRTIncludes.h
```diff
@@ -34,18 +43,21 @@ namespace winrt
...
+	using namespace ::winrt::Microsoft::UI::Xaml;
```
asklar, line 50: it would be good to standardize this straddling of system xaml and winui3 across our UI frameworks as I expect others need this too. I've done this already here: https://github.com/asklar/xamlstraddle

source/uwp/Renderer/dll/CppWinRTIncludes.h
carlos-zamora: `CppWinRTIncludes.h` is only included in the PCH. That accomplishes the same thing right?

carlos-zamora:
> 1. Rounded corners
>    
>    * **Problem**: The new UI doesn't completely take advantage of rounded corners, which means that we're not inheriting the new styling.
>    * **Fix**: I need to update the built-in styles to be `BasedOn` the `DefaultButtonStyle`. This way, we directly inherit things like rounded colors and (hopefully) background/foreground color.

Now fixed. Had to retemplate the control to take advantage of the default styling. The retemplated control XML was generated using VS, so it should be exactly what WinUI generally uses.

> 2. Color contrast
>    
>    * **Problem**: The a11y bug is specifically filed on insufficient contrast between the button and background color. That's still not solved!
>    * **Fix**: Aside from just using `BasedOn`, I'll have to be a bit more clever about how the styling fallback occurs. This may require a few code changes in `ActionHelpers.h`'s `HandleActionStyling()`.

Spoke with the accessibility team. There's two issues here:
1. the contrast should be between the button border and the surface background
2. we're not doing anything different here since we're just using the default button styling.
They're asking WinUI what the deal is there. At this point, the bug is either resolved-external to WinUI or we need to perform custom styling on the button (which doesn't sound right).

source/uwp/NuGet/AdaptiveCards.Rendering.Uwp.nuspec
beervoley: https://github.com/microsoft/AdaptiveCards/blob/main/custom.props
but it will first be updated in the release branch and then merged to main
looks like we forgot to update it last time

source/uwp/Renderer/lib/AdaptiveCardRendererComponent.cpp
beervoley: @carlos-zamora @licanhua not sure how react-native did it but in our case I assume we will want to provide some guidance on how to add `WinUI2` resources to the dictionaries we're using :)

source/uwp/Renderer/dll/CppWinRTIncludes.h
beervoley: @asklar we're following the same pattern `WinUI2` is using https://github.com/microsoft/microsoft-ui-xaml/blob/main/dev/inc/CppWinRTIncludes.h

carlos-zamora:
The latest commit has me stumped on a relatively minor issue. For whatever reason, buttons created by "Action.OpenUrl" don't adhere to the new styling. This can be seen in the images below where the "Action.Submit" buttons have rounded corners inherited from the WinUI styling, but the "Action.OpenUrl" buttons do not.

image

<details>
<summary>Action.OpenUrl Card Payload</summary>

```
{
	"type": "AdaptiveCard",
	"version": "1.0",
	"body": [
		{
			"type": "TextBlock",
			"text": "Publish Adaptive Card Schema",
			"weight": "bolder",
			"size": "medium",
			"wrap": true,
			"style": "heading"
		}
	],
	"actions": [
		{
			"type": "Action.OpenUrl",
			"title": "Default",
			"url": "https://adaptivecards.io",
"style": "default"
		},
{
			"type": "Action.OpenUrl",
			"title": "Positive",
			"url": "https://adaptivecards.io",
"style": "positive"
		},
{
			"type": "Action.OpenUrl",
			"title": "Destructive",
			"url": "https://adaptivecards.io",
"style": "destructive"
		}
	]
}

```
</details>

image

<details>
<summary>Action.Submit Card Payload</summary>

```
{
	"type": "AdaptiveCard",
	"version": "1.0",
	"body": [
		{
			"type": "TextBlock",
			"text": "Publish Adaptive Card Schema",
			"weight": "bolder",
			"size": "medium",
			"wrap": true,
			"style": "heading"
		}
	],
	"actions": [
		{
			"type": "Action.Submit",
			"title": "Default",
"style": "default"
		},
{
			"type": "Action.Submit",
			"title": "Positive",
"style": "positive"
		},
{
			"type": "Action.Submit",
			"title": "Destructive",
"style": "destructive"
		}
	]
}

```
</details>

jwoo-msft:
closing it for now.

PR #7209, opened 2022-04-11
[NodeJS Templating] Expose adaptivecards-templating version

anna-dingler:
# Description

Expose the `adaptivecards-templating` library version by adding a variable `$_acTemplateVersion` to `IEvaluationContext` and `TemplateObjectMemory`. `$_acTemplateVersion` holds a JSON with versioning info.

# Sample Card

```
{
    "type": "AdaptiveCard",
    "version": "1.0",
    "body": [
        {
            "$when": "${$_acTemplateVersion.major < 2}",
            "type": "TextBlock",
            "text": "Uhoh! Time to update your copy of `adaptivecards-templating`!"
        },
        {
            "$when": "${$_acTemplateVersion.major >= 2}",
            "type": "TextBlock",
            "text": "`adaptivecards-templating` is up to date!"
        }
    ]
}
```

# How Verified

Tested manually on AdaptiveCards designer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7209)


source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -47,6 +50,34 @@ class EvaluationContext {
     set $data(value: any) {
         this._$data = value;
     }
+
+    generateVersionJson() {
```
licanhua, line 54: This may not work. Customer may use templating like this:
```
<script src="https://unpkg.com/adaptivecards-templating/dist/adaptivecards-templating.min.js"></script>
```

You may have to hard coded it. also verify the version is matching with what in package.json in CI pipeline.

anna-dingler: `$_acTemplateVersion` is working with scripts. I tested with the following files included in an html file:

```
<script src="https://unpkg.com/markdown-it@8.4.0/dist/markdown-it.js"></script>
<script src="https://unpkg.com/adaptivecards/dist/adaptivecards.min.js"></script>
<script type="text/javascript" src="browser.js"></script>
<script type="text/javascript" src="adaptivecards-templating.min.js"></script>
```

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -47,6 +50,34 @@ class EvaluationContext {
...
+        let pkg = require('./../package.json');
+        let version = pkg.version;
```
licanhua, line 56: nit: use `const` if no change on the variable

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -47,6 +50,34 @@ class EvaluationContext {
...
+        let versionSplit = version.split('.');
+
+        let build;
```
licanhua, line 59: nit:
```
const patchSplit = versionSplit[2].split('-');
return {
            "major": parseInt(versionSplit[0]),
            "minor": parseInt(versionSplit[1]),
            "patch": parseInt(patchSplit[0]),
            "suffix": patchSplit[1]
        }
```

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -47,6 +50,34 @@ class EvaluationContext {
...
+        if (versionSplit[3]) {
+            build = versionSplit[3];
+        }
+
+        let patchSplit = versionSplit[2].split('-');
+
+        let suffix;
```
licanhua, line 66: nit: always assign default value if possible. For example `let suffix = undefined`

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -3,16 +3,19 @@
 import * as AEL from "adaptive-expressions";
 
 class EvaluationContext {
-    private static readonly _reservedFields = ["$data", "$when", "$root", "$index"];
+    private static readonly _reservedFields = ["$data", "$when", "$root", "$index", "$_acTemplateVersion"];
```
licanhua, line 6: Need to update document somewhere for new _acTemplateVersion

source/nodejs/adaptivecards-templating/src/template-engine.ts
licanhua, line 54: Add a unit test if possible

anna-dingler: @licanhua I added a basic unit test that fails if major, minor, or patch are not parsed correctly. The only other test I can think of is to directly check if all parts of the JSON are correct, but I believe that would require us to update the test whenever the version changes.

licanhua (approved):
You may update the document late. I don't know how to track it before we release this change

source/nodejs/adaptivecards-templating/src/template-engine.ts
anna-dingler: I created an issue to track updating documentation

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -49,6 +53,23 @@ class EvaluationContext {
...
+        const version = pkg.version;
+        const versionSplit = version.split('.');
+
+        let patchSplit = [];
+        if (versionSplit[2]) {
+            patchSplit = versionSplit[2].split('-');
+        }
+
+        return {
+            "major": parseInt(versionSplit[0]),
```
jwoo-msft, line 67: nit: can you add a check with a regex to ensure the whole of input string for the version is well-formed?

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -49,6 +53,23 @@ class EvaluationContext {
     set $data(value: any) {
         this._$data = value;
     }
+
+    generateVersionJson() {
+        const version = pkg.version;
+        const versionSplit = version.split('.');
+
+        let patchSplit = [];
```
jwoo-msft, line 61: nit: null, nullptr or None might be better candidates.

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -49,6 +53,23 @@ class EvaluationContext {
...
+        if (versionSplit[2]) {
+            patchSplit = versionSplit[2].split('-');
```
jwoo-msft, line 63: instead of using `2`, could you use a constant variable with descriptive name, so we can update the value in one place and understands what this value means?

jwoo-msft (changes requested):

source/nodejs/adaptivecards-templating/src/template-engine.ts
beervoley, line 63: is this to account for something like `2.3.0-alpha.0`?

source/nodejs/tests/unit-tests/template-test-resources/version-template.json
```diff
@@ -0,0 +1,21 @@
...
+        {
+            "$when": "${$_acTemplateVersion.minor >= 0}",
+            "type": "TextBlock",
+            "text": "$_acTemplateVersion.minor >= 0"
+        },
+        {
+            "$when": "${$_acTemplateVersion.patch >= 0}",
+            "type": "TextBlock",
+            "text": "$_acTemplateVersion.patch >= 0"
+        }
```
beervoley, line 19: it could be nice to have some conditions that fail as well:)

source/nodejs/adaptivecards-templating/src/template-engine.ts
anna-dingler: Yes, it is so we can separate the patch and suffix

source/nodejs/tests/unit-tests/template-test-resources/version-template.json
anna-dingler: Since we pull the version directly from `package.json`, how would we force a failing condition?

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -49,6 +53,23 @@ class EvaluationContext {
     set $data(value: any) {
         this._$data = value;
     }
+
+    generateVersionJson() {
+        const version = pkg.version;
+        const versionSplit = version.split('.');
+
+        let patchSplit = [];
+        if (versionSplit[2]) {
```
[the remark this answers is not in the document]
anna-dingler: From what I understand, typescript returns `undefined` when the index is out of bounds

source/nodejs/tests/unit-tests/template-test-resources/version-template.json
beervoley: Oh, I just mean something like
` "$when": "${$_acTemplateVersion.patch < 0}"` - to make sure that `TextBlock` is not rendered in this case :)

beervoley: @paulcam206 @jwoo-msft it would also be nice if we could set an arbitrary version in `package.json` (or some other way) to test proper handling of improperly structured version string :) but I assume that would be unlikely?

source/nodejs/adaptivecards-templating/src/template-engine.ts
anna-dingler: Is it necessary to check with a regex since we set the version ourselves?

source/nodejs/adaptivecards-templating/src/template-engine.ts
anna-dingler: Why is it better to set to null?

source/nodejs/tests/unit-tests/template-test-resources/version-template.json
anna-dingler: I just added two new unit tests :)

source/nodejs/adaptivecards-templating/src/template-engine.ts
jwoo-msft: sure, it's reasonable. but don't you want to verify that input string is in the correct order? we are assuming that input string is in the certain format and access them accordingly. 
we can have a new valid versioning format that may not confirm to this assumption in the future. 
if you have a regex check, when regex fails, other maintainers will be able to see what's expected format & quickly reason about it.
and regex check ensures that we don't get weird errors we might get from something like 'xxx.xxx.xxx.xxx' 
as i've stated, it's nit.

jwoo-msft: if you create a regex, not only that you can check the validity of the input, but you can also extract the value using it; just make sure you group them using '()'

licanhua: I think the existing logic is enough:
Keep it simple, and make it crash/test case can detect it if the version doesn't meet 2.3.0-alpha format.


nodejs version rule is a little complex and we don't need to support all of them.
https://docs.npmjs.com/cli/v6/configuring-npm/package-json#version
https://github.com/isaacs/node-semver#usage

source/nodejs/adaptivecards-templating/src/template-engine.ts
licanhua: It's typescript and I think ` let patchSplit = []` is the correct way to avoid warnings, I believe you don't want to write code like
```
let patchSplit?: string[] = null
```
or 
```
let patchSplit: string[]  | undefined = undefined
```

beervoley (approved):
:shipit:

source/nodejs/adaptivecards-templating/src/template-engine.ts
anna-dingler: Thanks for all of the versioning info! We test to make sure major, minor and patch are always parsed correctly. The only one that's a little funny is the suffix since it's optional on our end

PR #7220, opened 2022-04-20
[Shared, NodeJS] Allow `today` as a valid value for Input.Date

### LINKED ISSUE #5142: [C#] Input.Date : Default to today's date?
Is it possible to assign today's date to Input.Date control? So that whenever user open the email, the control defaults to today's date?

anna-dingler:
# Related Issue

Fixes #5142 

# Description

If the `value` of Input.Date is `"today"`, the initial date should be the current date. Implemented within nodeJS platform and shared object model (to cover UWP, Android and iOS).

I check to see if `value` is `today`, and generate the current date to assign to value if true.

# Sample Card

```
{
    "type": "AdaptiveCard",
    "version": "1.5",
    "body": [
        {
            "type": "Input.Date",
            "value": "today"
        }
    ]
}
```

# How Verified

Manually tested on adaptivecards-site, UWP Visualizer, Android mobile app, and iOS Visualizer 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7220)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4388,6 +4392,22 @@ export class DateInput extends Input {
     get value(): string | undefined {
         return this._dateInputElement ? this._dateInputElement.value : undefined;
     }
+
+    currentDate(): string {
+        const currDate = new Date();
+        let month = (currDate.getMonth() + 1).toString();
```
licanhua, line 4398:
```js
  let month = (1 + currDate.getMonth()).toString().padStart(2, "0");
```

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4347,6 +4347,10 @@ export class DateInput extends Input {
         };
 
         if (this.defaultValue) {
+
+            if (this.defaultValue === "today") {
```
licanhua, line 4351: If there is no customer, I prefer to cancel this PR. It introduces inconsistency like in ElementProperties. 

image

It breaks other features, for example, the today is changed to `4/20` after I drag two new items into the designer.

image

licanhua: This feature is broken too.

image

expected
image

anna-dingler: I am seeing this feature work for me:

image

anna-dingler: @licanhua Changing the code to:

```
        if (this.defaultValue) {
            if (this.defaultValue === "today") {
                this._dateInputElement.value = this.currentDate();
            } else {
                this._dateInputElement.value = this.defaultValue;
            }
        }
```

appears to solve the first issue you pointed out

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4347,11 +4347,11 @@ export class DateInput extends Input {
         };
 
         if (this.defaultValue) {
```
licanhua, line 4349: Add @adambarlow and @JeanRoca from PM side to see if there is any customer interested in this change.

JeanRoca: @licanhua Unfortunately, there are no customers requesting this feature that I could find.

licanhua: FYI, I found user is able to set `today` as utc, local time stamp or special time zone like this without this feature

```json
        {
            "type": "Input.Time",
            "value": "${convertFromUTC(utcNow(), 'New Zealand Standard Time', 'HH:mm')}"
        },
        {
            "id": "test",
            "type": "Input.Date",
            "value": "${convertFromUTC(utcNow(), 'New Zealand Standard Time', 'yyyy-MM-dd')}",
        },

```

```
        {
            "type": "Input.Date",
            "value": "${formatDateTime(utcNow(), 'yyyy-MM-dd')}"
        },

```

beervoley: @licanhua This is good of course, but it doesn't solve the **locale** problem. The card author needs to now the time zone information of the user to set the date what is not possible in most cases :) 
Setting the value to Today would allow us to get the local time for the user.

PR #7432, opened 2022-05-02
[Templating] Realign platforms on invalid $when behavior

### LINKED ISSUE #6630: [Rendering] Discrepancy (in js and C# libraries) while rendering an AC template containing an invalid "$when" condition
### Target Platforms

NodeJS

### SDK Version

latest

### Application Name

Microsoft Search

### Problem Description

We observe that, when the Adaptive card contains an invalid $when condition (an example of which is given below)

Example:
"$when": "https://searchuxcdn.azureedge.net/designerapp/images/servicenow-icon.png!=\"\""

The C# Templating library throws an exception, when we try to bind the template with data -> adaptiveTemplate.Expand(source),  with the following exception message:  "String was not recognized as a valid Boolean." 

However, the JS library  treats it as a valid string and displays the result evaluating the condition to true.
Can this discrepancy be fixed ?

### Screenshots

_No response_

### Card JSON

```JSON
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "TextBlock",
            "size": "Medium",
            "weight": "Bolder",
            "text": "${title}",
            "$when": "https://searchuxcdn.azureedge.net/designerapp/images/servicenow-icon.png!=\"\""
        }
    ],
    "version": "1.5"
}
```

### Sample Code Language

C#

### Sample Code

_No response_

anna-dingler:
# Related Issue

Fixes #6630 

# Description

Realign expected behavior for invalid boolean expressions for $when on JS and .NET Templating platforms.

Scenarios:
- If the given string is not an expression (ex: `"$when": "notAnExpression"`)
- If the given expression is invalid/not found in data (ex: `"$when": "${invalidExpression}"`)

In both scenarios, the $when condition should evaluate to `false` (drop the element) and issue a warning.

# Sample Card

## Test 1 - Not an Expression

```
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "TextBlock",
            "size": "Medium",
            "weight": "Bolder",
            "text": "${title}",
            "$when": "NotAnExpression"
        }
    ],
    "version": "1.5"
}
```

## Test 2 - Invalid Expression No Data

```
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "TextBlock",
            "size": "Medium",
            "weight": "Bolder",
            "text": "${title}",
            "$when": "${InvalidExpression}"
        }
    ],
    "version": "1.5"
}
```

## Test 3 - Invalid Expression with Data

Card

```
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "TextBlock",
            "size": "Medium",
            "weight": "Bolder",
            "text": "${title}",
            "$when": "${InvalidExpression}"
        }
    ],
    "version": "1.5"
}
```

Data
```
{
   "test": "dummyData"
}
```

# How Verified

Verified manually with adaptivecards-site and WPF Visualizer. I also added test cases to both platforms.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7432)

### DIFF
--- a/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
+++ b/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
@@ -502,7 +502,8 @@ namespace AdaptiveCards.Templating
                                 Console.WriteLine($"WARN: Could not evaluate {returnedResult} because it is not an expression or the " +
                                     $"expression is invalid. The $when condition has been set to false by default.");
                                 
-                            } else
+                            }
+                            else
                             {
                                 whenEvaluationResult = returnedResult.WhenEvaluationResult;
                             }
--- a/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
+++ b/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
@@ -495,7 +495,17 @@ namespace AdaptiveCards.Templating
 
                         if (returnedResult.IsWhen)
                         {
-                            whenEvaluationResult = returnedResult.WhenEvaluationResult;
+                            if (returnedResult.WhenEvaluationResult == AdaptiveCardsTemplateResult.EvaluationResult.NotEvaluated)
+                            {
+                                // The when expression could not be evaluated, so we are defaulting the value to false
+                                whenEvaluationResult = AdaptiveCardsTemplateResult.EvaluationResult.EvaluatedToFalse;
+                                Console.WriteLine($"WARN: Could not evaluate {returnedResult} because it is not an expression or the " +
+                                    $"expression is invalid. The $when condition has been set to false by default.");
+                                
+                            } else
+                            {
+                                whenEvaluationResult = returnedResult.WhenEvaluationResult;
+                            }
                         }
                         else
                         {
@@ -667,6 +677,15 @@ namespace AdaptiveCards.Templating
             // this node is visited only when parsing was correctly done
             // [ '{', '$when', ':', ',', 'expression'] 
             var result = Visit(context.templateExpression());
+
+            if (!result.IsWhen)
+            {
+                // We know that this result was supposed to be IsWhen since it is called from VisitTemplateWhen
+                // We create a result with `IsWhen = false` if the expression is invalid
+                // Result will now correctly follow the rest of the IsWhen logic
+                result.IsWhen = true;
+            }
+
             return result;
         }
 
@@ -728,7 +747,17 @@ namespace AdaptiveCards.Templating
             var (value, error) = new ValueExpression(Regex.Unescape(predicate)).TryGetValue(data);
             if (error == null)
             {
-                return bool.Parse(value as string);
+                try
+                {
+                    return bool.Parse(value as string);
+                }
+                catch (System.FormatException)
+                {
+                    // If the expression didn't evaluate to a boolean, we need to return false
+                    Console.WriteLine($"WARN: Could not evaluate boolean expression because it could not be found in the provided data. " +
+                                    "The condition has been set to false by default.");
+                    return false;
+                }
             }
             return true;
         }
--- a/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
+++ b/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
@@ -499,6 +499,7 @@ namespace AdaptiveCards.Templating
                             {
                                 // The when expression could not be evaluated, so we are defaulting the value to false
                                 whenEvaluationResult = AdaptiveCardsTemplateResult.EvaluationResult.EvaluatedToFalse;
+                                // TODO: Expose this warning to caller - documented in issue #7433
                                 Console.WriteLine($"WARN: Could not evaluate {returnedResult} because it is not an expression or the " +
                                     $"expression is invalid. The $when condition has been set to false by default.");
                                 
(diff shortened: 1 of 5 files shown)


source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -413,8 +413,17 @@ export class Template {
                     if (!evaluationResult.error) {
                         whenValue = typeof evaluationResult.value === "boolean" && evaluationResult.value;
                     }
+                    
+                    if (!evaluationResult.value) {
+                        // Value was not found, and we should warn the client that the Expression was invalid
+                        console.warn(`WARN: Unable to parse the Adaptive Expression ${when}. The $when condition has been set to false by default.`);
```
licanhua, line 419: Don't use console to report error. dev can't see it

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
```diff
@@ -728,7 +748,17 @@ public static bool IsTrue(string predicate, JToken data)
             var (value, error) = new ValueExpression(Regex.Unescape(predicate)).TryGetValue(data);
             if (error == null)
             {
-                return bool.Parse(value as string);
+                try
```
licanhua, line 751: Personally, I prefer to just crash the app.
silent eats the exception may cause confusion for customers.

From me, it's a breaking change, and you need to align with PM.

source/nodejs/adaptivecards-templating/src/template-engine.ts
anna-dingler: That's a good call out. Is there another way to warn developers without crashing templating?

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
anna-dingler: We had a discussion thread about how the platforms should align on the original issue. JS templating handled the invalid input without crashing, whereas .NET crashed 

https://github.com/microsoft/AdaptiveCards/issues/6630#issuecomment-1113643312

anna-dingler: I spoke with @rahulamlekar to connect with the PM team, and he likes the proposed solution (set to false and warn the user).

source/nodejs/adaptivecards-templating/src/template-engine.ts
paulcam206: It doesn't look like either templating library currently exposes warnings/errors to callers. I think the right approach would be similar to how things are done in Adaptive Cards more broadly where we pass along a list of strings that we can add warnings to... and then make those strings available to the caller in the results we hand back. We'd obviously need to introduce that without making it a breaking change.

So for C#, I'd expect:
* add optional out parameters to the `AdaptiveCardTemplate.Expand` methods that's an array of warning strings

and for JS I'd expect:
* add an overload for the `Template.expand` method that takes an out parameter that's an array of warning strings (or alternatively, maybe an optional out parameter?)

if you think these changes fit nicely within this PR, feel free to make them. if you'd rather address them separately or think it's too much of a change to deal with right now, feel free to file a bug and reference that bug in code comments for this PR

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
```diff
@@ -495,7 +495,18 @@ public override AdaptiveCardsTemplateResult VisitObj([NotNull] AdaptiveCardsTemp
 
                         if (returnedResult.IsWhen)
                         {
-                            whenEvaluationResult = returnedResult.WhenEvaluationResult;
+                            if (returnedResult.WhenEvaluationResult == AdaptiveCardsTemplateResult.EvaluationResult.NotEvaluated)
+                            {
+                                // The when expression could not be evaluated, so we are defaulting the value to false
+                                whenEvaluationResult = AdaptiveCardsTemplateResult.EvaluationResult.EvaluatedToFalse;
+                                Console.WriteLine($"WARN: Could not evaluate {returnedResult} because it is not an expression or the " +
```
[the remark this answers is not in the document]
paulcam206: (see comment below)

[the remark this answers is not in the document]
anna-dingler: I've created a new issue for this here: https://github.com/microsoft/AdaptiveCards/issues/7433

source/nodejs/adaptivecards-templating/src/template-engine.ts
anna-dingler: I've created a new issue for this here: https://github.com/microsoft/AdaptiveCards/issues/7433

jwoo-msft: it would be nice if the addition of warning is addressed before we make next release.

jwoo-msft (approved):
it looks good, my only reservation is the implementation of warning. I will almost say it is necessary to have this implemented before we make next release since we are changing the behavior adding the warning will give the dev more flexibility to adapt to our changes.

anna-dingler:
> it looks good, my only reservation is the implementation of warning. I will almost say it is necessary to have this implemented before we make next release since we are changing the behavior, adding the warning will give the dev more flexibility to adapt to our changes.

Thanks for reviewing! I'm working on updating the warning implementation now, so I'll make that PR before making the next release.

PR #7437, opened 2022-05-05
[Templating] Expose warnings/errors to callers

### LINKED ISSUE #7433: [Templating] Expose errors/warnings to callers
# Platform

* .NET HTML
* .NET WPF
* JavaScript

# Issue

Templating libraries do not currently have a way to expose errors and warning to the callers. We need to implement a system similar to the implementation in the base AdaptiveCards project.

# Proposed solution

- C#
Add optional out parameters to the AdaptiveCardTemplate.Expand methods that's an array of warning strings

- JS
Add an overload for the Template.expand method that takes an out parameter that's an array of warning strings (or alternatively, maybe an optional out parameter?)

anna-dingler:
# Related Issue

Fixes #7433 

# Description

Warnings need to be exposed to the caller instead of logged in the console.

I added an optional parameter for an array of warning/error strings to the expand method of the JS library.

For the .NET library, I added an out parameter for the warning/error string array to the Expand methods. I also added an additional wrapper methods, that call the updated Expand methods to ensure this isn't a breaking change.

# Sample Card

These cards should produce warning strings if the proper expand method is called.

## Test 1 - Not an Expression

```
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "TextBlock",
            "size": "Medium",
            "weight": "Bolder",
            "text": "${title}",
            "$when": "NotAnExpression"
        }
    ],
    "version": "1.5"
}
```

## Test 2 - Invalid Expression No Data

```
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "TextBlock",
            "size": "Medium",
            "weight": "Bolder",
            "text": "${title}",
            "$when": "${InvalidExpression}"
        }
    ],
    "version": "1.5"
}
```

## Test 3 - Invalid Expression with Data

Card

```
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "TextBlock",
            "size": "Medium",
            "weight": "Bolder",
            "text": "${title}",
            "$when": "${InvalidExpression}"
        }
    ],
    "version": "1.5"
}
```

Data
```
{
   "test": "dummyData"
}
```

# How Verified

Verified with the unit tests added with this PR.

### DIFF
--- a/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardTemplate.cs
+++ b/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardTemplate.cs
@@ -75,8 +75,8 @@ namespace AdaptiveCards.Templating
         /// <returns>json as string</returns>
         public string Expand(EvaluationContext context, Func<string, object> nullSubstitutionOption = null)
         {
-            ArrayList<string> log;
-            return Expand(context, out log, nullSubstitutionOption);
+            ArrayList<string> errorLog;
+            return Expand(context, out errorLog, nullSubstitutionOption);
         }
 
         /// <summary>
@@ -90,7 +90,7 @@ namespace AdaptiveCards.Templating
         /// <para> Default behavior is leaving templated string unchanged</para>
         /// </remarks>
         /// <param name="context">provides data context</param>
-        /// <param name="log">stores the outputed log statements from parsing the template</param>
+        /// <param name="errorLog">stores the outputed log statements from parsing the template</param>
         /// <param name="nullSubstitutionOption">defines behavior when no suitable data is found for a template entry</param>
         /// <example>
         /// <code>
@@ -101,13 +101,13 @@ namespace AdaptiveCards.Templating
         /// </example>
         /// <seealso cref="EvaluationContext"/>
         /// <returns>json as string</returns>
-        public string Expand(EvaluationContext context, out ArrayList<string> log, Func<string, object> nullSubstitutionOption = null)
+        public string Expand(EvaluationContext context, out ArrayList<string> errorLog, Func<string, object> nullSubstitutionOption = null)
         {
         
             if (parseTree == null)
             {
                 // Create empty array list so that `out` parameter has a value
-                log = new ArrayList<string>();
+                errorLog = new ArrayList<string>();
                 return jsonTemplateString;
             }
 
@@ -129,9 +129,21 @@ namespace AdaptiveCards.Templating
 
             AdaptiveCardsTemplateResult result = eval.Visit(parseTree);
 
-            log = eval.getTemplateLog();
+            errorLog = eval.getTemplateLog();
 
             return result.ToString();
+        }
+
+        /// <summary>
+        /// Wrapper method to maintain functionality if caller decides not to use log files
+        /// </summary>
+        /// <param name="rootData">Serializable object or a string in valid json format that will be used as data context</param>
+        /// <param name="nullSubstitutionOption">Defines behavior when no suitable data is found for a template entry</param>
+        /// <returns></returns>
+        public string Expand(object rootData, Func<string, object> nullSubstitutionOption = null)
+        {
+            ArrayList<string> errorLog = new ArrayList<string>();
+            return Expand(rootData, out errorLog, nullSubstitutionOption);
         }
 
         /// <summary>
@@ -145,6 +157,7 @@ namespace AdaptiveCards.Templating
         /// <para> Default behavior is leaving templated string unchanged</para>
         /// </remarks>
         /// <param name="rootData">Serializable object or a string in valid json format that will be used as data context</param>
+        /// <param name="errorLog">stores the outputed log statements from parsing the template</param>
         /// <param name="nullSubstitutionOption">Defines behavior when no suitable data is found for a template entry</param>
         /// <example>
         /// <code>
@@ -154,13 +167,15 @@ namespace AdaptiveCards.Templating
         /// </example>
         /// <seealso cref="EvaluationContext"/>
         /// <returns>json as string</returns>
-        public string Expand(object rootData, Func<string, object> nullSubstitutionOption = null)
+        public string Expand(object rootData, out ArrayList<string> errorLog, Func<string, object> nullSubstitutionOption = null)
         {
-            // TODO: need to add error logging here - not so much in the template visitor??
-            // Caller will call AdaptiveCardTemplate.Expand - NOT from AdaptiveCardsTemplateVisitor
             var context = new EvaluationContext(rootData);
 
-            return Expand(context, nullSubstitutionOption);
+            errorLog = new ArrayList<string>();
+
+            string expand = Expand(context, out errorLog, nullSubstitutionOption);
+
+            return expand;
         }
     }
 }
--- a/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardTemplate.cs
+++ b/source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardTemplate.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 using Antlr4.Runtime;
+using Antlr4.Runtime.Misc;
 using Antlr4.Runtime.Tree;
 using Newtonsoft.Json;
 using System;
@@ -67,6 +68,18 @@ namespace AdaptiveCards.Templating
         }
 
         /// <summary>
+        /// Wrapper method to maintain functionality if caller decides not to use log files
+        /// </summary>
+        /// <param name="context">provides data context</param>
+        /// <param name="nullSubstitutionOption">defines behavior when no suitable data is found for a template entry</param>
+        /// <returns>json as string</returns>
+        public string Expand(EvaluationContext context, Func<string, object> nullSubstitutionOption = null)
+        {
+            ArrayList<string> log;
+            return Expand(context, out log, nullSubstitutionOption);
+        }
+
+        /// <summary>
         /// Bind data in <paramref name="context"/> to the instance of AdaptiveCardTemplate
         /// </summary>
         /// <remarks>
@@ -77,6 +90,7 @@ namespace AdaptiveCards.Templating
         /// <para> Default behavior is leaving templated string unchanged</para>
         /// </remarks>
         /// <param name="context">provides data context</param>
+        /// <param name="log">stores the outputed log statements from parsing the template</param>
         /// <param name="nullSubstitutionOption">defines behavior when no suitable data is found for a template entry</param>
         /// <example>
         /// <code>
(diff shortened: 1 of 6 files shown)


source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -531,9 +532,22 @@ export class Template {
      * @returns A value representing the expanded template. The type of that value
      *   is dependent on the type of the original template payload passed to the constructor.
      */
-    expand(context: IEvaluationContext): any {
+    expand(context: IEvaluationContext, errorLog?: String[]): any {
```
anna-dingler, line 535: This logic could also be done with an optional boolean, but I wasn't sure if we wanted devs to have the flexibility to pass in an existing array.

It would also be simplified with an out parameter, but I wasn't sure how to do this in typescript.

source/dotnet/Test/AdaptiveCards.Templating.Test/AdaptiveCards.Templating.Test.csproj
```diff
@@ -8,7 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="AdaptiveExpressions" Version="4.11.*" />
-    <PackageReference Include="Antlr4.Runtime.Standard" Version="4.9.1" />
+    <PackageReference Include="Antlr4.Runtime.Standard" Version="4.8.0" />
```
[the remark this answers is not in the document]
anna-dingler: The templating library uses version 4.8.0, so the higher version was causing conflicts.

I can also upgrade the templating library to 4.9.1, but I wasn't sure if this would cause errors

source/nodejs/adaptivecards-templating/src/template-engine.ts
licanhua, line 535: errorLog is a hacky way and I would not use it in production. I have two alternatives:
1. add a feature flag to Template. For example,  Template.treatInvalidWhenAsFalse. and throw exception if user set this flag to false (For example, used by designer) to report error. 3rd party use this flag to switch back to old logic too if don't want this feature. 
2. 
```
expand(context: IEvaluationContext,  errorHandler?: function(error, errorDetail) => bool): any
```

dev may handle the error like this:

```
expand(context: IEvaluationContext,  (error, detail) => {  console.log(detail); return false} );
```

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardTemplate.cs
```diff
@@ -132,11 +167,15 @@ public string Expand(EvaluationContext context, Func<string, object> nullSubstit
         /// </example>
         /// <seealso cref="EvaluationContext"/>
         /// <returns>json as string</returns>
-        public string Expand(object rootData, Func<string, object> nullSubstitutionOption = null)
+        public string Expand(object rootData, out ArrayList<string> errorLog, Func<string, object> nullSubstitutionOption = null)
```
licanhua, line 170: Don't change existing API. 

Also the API is hard to use. 
You may add new api like

```
public <string, error> Expand2(object rootData, Func<string, object> nullSubstitutionOption = null)
```

source/nodejs/adaptivecards-templating/src/template-engine.ts
anna-dingler: I agree that this felt hacky which is why I made this a draft.

From my understanding, I was supposed to be making a more generic error handling system that we could use in future scenarios. Should I be catering to the specific boolean case instead?

jwoo-msft (commented):
we have a lot of Expand methods, can you somehow consolidate it. at minimum, can you make sure the primary Expand method is at the prominent place, so devs would know which one is the primary one without much of a search?
and if you could, it would be nice if we can reduce the number ArrayList instantiation. So if we are going to change log data structure, we could do it at one place.

source/dotnet/Test/AdaptiveCards.Templating.Test/AdaptiveCards.Templating.Test.csproj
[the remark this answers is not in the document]
anna-dingler: The templating library uses version 4.8.0, so the higher version was causing conflicts.

I can also upgrade the templating library to 4.9.1, but I wasn't sure if this would cause errors

source/dotnet/Test/AdaptiveCards.Templating.Test/TestTransform.cs
```diff
@@ -6,6 +6,7 @@
 using System;
 using AdaptiveExpressions.Memory;
 using System.Collections.Generic;
+using Antlr4.Runtime.Misc;
```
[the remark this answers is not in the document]
anna-dingler: ArrayList is being imported from Antl4.Runtime.Misc

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardTemplate.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 using Antlr4.Runtime;
+using Antlr4.Runtime.Misc;
```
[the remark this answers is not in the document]
anna-dingler: ArrayList is being imported from Antl4.Runtime.Misc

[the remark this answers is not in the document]
jwoo-msft: thanks; please do make sure to add the correct library.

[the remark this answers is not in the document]
anna-dingler: Updated to use System.Collections

source/dotnet/Test/AdaptiveCards.Templating.Test/TestTransform.cs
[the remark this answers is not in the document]
anna-dingler: Updated to use System.Collections

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardTemplate.cs
anna-dingler: I used wrapper methods that kept the initial method headers so that the existing API is still usable. Would that be okay?

anna-dingler:
> we have a lot of Expand methods, can you somehow consolidate it. at minimum, can you make sure the primary Expand method is at the prominent place, so devs would know which one is the primary one without much of a search? and if you could, it would be nice if we can reduce the number ArrayList instantiation. So if we are going to change log data structure, we could do it at one place.

I grouped the wrapper methods together and left additional comments to the Expand methods to make it clearer. I also used `_` to indicate a discarded out parameter to cut down on ArrayList instantiation.

Should I be committing the autogenerated .md files fot the .NET library?

source/nodejs/adaptivecards-templating/src/template-engine.ts
licanhua: I did simple research on JS implementation. maybe it's easy to add new stuff to IEvaluationContext or Template to get the support the error log.

```
export interface IEvaluationContext {
    /**
     * The root data object the template will bind to. Expressions that refer to $root in the template payload
     * map to this field. Initially, $data also maps to $root.
     */
    $root: any,
    logEvaluationError? : (error) => void,
    featureflag: true,
}

```

or 
```
class Template {
...
    expand(context: IEvaluationContext): any;
    getEvaluationError(): string;
}
```

anna-dingler: I like the idea of collecting the errors in the template object so that developers can get them after expanding.

@jwoo-msft what do you think?

jwoo-msft: @anna-dingler that makes sense to me; AdaptiveCards parser works that way, too.

I think adding it to the Template instance makes more sense to me since EvaluationContext's role is more specific.

anna-dingler: @licanhua @jwoo-msft  Thanks for the help here! Just updated it - let me know what you all think!

source/nodejs/adaptivecards-templating/src/template-engine.ts
```diff
@@ -533,7 +535,14 @@ export class Template {
...
         this._context = new EvaluationContext(context);
-
         return this.internalExpand(this._preparedPayload);
     }
+
+    /**
+     * Getter method for the array of error strings
+     * @returns An array storing any errors that occurred while expanding the template
+     */
+    public getTemplateExpansionErrors(): string[] {
```
licanhua, line 545: I really want to see the message is show in designer

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardTemplate.cs
```diff
@@ -135,8 +141,20 @@ public string Expand(EvaluationContext context, Func<string, object> nullSubstit
...
             var context = new EvaluationContext(rootData);
-
             return Expand(context, nullSubstitutionOption);
         }
+
+        /// <summary>
+        /// Getter method for the array of error strings
+        /// </summary>
+        /// <returns>ArrayList</returns>
+        public ArrayList GetTemplateExpansionErrors()
```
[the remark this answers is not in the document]
anna-dingler: @licanhua Do we only want the warnings from the last expansion in the JS library as well?

source/nodejs/adaptivecards-templating/src/template-engine.ts
anna-dingler: Created an issue, and will resolve in a separate PR.

https://github.com/microsoft/AdaptiveCards/issues/7442

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardTemplate.cs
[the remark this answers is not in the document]
licanhua: yes

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardsTemplateVisitor.cs
```diff
@@ -336,8 +349,20 @@ public override AdaptiveCardsTemplateResult VisitValueTemplateExpression([NotNul
...
 
+            bool isTrue = false;
+
+            try
+            {
+                isTrue = IsTrue(result.Predicate, dataContext.token);
+            }
+            catch (System.FormatException)
+            {
+                templateVisitorErrors.Add($"WARN: Could not evaluate boolean expression because it could not be found in the provided data. " +
```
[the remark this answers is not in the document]
anna-dingler: Good callout - just updated!

source/dotnet/Library/AdaptiveCards.Templating/AdaptiveCardTemplate.cs
```diff
@@ -4,6 +4,7 @@
 using Antlr4.Runtime.Tree;
 using Newtonsoft.Json;
 using System;
+using System.Collections;
```
licanhua, line 8: Don't forget to update document for this breaking change

anna-dingler: I'll be sure to note this in the release notes. Thanks!

PR #7473, opened 2022-05-23
[Designer] Implement Designer as PWA

### LINKED ISSUE #7471: Make Designer installable as a PWA.

### LINKED ISSUE #7472: Stretch goals: Make designer work offline when user installed as a PWA.

anna-dingler:
# Related Issue

Fixes #7471 
Fixes #7472 

# Description

Added implementation for the designer to be installable as a PWA.

## To install

`npx lerna run release`
`npm start`

When you navigate to the designer page, the option to install the page should be available.

## Changes

`package.json`

- Updating scripts to run `npx workbox-cli injectManifest workbox-config.json` while building. This generates a list of files to be precached and injects them into the service worker

`service-worker.js`

- Uses workbox to cache elements from the designer
- Caching strategies descriptions: https://developer.chrome.com/docs/workbox/modules/workbox-strategies/#stale-while-revalidate

`head.ejs`

- Register the manifest for the designer page

`designer.ejs`

- Hide the header if we are viewing the PWA
- Register the service worker

`manifest.json`

- Defines meta data for the PWA

`workbox-config.json`
- Defines the files to be precached

# How Verified

Tested Manually

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7473)


source/nodejs/adaptivecards-site/package.json
```diff
@@ -4,12 +4,13 @@
...
-		"build": "hexo generate",
+		"build": "hexo generate && npx workbox-cli injectManifest workbox-config.json",
 		"release-deps": "npx lerna run --scope @microsoft/adaptivecards-site --include-dependencies release --stream",
 		"generate-md-ac": "node node_modules/marked-schema/bin/marked-schema ../../../schemas/adaptive-card.json ./schema-explorer-toc.yml -r AdaptiveCard",
 		"generate-md-hc": "node node_modules/marked-schema/bin/marked-schema ../../../schemas/host-config.json ./schema-hostconfig-toc.yml -r AdaptiveCardConfig",
-		"release": "hexo clean && hexo generate",
-		"start": "hexo server"
+		"release": "hexo clean && hexo generate && npx workbox-cli injectManifest workbox-config.json",
+		"start": "hexo server",
+		"generate-sw": "npx workbox-cli injectManifest workbox-config.json"
```
beervoley, line 13: use this task in `build` and `release` to avoid code repetition :)

source/nodejs/adaptivecards-site/service-worker.js
```diff
@@ -0,0 +1,109 @@
+// Used pwa-builder as reference: https://github.com/pwa-builder/pwabuilder-serviceworkers/blob/master/serviceWorker5/pwabuilder-sw.js
+importScripts(
```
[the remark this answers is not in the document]
anna-dingler: @aarongustafson I registered the service worker and manifest within the designer html file. For the service worker, I set the defined scope to the designer directory at registration, but have it stored a directory up so that it can access the `content` folder. Do you think this is the correct approach?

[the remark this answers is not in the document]
aarongustafson: I’m not super familiar with Hexo and how this will all come out in the build, but I think that sounds right. Ultimately you want it to be in the directory you want it to govern. If it need to be able to operate on all fines in the site, the root is the right place for it. If you only want it to affect the Designer, it might make more sense to put it there so it doesn’t interfere with what’s happening outside of that directory.

Honestly, there’s probably a lot of good the SW can do for the site as a whole, so the root is probably the right place, even if you don’t add all those rules right now.

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/designer.ejs
```diff
@@ -45,8 +45,27 @@
...
+			display: none;
+		}
+	}
 </style>
 
+<script type="text/javascript">
+	// Register the service worker for PWA
+	if ('serviceWorker' in navigator) {
+		window.addEventListener('load', () => {
+			navigator.serviceWorker.register('/../service-worker.js', {scope: '/designer/'}).then(registration => {
```
aarongustafson, line 60:
```suggestion
			navigator.serviceWorker.register('/service-worker.js', {scope: '/designer/'}).then(registration => {
```

You could either start with / (for the root) or ../ to move to the root from the subdirectory. I tend to prefer the former (site root relative links).

source/nodejs/adaptivecards-site/package.json
anna-dingler: Good call out! Just updated these

source/nodejs/adaptivecards-site/service-worker.js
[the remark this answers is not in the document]
anna-dingler: @licanhua I add a minified version of the service worker

anna-dingler:
> Hi @anna-dingler. Thanks for helping make the AdaptiveCards JS renderer + tooling better. As additional verification, once the JS build succeeds, please go to the [test site](https://adaptivecardsci.z5.web.core.windows.net/pr/7473) to test out your website/designer changes.

@paulcam206 I'm running into a weird issue with this test site, and wanted to get your thoughts on it.

When testing locally and on the published version, the designer page is referenced with `<root-site>/designer/`, but on the test site it is referenced with `<root-site>/pr/pr#/designer/`.

Since the path is different we have two issues:
1. The `manifest.json` and `service-worker.js` files cannot be found during registration
2. The url paths within `manifest.json` need to be dynamic. Without a dynamic manifest, the `start_url` and `icon src` do not include the `pr/pr#` and the urls cannot be found.

I was able to solve (1) by prefixing the paths with `<%- config.root %>`. The solution to (2) was much less eloquent and required writing the manifest file within a `<script>`.

These changes can be found on my test PR: https://github.com/microsoft/AdaptiveCards/pull/7484

@beervoley and I were wondering if these changes were worth the additional complexity since they only seem to affect our test PR site. What do you think?

source/nodejs/adaptivecards-site/package.json
```diff
@@ -4,12 +4,13 @@
 	"private": true,
 	"scripts": {
 		"clean": "hexo clean",
-		"build": "hexo generate",
+		"build": "hexo generate && npm run generate-sw",
```
licanhua, line 7: do you know why this failed?
https://adaptivecardsci.z5.web.core.windows.net/pr/7473/designer
then ctrl+shift+i to developer mode

image

anna-dingler: @licanhua This error is outlined in the comment above: https://github.com/microsoft/AdaptiveCards/pull/7473#issuecomment-1140010232

I created a fix for it in another PR (https://github.com/microsoft/AdaptiveCards/pull/7484), but we aren't sure if we want to include it

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/designer.ejs
```diff
@@ -45,8 +45,54 @@
...
+		}
+	}
 </style>
 
+<script type="text/javascript">
+
+	const startUrl = window.location.origin + '<%- config.root %>designer/';
+	const iconUrl = window.location.origin + '<%- config.root %>content/icons_blue/blue-512.png';
+
+	const manifest = {
```
licanhua, line 61: Can you add comments in code that why you need to generate manifest manually? It will be easy for the next developer to take over on it

source/nodejs/adaptivecards-site/service-worker.js
```diff
@@ -0,0 +1,109 @@
...
+  'https://storage.googleapis.com/workbox-cdn/releases/6.4.1/workbox-sw.js'
```
paulcam206, line 3: I think we'll want to pull this from a different CDN, or perhaps we should host it directly on our site?

paulcam206 (changes requested):
looks great! just need to pull from a different CDN 👍

source/nodejs/adaptivecards-site/service-worker.js
```diff
@@ -0,0 +1,109 @@
...
+    cacheName: HTML_CACHE,
+    plugins: [
+      new workbox.expiration.ExpirationPlugin({
+        maxEntries: 10,
+      })
+    ]
+  })
+);
+
+workbox.routing.registerRoute(
```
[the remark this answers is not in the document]
anna-dingler: Good call, I made it a function

licanhua (approved):

source/nodejs/adaptivecards-site/package.json
```diff
@@ -4,12 +4,13 @@
 	"private": true,
 	"scripts": {
 		"clean": "hexo clean",
-		"build": "hexo generate",
```
licanhua, line 7: minor: can you make it default to window maximized? It's a little weird me that I clicked preview, and I didn't see it

image

source/nodejs/adaptivecards-site/service-worker.js
anna-dingler: @paulcam206 I updated to use local workbox files per this documentation: https://developer.chrome.com/docs/workbox/modules/workbox-sw/#using-local-workbox-files-instead-of-cdn

Let me know what you think!

source/nodejs/adaptivecards-site/package.json
anna-dingler: @licanhua @aarongustafson Do you have any ideas for this? I tried using `viewport-fit=cover` in the html and `"display": "fullscreen"` in the manifest, but neither made a change.

I think the fix would be relatively straight forward in JS, but we don't want the screen to visibly resize when the app loads.

source/nodejs/adaptivecards-site/service-worker.js
```diff
@@ -0,0 +1,68 @@
+// Used pwa-builder as reference: https://github.com/pwa-builder/pwabuilder-serviceworkers/blob/master/serviceWorker5/pwabuilder-sw.js
+importScripts('/workbox/workbox-v6.5.3/workbox-sw.js');
```
anna-dingler, line 2: @paulcam206 By hosting these files on the website, we are hitting the `pr/pr#` url issue on the test site again. I don't think generating the service-worker dynamically is a good solution here, but we don't have access to the hexo url variable or `window` to grab the root of the url.

Do you have any ideas for resolving this?

paulcam206: does `<%- config.root %>` not work here?

paulcam206: alternatively, what happens if you refer to the path without the initial `/`?

anna-dingler: That worked, also fixed the manifest issue. Thanks!

beervoley (approved):

beervoley:
@anna-dingler @paulcam206 I see massive changes to `package-lock.json`. I'm not sure we want to check them in?

paulcam206:
> @anna-dingler @paulcam206 I see massive changes to `package-lock.json`. I'm not sure we want to check **them all** in?

I took a look through the lock file and it looks okay. The reason this is happening is us officially adding `workbox-cli` to our `package.json` `devDependencies` (before it was being pulled down and invoked on-demand).

paulcam206 (approved):
:shipit:

source/nodejs/adaptivecards-site/package.json
anna-dingler: I opened an issue for this: https://github.com/microsoft/AdaptiveCards/issues/7498

PR #7474, opened 2022-05-23
[Android] Escape character not removed from TextBlock

karthikgopal36:
# Description

Escape character is not removed from TextBlock. 

For e.g : \\*text\\* should be displayed as *text* but displayed as  \*text\* in Mobile

# Sample Card

```
{
  "type": "AdaptiveCard",
  "version": "1.2",

  "body": [

	{

	  "type": "TextBlock",

	  "text": "Text with stars to make it italic: \\*text\\*",

	  "wrap": true

	}
  ],

  "actions": [],

  "version": "1.4"


}
```

# How Verified

before           |  after
:-------------------------:|:-------------------------:
screenshot-1653330462137  |  screenshot-1653330348771

### DIFF
--- a/source/shared/cpp/ObjectModel/MarkDownBlockParser.cpp
+++ b/source/shared/cpp/ObjectModel/MarkDownBlockParser.cpp
@@ -138,7 +138,7 @@ EmphasisParser::EmphasisState EmphasisParser::MatchText(EmphasisParser& parser, 
     }
     else
     {
-        if (isEmphasisToken && parser.m_lookBehind == DelimiterType::Escape)
+        if (parser.m_lookBehind == DelimiterType::Escape)
         {
             // remove escape char from stream
             token.pop_back();


jwoo-msft (approved):

jwoo-msft (commented):
Can you update screenshot?
Your screenshot doesn't show this PR fixed the issue.

source/shared/cpp/ObjectModel/MarkDownBlockParser.cpp
```diff
@@ -138,7 +138,7 @@ EmphasisParser::EmphasisState EmphasisParser::MatchText(EmphasisParser& parser,
     }
     else
     {
-        if (isEmphasisToken && parser.m_lookBehind == DelimiterType::Escape)
+        if (parser.m_lookBehind == DelimiterType::Escape)
```
jwoo-msft, line 141: does escape char works?
looks like, these changes removing all of the escape char.

jwoo-msft (changes requested):

jwoo-msft (changes requested):
@karthikgopal36 
this pr isn't correct.

```
\\*hello\\*
```
should output \\*hello\\*

I created this PR to address the issue. https://github.com/microsoft/AdaptiveCards/pull/7489

beervoley:
closing this one since it since #7489 is fixing this issue.

Issue #7479, opened 2022-05-25
[Rendering] ReactNative cannot register custom column


feiyin0719:
### Target Platforms

Other

### SDK Version

2.2.24

### Application Name

Test app

### Problem Description

in ColumnSet, it use Column directly , so when we register my custom column, it will not work

### Card JSON

```JSON
N/A
```

paulcam206:
We don't support registering custom `Column`s. If you want to override the behavior of `Column`, you need to register a custom `ColumnSet`.

PR #7482, opened 2022-05-25
[ReactNative] fix ColumnSet use Column directly, it will cause cannot register custom Column

feiyin0719:
# Related Issue
[#7479 ]

# Description

in ColumnSet, it use Column directly , if we have registered custom Column, it will not work

# Sample Card

# How Verified
1. open ReactNative column

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7482)


paulcam206:
As per comment on bug, the correct approach to customizing `Column` behavior is to override `ColumnSet` itself.

Issue #7037, opened 2022-02-08
[InputForm.json]: Talkback does not correctly announce the format in which the user should enter the phone number.


jwoo-msft:
https://microsoft.visualstudio.com/OS/_workitems/edit/38003881

PR #7493, opened 2022-06-07
[Accessibility] Change phone number validation

almedina-ms:
# Related Issue

Fixes https://microsoft.visualstudio.com/OS/_workitems/edit/38003881

Fixes https://github.com/microsoft/AdaptiveCards/issues/7037

# Description

Hyphens are not being read making it hard for people who rely on talkback to fill the field on a first pass. According to this article https://www.deque.com/blog/dont-screen-readers-read-whats-screen-part-1-punctuation-typographic-symbols/ the hyphen '-' symbol is not recognized by all screen readers so the regex had to be changed to allow for different ways of inputting a phone number

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7493)


samples/v1.3/Scenarios/InputFormWithLabels.json
```diff
@@ -51,9 +51,9 @@
 						{
 							"type": "Input.Text",
 							"id": "myTel",
-							"label": "Phone Number (xxx-xxx-xxxx)",
-							"regex": "^[0-9]{3}-[0-9]{3}-[0-9]{4}$",
-							"errorMessage": "Invalid phone number. Use the specified format: 3 numbers, hyphen, 3 numbers, hyphen and 4 numbers",
+							"label": "Phone Number (xxx xxx xxxx)",
+							"regex": "^\\(?([0-9]{3})\\)?[-.\\s]?([0-9]{3})[-.\\s]?([0-9]{4})$",
```
licanhua, line 55: Is xxx.xxx.xxxx expected?

beervoley: what formats are we planning to support? for example:
123-456-7890
(123) 456-7890
123 456 7890
123.456.7890
+91 (123) 456-7890

do we plan to support stuff like this? if yes, then this regex is probably not enough?:)

paulcam206 (approved):

PR #7509, opened 2022-06-10
[Spec] Plan for updating the FactSet element

almedina-ms:
# Related Issue

Proposal for a long lasting fix for https://microsoft.visualstudio.com/OS/_workitems/edit/37990023

# Description

Add details of how to update the FactSet element to provide a general better experience for car authors

### DIFF
--- a/specs/elements/FactSetv2.md
+++ b/specs/elements/FactSetv2.md
@@ -1,0 +1,105 @@
+# FactSet renewal
+
+## Justification
+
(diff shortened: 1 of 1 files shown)


almedina-ms:
> If FactSet's title and value have all of the properties of TextBlock, we can also use TextBlocks as values of FactSet's title and value.
> 
> With this change, I'm wondering what benefits cards authors get when they can use Tables with TextBlocks.

I think factset cover only the scenario where you want a 2 by N table

jwoo-msft:
> > If FactSet's title and value have all of the properties of TextBlock, we can also use TextBlocks as values of FactSet's title and value.
> > With this change, I'm wondering what benefits cards authors get when they can use Tables with TextBlocks.
> 
> I think factset cover only the scenario where you want a 2 by N table

Yes, you are correct. I like the changes. My humble suggestion is we can also consider factoring out text properties and create a new property that holds all of the text properties as we do with Background Property.
And add the property to TextBlock and FactSet.

Issue #7436, opened 2022-05-05
Make the Designer more accessible


carlos-zamora:
There's a good amount of accessibility work that can be done on the designer. This issue is intended to track that work.

To start, the "drag" and "remove" controls need to be accessible and actually usable. See https://github.com/microsoft/AdaptiveCards/issues/7067 for more details.

Unfortunately, the work required to do that is no small task. Pankaj and I wrote a proposal for a designer overhaul. That can be found [here](https://microsoft.sharepoint.com/teams/AdaptiveCards/_layouts/15/Doc.aspx?sourcedoc={ef9a3136-2fec-4955-b14d-0507fce791ed}&action=edit&wd=target%28Accessibility.one%7C32f608f3-b7db-425d-9f46-568eae5be2a2%2FDesigner%20Overhaul%7C9c775d7a-cecb-4433-a9f5-f521c2090582%2F%29&wdorigin=703). For simplicity, I've copied the summary below...

<details>
<summary>Designer Overhaul Summary</summary>

- There are various aspects of the designer that are only available to mouse users 
- Making these features available to keyboard users will require a refactor of the designer, which should involve 
   - Rendering every element of the card upon initialization, even if the element is initially hidden 
   - In designer mode, only the peers should be focusable whereas the elements themselves should be not-focusable 
      - Furthermore, all commands available to the peer should be invokable via keyboard (general commands like drag/drop, remove; as well as specific commands like 'Add a column') 
   - When new elements are added to the card, or when existing elements are moved around, the elements should be inserted into the html tree in such a way as to mimic their order in the card itself 
      - If this requires a complete re-rendering of the card every time such an action is invoked, then extra steps must be taken to ensure focus does not get lost 
- Aside from the preview card itself, the other panes in the designer (i.e. 'Card Elements', 'Card Structure' and 'Element Properties') also need to be improved to support keyboard users 
   - Since these panes are quite linked with each other, in particular the 'Card Structure' and 'Element Properties' panes, there should exist an easy way for keyboard users to switch between these panes 
   - All elements in these panes should be accessible/activatable by keyboard 
      - i.e. inserting an element into the preview card from the 'Card Elements' pane should be possible via keyboard instead of only via dragging onto the preview card 
      - i.e. all elements in the 'Card Structure' pane should be accessible via keyboard 

</details>

As a bonus, I strongly suggest that whoever begins to work on making the designer more accessible also keep an eye on the following queries:
[...]

PR #7159, opened 2022-03-21
[DRAFT] Make the designer page keyboard accessible

PankajBhojwani:
WIP of our branch where we are working on updating the designer page to be keyboard accessible

Work items:
- [x] focus should be able to land on a peer
- [x] remove control should be keyboard activatable
- [ ] drag control should be keyboard activatable
- [x] bind control should be keyboard activatable
- [x] if the peer is wrapping an action, the action should still be invokable
- [x] we need some way to communicate the keybindings to the user

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7159)

PR #7514, opened 2022-06-10
[Designer Accessibility] Insert peers in the DOM tree in the correct order

anna-dingler:
# Related Issue

Progress towards #7436 

# Description

Instead of inserting designer peers to the bottom of the HTML tree, we must insert designer peers to mimic the order of the card.

To do so, I added a new method to find the correct placement before adding the new element to the designer surface.

# How Verified

Verified manually on the adaptive cards website.

# Outstanding Issues

- `Media` element does not always update the heigh of the entire card correctly, so some of the peers overlap. I was able to repro on older version of the site as well.
- Input elements should not be able to get focus outside of preview mode (ex: checkbox within `Input.Toggle` is still focusable)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7514)


source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -839,6 +843,32 @@ export class CardDesignerSurface {
         }
     }
 
+    // Outstanding issue: cannot drop an element into a ColumnSet
+    // Find the index of the designer surface where the dragged peer should be dropped
+    calculateDraggedPeerIndex(currentRoot: DesignerPeers.DesignerPeer): boolean {
+        // Need to be able to access the index of the dragged peer in the container
+        if (this._draggedPeer instanceof DesignerPeers.CardElementPeer) {
```
jwoo-msft, line 850: what is CardElementPeer?

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -839,6 +843,32 @@ export class CardDesignerSurface {
...
+            for (let i = 0; i < currentRoot.getChildCount(); i++) {
```
jwoo-msft, line 851: what is currentRoot?

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -839,6 +843,32 @@ export class CardDesignerSurface {
         }
     }
 
+    // Outstanding issue: cannot drop an element into a ColumnSet
+    // Find the index of the designer surface where the dragged peer should be dropped
+    calculateDraggedPeerIndex(currentRoot: DesignerPeers.DesignerPeer): boolean {
```
jwoo-msft, line 848: nit: the name doesn't seem to match what it does.

it returns a boolean value, but the method name indicates it returns an integral value.

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
anna-dingler: `currentRoot` is the element we are currently recursing through. It is initially the entire card, and as we come across additional containers, we count the elements in those as well.

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -1384,14 +1384,24 @@ export abstract class DesignerPeer extends DraggableElement {
...
 
-    addElementsToDesignerSurface(designerSurface: HTMLElement, processChildren: boolean = false) {
-        designerSurface.appendChild(this.renderedElement);
-
+    addElementsToDesignerSurface(designerSurface: HTMLElement, processChildren: boolean = false, index: number = -1): number {
         if (processChildren) {
-            for (let i = 0; i < this.getChildCount(); i++) {
-                this.getChildAt(i).addElementsToDesignerSurface(designerSurface, processChildren);
+            let neighbor = designerSurface.children.item(index + 1);
+            designerSurface.insertBefore(this.renderedElement, neighbor);
```
jwoo-msft, line 1390: it's weird that js doesn't have a way to insert with index.

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
anna-dingler: `CardElementPeer` is the peer associated to a `CardElement` (to the best of my knowledge). Ensuring that is is a `CardElementPeer` lets us get the index of the `CardElement` in its container.

`draggedPeer` is the peer corresponding to the element that is being dragged around the card (either a new element or an existing element that the user is moving). We are trying to find the proper index to insert this peer at.

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -839,6 +843,32 @@ export class CardDesignerSurface {
...
+
+                if (child.getChildCount() > 0) {
+                    // Exit the method if recursive call found the element
+                    if (this.calculateDraggedPeerIndex(child)) {
+                        return true;
+                    }
+                }
+            }
+        }
+        return false;
```
jwoo-msft, line 869: is this._draggedPeerIndex valid value when the method returns with false?

source/nodejs/adaptivecards-designer/src/designer-peers.ts
anna-dingler: Attempting to refactor to use https://developer.mozilla.org/en-US/docs/Web/API/Element/after so that we don't have to find the index

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -6018,6 +6018,9 @@ class ActionCollection {
                 action.render();
 
                 if (action.renderedElement) {
+                    if (_isDesignMode) {
+                        action.renderedElement.tabIndex = -1;
+                    }
```
paulcam206, line 6023: this is goodness. _however_ it means that we need a keyboard-accessible way to expand an `Action.ShowCard` in design mode (feel free to file an issue to revisit this later if that makes more sense to you)

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -2255,7 +2276,10 @@ export class ColumnSetPeer extends TypedCardElementPeer<Adaptive.ColumnSet> {
 
                         this.cardElement.addColumn(column);
 
-                        this.insertChild(CardDesignerSurface.cardElementPeerRegistry.createPeerInstance(this.designerSurface, this, column));
+                        let newPeer = CardDesignerSurface.cardElementPeerRegistry.createPeerInstance(this.designerSurface, this, column);
+                        newPeer.insertAfterNeighbor = true;
+
+                        this.insertChild(newPeer);
```
paulcam206, line 2282: this pattern feels like the `insertAfterNeighbor` state should be passed into `createPeerInstance` or `insertChild`... 

looking a little closer, is this particular to some collection-type elements? all collection-type elements? perhaps we should be applying it based on `DesignerPeerCategory` or a parameter added to `DesignerPeerRegistry.registerPeer()`?

https://github.com/microsoft/AdaptiveCards/blob/7e2306219d368a1e2ec3b641c917b9d0c4a34ef3/source/nodejs/adaptivecards-designer/src/card-designer-surface.ts#L105-L113

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -839,6 +863,39 @@ export class CardDesignerSurface {
         }
     }
 
+    // Find the element directly above the dragged element
+    getPeerHTMLNeighbor(peer: DesignerPeers.DesignerPeer): HTMLElement {
```
paulcam206, line 867: super nitpicky, but this should probably be named `getPeerDOMNeighbor` :)

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -1384,14 +1386,25 @@ export abstract class DesignerPeer extends DraggableElement {
         return result;
     }
 
-    addElementsToDesignerSurface(designerSurface: HTMLElement, processChildren: boolean = false) {
```
paulcam206, line 1387: out of curiosity, under what circumstances would we process children (or not) prior to this change? are those scenarios still applicable?

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -407,16 +407,38 @@ export class CardDesignerSurface {
...
-            peer.addElementsToDesignerSurface(this._designerSurface);
+            peer.addElementsToDesignerSurface(this._designerSurface, neighbor);
 
+            // If we process children, we need to make sure we track the child elements here as well
+            // Test scenario: when we add a row to a table, the TableRow is a container, so the children of the container must be added to the right spot as well
+            // TODO: it might be worth checking the insertAfterNeighbor variable here as well?
+            let currentNeighbor = peer;
             for (let i = 0; i < peer.getChildCount(); i++) {
-                this.addPeer(peer.getChildAt(i));
+                let currentPeer = peer.getChildAt(i);
```
paulcam206, line 430: here and elsewhere in this file

```suggestion
                const currentPeer = peer.getChildAt(i);
```

a best practice (that wasn't being followed before) is to use `const` when you aren't assigning a variable again in the same method -- it's a code correctness thing that will prevent accidental assignments in the future (and establishes author intent around how the variable is supposed to be used). I should note that unlike other languages, JS doesn't enforce constness beyond just assignments, so the following works without error (and wouldn't compile in the equivalent C++):

```javascript
const someObj = { a: "string A", b: "string B" };
someObj.a = "this is perfectly legal";
```

source/nodejs/adaptivecards-designer/src/designer-peers.ts
anna-dingler: From what I could tell, we didn't process the children in two scenarios:

1. Initial rendering
2. When a new peer is added (not via drop)

For the initial load, the behavior is still the same because we do not pass a neighbor.

When a new peer is added, we do not pass a neighbor _unless_ the peer was added via an element action (ex: add a column).

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
anna-dingler: Good call out! I'll go through and update these :)

source/nodejs/adaptivecards-designer/src/designer-peers.ts
anna-dingler: I added `insertAfterNeighbor` so that we could handle adding new elements from an element action. So it is only applicable for `ImageSet`, `ColumnSet`, and `Table`

anna-dingler: The `onPeerAdded` callback is overloaded and is called whenever an element is being dragged over the card (**before** it is dropped) and when we add a new element via element action

https://github.com/anna-dingler/AdaptiveCards/blob/accessibilityUpdates/source/nodejs/adaptivecards-designer/src/card-designer-surface.ts#L409-L418

source/nodejs/adaptivecards/src/card-elements.ts
anna-dingler: Just brough in some changes from [this PR](https://github.com/microsoft/AdaptiveCards/pull/7159) which should fix this!

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -3234,8 +3267,13 @@ export class TablePeer extends TypedCardElementPeer<Adaptive.Table> {
...
 
-                        this.insertChild(CardDesignerSurface.cardElementPeerRegistry.createPeerInstance(this.designerSurface, this, row));
-                        this.updateChildren();
+                        const newPeer = CardDesignerSurface.cardElementPeerRegistry.createPeerInstance(this.designerSurface, this, row);
+                        newPeer.insertAfterNeighbor = true;
+
+                        this.insertChild(newPeer);
+
+                        // TODO: I'm not sure if this call is necessary
+                        // this.updateChildren();
```
anna-dingler, line 3276: Does anyone know why we call `updateChildren()` here?

source/nodejs/adaptivecards-designer/src/designer-peers.ts
anna-dingler: @paulcam206 I might need more context here, but I don't think we can assume that we should determine `insertAfterNeighbor` based on peer type. When we drag an element over the card, `onPeerAdded` is called several times, and we don't want to find the neighbor each time.

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -388,7 +388,7 @@ export class CardDesignerSurface {
         this._cardHost.appendChild(renderedCard);
     }
 
-    private addPeer(peer: DesignerPeers.DesignerPeer) {
+    private addPeer(peer: DesignerPeers.DesignerPeer, neighbor: HTMLElement = undefined) {
```
licanhua, line 391: The code is hard to understand. Is it possible to use signature `addPeer(peer: DesignerPeers.DesignerPeer, bool insertAfterNeighbor)` and then calculate the neighbor during adding? (I don't know if it is easier or harder to read)

anna-dingler: Just updated and it's definitely easier to read - great suggestion! Let me know what you think

licanhua (approved):

source/nodejs/adaptivecards-designer/src/designer-peers.ts
anna-dingler: @dclaux Do you have any info about this?

dclaux:
There are two choices to make navigation between DOM elements via the keyboard (or narration) happen in the intended order:
- Either you rely on the order in which the elements are inserted into the DOM, and you don't use `tabindex` at all (except if something should explicitly not be focusable)
- Or you entirely rely on `tabindex`, and the order in which the elements are inserted into the DOM doesn't matter

It definitely seems to me that the second option would yield a much simpler implementation.

anna-dingler:
> > I'm not sure. This seems quite complicated. Wouldn't setting `tabindex` on each peer (in the right order) be enough?
> 
> It's very interesting. Before dive into it, make sure it will not break any rule of [tabindex](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex)
> 
> image

I've pinging @carlos-zamora to see if he has any context about this :)

anna-dingler:
> There are two choices (that I know of) to make navigation between DOM elements via the keyboard (or narration) happen in the intended order:
> 
> * Either you rely on the order in which the elements are inserted into the DOM, and you don't use `tabindex` at all (except if something should explicitly not be focusable)
> * Or you entirely rely on `tabindex`, and the order in which the elements are inserted into the DOM doesn't matter
> 
> It definitely seems to me that the second option would yield a much simpler implementation. That said, I could be totally wrong.

I spoke to @carlos-zamora (relaying the info here), and I think having the DOM tree in the correct order makes sense. 

If we used a non-zero `tabIndex`, each element on the page would need to be assigned an index. We would also need to update `tabIndex` for each element every time we insert or remove elements; so the implementation wouldn't be simplified by much.

Additionally, it is best practice to use `tabIndex=0` and have the elements follow the natural order.

source/nodejs/adaptivecards-designer/src/designer-peers.ts
anna-dingler: I'm leaving this line commented out since I don't see a behavior difference in the designer and other similar elements (`ColumnSet`, `ImageSet`) do not call `updateChildren()`

compulim:
@anna-dingler thanks Anna for making the right choice. Having the correct DOM tree order makes much more sense as Adaptive Cards is an UI component and live in other apps.

If Adaptive Cards is its own app, yes, you can control `tabindex` and put whatever number in.

But Adaptive Cards is a UI component and hosted in apps that AC has no control/influence over. We should not use `tabindex >= 1` because it will certainly conflict with the hosting app.

PR #7654, opened 2022-07-09
[Designer] Add Carousel to Designer

anna-dingler:
# Description

Add carousel support to the designer

# Sample Card

```
{
    "type": "AdaptiveCard",
    "version": "1.6",
    "body": {
        "type": "Carousel",
        "timer": 5000,
        "pages": [
                {
                    "type": "CarouselPage",
                    "id": "firstCarouselPage",
                    "selectAction": {
                        "type": "Action.OpenUrl",
                        "title": "Click for more information about the first carousel page!",
                        "url": "https://adaptivecards.io/"
                    },
                    "items": [
                    {
                        "type": "Image",
                        "url": "https://adaptivecards.io/content/cats/1.png",
                        "size": "medium"
                    }
                ]
        },
        {
            "type": "CarouselPage",
            "id": "theSecondCarouselPage",
            "items": [
                {
                    "type": "Image",
                    "url": "https://adaptivecards.io/content/cats/2.png",
                    "size": "medium"
                }
            ]
        },
        {
            "type": "CarouselPage",
            "id": "last-carousel-page",
            "items": [
                {
                    "type": "Image",
                    "url": "https://adaptivecards.io/content/cats/3.png",
                    "altText": "That's a cool cat!",
                    "size": "medium"
                }
            ]
        }
        ]
    },
    "actions": [
        {
            "type": "Action.OpenUrl",
        }
    ]
}
```

# How Verified

Tested manually on the adaptivecards site

Carousel-1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7654)


source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -181,6 +202,31 @@ export class Carousel extends Container {
         }
     }
 
+    isCurrentPage(page: CarouselPage): boolean {
```
jwoo-msft, line 205: you don't need to do this, you can use realIndex.

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -268,20 +314,23 @@ export class Carousel extends Container {
             "swiper-button-prev",
             "ac-carousel-left"
         );
+        prevElementDiv.style.zIndex = "20";
```
jwoo-msft, line 317: what is this for?

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -372,7 +424,15 @@ export class Carousel extends Container {
             keyboard: {
                 enabled: true,
                 onlyInViewport: true
-            }
+            },
```
jwoo-msft, line 427: can you use [this pr](https://github.com/microsoft/AdaptiveCards/pull/7656) for the event handler?

jwoo-msft (commented):
there are some overlaps between your work and [this pr](https://github.com/microsoft/AdaptiveCards/pull/7656).
can you review the PR?

source/nodejs/adaptivecards/src/carousel.ts
anna-dingler: Good call out - thanks!

source/nodejs/adaptivecards/src/carousel.ts
anna-dingler: The carousel peers cover the navigation arrows, so updating the zIndex places them in front.

source/nodejs/adaptivecards/src/carousel.ts
anna-dingler: Yes, but I think we need an additional event if we use the navigation dots to switch between non-neighboring pages

source/nodejs/adaptivecards/src/carousel.ts
jwoo-msft: @anna-dingler,
do make sure, you've updated. I experienced the same until i cleared the web browser cache.

source/nodejs/adaptivecards/src/carousel.ts
jwoo-msft: @anna-dingler, the pr has the pagenation (via 'dot') event.

source/nodejs/adaptivecards/src/carousel.ts
jwoo-msft: FYI, https://github.com/nolimits4web/swiper/issues/1696

source/nodejs/adaptivecards-designer/src/designer-peer-treeitem.ts
```diff
@@ -58,4 +59,28 @@ export class DesignerPeerTreeItem extends BaseTreeItem {
...
         return this.owner.getChildAt(index).treeItem;
     }
+
+    protected click(e: MouseEvent) {
+        super.click(e);
+
+        let currentPeer = this.owner;
+
+        while (currentPeer) {
+            if (currentPeer instanceof CarouselPagePeer) {
```
dclaux, line 69: I don't think `DesignerPeerTreeItem` shoul have logic that depends on any specific peer type. That logic should be implemented in the specific peer class.

Here, what you want is when the peer for a carousel page is clicked then that page should be made visible. This extends to all peers: when a peer is selected, its associated card element should be brought into view. So what you should do is add a `bringCardElementIntoView(): boolean` method to `CardElementPeer` that returns `false` (meaning "I didn't do anything") in the base implementation. You ovveride it in `CarouselPagePeer` with the logic you currently have here, and you return `true` when you have actually slid to another page. And here, you can change the logic to simply:
```typescript
while (currentPeer) {
    if (currentPeer.bringIntoView()) {
        break;
    }

    currentPeer = currentPeer.parent;
}
```

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -807,6 +810,10 @@ export class CardDesignerSurface {
 
     startDrag(peer: DesignerPeers.DesignerPeer) {
         if (!this.draggedPeer) {
+			if (peer instanceof DesignerPeers.CarouselPeer) {
+				this.reassignCardElementToCarousel(peer);
```
dclaux: I'm not getting this. If there is no dragged peer and the peer that the user is starting to drag is a `CarouselPeer`, then  move all of the root peer's children that are not `CarouselPeers` into the first page of the Carousel being dragged?

That doesn't seem to make much sense to me, but I might be missing something.

dclaux (changes requested):

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
anna-dingler: For now, carousel is a singleton element. When you add a new carousel, the existing card elements are added to the first carouselPage so we maintain this behavior.

dclaux: I see. Fair enough.

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -118,6 +122,9 @@ export class Carousel extends Container {
 
     private _pages: CarouselPage[] = [];
     private _renderedPages: CarouselPage[];
+    private _carouselPageContainer: HTMLElement;
+
+    currentIndex: number = 0;
```
licanhua, line 127: Is it `private`? or add a get function with private _currentIndex?
Is it only be used in designer mode, possible a name like currentIndexInDesignerMode

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -147,6 +158,16 @@ export class Carousel extends Container {
         return this._pages[index];
     }
 
+    addPage(page: CarouselPage) {
+        if (!page.parent) {
+            this._pages.push(page);
+
+            page.setParent(this);
+        } else {
+            throw new Error(Strings.errors.pageAlreadyBelongsToAnotherCarousel());
```
licanhua, line 167: why do we have this error? If I understand correctly, we only support one carousel.
what's the problem if we do reparent?

source/nodejs/adaptivecards/src/carousel.ts
licanhua, line 317: Can you add a comment on it?
Also will it only applies to design mode?

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -394,6 +398,18 @@ export class CardDesignerSurface {
 
             peer.render();
 
+            if (peer instanceof DesignerPeers.CardElementPeer) {
+                if (!peer.isVisible()) {
```
licanhua, line 402: nit:
```
peer.renderedElement.style.display =  peer.isVisible() ? "initial" : "none";
```

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -624,6 +640,11 @@ export class CardDesignerSurface {
 
                 this.tryDrop({ x: e.x - clientRect.left, y: e.y - clientRect.top }, this.draggedPeer);
             }
+
+            if (this._needsLayoutUpdate) {
```
licanhua, line 644: Can you add comments in code?

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -933,6 +960,27 @@ export class CardDesignerSurface {
...
     }
 
+    reassignCardElementToCarousel(carouselPeer: DesignerPeers.CarouselPeer) {
+
+        const carouselPage = carouselPeer.getChildAt(0);
+
+        if (carouselPage instanceof DesignerPeers.CardElementPeer) {
+
+            // Can't modify the list as we loop through it
+            let children = [];
```
licanhua, line 970: nit: const children = []

source/nodejs/adaptivecards-designer/src/designer-peer-treeitem.ts
anna-dingler: Just incorporated this change :)

source/nodejs/adaptivecards/src/carousel.ts
anna-dingler: Will do. I'll also update so it is only applied in design mode

source/nodejs/adaptivecards/src/carousel.ts
anna-dingler: Good point - I'll remove this conditional :)

source/nodejs/adaptivecards/src/carousel.ts
anna-dingler: We also use currentIndex within `SwiperOptions` to assign the carousel's initial index, so I'm not sure renaming it makes sense.

We re-render the carousel whenever we switch pages in the designer, so this makes sure we save the state.
The default is 0 so it shouldn't affect the carousel in general, but it is a good start if we ever want to add `currentPageIndex` to the schema.

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -933,6 +965,27 @@ export class CardDesignerSurface {
...
     }
 
+    reassignCardElementToCarousel(carouselPeer: DesignerPeers.CarouselPeer) {
+
+        const carouselPage = carouselPeer.getChildAt(0);
+
+        if (carouselPage instanceof DesignerPeers.CardElementPeer) {
+
+            // Can't modify the list as we loop through it
+            const children = [];
```
licanhua, line 975:
```
const nonCarouselChildren = this.rootPeer.children.filter((child) => !(child instanceof DesignerPeers.CarouselPeer}));
nonCarouselChildren.foreach((e) => {carouselPage.tryAdd(e);});
```

or 
```
this.rootPeer.children.filter((child) => !(child instanceof DesignerPeers.CarouselPeer})).foreach((e) => {carouselPage.tryAdd(e);});
```

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -933,6 +965,27 @@ export class CardDesignerSurface {
...
+
+            for (let child of this._rootPeer.children) {
+                children.push(child);
+            }
+
+            for (let child of children) {
+                if (!(child instanceof DesignerPeers.CarouselPeer)) {
+                    carouselPage.tryAdd(child);
```
licanhua, line 983: Add a comment to code like `tryAdd remove element from the children, then append it to the last (possible wrong description here). and this function use this feature to shuffle the order`

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -3313,4 +3334,123 @@ export class TablePeer extends TypedCardElementPeer<Adaptive.Table> {
...
+        const carousel = this.cardElement.parent as Adaptive.Carousel;
+
+        if (carousel) {
+            const containerClientRect = carousel.carouselPageContainer.getBoundingClientRect();
+
+            const leftOffset = carousel.renderedElement.offsetLeft + carousel.carouselPageContainer.offsetLeft;
+
+            boundingRect.right = leftOffset + boundingRect.width;
+            boundingRect.left = leftOffset;
+            boundingRect.bottom = boundingRect.top + containerClientRect?.height;
```
licanhua, line 3430: when containerClientRect is undefined or null?

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -3313,4 +3334,123 @@ export class TablePeer extends TypedCardElementPeer<Adaptive.Table> {
...
+        const boundingRect = super.getBoundingRect();
+
+        const carousel = this.cardElement.parent as Adaptive.Carousel;
+
+        if (carousel) {
+            const containerClientRect = carousel.carouselPageContainer.getBoundingClientRect();
+
+            const leftOffset = carousel.renderedElement.offsetLeft + carousel.carouselPageContainer.offsetLeft;
+
+            boundingRect.right = leftOffset + boundingRect.width;
```
licanhua, line 3428: shift the order and remove leftOffset;

boundingRect.left = carousel.renderedElement.offsetLeft + carousel.carouselPageContainer.offsetLeft;
boundingRect.right = boundingRect.left + boundingRect.width;

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -3313,4 +3334,123 @@ export class TablePeer extends TypedCardElementPeer<Adaptive.Table> {
...
+        const parentCarousel = (this.parent as CarouselPeer);
+
+        return (parentCarousel.cardElement as Adaptive.Carousel).isCurrentPage(this.cardElement as Adaptive.CarouselPage);
+    }
+
+    bringCardElementIntoView(): boolean {
+        const carouselPeer = this.parent as CarouselPeer;
+
+        if (carouselPeer) {
+
```
licanhua, line 3446: nit: remove empty line after if

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
```diff
@@ -147,6 +147,7 @@ export const adaptiveCardSchema =
           "$ref": "#/definitions/Actions"
         },
         "body": {
```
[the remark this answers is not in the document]
anna-dingler: I updated the render method to optionally restore the selected state.

The only issue I ran into was that you need to refresh the layout to update peer location again. I expanded the pointer event listener to watch the entire window instead of just the designer surface, so it is slightly better now - but not the ideal solution.

Carousel-3

source/nodejs/adaptivecards-designer/src/designer-peers.ts
anna-dingler: I added a null check here

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
```diff
@@ -147,6 +147,7 @@ export const adaptiveCardSchema =
           "$ref": "#/definitions/Actions"
```
licanhua, line 147: 1. We should be restricted to only one carousel.
2. After I click 'Preview' button, then all the changes are gone after I added 2 or more carousel.

image

image

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
```diff
@@ -147,6 +147,7 @@ export const adaptiveCardSchema =
...
         },
```
licanhua, line 148: Add more than one CarouselPages, then click `Preview` and uncheck `Preview`, `add a page` button disappeared.
image

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
```diff
@@ -147,6 +147,7 @@ export const adaptiveCardSchema =
...
         "body": {
+          "type": ["object", "array"],
           "description": "The Card Elements to show in the primary card region",
```
licanhua, line 151: Add a Carousel, then click x to delete it, we may need to convert the content back to normal AC card instead of an weird state

image

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
anna-dingler: Is the carousel selectable in this state? I believe losing state when we switch between preview and design mode is the intended behavior

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
licanhua, line 147: @JeanRoca 
From my understanding, we only support one Carousel. 
1. So the best user experience is that we have a button to convert an AC card to Carousel or convert it back to a normal card.
So a Carousel in Containers menu doesn't make sense to me.
image

2. When the AC card is Carousel, we add CarouselPage to the contains, then user can drag the page to Carousel.
image

anna-dingler: I'll look into this!

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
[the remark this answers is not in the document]
anna-dingler: @JeanRoca @KesemZ @paulcam206 What do you think?

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
anna-dingler: The Carousel control mimics the behavior of similar elements that only allow one type of child (Table and TableRow, ColumnSet and Column, ImageSet and Image).

For #1, what would the flow be if someone had a carousel with multiple pages w/ content, and converted it back to an AdaptiveCard?

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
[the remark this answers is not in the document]
JeanRoca: @anna-dingler @licanhua Yes I think adding it to the page would make more sense as each page is technically one AC

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
JeanRoca: @licanhua @anna-dingler we recently got an ask of allowing certain elements (like text blocks for example) to be part of the parent Adaptive Card and not part of the carousel page. This would enable a behavior where the title remains static and paginating through the carousel does not move the title out of place.

If we went with the proposed solution in 1 above, would that cause challenges when/if we try to enable the static element functionality later?

anna-dingler: @JeanRoca I think it depends how complex the static element would be. If it was just a TextBlock (or something simple), the carousel could potentially have a `title` option. However, if we want more complex static elements, it could get tricky. 

For instance, say you converted the card to a carousel, then dragged all of the static content to the card. Accidentally clicking the carousel button and then re-enabling it would move all of the static content into the carousel.

KesemZ: Can we prevent adding carousel within a carousel? that might be easier for us to maintain and also prevent the user from creating a schema to a bad experience...

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
KesemZ: @anna-dingler - i agree. the current behavior (hiding the "add a page" on preview mode) is as expected. this is consistent with the other design/preview workflows

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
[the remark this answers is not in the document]
KesemZ: @anna-dingler @licanhua  - agree with adding it to the page. it also keeps the action within the same UI context

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
KesemZ: not sure what "normal AC" means. I believe that if you delete the carousel - the *whole* carousel (including any pages) should be deleted. IF it makes sense - it might be good to give some warning as the user may have intended to delete just the page and not the whole carousel.

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
KesemZ: The carousel _is_ container so i agree with the current implementation of it being part of the containers, but we should prevent the user from adding a carousel within a page, if possible. 

@anna-dingler and @JeanRoca - please discuss @JeanRoca's comment with @jwoo-msft. I believe that the experience that @JeanRoca is describing will be properties on the carousel implementation (so the carousel won't take the whole AC space) but this can also be implemented as two types of carousels.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -6892,7 +6902,14 @@ export class Container extends ContainerBase {
     protected internalToJSON(target: PropertyBag, context: SerializationContext) {
         super.internalToJSON(target, context);
 
-        context.serializeArray(target, this.getItemsCollectionPropertyName(), this._items);
+        const collectionPropertyName = this.getItemsCollectionPropertyName();
+
+        if ((this._items.length === 1) && (this._items[0].getElementSingletonBehavior() === ElementSingletonBehavior.Only)) {
```
jwoo-msft, line 6907: why are you checking if the length is 1?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -6866,20 +6870,26 @@ export class Container extends ContainerBase {
             if (typeName) {
                 const registration = context.elementRegistry.findByName(typeName);
                 if (registration?.singletonBehavior !== ElementSingletonBehavior.NotAllowed) {
-                    jsonItems = [jsonItems];
-                    parsingSingletonObject = true;
+                    const element = context.parseElement(
```
jwoo-msft, line 6873: what is this for?

source/nodejs/adaptivecards/src/card-elements.ts
anna-dingler: To confirm that this is only one element in the array

source/nodejs/adaptivecards/src/card-elements.ts
anna-dingler: I strengthened the singleton behavior here. The third argument is the forbiddenElements array, which should contain "Carousel" _unless_ the parent container can host singleton elements, and the carousel is the only element in the container.

This behavior will need to change for the updated carousel you're implementing - happy to remove these changes

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
anna-dingler: Would we also want to prevent multiple carousels within one card (assuming we remove some of the singleton behavior)

source/nodejs/adaptivecards/src/card-elements.ts
anna-dingler: We at least need a null check before calling `getElementSingletonBehavior()`. The length check ensures that there is only one element in the provided array.

We could strengthen the singleton behavior here so that the edge case we avoid (ex: `_items = ["carousel", "textblock, ...]` never occurs

source/nodejs/adaptivecards/src/carousel.ts
anna-dingler: @jwoo-msft I tried to refactor to use `realIndex`, but in this method, `realIndex` held the same value as `activeIndex`. 

I was able to get the correct value for `realIndex` in the event handlers though.

source/nodejs/adaptivecards/src/carousel.ts
anna-dingler: @jwoo-msft I refactored to use the event handlers from your PR. The timing is slightly more awkward because the swipe event was called more frequently.

I think using these event handlers will be good though if there is a better solution for updating peer location. I'm hoping to talk to @paulcam206 about this next week.

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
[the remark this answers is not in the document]
KesemZ: It makes sense that the actions will be within the pages. The only interaction that is related to the carousel itself is switching pages

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
paulcam206: I agree with @KesemZ here on the scope of deletion behavior. It's worth noting that the `Undo` button in this situation works perfectly. This behavior is also in line with what happens if you delete a toplevel container -- all of its children will also be removed.

My only slight nitpick here is that it's hard (impossible?) to select a non-empty `CarouselPage` on the designer surface using the mouse. Works fine with keyboard or using mouse in concert with the Card Structure panel. It's pretty corner case-y so I don't think it's higher than a P3 :)

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -111,6 +111,9 @@ export class CardElementPeerRegistry extends DesignerPeerRegistry<CardElementTyp
         this.registerPeer(Adaptive.Table, DesignerPeers.TablePeer, DesignerPeerCategory.Containers, "acd-icon-table");
         this.registerPeer(Adaptive.TableRow, DesignerPeers.TableRowPeer, DesignerPeerCategory.Containers, "acd-icon-tableRow");
         this.registerPeer(Adaptive.TableCell, DesignerPeers.TableCellPeer, DesignerPeerCategory.Containers, "acd-icon-tableCell");
+        this.registerPeer(Adaptive.Carousel, DesignerPeers.CarouselPeer, DesignerPeerCategory.Containers, "acd-icon-carousel");
+        this.registerPeer(Adaptive.CarouselPage, DesignerPeers.CarouselPagePeer, DesignerPeerCategory.Containers, "acd-icon-carouselPage");
+        // TODO: what is the carousel config class?
```
paulcam206, line 116: does this TODO need to be made into a bug, or should it be removed?

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -437,10 +439,12 @@ export class NumberPropertyEditor extends SingleInputPropertyEditor {
 
     protected createInput(context: PropertySheetContext): Adaptive.Input {
         let input = new Adaptive.NumberInput();
-        input.defaultValue = this.getPropertyValue(context);
+
+        input.defaultValue = this.getPropertyValue(context) ? this.getPropertyValue(context) : this.defaultValue;
```
paulcam206, line 443: Avoids calling `getPropertyValue()` twice :)

```suggestion
        input.defaultValue = this.getPropertyValue(context) ?? this.defaultValue;
```

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -3317,4 +3385,135 @@ export class TablePeer extends TypedCardElementPeer<Adaptive.Table> {
...
+    isDraggable(): boolean {
+        return false;
+    }
+
+    getBoundingRect(): Rect {
+        const boundingRect = super.getBoundingRect();
+
+        const carousel = this.cardElement.parent as Adaptive.Carousel;
+
+        if (carousel && carousel.carouselPageContainer) {
```
paulcam206, line 3483: perhaps a little simpler

```suggestion
        if (carousel?.carouselPageContainer) {
```

paulcam206 (approved):
this all looks great! only had a couple of nits to pick 👍

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
anna-dingler: @paulcam206 Did you have any thoughts about this?

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
anna-dingler: @paulcam206 It was more of a question - I wasn't sure what it was used for, so wanted to make sure I wasn't missing anything

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
paulcam206: I think I get @licanhua's argument, but as @anna-dingler notes, `Carousel`'s behavior most closely maps to `ColumnSet`. `Carousel` is a container solely of `CarouselPage` elements (as `ColumnSet` only has `Column` children).

The thing that makes `Carousel` seem odd is the fact that it's a singleton element. I think if anything this points out that we should have a better answer for documentation in the Designer (and make sure we have good docs for `Carousel`).

paulcam206: At any rate, I think this model makes sense for now...

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -1196,6 +1225,9 @@ export abstract class DesignerPeer extends DraggableElement {
         }
 
         element.style.position = "absolute";
+        // Issue: I believe this stops us from interacting with the elements - this could be okay since it shouldn't be an issue in preview mode
+        // Ensure that the peer is in front of the card
+        element.style.zIndex = "10";
```
paulcam206, line 1230: Not sure if this is the responsible code, but I found a z-order issue:

zorder

anna-dingler: Resolved the zIndex issue, but still need to look into why tryDrop isn't working correctly :)

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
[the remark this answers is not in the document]
anna-dingler: @paulcam206 @licanhua I looked into this more, and I'm unsure how we want to move the actions to the CarouselPage. From my understanding, containers only have a `selectAction` property, but that is not the behavior we want here. 

Can an individual element have `actions[]`? If not, I don't think we want to move the actions to the page

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
anna-dingler: @paulcam206 We can fix this by adding padding to the carousel pages. Not sure if we want to enforce that styling though, so I didn't include any changes

I can also file an issue for this one since its P3 :)

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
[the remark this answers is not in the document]
paulcam206: > @paulcam206 @licanhua I looked into this more, and I'm unsure how we want to move the actions to the CarouselPage. From my understanding, containers only have a `selectAction` property, but that is not the behavior we want here.
> 
> Can an individual element have `actions[]`? If not, I don't think we want to move the actions to the page

You would do this by using an [`ActionSet`](https://adaptivecards.io/explorer/ActionSet.html).

However, I don't think this change is necessary. It feels like there's a 50/50 chance that the actions on the original card would make sense across all pages in the `Carousel`. Consider a card with an image and description of a book with a "Buy Now" button, or an expense report card with "Approve" and "Reject" buttons... both would be good candidates for promotion to `Carousel`-based experiences sharing card-level actions. I think that that's true in general if pages in the `Carousel` represent homogenous items.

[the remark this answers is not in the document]
anna-dingler: Since `ActionSet` is a forbidden element of a carousel, we won't be able to implement this anyway. Resolving this :)

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
anna-dingler: I discussed this with @KesemZ, and we are going to stick with the current implementation for now since it is consistent with the current design and would be a major refactor. We can circle back to this when we make more overall changes to the designer.

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -3317,4 +3413,146 @@ export class TablePeer extends TypedCardElementPeer<Adaptive.Table> {
...
+        return false;
+    }
+
+    isDraggable(): boolean {
+        return false;
+    }
+
+    attachOnPageChange() {
+        (this.cardElement as Adaptive.Carousel).onPageChanged = (activeIndex: number, realIndex: number) => {
+            const carouselElement = this.cardElement as Adaptive.Carousel;
```
jwoo-msft, line 3471: how do you use the onPageChanged?
why  do you update pages here?

jwoo-msft (changes requested):
exposing swiper looks scary.
please consider moving some of logic to carousel, so you don't have to expose swiper

source/nodejs/adaptivecards-designer/src/designer-peers.ts
anna-dingler: Update: plan is to disable loop mode while in the design mode, and handle the rerendering with the carousel event

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -109,9 +133,9 @@ export class Carousel extends Container {
     set timer(value: number | undefined) {
         if (value && value < this.hostConfig.carousel.minAutoplayDelay) {
             console.warn(Strings.errors.tooLittleTimeDelay);
-            this.timer = this.hostConfig.carousel.minAutoplayDelay;
+            this.setValue(Carousel.timerProperty, this.hostConfig.carousel.minAutoplayDelay);
```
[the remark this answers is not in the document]
anna-dingler: Using `this.timer` created an infinite loop, so I changed it to `setValue`. I'm not sure what the intended behavior of `@property` is though.

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -448,6 +496,22 @@ export class Carousel extends Container {
         );
     }
 
+	slideTo(index: number) {
```
jwoo-msft, line 499: nit: spacing

jwoo-msft (approved):
looks great!

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -394,6 +404,22 @@ export class CardDesignerSurface {
 
             peer.render();
 
+            if (peer instanceof DesignerPeers.CardElementPeer) {
+                peer.renderedElement.style.display =  peer.isVisible() ? "initial" : "none";
+
+                // If we have a card element that we want to be selected after rendering, save the corresponding peer
+                if (this._persistentSelectedCardElement && this._persistentSelectedCardElement == peer.cardElement) {
```
licanhua, line 411: use `===`

```
_persistentSelectedCardElement === peer.cardElement
```

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -621,6 +647,7 @@ export class CardDesignerSurface {
 
                 this._dragVisual.style.width = renderedCardObjectRect.width + "px";
                 this._dragVisual.style.height = renderedCardObjectRect.height + "px";
+                this._dragVisual.style.zIndex = "500";
```
licanhua, line 650: Can you add comment in code why do you change zIndex and why 500 here

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
paulcam206: yeah, I'd just file a bug and move on :)

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -949,6 +1015,49 @@ export class CardDesignerSurface {
...
+                });
+
+            // If the carousel was not dropped on the root peer, we should reassign it
+            peerToSearch.children.filter((child) => 
+                (child instanceof DesignerPeers.CarouselPeer))
+                .forEach((e) => {
+                    if (e.parent !== this._rootPeer) {
+                        this._rootPeer.tryAdd(e);
+                    }
+                })
```
paulcam206, line 1056: perhaps the pickiest nit I've ever picked 🥇 

```suggestion
                });
```

source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -1963,24 +2020,34 @@ export class CardElementPeer extends DesignerPeer {
...
-            return new Rect(
+            returnRect = new Rect(
                 cardElementBoundingRect.top - designSurfaceOffset.y,
                 cardElementBoundingRect.right - designSurfaceOffset.x,
                 cardElementBoundingRect.bottom - designSurfaceOffset.y,
                 cardElementBoundingRect.left - designSurfaceOffset.x
             );
         }
+
+        if (this.designerSurface.containsCarousel && this.designerSurface.currentCarouselPage && !(this instanceof AdaptiveCardPeer) && !(this instanceof CarouselPeer) && !(this instanceof CarouselPagePeer)) {
```
paulcam206, line 2043: feels like this conditional might warrant a comment :)

paulcam206 (approved):
just a couple super minor things -- looks great! ✅

source/nodejs/adaptivecards-designer/src/adaptive-card-schema.ts
anna-dingler: Filed a bug for this issue https://github.com/microsoft/AdaptiveCards/issues/7737

paulcam206 (approved):

PR #7709, opened 2022-07-25
[Accessibility] Set focus to first element in card for Action.ShowCard

### LINKED ISSUE #7689: [Designer] [Tools- Adaptive card editor]: Keyboard focus is not moving on the calendar combo box after activating the "Set due date".
https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/1916728

anna-dingler:
# Related Issue

Fixes #7689 

# Description

When the user show's a card, the focus remains on the ShowCard action instead of moving to the first focusable element in the now visible card.

Fixed by recursively searching for the first focusable element (input or action) within the card and focusing on it.

# How Verified

Verified manually on the adaptivecards-site.

# Outstanding Issue

How do we want this to behave with other actions? For example, if we have two actions (`ShowCard` and `OpenUrl`) in that order, selecting show card will focus on an element within the card, and `OpenUrl` will never have focus.

ShowCardBug

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7709)

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -5739,6 +5739,19 @@ class ActionCollection {
 
             raiseInlineCardExpandedEvent(action, true);
         }
+
+        if (!this._owner.isDesignMode()) {
+            for (let i = 0; i < action.card.getItemCount(); i++) {
+                const element = this.getNearestFocusableElement(action.card.getItemAt(i));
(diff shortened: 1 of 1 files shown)


paulcam206 (approved):

anna-dingler:
Closing this PR since the original bug was closed.

PR #7725, opened 2022-07-28
[Designer Accessibility] Add icon for remove button

### LINKED ISSUE #7693: [Designer] [Tools- Card Editor -Element Properties]: Screen Reader is announcing incorrect name and role for 'Delete' button.
https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/1898859

anna-dingler:
# Related Issue

Fixes #7693 

# Description

For the delete button, we assigned the `title` to be `X`. This caused the aria-label to be `X`, causing the issues with screen readers.

I updated the delete button to have an icon instead, and aria-label is now set properly.

# How Verified

Verified manually on the AdaptiveCards site.

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7725)


source/nodejs/adaptivecards-designer/src/designer-peers.ts
```diff
@@ -906,7 +906,7 @@ class NameValuePairPropertyEditor extends PropertySheetEntry {
                 valueColumn.addItem(textInput);
 
                 let removeAction = new Adaptive.SubmitAction();
-                removeAction.title = "X";
+                removeAction.iconUrl = require("svg-url-loader!../assets/xmark.svg");
```
licanhua, line 909: I think you want something displayed as `X`, but announced as `Delete`.
this doesn't sound good to me:
1. why do you use svg not a simple png?
2. `Delete` is not in the accessibility area.

I'm wondering if you can use aria-label `removeAction.aria-label='X'; removeAction.title='Delete'; removeAction.aria-description="Delete"`

anna-dingler: So right now, the screen reader announces "X Button Remove" and has the labels and descriptions you noted above. If this is acceptable, I can close this out.

I didn't think we wanted to have the screen reader announce "X" at all which is why I made this change. I also thought an icon seemed like good practice since didn't really mean the letter X here.

For your questions:

1. I used an svg to account for different screen sizes
2. what do you mean by `Delete` is not in the accessibility area?

licanhua: I meant to use [aria-label label](https://www.w3docs.com/snippets/html/what-is-aria-label-and-how-to-use-it.html#:~:text=What%20is%20aria-label%20The%20aria-label%20attribute%20specifies%20a,which%20is%20otherwise%20anonymous%2C%20to%20an%20HTML%20element.). For example:    `<button aria-label="Close" onclick="myDialog.close()">X</button>`

paulcam206: > I meant to use [aria-label label](https://www.w3docs.com/snippets/html/what-is-aria-label-and-how-to-use-it.html#:~:text=What%20is%20aria-label%20The%20aria-label%20attribute%20specifies%20a,which%20is%20otherwise%20anonymous%2C%20to%20an%20HTML%20element.). For example: `<button aria-label="Close" onclick="myDialog.close()">X</button>`

This is the existing behavior - the problem (as @anna-dingler noted) is that the screen reader reads "[button content] Button [aria label]", or in this case, "X Button Remove", which isn't accurate.

paulcam206: I should also note (agree that it's not obvious just from this diff) that `tooltip`s get promoted to `aria-label`s.

paulcam206 (approved):

source/nodejs/adaptivecards-designer/src/designer-peers.ts
anna-dingler: I believe the logic only promotes the `tooltip` to `aria-label` _if_ the `title` is blank. If we have a `title`, `tooltip` is promoted to `aria-description` and `title` is promoted to `aria-label`. So "[aria-label] Button [aria-description]" is being read.

I think this behavior makes sense because you would typically want to promote the title to the `aria-label` if you assigned it text.

If we change it to an icon and leave `title` null, the `tooltip` is properly promoted to `aria-label`. I like the idea of making it an icon because we never wanted the _text_ "x", we wanted the _visual_ of the x.

@licanhua @paulcam206 what do you think?

paulcam206: I agree 👍

licanhua (approved):

PR #7728, opened 2022-07-29
implemented header

### LINKED ISSUE #7529: [AC New Features] Static Header for Carousel
## Description
The horizontal carousel currently does not support a way for users to add a header to their carousel that remains static even after paginating through the carousel. A use case for this would be a card author wanting to use the carousel to display related content on different pages of the carousel. Instead of the same title paginating with the content in the carousel, we should enable something like this:
Carousel

In this example one case observe that the carousel header remains in a fixed position while the carousel content (in this case only images, but could be **any** supported carousel element) paginates.

## Describe the solution you'd like
Capability | Priority
-- | --
Allow text to remain static above carousel content serving as a static carousel header | Must
Support multiple lines in header | Must
When a header is used with a carousel the carousel arrows need to be vertically centered with the carousel content (the content that is paginating) and **not the whole container** | Must
The user must be able to add some spacing between the header and the carousel content | Must
Ability to style header text on a per line basis (size, weight, isSubtle, accent) | Should

## Implementation Platforms

- [X] JS
- [ ] .NET WPF
- [ ] Android
- [ ] iOS
- [ ] UWP

jwoo-msft:
# Related Issue

Fixed #7529

# Description

Added Header according to #7529
* Multiple lines of TextBlock can be added and text's size, fonts and style properties of each TextBlocks are configurable.
* Implied type is supported (see the sample)
* spacing between the header and carousel pages are configurable according to the spec
* isVisible not supported since Action.ToggleVisibility is not supported in Carousel.

Screen Shot 2022-08-16 at 4 52 49 PM
Screen Shot 2022-08-16 at 4 52 11 PM

# Sample Card

CarouselHeader.json

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7728)


JeanRoca:
Based on how the schema was implemented in https://github.com/microsoft/AdaptiveCards/pull/6667 I thought that the Carousel was only allowed to show if it was the **only** element inside of body as specified with the `body: { }` syntax instead of `body: [ ]`.

I just did some testing and currently it seems that other elements **are allowed** to be on an Adaptive Card with a Carousel. Here is an example:
```
{
    "type": "AdaptiveCard",
    "version": "1.6",
    "body": [
        {
            "type": "Container",
            "items": [
                {
                    "type": "TextBlock",
                    "wrap": true,
                    "text": "These are cat pictures",
                    "style": "heading",
                    "fontType": "Default",
                    "size": "ExtraLarge",
                    "weight": "Bolder"
                }
            ]
        },
        {
            "type": "Carousel",
            "timer": 5000,
            "pages": [
                {
                    "type": "CarouselPage",
                    "id": "firstCarouselPage",
                    "selectAction": {
                        "type": "Action.OpenUrl",
                        "title": "Click for more information about the first carousel page!",
                        "url": "https://adaptivecards.io/"
                    },
                    "items": [
                        {
                            "type": "Image",
                            "url": "https://adaptivecards.io/content/cats/1.png",
                            "altText": "Cat",
                            "size": "Medium"
                        }
                    ],
                    "rtl": false
                },
                {
                    "type": "CarouselPage",
                    "id": "theSecondCarouselPage",
                    "items": [
                        {
                            "type": "Image",
                            "url": "https://adaptivecards.io/content/cats/2.png",
                            "altText": "Cat with bandana",
                            "size": "Medium"
                        }
                    ],
                    "rtl": false
                },
                {
                    "type": "CarouselPage",
                    "id": "last-carousel-page",
                    "items": [
                        {
                            "type": "Image",
                            "url": "https://adaptivecards.io/content/cats/3.png",
                            "altText": "That's a cool cat!",
                            "size": "Medium"
                        }
                    ],
                    "rtl": false
                }
            ],
            "rtl": false
        }
    ],
    "actions": [
        {
            "type": "Action.OpenUrl",
            "title": "See more",
        },
        {
            "type": "Action.OpenUrl",
            "title": "Another action",
        }
    ]
}
```

On the [preview designer site](https://adaptivecardsci.z5.web.core.windows.net/designer) this renders:
Carousel
 
This **can** achieve the same outcome as a header, but it can also lead to some broken experiences like:
Carousel2

 @paulcam206 any idea why this is rendering in the Designer even though the schema should not be allowing this?

I am not sure what other broken experiences might arise from this behavior so I think @jwoo-msft change in this PR should be what we go with.

anna-dingler:
@JeanRoca I ran into this as well when I made my carousel changes. My open PR (https://github.com/microsoft/AdaptiveCards/pull/7654) should only allow the carousel elements in an object body `{}`.

You can try these out on the test site for my PR - https://adaptivecardsci.z5.web.core.windows.net/pr/7654

paulcam206:
I have a spec question about accessibility -- in schema 1.5 we added the `style` property to `TextBlock`, which allows a card author to indicate that a given `TextBlock` serves as a section header. Under the hood, this is wired up to a11y properties so that it's exposed property to AT. It doesn't look like we extended this same functionality to `RichTextBlock`, presumably because `RichTextBlock` is probably a bit too featureful for use as document section headers (though I don't remember the discussion around this well enough to know if that was the argument). 

It feels to me like it should be possible (or even default behavior?) for the element in the header here to be marked as a header?

@JeanRoca / @jwoo-msft

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -241,6 +247,15 @@ export class Carousel extends Container {
         const containerForAdorners: HTMLElement = document.createElement("div");
         containerForAdorners.className = this.hostConfig.makeCssClassName("ac-carousel-container");
 
+        if (this.header) {
+            this.header.setParent(this);
+            let renderedHeader = this.header?.render();
```
paulcam206, line 252: nitpick: We already checked for null above, so we should have to check again. Also, `renderedHeader` could be `const`.

```suggestion
            const renderedHeader = this.header.render();
```

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -241,6 +247,15 @@ export class Carousel extends Container {
...
+            if (renderedHeader) {
+                Utils.appendChild(cardLevelContainer, renderedHeader);
+                renderedHeader.className = this.hostConfig.makeCssClassName("ac-carousel-header");
```
paulcam206, line 255: genuinely asking, because I haven't really thought it through -- should we be _replacing_ the `ac-richTextBlock` `className` with `ac-carousel-header`?

paulcam206 (changes requested):
code and sample look good (one minor nit and a small question), but would like to close on spec issue with JP :)

jwoo-msft:
> @JeanRoca I ran into this as well when I made my carousel changes. My open PR (#7654) should only allow the carousel elements in an object body `{}`.
> 
> You can try these out on the test site for my PR - https://adaptivecardsci.z5.web.core.windows.net/pr/7654

Yes, this pr is assuming 7654 pr is in.

jwoo-msft:
> I have a spec question about accessibility -- in schema 1.5 we added the `style` property to `TextBlock`, which allows a card author to indicate that a given `TextBlock` serves as a section header. Under the hood, this is wired up to a11y properties so that it's exposed property to AT. It doesn't look like we extended this same functionality to `RichTextBlock`, presumably because `RichTextBlock` is probably a bit too featureful for use as document section headers (though I don't remember the discussion around this well enough to know if that was the argument).
> 
> It feels to me like it should be possible (or even default behavior?) for the element in the header here to be marked as a header?
> 
> @JeanRoca / @jwoo-msft

TextBlock makes senses to me.

JeanRoca:
> I have a spec question about accessibility -- in schema 1.5 we added the `style` property to `TextBlock`, which allows a card author to indicate that a given `TextBlock` serves as a section header. Under the hood, this is wired up to a11y properties so that it's exposed property to AT. It doesn't look like we extended this same functionality to `RichTextBlock`, presumably because `RichTextBlock` is probably a bit too featureful for use as document section headers (though I don't remember the discussion around this well enough to know if that was the argument).
> 
> It feels to me like it should be possible (or even default behavior?) for the element in the header here to be marked as a header?
> 
> @JeanRoca / @jwoo-msft

I'm sorry I don't think I understand this entirely.

Is the suggestion here to use TextBlock instead of RichTextBlock in order to take advantage of the a11y features that were added in 1.5?

What are the disadvantages of going with RichTextBlock over TextBlock?

jwoo-msft:
> > I have a spec question about accessibility -- in schema 1.5 we added the `style` property to `TextBlock`, which allows a card author to indicate that a given `TextBlock` serves as a section header. Under the hood, this is wired up to a11y properties so that it's exposed property to AT. It doesn't look like we extended this same functionality to `RichTextBlock`, presumably because `RichTextBlock` is probably a bit too featureful for use as document section headers (though I don't remember the discussion around this well enough to know if that was the argument).
> > It feels to me like it should be possible (or even default behavior?) for the element in the header here to be marked as a header?
> > @JeanRoca / @jwoo-msft
> 
> I'm sorry I don't think I understand this entirely.
> 
> Is the suggestion here to use TextBlock instead of RichTextBlock in order to take advantage of the a11y features that were added in 1.5?
> 
> What are the disadvantages of going with RichTextBlock over TextBlock?

TextBlock has style property, and the one of the style property is heading. if the style is set heading, the renderers set heading as a11y role for the text.

JeanRoca:
> > > I have a spec question about accessibility -- in schema 1.5 we added the `style` property to `TextBlock`, which allows a card author to indicate that a given `TextBlock` serves as a section header. Under the hood, this is wired up to a11y properties so that it's exposed property to AT. It doesn't look like we extended this same functionality to `RichTextBlock`, presumably because `RichTextBlock` is probably a bit too featureful for use as document section headers (though I don't remember the discussion around this well enough to know if that was the argument).
> > > It feels to me like it should be possible (or even default behavior?) for the element in the header here to be marked as a header?
> > > @JeanRoca / @jwoo-msft
> > 
> > 
> > I'm sorry I don't think I understand this entirely.
> > Is the suggestion here to use TextBlock instead of RichTextBlock in order to take advantage of the a11y features that were added in 1.5?
> > What are the disadvantages of going with RichTextBlock over TextBlock?
> 
> @JeanRoca TextBlock has style property, and the one of the style property is heading. if the style is set heading, the renderers set heading as a11y role for the text.

I see thank you for clarifying @jwoo-msft . Based on my understanding here, it seems we have two options:

Option 1:
The supported element for a carousel header is a TextBlock.
With this option, we have access to the style property of "Header" that sets the a11y role.

Option 2:
The supported element for a carousel header is a RichTextBlock.
With this option, we gain the ability for inline styling text, but lose the ability to set the Style property of "Header" that sets the a11y role.

I think going with option 1 makes sense here. The only thing I am concerned about is the following scenario:
Carousel2

I think that this scenario would only be possible if RichTextBlock is used as different lines have different styles. The main functionality of having a header can be achieved with just TextBlock and this will solve our customer's request. **Allowing for multi-line styling is currently a nice to have.**

Is there any way we could achieve a similar outcome by only using TextBlock?

jwoo-msft:
> > > > I have a spec question about accessibility -- in schema 1.5 we added the `style` property to `TextBlock`, which allows a card author to indicate that a given `TextBlock` serves as a section header. Under the hood, this is wired up to a11y properties so that it's exposed property to AT. It doesn't look like we extended this same functionality to `RichTextBlock`, presumably because `RichTextBlock` is probably a bit too featureful for use as document section headers (though I don't remember the discussion around this well enough to know if that was the argument).
> > > > It feels to me like it should be possible (or even default behavior?) for the element in the header here to be marked as a header?
> > > > @JeanRoca / @jwoo-msft
> > > 
> > > 
> > > I'm sorry I don't think I understand this entirely.
> > > Is the suggestion here to use TextBlock instead of RichTextBlock in order to take advantage of the a11y features that were added in 1.5?
> > > What are the disadvantages of going with RichTextBlock over TextBlock?
> > 
> > 
> > @JeanRoca TextBlock has style property, and the one of the style property is heading. if the style is set heading, the renderers set heading as a11y role for the text.
> 
> I see thank you for clarifying @jwoo-msft . Based on my understanding here, it seems we have two options:
> 
> Option 1: The supported element for a carousel header is a TextBlock. With this option, we have access to the style property of "Header" that sets the a11y role.
> 
> Option 2: The supported element for a carousel header is a RichTextBlock. With this option, we gain the ability for inline styling text, but lose the ability to set the Style property of "Header" that sets the a11y role.
> 
> I think going with option 1 makes sense here. The only thing I am concerned about is the following scenario: Carousel2 [ Carousel2 ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif) [ ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> 
> I think that this scenario would only be possible if RichTextBlock is used as different lines have different styles. The main functionality of having a header can be achieved with just TextBlock and this will solve our customer's request. **Allowing for multi-line styling is currently a nice to have.**
> 
> Is there any way we could achieve a similar outcome by only using TextBlock?

@JeanRoca 
TextBlock's size applies to all of the text. MarkDown can be used to set bold and italic in section of text and that's about it.
If the new scenario is to be supported, 'header' property should be changed to different name, and instead of having a single TextBlock, we need a container. 
for example, instead of 'header', change it to  something like 'staticContent' and have its value to be a Container.
We could also add style to TextRun, and that would impact other renders meaning they are needed to be updated.
Can you gather all the requirement regardless how important they are?

JeanRoca:
> > > > > I have a spec question about accessibility -- in schema 1.5 we added the `style` property to `TextBlock`, which allows a card author to indicate that a given `TextBlock` serves as a section header. Under the hood, this is wired up to a11y properties so that it's exposed property to AT. It doesn't look like we extended this same functionality to `RichTextBlock`, presumably because `RichTextBlock` is probably a bit too featureful for use as document section headers (though I don't remember the discussion around this well enough to know if that was the argument).
> > > > > It feels to me like it should be possible (or even default behavior?) for the element in the header here to be marked as a header?
> > > > > @JeanRoca / @jwoo-msft
> > > > 
> > > > 
> > > > I'm sorry I don't think I understand this entirely.
> > > > Is the suggestion here to use TextBlock instead of RichTextBlock in order to take advantage of the a11y features that were added in 1.5?
> > > > What are the disadvantages of going with RichTextBlock over TextBlock?
> > > 
> > > 
> > > @JeanRoca TextBlock has style property, and the one of the style property is heading. if the style is set heading, the renderers set heading as a11y role for the text.
> > 
> > 
> > I see thank you for clarifying @jwoo-msft . Based on my understanding here, it seems we have two options:
> > Option 1: The supported element for a carousel header is a TextBlock. With this option, we have access to the style property of "Header" that sets the a11y role.
> > Option 2: The supported element for a carousel header is a RichTextBlock. With this option, we gain the ability for inline styling text, but lose the ability to set the Style property of "Header" that sets the a11y role.
> > I think going with option 1 makes sense here. The only thing I am concerned about is the following scenario: Carousel2
> >       
> >         [
> >           
> >             Carousel2
> >           
> >         ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> >         
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> >         
> >        [ Carousel2 ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> >       
> >         [
> >           
> >             Carousel2
> >           
> >         ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> >         
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> >         
> >        [ ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> > I think that this scenario would only be possible if RichTextBlock is used as different lines have different styles. The main functionality of having a header can be achieved with just TextBlock and this will solve our customer's request. **Allowing for multi-line styling is currently a nice to have.**
> > Is there any way we could achieve a similar outcome by only using TextBlock?
> 
> @JeanRoca TextBlock's size applies to all of the text. MarkDown can be used to set bold and italic in section of text and that's about it. If the new scenario is to be supported, 'header' property should be changed to different name, and instead of having a single TextBlock, we need a container. for example, instead of 'header', change it to something like 'staticContent' and have its value to be a Container. We could also add style to TextRun, and that would impact other renders meaning they are needed to be updated. Can you gather all the requirement regardless how important they are?

@jwoo-msft I went ahead and added the requirements we have in #7529. Let me know if you have any questions on those and I'll be happy to help.

jwoo-msft:
> > > > > > I have a spec question about accessibility -- in schema 1.5 we added the `style` property to `TextBlock`, which allows a card author to indicate that a given `TextBlock` serves as a section header. Under the hood, this is wired up to a11y properties so that it's exposed property to AT. It doesn't look like we extended this same functionality to `RichTextBlock`, presumably because `RichTextBlock` is probably a bit too featureful for use as document section headers (though I don't remember the discussion around this well enough to know if that was the argument).
> > > > > > It feels to me like it should be possible (or even default behavior?) for the element in the header here to be marked as a header?
> > > > > > @JeanRoca / @jwoo-msft
> > > > > 
> > > > > 
> > > > > I'm sorry I don't think I understand this entirely.
> > > > > Is the suggestion here to use TextBlock instead of RichTextBlock in order to take advantage of the a11y features that were added in 1.5?
> > > > > What are the disadvantages of going with RichTextBlock over TextBlock?
> > > > 
> > > > 
> > > > @JeanRoca TextBlock has style property, and the one of the style property is heading. if the style is set heading, the renderers set heading as a11y role for the text.
> > > 
> > > 
> > > I see thank you for clarifying @jwoo-msft . Based on my understanding here, it seems we have two options:
> > > Option 1: The supported element for a carousel header is a TextBlock. With this option, we have access to the style property of "Header" that sets the a11y role.
> > > Option 2: The supported element for a carousel header is a RichTextBlock. With this option, we gain the ability for inline styling text, but lose the ability to set the Style property of "Header" that sets the a11y role.
> > > I think going with option 1 makes sense here. The only thing I am concerned about is the following scenario: Carousel2
> > >       
> > >         [
> > >           
> > >             Carousel2
> > >           
> > >         ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> > >         
> > >         
> > >           
> > >             
> > >               
> > >             
> > >             
> > >               
> > >               
> > >             
> > >           
> > >           [
> > >             
> > >               
> > >             
> > >           ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> > >         
> > >       
> > > ```
> > >     [
> > >       
> > >         Carousel2
> > >       
> > >     ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> > >     
> > >     
> > >       
> > >         
> > >           
> > >         
> > >         
> > >           
> > >           
> > >         
> > >       
> > >       [
> > >         
> > >           
> > >         
> > >       ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> > >     
> > >    [ Carousel2 ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> > >   
> > >     [
> > >       
> > >         Carousel2
> > >       
> > >     ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> > >     
> > >     
> > >       
> > >         
> > >           
> > >         
> > >         
> > >           
> > >           
> > >         
> > >       
> > >       [
> > >         
> > >           
> > >         
> > >       ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> > >     
> > >    [ ](https://user-images.githubusercontent.com/33772802/182461142-f3d53844-105e-4928-a7f3-ed1d50925eb3.gif)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I think that this scenario would only be possible if RichTextBlock is used as different lines have different styles. The main functionality of having a header can be achieved with just TextBlock and this will solve our customer's request. **Allowing for multi-line styling is currently a nice to have.**
> > > Is there any way we could achieve a similar outcome by only using TextBlock?
> > 
> > 
> > @JeanRoca TextBlock's size applies to all of the text. MarkDown can be used to set bold and italic in section of text and that's about it. If the new scenario is to be supported, 'header' property should be changed to different name, and instead of having a single TextBlock, we need a container. for example, instead of 'header', change it to something like 'staticContent' and have its value to be a Container. We could also add style to TextRun, and that would impact other renders meaning they are needed to be updated. Can you gather all the requirement regardless how important they are?
> 
> @jwoo-msft I went ahead and added the requirements we have in #7529. Let me know if you have any questions on those and I'll be happy to help.

love it

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -278,6 +286,24 @@ export class Carousel extends Container {
         const containerForAdorners: HTMLElement = document.createElement("div");
         containerForAdorners.className = this.hostConfig.makeCssClassName("ac-carousel-container");
 
+        if (this.header) {
+            for (let adaptiveTextBlock of this.header as Array<CardElement>)
+            {
+                adaptiveTextBlock.setParent(this);
+                let renderedHeader = adaptiveTextBlock.render();
```
paulcam206, line 293: nitpicking, but the `let`s here look like they could be `const` instead :)

paulcam206 (approved):

PR #7710, opened 2022-07-25
[Designer Accessibility] Make new card dialog more keyboard accessible

anna-dingler:
# Related Issue

Fixes #7687 

# Description

When creating a new card on the designer, you should be able to select a new card (blank or sample) with the enter key. We needed keyboard events in addition to the existing onClick events.

# How Verified

Verified manually on the adaptivecards-site

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7710)

PR #7732, opened 2022-07-29
[Designer] Make new card lists accessible with arrow keys

### LINKED ISSUE #7699: [Tools- Adaptive card editor]: Cards list items present under the "Pick a sample as a starting point" dialog box are not accessible with arrow keys.[Rendering] Bug Title Here
### Target Platforms

Other

### SDK Version

105.0.1300.0 

### Application Name

Adaptive Cards

### Problem Description

**Test Environment:**
OS Version: Dev (OS build 21390.1)
Browser Version: 93.0.916.1 (Official build) dev(64-bit)
URL: https://adaptivecards.io/designer/

**Repro Steps:**
1. Launch the URL: [Adaptive Cards (adaptivecards-ci.azurewebsites.net)](http://adaptivecards-ci.azurewebsites.net/)
2. Navigate to header control ‘Designer’ using tab key and activate it through enter key.
3. Navigate to 'New Card' button using the TAB key & activate it.
4. Observe the issue on the list items.

Actual Result:
Cards list items present under the "Pick a sample as a starting point" dialog box are not accessible with arrow keys. Cards list items are navigable with tab keys not but arrow keys.

**Expected Result:**
Cards list items present under the "Pick a sample as a starting point" dialog box should be accessible with arrow keys not with tab keys.

**User Experience:**
PWD users would be confused and try to access the cards list items with arrow keys and not able to access the cards if the cards list items are not accessible with arrow keys.

**MAS Reference:**
https://liquid.microsoft.com/Web/Object/Read/MS.Accessibility/Requirements/02.01.01

### Screenshots

MAS2 1 1_Cards list items are not accessible with arrow keys

https://user-images.githubusercontent.com/96283152/179941163-9e7a9451-d377-495a-a99c-0ef9bf058355.mp4

### Card JSON

```JSON
NA
```

### Sample Code Language

_No response_

### Sample Code

_No response_

anna-dingler:
# Related Issue

Fixes #7699

# Description

Added support for new card options to be accessible with arrow keys

# How Verified

Verified manually on the Adaptive Cards designer

ArrowNavigation

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7732)


source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -696,11 +694,9 @@ export class CardDesigner extends Designer.DesignContext {
                                     dialog.close()
                                     this.launchImagePopup()
                                 },
-                                onKeyEvent: (e: KeyboardEvent) => {
-                                    if (e.key === ControlConstants.keys.enter) {
-                                        dialog.close();
-                                        this.launchImagePopup();
-                                    }
+                                onKeyEnterEvent: (e: KeyboardEvent) => {
```
licanhua, line 697: why do you need this change? Will customer be impacted and a breaking change?

anna-dingler: I added this to avoid adding the arrow key handlers in this method as well. And so that if users decide to override the key event, the arrow handling will still work.

This isn't a breaking change - I just added `onKeyEvent` in this PR (https://github.com/microsoft/AdaptiveCards/pull/7710) so it hasn't shipped

licanhua (approved):

paulcam206 (changes requested):
Code looks great (no notes!), but some behavioral changes we should consider:

* left/right arrow should also work
  * min bar is that left/up go previous and right/down go next
  * stretch goal would be for up/down to navigate appropriately (e.g. if we're displaying three sample cards per row, and the user has focused the middle card on a row and then presses down, the middle card on the next row should be focused). I should note that I'm not even sure if this is easy/possible, so feel free to punt
* we should allow arrowing from the top section (Blank/From Schema) to the "Explore" section

anna-dingler:
> Code looks great (no notes!), but some behavioral changes we should consider:
> 
> * left/right arrow should also work
>   
>   * min bar is that left/up go previous and right/down go next
>   * stretch goal would be for up/down to navigate appropriately (e.g. if we're displaying three sample cards per row, and the user has focused the middle card on a row and then presses down, the middle card on the next row should be focused). I should note that I'm not even sure if this is easy/possible, so feel free to punt
> * we should allow arrowing from the top section (Blank/From Schema) to the "Explore" section

@paulcam206 I just enabled the left/right arrow keys and updated it so that you can navigate between multiple lists.

Let me know what you think!

source/nodejs/adaptivecards-designer/src/open-sample-dialog.ts
```diff
@@ -51,12 +53,31 @@ class OpenSampleItem {
...
-        element.onkeyup = this.props.onKeyEvent ?? (
-            (e) => {
-                if (e.key === Constants.keys.enter) {
-                    this.onCardSelected();
-                }
-            })
+        element.onkeyup = (e: KeyboardEvent) => {
+            switch (e.key) {
+                case Constants.keys.enter:
+                    if (this.props.onKeyEnterEvent) {
```
licanhua, line 59: Can you add a comment that when will onKeyEnterEvent is not set?

source/nodejs/adaptivecards-designer/src/open-sample-dialog.ts
```diff
@@ -171,6 +214,22 @@ export class OpenSampleDialog extends Dialog {
                 this._output = output;
                 this.close();
             }
+            itemElement.onNavigateToNextList = () => {
+                // Find the index of the current list
+                const currentIndex = this._listElements.indexOf(listElement);
```
licanhua, line 219: I assume currentIndex is always `>=0`, otherwise this code is problematic.
If listElement is not in the list, then currentIndex is -1.

licanhua (approved):

source/nodejs/adaptivecards-designer/src/open-sample-dialog.ts
anna-dingler: Good call - I'll add in a check here just to make sure nothing weird happens! But theoretically, yes the `currentIndex` should be `>=0` since we generate `_listElements` on line 208

anna-dingler: I added the check to make sure current index is not `-1` and included an upper bounds check.

licanhua (approved):

licanhua (approved):

paulcam206 (approved):
:shipit:

Issue #7540, opened 2022-06-15
Preview all modes (s/m/l, light/dark/high contrast) at the same time


KesemZ:

anna-dingler:
@KesemZ 

Do we want the ability to enable/disable this feature?
If we always have this feature enabled, #7538 might not be necessary.

However, I do think it would be fair to disable all 6 views at once since that would be a crowded UI.

KesemZ:
This feature should always be enabled. we can close #7538 as duplicate

anna-dingler:
# Proposal

The customer wants to be able to view all 9 versions of the card from one container. Since all 9 options will not fit on one display, I propose we display all of the card sizes (small, medium, and large) and have a drop down menu to select the color mode (light, dark, and high contrast).

Here is an example of what it would look like in light mode:
<img width="931" alt="MicrosoftTeams-image" src="https://user-images.githubusercontent.com/98650930/180281709-2f99a55c-04b3-47c6-a490-c5425dd902d2.png">

licanhua:
I don't think you can do HC preview. So we only have 6.

For any `Berlin` conf, after user clicks `Preview` button, we may have a group preview combobox which includes selections: `dark group`, `light group` and `dark & light group`, and `default`. In `none`, .

![image](https://user-images.githubusercontent.com/6290692/180285840-9863d7ee-3f44-4468-be83-d3e6ec6e8dd9.png)

KesemZ:
@licanhua  - the ask was to consolidate these, though I'm not sure if the purpose is actually to show all sizes at once, as in the proposal above (which actually makes sense to me as a user) or just get rid of the clutter in the drop down where you have Berlin-dark-small, Berlin-light-small etc. 

@anna-dingler  - can you please run the proposal by the PMs ? If the are good with it then so am I :)

anna-dingler:
@KesemZ  I ran it by @kat-y this morning. @adambarlow @rahulamlekar @JeanRoca what do you think?

licanhua:
consolidation is hard. @anna-dingler can you post screenshot in design mode (not preview) for your design?

anna-dingler:
> consolidation is hard. @anna-dingler can you post screenshot in design mode (not preview) for your design?

@licanhua I assumed the design mode would also display all three sizes so it would essentially look the same. If this is too difficult, I would add another drop-down menu to select the size, and then preview mode would display all three at once.

@paulcam206 do you have any context here?

[...]

PR #7741, opened 2022-08-03
[Designer] Add theme and container size drop-down menus

anna-dingler:
# Related Issue

Progress towards #7540 

# Description

Add infrastructure for including `ContainerSize` and `ColorTheme` menus for individual containers.

If a container has the feature enabled, there will be two new drop-down menus to select the container size and theme.

When you select a size or theme from the menus, the respective properties are updated in the Host Data Payload Editor. If the host data payload is updated, we reflect the change in the drop-down menus as well.

# How Verified

Verified manually on the AdaptiveCards Designer site.

# Future Work

1. Allow display of all size options at once. There will be a fourth option, `View all`, for the container size choice picker.
Open question: How will this behave with the designer surface?
3. Consolidate `Microsoft Teams` and `Cortana Skills` containers

# Outstanding discussion

- We can update the data bind variable names if we do not want to use `size` and `theme`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7741)


source/nodejs/adaptivecards-designer/src/adaptivecards-designer.ts
```diff
@@ -7,18 +7,18 @@ export * from "./adaptivecards-designer-standalone";
 export * from "./containers/index";
 
 export const defaultMicrosoftHosts: containers.HostContainer[] = [
```
licanhua, line 9: nit: Personally, I would make the two added-on to use as less as space as possible.
1. remove the Seperator in between
2. `Container Size` and `Theme` is enough. If you think it's not obvious, you can add tooltip to provide more information

image

jwoo-msft (commented):
Can you make the new size & theme menu always visible even though the host app doesn't have sizes and theme?
Screen Shot 2022-08-17 at 10 56 54 AM
Screen Shot 2022-08-17 at 10 57 31 AM

It was hard to notice the change until I used berlin container.
I think putting the two menu options together with host app selection menu option as a group will improve a11y since users don't have to move through all of the other buttons to change size and etc.

source/nodejs/adaptivecards-designer/src/containers/berlin/berlin-container.ts
```diff
@@ -69,21 +59,28 @@ export class BerlinContainer extends HostContainer {
...
-        return containers;
+    get targetVersion(): Adaptive.Version {
+        return Adaptive.Versions.v1_6;
     }
 
     get isFixedHeight(): boolean {
         return true;
     }
+
+    set containerSize(value: ContainerSize) {
```
licanhua, line 70: I'm wondering if we made the host data editor read-only.
In reality, user can't make any change on it.

image

anna-dingler:
> Can you make the new size & theme menu always visible even though the host app doesn't have sizes and theme? Screen Shot 2022-08-17 at 10 56 54 AM Screen Shot 2022-08-17 at 10 57 31 AM
> 
> It was hard to notice the change until I used berlin container. I think putting the two menu options together with host app selection menu option as a group will improve a11y since users don't have to move through all of the other buttons to change size and etc.

@jwoo-msft I can definitely move the drop down menus to be directly after the host container selection menu.

In terms of visibility, I was following the behavior of our other menus such as `Emulate device:` . If we make them always visible, do you think we should disable the control when a host doesn't have size or theme?

source/nodejs/adaptivecards-designer/src/containers/berlin/berlin-container.ts
anna-dingler: @licanhua Is there any scenario where the host would want to use the designer?

source/nodejs/adaptivecards-designer/src/containers/host-container.ts
```diff
@@ -80,4 +91,40 @@ export abstract class HostContainer {
     get enableDeviceEmulation(): boolean {
         return false;
     }
+
+    // if various containers support different sizes in the future, we can override this method
+    static get supportedContainerSizes(): string[] {
```
licanhua, line 96: remove `static get` and move it to berlin container. don't need place holder for future. For example, the container size is only meaningful for Berlin.

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -431,8 +431,38 @@ export class CardDesigner extends Designer.DesignContext {
...
-            this.activeDeviceEmulationChanged()
+            this._deviceEmulationChoicePicker.isEnabled = !!this.hostContainer.enableDeviceEmulation;
+            this.activeDeviceEmulationChanged();
+        }
+
+        if (this._containerSizeChoicePicker) {
+            this._containerSizeChoicePicker.isEnabled = !!this.hostContainer.supportsMultipleSizes;
+
+            // Update the host parameter data with the value of the choice picker
+            if (this._containerSizeChoicePicker.isEnabled && this._sampleHostData) {
```
licanhua, line 442: Personally, I think we are running into `bad smell`. Each time, we are adding a new control into the toolbar(deviceemulation first, then size and theme) for any container, then it adds everything to the base host-container class even other containers doesn't need it.
Is it possible to reverse the dependencies? For example, have a function:
hostContainer.UpdateToolBar() instead of 
 `if (hostContainer.supportsMultipleSizes....`. 

In this way, we don't need to keep `this._containerSizeChoicePicker.isEnabled` and it's just an internal state of berlin container.
Also doesn't need to make change on host-container most of time.

jwoo-msft:
> > Can you make the new size & theme menu always visible even though the host app doesn't have sizes and theme? Screen Shot 2022-08-17 at 10 56 54 AM Screen Shot 2022-08-17 at 10 57 31 AM
> > It was hard to notice the change until I used berlin container. I think putting the two menu options together with host app selection menu option as a group will improve a11y since users don't have to move through all of the other buttons to change size and etc.
> 
> @jwoo-msft I can definitely move the drop down menus to be directly after the host container selection menu.
> 
> In terms of visibility, I was following the behavior of our other menus such as `Emulate device:` . If we make them always visible, do you think we should disable the control when a host doesn't have size or theme?

that's good point, as long as the menus are together, I don't think making them always visible necessary. my main concern is that after host app is chosen, it's hard to noticed that the new menu are added presently. 
if we make them always visible. would it make sense to show no entries?

source/nodejs/adaptivecards-designer/src/containers/berlin/berlin-container.ts
licanhua: I don't see any except that the one who owns the host and want to prototype the host data feature.

It's reasonable for me to remove `host data editor` or minimize it by default since the toolbar already includes duplicated information.

image

anna-dingler: That's a good call out. I like the idea of minimizing it especially since 3 data editors looks a little cluttered.

I also think its fair to note that adaptivecards developers could end up using `$host` parameters for other functionality as well. Since the designer and the templating library aren't bound to the widgets platform, I'm not sure if we should limit the functionality of the JSON editor to only use size and theme

anna-dingler:
> > > Can you make the new size & theme menu always visible even though the host app doesn't have sizes and theme? Screen Shot 2022-08-17 at 10 56 54 AM Screen Shot 2022-08-17 at 10 57 31 AM
> > > It was hard to notice the change until I used berlin container. I think putting the two menu options together with host app selection menu option as a group will improve a11y since users don't have to move through all of the other buttons to change size and etc.
> > 
> > 
> > @jwoo-msft I can definitely move the drop down menus to be directly after the host container selection menu.
> > In terms of visibility, I was following the behavior of our other menus such as `Emulate device:` . If we make them always visible, do you think we should disable the control when a host doesn't have size or theme?
> 
> that's good point, as long as the menus are together, I don't think making them always visible necessary. my main concern is that after host app is chosen, it's hard to noticed that the new menu are added presently. if we make them always visible. would it make sense to show no entries?

I think showing no entries might be more confusing while using other containers. We could show something along the lines of `No options available` or disable the control to avoid confusion there?

jwoo-msft:
> > > > Can you make the new size & theme menu always visible even though the host app doesn't have sizes and theme? Screen Shot 2022-08-17 at 10 56 54 AM Screen Shot 2022-08-17 at 10 57 31 AM
> > > > It was hard to notice the change until I used berlin container. I think putting the two menu options together with host app selection menu option as a group will improve a11y since users don't have to move through all of the other buttons to change size and etc.
> > > 
> > > 
> > > @jwoo-msft I can definitely move the drop down menus to be directly after the host container selection menu.
> > > In terms of visibility, I was following the behavior of our other menus such as `Emulate device:` . If we make them always visible, do you think we should disable the control when a host doesn't have size or theme?
> > 
> > 
> > that's good point, as long as the menus are together, I don't think making them always visible necessary. my main concern is that after host app is chosen, it's hard to noticed that the new menu are added presently. if we make them always visible. would it make sense to show no entries?
> 
> I think showing no entries might be more confusing while using other containers. We could show something along the lines of `No options available` or disable the control to avoid confusion there?

either options sounds good.

source/nodejs/adaptivecards-designer/src/card-designer.ts
anna-dingler: I'm not sure if this is possible because the toolbar is not contained within the `Container` classes. There's not much that the container itself "owns" unless we want to do a major refactor.

@paulcam206 Do you have any ideas here?

source/nodejs/adaptivecards-designer/src/containers/berlin/berlin-container.ts
paulcam206: I agree with Anna here -- `$host` parameters aren't required to be coming from the Widget Dashboard, but is rather a generic data-passing mechanism

paulcam206:
> @jwoo-msft I can definitely move the drop down menus to be directly after the host container selection menu.
> 
> In terms of visibility, I was following the behavior of our other menus such as `Emulate device:` . If we make them always visible, do you think we should disable the control when a host doesn't have size or theme?

I think it can make sense to either show/hide _or_ enable/disable, but I think it depends on the broader experience. Does the theme picker work to switch between what we currently refer to as "Microsoft Teams - Light" and "Microsoft Teams - Dark"? If it _does_, then it certainly at least makes sense to have the theme dropdown always visible and just disabled for containers that don't have a light/dark theme pair. If the picker _doesn't_ work, should it? If not, we should only be showing the dropdowns while displaying containers to which they apply.

source/nodejs/adaptivecards-designer/src/containers/berlin/berlin-container.ts
anna-dingler: Resolving since the $host parameter data editor is needed for non-Widget users.

anna-dingler:
> > @jwoo-msft I can definitely move the drop down menus to be directly after the host container selection menu.
> > In terms of visibility, I was following the behavior of our other menus such as `Emulate device:` . If we make them always visible, do you think we should disable the control when a host doesn't have size or theme?
> 
> I think it can make sense to either show/hide _or_ enable/disable, but I think it depends on the broader experience. Does the theme picker work to switch between what we currently refer to as "Microsoft Teams - Light" and "Microsoft Teams - Dark"? If it _does_, then it certainly at least makes sense to have the theme dropdown always visible and just disabled for containers that don't have a light/dark theme pair. If the picker _doesn't_ work, should it? If not, we should only be showing the dropdowns while displaying containers to which they apply.

I believe the plan is to use the theme picker for "Microsoft Teams" in the future (not implemented in this PR), so I'll enable/disable the control.

source/nodejs/adaptivecards-designer/src/card-designer.ts
anna-dingler: @licanhua I've filed a ticket for this here (https://github.com/microsoft/AdaptiveCards/issues/7773) so that we can refactor all of the host container classes at the same time.

source/nodejs/adaptivecards-designer/src/containers/host-container.ts
```diff
@@ -3,17 +3,28 @@
 import { GlobalSettings, GlobalRegistry, CardObjectRegistry, CardElement, Action, HostConfig, SerializationContext, Version, Versions } from "adaptivecards";
 import * as hostConfig from "../hostConfigs/sample.json";
 
+export enum ContainerSize {
+    Small = "Small",
```
licanhua, line 7: Please make sure you are using lowercases for large/medium/small in $host

image

source/nodejs/adaptivecards-controls/src/inputwithpopup.ts
```diff
@@ -177,11 +177,14 @@ export abstract class InputWithPopup<TPopupControl extends PopupControl, TValue>
     private _popupControl: TPopupControl;
     private _placeholderText: string;
     private _value: TValue;
```
licanhua, line 179: Can we remove Theme and Size from other containers? 
Size only means something to Berlin.
Theme is a little tricky, but we should not add it until you merge dark/light their contain into one.

image

source/nodejs/adaptivecards-designer/src/containers/host-container.ts
anna-dingler: @licanhua I moved this method to BerlinContainer class. I kept it static though to avoid interating through `_hostContainers` to find the Berlin instance while creating the toolbar.

source/nodejs/adaptivecards-controls/src/inputwithpopup.ts
anna-dingler: @jwoo-msft @paulcam206 What do you think?

JeanRoca: I like how Teams and Outlook have their "Emulate Device" only show up on their containers. I agree that other host apps like Viva and Bot Framework wont need a size dropdown and this should be mainly for the berlin containers.

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -431,8 +430,38 @@ export class CardDesigner extends Designer.DesignContext {
...
-            this.activeDeviceEmulationChanged()
+            this._deviceEmulationChoicePicker.isHidden = !(!!this.hostContainer.enableDeviceEmulation);
+            this.activeDeviceEmulationChanged();
+        }
+
+        if (this._containerSizeChoicePicker) {
+            this._containerSizeChoicePicker.isHidden = !(!!this.hostContainer.supportsMultipleSizes);
+
+            // Update the host parameter data with the value of the choice picker
+            if (this._containerSizeChoicePicker.isEnabled && this._sampleHostData) {
```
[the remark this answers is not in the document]
anna-dingler: I chose `isEnabled` to match aria-disabled and so we have 3 states (shown & active, shown & disabled, and hidden). Does that work?

source/nodejs/adaptivecards-controls/src/inputwithpopup.ts
anna-dingler: To clarify, the size dropdown has already been updated and is only displayed for the Berlin container. The theme container is visible on all of the containers with the idea that Microsoft Teams, Cortana Skills, and Viva Connections will all use it (to be added in the my next PR).

paulcam206 (approved):
some minor comments/optimizations, but looks good :)

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -431,20 +430,20 @@ export class CardDesigner extends Designer.DesignContext {
...
 
             // Update the host parameter data with the value of the choice picker
             if (this._containerSizeChoicePicker.isEnabled && this._sampleHostData) {
                 this.updateHostDataSizeProperty();
 
                 // If the container properties do not align with the choice picker, update the property and recreate the designer surface
-                if (this._containerSizeChoicePicker.value !== this._hostContainer.containerSize) {
-                    this._hostContainer.containerSize = this._containerSizeChoicePicker.value as ContainerSize;
+                if (this._containerSizeChoicePicker.value !== (this._hostContainer as BerlinContainer).containerSize) {
+                    (this._hostContainer as BerlinContainer).containerSize = this._containerSizeChoicePicker.value as ContainerSize;
```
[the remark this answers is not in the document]
anna-dingler: There would be an issue if `this._hostContainer` isn't a `BerlinContainer`, so I'll add an `instanceof` check.

We shouldn't hit that scenario because `this._containerSizeChoicePicker.isEnabled` should only set if hostContainer is a BerlinContainer, but good to have the check in case we change the implementation.

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -431,20 +430,20 @@ export class CardDesigner extends Designer.DesignContext {
         this.recreateDesignerSurface();
 
         if (this._deviceEmulationChoicePicker) {
-            this._deviceEmulationChoicePicker.isEnabled = !!this.hostContainer.enableDeviceEmulation;
+            this._deviceEmulationChoicePicker.isHidden = !(!!this.hostContainer.enableDeviceEmulation);
```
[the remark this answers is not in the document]
anna-dingler: that's a good point. Can customers create custom containers?

The only reason I could think of needing `!!` is if a custom implementation returned `null` for `enableDeviceEmulation`, but would that matter since null evaluates to falsey?

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -859,6 +870,79 @@ export class CardDesigner extends Designer.DesignContext {
...
+
+        // If any of the containers support size or theme, we should go ahead and create the choice pickers
+        for (let container of this._hostContainers) {
+            if (container.supportsMultipleThemes) {
+                supportsTheme = true;
+            }
+            if (container.supportsMultipleSizes) {
+                supportsSize = true;
+            }
+        }
```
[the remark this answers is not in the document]
anna-dingler: This looks good to me - nice suggestion!

source/nodejs/adaptivecards-controls/src/inputwithpopup.ts
anna-dingler: Resolving and created an issue for using the drop-down menu for all containers with a light and dark theme: https://github.com/microsoft/AdaptiveCards/issues/7813

anna-dingler:
Confirmed that both of the drop-down menus will not be present when we exclude the berlin container.

PR #7751, opened 2022-08-09
Carousel Additional Features

jwoo-msft:
# Description
Opened for design discussion of #7529
Although it's added on top Carousel spec, it's not going to be added to the spec as written; only the final version that's written as spec will be added.

# Current State of Discussion
Deciding between option 2b and 3 on header.
Option 2 is strongly considered for spacing.


specs/DesignDiscussions/Carousel.md
```diff
@@ -118,3 +118,69 @@ All elements not mentioned above are allowed inside a carousel.
 
 - [ ] Should custom elements (card elements **and** actions) be allowed in carousel? Controllable by custom element author?
 - [ ] Should we rethink allowing carousel cards inside of an `Action.ShowCard`?
+
+# Header with Per Line Text Style
+Supporting different text styling such as different fonts per line of text is not a must for this feature but we have to consider it as customers may request for it in the future.
```
licanhua, line 123: From @KesemZ 
I'm in favor of option 1 or b1. Let's verify with PMs if the different styles for each line is a requirement. I believe that even if it's not required now, it will be so it's better if we go with b1 and be ahead with the implementation.
"Must implement for all of the renders at some point" - is it's just for the carousel? the carousel is not implemented in other renderers anyways (or is it?....), so can we refactor/add the title when we'll implement the carousel for other renderers? regardless, i believe that there's no urgency there as the title ask is for the JS renderer only at this point.

Spacing - is the use of "\n" and "\r" common and valid for all cases when the user needs multiple lines? if yes then option 1 is probably ok. otherwise we should probably go with option 2

licanhua: From @JeanRoca 
Thank you for creating the proposal. For this specific feature I believe we should:

1. Make as little schema changes as possible
2. Build in a future proof way that saves our team from having to make changes to this in the future and potentially introducing breaking changes

I don't think we should go with an approach that will cause the a11y team to file a bug on us in the future. 

**If we were to choose option 1 and only allow TextBlock, would a user creating a header with multiple lines be able to specify a specific line in the TextBlock to apply the a11y header role to or would it be applied to the entire TextBlock?** If it is applied to the entire TextBlock, then I think it is the same experience as the a11y header role getting applied to all of the TextRuns in RichTextBlock. Ultimately, the a11y header role is only an issue if the user has multiple lines of text in the header and they only want this on a specific line. **If this is not possible with TextBlock, then I think we should go with option 2b as the con listed in 2b would also apply to option 1.**

On the subject of spacing I mainly want to make sure that the spacing property on the Carousel is respected and capable of pushing the Carousel content down if the user desires this. 

For example, let's assume the following screenshot has set the spacing property on the Carousel element to "small".
image

The expectation for this request is that if the user sets spacing to be "large" in the previous example the Carousel content would move down more giving more space to the header. On the other hand if the user specifies spacing to be "none" then the Carousel would appear closer to the header. I believe this is already implemented as it works for every other container/element we have, but wanted to make sure it was called out just in case the Carousel worked differently.

licanhua: I prefer header as just a container:

we may need to support footer someday. which may add like button and other stuff
we may add icon, http link or other things to header.

licanhua: From @KesemZ 
@jwoo-msft - with the answer we got from the accessibility team, sounds like the best implementation is to have 3 seperate headers (with 3 separate descriptions and styles) - it could be
header
Title
sub-title

Will it make sense to implement it this way?

jwoo-msft: @KesemZ 
That is not currently possible in AdaptiveCards. We don't support header levels. Text is either header or not. We have a plan for supporting it in the future. 2a is the closet option we have. It can designate TextRun to be Header. For example,
A TextRun
B TextRun
C TextRun
We can set B TextRun as a header as an example.

paulcam206: > @jwoo-msft - with the answer we got from the accessibility team, sounds like the best implementation is to have 3 seperate headers (with 3 separate descriptions and styles) - it could be header Title sub-title
> 
> Will it make sense to implement it this way?

I know I'm kinda jumping in the middle of this conversation, but @JeanRoca -- do you know what specific scenarios are desired here? Is this an ask to satisfy one partner? more than one?

It will certainly influence which option is appropriate:
* If we definitely need to support more than one line, and those lines conceptually follow a set pattern (e.g. header/title/subtitle):
  * The schema should define and enforce those semantics
  * In JSON-speak, we would likely want a `header` property on `Carousel` that contains `header`, `title`, and `subtitle` properties, each being a `TextBlock`
  * We would need to decide what behavior makes sense (e.g. is it okay to supply a `subtitle`, but no `title` or `header`?)
* If we definitely need to support more than one line of text, and that text does _not_ follow a set schema:
  * We need to support more than one text element (be it `TextBlock` or `RichTextBlock`)
  * We should consider making `Carousel`'s `header` property an array of `TextBlock`s
* Are there asks to support non-text elements (e.g. `Image`) here? If so, we again need to decide if the asks follow a pattern, and decide accordingly.

My guess is that UX designers will take as much as we're willing to give them, and would love for this to be a generic container. However, I think we should resist the urge to enable that amount of variability -- it would be easy to author really bad cards.

I also think that using `RichTextBlock` here makes it too easy to make bad cards (each letter can be in a different font/color/strikethrough/highlight and can have a different `selectAction`). Enabling _limited_ styling via `TextBlock` and markdown seems more reasonable - italics/bold/hyperlinking is okay, but color/font/actions are at the line level.

If we don't know what schema should be applied to lines in the header, then I would propose either option 1 (if we're okay with the one-line limitation) or a 4th option: `Carousel` has a `header` property that's an array of `TextBlock`s.

We can always expand functionality later, but it's difficult to take something back once it's out.

JeanRoca: > > @jwoo-msft - with the answer we got from the accessibility team, sounds like the best implementation is to have 3 seperate headers (with 3 separate descriptions and styles) - it could be header Title sub-title
> > Will it make sense to implement it this way?
> 
> I know I'm kinda jumping in the middle of this conversation, but @JeanRoca -- do you know what specific scenarios are desired here? Is this an ask to satisfy one partner? more than one?
> 
> It will certainly influence which option is appropriate:
> 
> * If we definitely need to support more than one line, and those lines conceptually follow a set pattern (e.g. header/title/subtitle):
>   
>   * The schema should define and enforce those semantics
>   * In JSON-speak, we would likely want a `header` property on `Carousel` that contains `header`, `title`, and `subtitle` properties, each being a `TextBlock`
>   * We would need to decide what behavior makes sense (e.g. is it okay to supply a `subtitle`, but no `title` or `header`?)
> * If we definitely need to support more than one line of text, and that text does _not_ follow a set schema:
>   
>   * We need to support more than one text element (be it `TextBlock` or `RichTextBlock`)
>   * We should consider making `Carousel`'s `header` property an array of `TextBlock`s
> * Are there asks to support non-text elements (e.g. `Image`) here? If so, we again need to decide if the asks follow a pattern, and decide accordingly.
> 
> My guess is that UX designers will take as much as we're willing to give them, and would love for this to be a generic container. However, I think we should resist the urge to enable that amount of variability -- it would be easy to author really bad cards.
> 
> I also think that using `RichTextBlock` here makes it too easy to make bad cards (each letter can be in a different font/color/strikethrough/highlight and can have a different `selectAction`). Enabling _limited_ styling via `TextBlock` and markdown seems more reasonable - italics/bold/hyperlinking is okay, but color/font/actions are at the line level.
> 
> If we don't know what schema should be applied to lines in the header, then I would propose either option 1 (if we're okay with the one-line limitation) or a 4th option: `Carousel` has a `header` property that's an array of `TextBlock`s.
> 
> We can always expand functionality later, but it's difficult to take something back once it's out.

This is great insight Paul I agree with you here. This is a request coming from 1 partner and the requirements they requested can be found in #7529. No current asks for supporting images in the header. It's important to note that even though this is only requested by one partner at the moment, it will likely be used by anyone using the Carousel.

I see now how RichTextBlock can lead people into designing bad cards. Given this, I believe using just TextBlock makes the most sense. By one-line limitation do you mean a single TextBlock? Regardless of which option we choose we must support multiple lines but this should be achievable by having a single TextBlock that wraps. I re-shared mocks of the experience we are looking to enable in our internal chat for reference :)

jwoo-msft: @JeanRoca
How strongly does the customer wants the ability to style individual line? If there wasn't an actual request for it. We should go for option 1. 

@licanhua, do you have strong inclination for option 3?

jwoo-msft: @paulcam206,
looking at the our sample, I noticed that we have been using an array of TextBlock to set up a header in our AdaptiveCard.

Screen Shot 2022-08-10 at 1 37 35 PM
Screen Shot 2022-08-10 at 1 38 39 PM

In comparison to RichTextBlock where styling is more targeted to section of text, an array of TextBlock makes sense.
Screen Shot 2022-08-10 at 1 41 06 PM

If no one objects, I will go ahead with option 4 of array of TextBlock as header.

KesemZ: @jwoo-msft  - yes - I like this option the best.

specs/DesignDiscussions/Carousel.md
```diff
@@ -108,6 +108,49 @@ In order to meet our schedule, we need to avoid some common pain points that wou
...
+As a result, author can shortcircuits specifying the type. When this happens, renderers shall assume the type `TextBlock` and renders the content accordingly.
+Implied type shall be remained `TextBlock` even when additional types are added to the list of supported types in the future.
+```
+{
+  "type": "Carousel",
+  "header": [
+    {
+      "text": "cat picture",
+      "isSubtle": true,
+	  "size": "small"
```
licanhua, line 132: nit: formatting and not aligned correctly

specs/DesignDiscussions/Carousel.md
```diff
@@ -108,6 +108,49 @@ In order to meet our schedule, we need to avoid some common pain points that wou
...
+Author of card can designate a TextBlock in the array to be a11y header by setting its `style` property to `heading`. 
+Renderers shall apply a11y heading role according to their respective a11y options. 
+
+Header supports an implied type. The implied type of `header` is `TextBlock`.
+As a result, author can shortcircuits specifying the type. When this happens, renderers shall assume the type `TextBlock` and renders the content accordingly.
+Implied type shall be remained `TextBlock` even when additional types are added to the list of supported types in the future.
+```
+{
+  "type": "Carousel",
+  "header": [
```
licanhua, line 128: If we need to support rtl,padding,spacing,isVisible and all other properties a container have for the header, you may need to change the header to a container object instead of a list of text controls

specs/DesignDiscussions/Carousel.md
jwoo-msft: @KesemZ thanks!

specs/DesignDiscussions/Carousel.md
jwoo-msft: Hi @licanhua ,
`header` behaves similar to `body` and `action` of `AdaptiveCard`. Elements in `header` inherits rtl, iVisible, spacing and padding of Carousel as `action` and `body` does with `AdaptiveCard`. Author can also configure them for each element in `header`.
In the future, we can add Container as an allowable elements if there is need for it, then authors can add TextBlocks to the Container.

licanhua: @JeanRoca, @paulcam206 and @jwoo-msft 
Let's make padding as an example.
Assume orange is the head area, and not colored area is the Carousel.

The padding setting of Carousel applies to the green area.
But we don't have setting for the yellow part because we can't set padding to body and header separately.

image

jwoo-msft: @licanhua,

AdaptiveCards do not let card author to set padding explicitly. padding is applied implicitly by style rule. Another example of `padding` is an enumeration value of spacing which is applied between adaptive elements.

licanhua: I leave this to @paulcam206 and @JeanRoca since they have better understanding on customer requirements and the product. 
Padding here is just an example to illustrate the problem. The question is: will we need to set proprieties to header area and carousel area separately with different value? 

For example (It's a fiction case and don't make too much sense in reality), carousel is set to rtl=false, while customer want header area be rtl=true. Because we can't set rtl on header itself, we have to set rtl=true on all elements in the header.

jwoo-msft: resolved in the design discussion meeting.

licanhua (approved):

jwoo-msft:
merged without android pr passing since it's a change to spec.

PR #7757, opened 2022-08-10
[Designer] Update designer service worker

anna-dingler:
# Description

When a new service worker is installed, we need to make sure that it claims all existing clients on activation and does not wait to take over for the prior service worker.

Otherwise, the new service worker would not take over until prior clients are closed/refreshed.

See documentation here:
https://developer.mozilla.org/en-US/docs/Web/API/Clients/claim
https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/skipWaiting

The caching strategy was also changed from StaleWhileRevalidate to NetworkFirst.

# How Verified

Tested on the Adaptive Cards Designer.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7757)


source/nodejs/adaptivecards-site/workbox-config.json
```diff
@@ -8,7 +8,6 @@
 		"samples/*.html",
 		"schemas/**",
 		"**/*.css",
-		"designer/*",
```
anna-dingler, line 11: @paulcam206 I believe removing `designer/*` from the precache resolved the hashing issue we were talking about. Are there any other assets here that you think should be removed from the precache as well?

source/nodejs/adaptivecards-site/service-worker.js
```diff
@@ -5,6 +5,14 @@ workbox.setConfig({
   modulePathPrefix: "node_modules/workbox/"
 });
 
+self.addEventListener("install", (e) => {
+     self.skipWaiting();
+})
+
+self.addEventListener("activate", (e) => {
+    e.waitUntil(clients.claim());
```
licanhua, line 13: Look like it's a nice to have but not a must have.

I'm wondering what's penalty here. With the new change, and if claim is not called, will the service worker check the network more often?

```
When a service worker is initially registered, pages won't use it until they next load. The claim() method causes those pages to be controlled immediately. Be aware that this results in your service worker controlling pages that loaded regularly over the network, or possibly via a different service worker.
```

anna-dingler: @licanhua From my understanding, this piece of code ensures that you don't need to refresh a page in order for the service worker to take over. I'm not sure what the behavior is if you call `skipWaiting` without `claim`.

I spoke to Aaron, and he said the major concern with this is if your activate or install events include major data structure changes. I don't believe we have any changes that would cause an issue, so I went ahead and added this piece.

That being said, the caching strategy updates greatly improved the service worker, and this might not be necessary.

@aarongustafson Do you have any thoughts here?

aarongustafson: As I understand it, without `claim()`, you run the risk of a page potentially running without an active SW or with an out of date one. So if someone had two tabs open to the Designer, the first one might trigger a SW update and activation, taking over Tab 1, but Tab 2 would still be using the old SW until refreshed. Claiming all clients immediately on activation ensures any currently runnign tabs & windows all get the (new) SW ASAP.

source/nodejs/adaptivecards-site/workbox-config.json
paulcam206: I can't think of anything else here :)

paulcam206 (approved):

source/nodejs/adaptivecards-site/service-worker.js
anna-dingler: @licanhua

licanhua: Likely perf is not a concern here. Look's good to me

licanhua (approved):

PR #7762, opened 2022-08-17
[Designer] Allow data binding for all properties in preview mode

### LINKED ISSUE #6631: [Feature Request] support template bool for rtl and other boolean properties.
### Problem Statement

$ is not a valid usage in JSON, so it has problem to convert the template data to bool directly.
```
"rtl": ${bool($rtl)}
```

### Proposed solution
Data binding works as expected as requested in this feature request. To see the correct usage, see [4350](https://github.com/microsoft/AdaptiveCards/issues/4350). Designer needs to be updated so that a serialization doesn't happen before template expansion because templates are stripped off if the expected type are not string.

_No response_

anna-dingler:
# Related Issue

Fixes #6631 

# Description

Added an optional parameter to the `getValue` function that allows the user to get a property from `_rawProperties` instead of `_propertyBag`.

We pass `GlobalSettings.enableFullJsonRoundTrip` to `getValue` when generating JSON from the card.

Note: Data binding will only work while in preview mode.

# Sample Card

```
{
    "type": "AdaptiveCard",
    "version": "1.5",
    "body": [
        {
            "type": "TextBlock",
            "text": "Image test:",
            "wrap": true
        },
        {
            "type": "Image",
            "url": "https://pbs.twimg.com/profile_images/3647943215/d7f12830b3c17a5a9e4afcc370e3a37e_400x400.jpeg",
            "width": "${imageSize}",
            "height": "${imageSize}"
        }
    ]
}
```

Data JSON
```
{
    "imageSize": "100px"
}
```

# How Verified

Verified manually on the AdaptiveCards site

# Questions

Are there any other cases where we use `GlobalSettings.enableFullJsonRoundTrip` that I should be aware of?

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7762)

### DIFF
--- a/.pipelines/android-ci.yml
+++ b/.pipelines/android-ci.yml
@@ -60,9 +60,11 @@ steps:
    echo "Emulator started"
    
   displayName: Creation of emulator for testing
+  enabled: false
 
 - task: Gradle@2
   displayName: gradlew test
+  enabled: false
   inputs:
     wrapperScript: source/android/gradlew
     cwd: source/android
@@ -75,11 +77,13 @@ steps:
    adb shell ls /storage/emulated/0/
    adb pull /storage/emulated/0/Screenshots/
   displayName: 'Pull screenshots script'
+  enabled: false
   workingDirectory: $(Build.SourcesDirectory)
   condition: succeededOrFailed()
 
 - task: PublishBuildArtifacts@1
   displayName: 'Publish images from Screeenshots folder'
+  enabled: false
   inputs:
     PathtoPublish: '$(build.SourcesDirectory)/Screenshots'
   condition: succeededOrFailed()
--- a/.pipelines/android-ci.yml
+++ b/.pipelines/android-ci.yml
@@ -68,6 +68,7 @@ steps:
     cwd: source/android
     tasks: connectedAndroidTest
     jdkVersionOption: 1.11
+  enabled: false
 
 - script: |
    mkdir Screenshots
--- a/.pipelines/android-ci.yml
+++ b/.pipelines/android-ci.yml
@@ -68,7 +68,6 @@ steps:
     cwd: source/android
     tasks: connectedAndroidTest
     jdkVersionOption: 1.11
-  enabled: false
 
 - script: |
    mkdir Screenshots
--- a/source/nodejs/adaptivecards/src/serialization.ts
+++ b/source/nodejs/adaptivecards/src/serialization.ts
@@ -1086,20 +1086,20 @@ export abstract class SerializableObject {
         }
     }
 
-    protected getValue(prop: PropertyDefinition, overrideProperty: boolean = false): any {
+    protected getValue(prop: PropertyDefinition, shouldOverrideProperty: boolean = false): any {
         const name = prop.getInternalName();
         if (this._propertyBag.hasOwnProperty(name)) {
-            const currentValue = this._propertyBag[name];
+            const propertyBagValue = this._propertyBag[name];
 
-            if (overrideProperty && this._rawProperties.hasOwnProperty(name)) {
-                const overrideValue = this._rawProperties[name];
+            if (shouldOverrideProperty && this._rawProperties.hasOwnProperty(name)) {
+                const rawPropertyValue = this._rawProperties[name];
 
-                if (currentValue !== overrideValue && typeof overrideValue === "string" && overrideValue.includes("${")) {
-                    return overrideValue;
+                if (propertyBagValue !== rawPropertyValue && typeof rawPropertyValue === "string" && rawPropertyValue.includes("${")) {
+                    return rawPropertyValue;
                 }
             }
-            return currentValue;
-        } else if (overrideProperty && this._rawProperties.hasOwnProperty(name)) {
+            return propertyBagValue;
+        } else if (shouldOverrideProperty && this._rawProperties.hasOwnProperty(name)) {
             return this._rawProperties[name];
         }
         return prop.defaultValue;
--- a/source/nodejs/adaptivecards/src/serialization.ts
+++ b/source/nodejs/adaptivecards/src/serialization.ts
@@ -1088,19 +1088,11 @@ export abstract class SerializableObject {
 
     protected getValue(prop: PropertyDefinition, shouldOverrideProperty: boolean = false): any {
         const name = prop.getInternalName();
-        if (this._propertyBag.hasOwnProperty(name)) {
-            const propertyBagValue = this._propertyBag[name];
 
-            if (shouldOverrideProperty && this._rawProperties.hasOwnProperty(name)) {
-                const rawPropertyValue = this._rawProperties[name];
-
-                if (propertyBagValue !== rawPropertyValue && typeof rawPropertyValue === "string" && rawPropertyValue.includes("${")) {
-                    return rawPropertyValue;
-                }
-            }
-            return propertyBagValue;
-        } else if (shouldOverrideProperty && this._rawProperties.hasOwnProperty(name)) {
-            return this._rawProperties[name];
+		if (shouldOverrideProperty && this._rawProperties.hasOwnProperty(name)) {
+			return this._rawProperties[name];
+		} else if (this._propertyBag.hasOwnProperty(name)) {
+			return this._propertyBag[name];
         }
         return prop.defaultValue;
     }
--- a/source/nodejs/adaptivecards/src/serialization.ts
+++ b/source/nodejs/adaptivecards/src/serialization.ts
@@ -1086,10 +1086,23 @@ export abstract class SerializableObject {
         }
     }
 
-    protected getValue(prop: PropertyDefinition): any {
-        return this._propertyBag.hasOwnProperty(prop.getInternalName())
-            ? this._propertyBag[prop.getInternalName()]
-            : prop.defaultValue;
+    protected getValue(prop: PropertyDefinition, overrideProperty: boolean = false): any {
+        const name = prop.getInternalName();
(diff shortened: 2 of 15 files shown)


jwoo-msft (commented):
Hi @anna-dingler ,
This doesn't work for boolean property.
for example, if I want to use data binding for wrap, it doesn't work.
```
{
    "type": "AdaptiveCard",
    "version": "1.5",
    "body": [
        {
            "type": "TextBlock",
            "text": "Image test:",
            "wrap": "${shouldWrap}"
        },
        {
            "type": "Image",
            "url": "https://pbs.twimg.com/profile_images/3647943215/d7f12830b3c17a5a9e4afcc370e3a37e_400x400.jpeg",
            "width": "${imageSize}",
            "height": "${imageSize}"
        }
    ]
}
```
```
{
    "imageSize": "100px",
    "shouldWrap": false
}
```

anna-dingler:
> {
>     "imageSize": "100px",
>     "shouldWrap": false
> }

@jwoo-msft I tested tried those payloads, and it worked for me. We're you using preview mode?

I'll update the description, but the data binding only works in preview mode.

jwoo-msft:
> {
>     "imageSize": "100px",
>     "shouldWrap": false
> }



> 
no, it doesn't work for me in preview mode. can you share the pr site you used?

jwoo-msft:
> > {
> > "imageSize": "100px",
> > "shouldWrap": false
> > }
> 
> > 
> 
> no, it doesn't work for me in preview mode. can you share the pr site you used?



> > {
> > "imageSize": "100px",
> > "shouldWrap": false
> > }
> 
> > 
> 
> no, it doesn't work for me in preview mode. can you share the pr site you used?

it works now; could be caching issue on my side.

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -1086,10 +1086,23 @@ export abstract class SerializableObject {
...
-    protected getValue(prop: PropertyDefinition): any {
-        return this._propertyBag.hasOwnProperty(prop.getInternalName())
-            ? this._propertyBag[prop.getInternalName()]
-            : prop.defaultValue;
+    protected getValue(prop: PropertyDefinition, overrideProperty: boolean = false): any {
+        const name = prop.getInternalName();
+        if (this._propertyBag.hasOwnProperty(name)) {
+            const currentValue = this._propertyBag[name];
+
+            if (overrideProperty && this._rawProperties.hasOwnProperty(name)) {
```
jwoo-msft, line 1094: can you change overrideProperty to convey its boolean type?

for example, it can be shouldOverrideProperty, it makes the meaning much clear.

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -1086,10 +1086,23 @@ export abstract class SerializableObject {
...
+                const overrideValue = this._rawProperties[name];
+
+                if (currentValue !== overrideValue && typeof overrideValue === "string" && overrideValue.includes("${")) {
```
jwoo-msft, line 1097: i don't like using the template token, "${" directly. the token can be changed and more than one token.
also if we allow override, i don't think we really need to check if it's template or not.

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -1086,10 +1086,23 @@ export abstract class SerializableObject {
...
-        return this._propertyBag.hasOwnProperty(prop.getInternalName())
-            ? this._propertyBag[prop.getInternalName()]
-            : prop.defaultValue;
+    protected getValue(prop: PropertyDefinition, overrideProperty: boolean = false): any {
+        const name = prop.getInternalName();
+        if (this._propertyBag.hasOwnProperty(name)) {
+            const currentValue = this._propertyBag[name];
+
+            if (overrideProperty && this._rawProperties.hasOwnProperty(name)) {
+                const overrideValue = this._rawProperties[name];
```
jwoo-msft, line 1095: nit: would rawPropertyValue makes more sense than overrideValue?

source/nodejs/adaptivecards/src/serialization.ts
anna-dingler: I think this should be fine. My only concern is for properties where `propertyBag` and `rawProperties` hold different values and there's not templating (ex: enums).

I'm not seeing any issues, but it seems like this could cause some unforeseen errors. Any thoughts?

jwoo-msft: that should be fine. your change is applied during serialization, and we want full round trip.
during deserialization, bad values will get good default values.
you could consider

if (shouldOverride)
elif (has PropertyBagValue)

return property default.

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -1086,10 +1086,23 @@ export abstract class SerializableObject {
         }
     }
 
-    protected getValue(prop: PropertyDefinition): any {
-        return this._propertyBag.hasOwnProperty(prop.getInternalName())
-            ? this._propertyBag[prop.getInternalName()]
-            : prop.defaultValue;
+    protected getValue(prop: PropertyDefinition, shouldOverrideProperty: boolean = false): any {
```
paulcam206, line 1089: minor question/nitpick -- is there any other scenario outside of when `GlobalSettings.enableFullJsonRoundTrip === true` you can envision where we would the `shouldOverrideProperty === true` ?

if so, I'm curious about it (this isn't supposed to be a "gotcha" question or me using the Socratic Method)... if there's no such scenario, then consider splitting out the `_rawProperties`-accessing version of this method into a new one (`getRawValue`, perhaps? could it be `private`?) and have `setValue` call it explicitly when `GlobalSettings.enableFullJsonRoundTrip` is `true`.

this helps eliminate confusion for future devs working in the area -- otherwise they may need to figure out what this `shouldOverrideProperty` parameter does in order to determine if they should supply it or not...

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -1086,10 +1086,23 @@ export abstract class SerializableObject {
...
+        const name = prop.getInternalName();
+        if (this._propertyBag.hasOwnProperty(name)) {
+            const propertyBagValue = this._propertyBag[name];
+
+            if (shouldOverrideProperty && this._rawProperties.hasOwnProperty(name)) {
+                const rawPropertyValue = this._rawProperties[name];
+
+                if (propertyBagValue !== rawPropertyValue && typeof rawPropertyValue === "string" && rawPropertyValue.includes("${")) {
```
paulcam206, line 1097: I haven't stepped through any of this code, so I don't really know, but is there any potential harm from this `.includes("${")` being overly broad?

consider when `rawPropertyValue === "\${I can put anything here}"` (note the `\`)

source/nodejs/adaptivecards/src/serialization.ts
anna-dingler: I included `shouldOverrideProperty` because we don't want to access `_rawProperties` in every scenario even when `GlobalSettings.enableFullJsonRoundTrip === true`.

For instance, when the card is rendered, we call `getValue`. If we use the `_rawProperties` value, we get type errors (enums, templating expressions).

I only wanted `shouldOverrideProperty === GlobalSettings.enableFullJsonRoundTrip` when converting the card to JSON so that this piece of code in the designer gets a JSON with the expressions.

https://github.com/microsoft/AdaptiveCards/blob/4e987c5fbdd9725f9b769efc1c62af022a14e945/source/nodejs/adaptivecards-designer/src/card-designer-surface.ts#L292-L324

source/nodejs/adaptivecards/src/serialization.ts
anna-dingler: I actually had another question regarding this piece of code.

Joseph suggested that I simplify the method to always get the value from `_rawProperties` if `shouldOverrideProperty` and remove the additional checks here.

That seems to be working fine, but it made me wonder why the `string` values are being accepted by the property `toJSON` methods.

For instance, why doesn't `BoolProperty` throw an error here when `value` is a `string`?
https://github.com/microsoft/AdaptiveCards/blob/4e987c5fbdd9725f9b769efc1c62af022a14e945/source/nodejs/adaptivecards/src/serialization.ts#L432-L438

source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -1086,10 +1086,15 @@ export abstract class SerializableObject {
...
     }
 
-    protected getValue(prop: PropertyDefinition): any {
-        return this._propertyBag.hasOwnProperty(prop.getInternalName())
-            ? this._propertyBag[prop.getInternalName()]
-            : prop.defaultValue;
+    protected getValue(prop: PropertyDefinition, shouldOverrideProperty: boolean = false): any {
+        const name = prop.getInternalName();
+
+		if (shouldOverrideProperty && this._rawProperties.hasOwnProperty(name)) {
```
jwoo-msft, line 1092: nit: spacing

jwoo-msft (approved):

.pipelines/android-ci.yml
```diff
@@ -68,6 +68,7 @@ steps:
     cwd: source/android
     tasks: connectedAndroidTest
     jdkVersionOption: 1.11
+  enabled: false
```
paulcam206, line 71: probably best to do this as a separate PR

paulcam206 (approved):

PR #7774, opened 2022-08-24
Enable data binding for the designer

anna-dingler:
# Related Issue
Fixes https://github.com/microsoft/AdaptiveCards/issues/6631

# Description
Updated `setValue` and serialization methods so that values are not deleted from the `propertyBag` if `GlobalSettings.enableFullJsonRoundTrip === true`.

Note: Data binding will only work while in preview mode.

# Sample Card

```
{
    "type": "AdaptiveCard",
    "version": "1.5",
    "body": [
        {
            "type": "TextBlock",
            "text": "Image test:",
            "wrap": true
        },
        {
            "type": "Image",
            "url": "https://pbs.twimg.com/profile_images/3647943215/d7f12830b3c17a5a9e4afcc370e3a37e_400x400.jpeg",
            "width": "${imageSize}",
            "height": "${imageSize}"
        }
    ]
}
```
# Data JSON

```
{
    "imageSize": "100px"
}
```
# How Verified
Verified manually on the AdaptiveCards site

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7774)

### DIFF
--- a/source/nodejs/adaptivecards/src/serialization.ts
+++ b/source/nodejs/adaptivecards/src/serialization.ts
@@ -197,7 +197,9 @@ export abstract class BaseSerializationContext {
             propertyValue === undefined ||
             propertyValue === defaultValue
         ) {
-            delete target[propertyName];
+            if (!GlobalSettings.enableFullJsonRoundTrip) {
+                delete target[propertyName];
+            }
         } else {
             target[propertyName] = propertyValue;
         }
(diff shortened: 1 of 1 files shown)


source/nodejs/adaptivecards/src/serialization.ts
```diff
@@ -197,7 +197,9 @@ export abstract class BaseSerializationContext {
             propertyValue === undefined ||
             propertyValue === defaultValue
         ) {
-            delete target[propertyName];
+            if (!GlobalSettings.enableFullJsonRoundTrip) {
```
anna-dingler, line 200: For these methods, I mimicked the behavior of `serializeValue`. Is there a reason why we wouldn't want to move this check to be one level up?

The only difference is that we would be reassigning `target[propertyName]`

paulcam206 (approved):

jwoo-msft (approved):

Issue #7442, opened 2022-05-09
[Designer] Display templating expansion warnings


anna-dingler:
# Platform

Designer

# Issue

With the latest templating updates, we collect warnings when a boolean expression was not given correctly and the value is defaulted.

These warnings should be displayed in the designer to help card authors.

PR #7798, opened 2022-08-31
[Templating Feature] Update templating library to output an object

### LINKED ISSUE #7779: [Templating]
Templating currently doesn't work for '[' or '{'

This is a critical issue as value of "body" can be either an object or an array, e.g. Carousel.

If we want to enable a scenario where card authors maintain a single card, they should be able to use conditional logic where array type and object type specifier can be specified.

anna-dingler:
# Related Issue

Fixes #7779 

# Description

The JS templating library can now accept an array of AdaptiveCards as the template, and return an object. 

## Changes in templating

After expanding the template, we check if the remaining object is an array. If the object is an array of length 1, we return the AdaptiveCard. If the array is of length 0 or greater than 1, we log an error and return undefined.

## Changes in Designer

If the given card payload is an array of AdpativeCards, we disable design mode and only show the rendered card in preview mode. This is because templating is only applied in preview mode.

# Sample Card

```JSON
[
    {
        "type": "AdaptiveCard",
        "version": "1.5",
        "$when": "${test}",
        "body": [
            {
                "type": "TextBlock",
                "text": "To be displayed if test === true",
                "wrap": true
            }
        ]
    },
    {
        "type": "AdaptiveCard",
        "version": "1.5",
        "$when": "${!test}",
        "body": [
            {
                "type": "TextBlock",
                "text": "To be displayed if test === false",
                "wrap": true
            }
        ]
    }
]
```

Data
```JSON
{
    "test": false
}
```

# How Verified

1. Added three templating unit-test (simple TextBlock, Container vs. Carousel, failure to test template error).
2. Verified manually in the designer.
3. Verified the new UI adjusts to high contrast.

# Future work
Templating expansion warnings/errors need to be displayed in the designer (#7442)


source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -449,9 +457,21 @@ export class CardDesigner extends Designer.DesignContext {
     }
 
     private targetVersionChanged() {
+        // Question: is there a specific reason that we get the payload from the designerSurface first?
```
anna-dingler, line 460: Does anyone know why we do this?

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -208,6 +208,7 @@ export class CardDesignerSurface {
     private _persistentSelectedCardElement: CardElement;
     private _containsCarousel: boolean = false;
     private _currentCarouselPage: DesignerPeers.CarouselPagePeer;
+    private _currentPayload: any;
```
licanhua, line 211: object|undefined?

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -106,8 +110,12 @@ export class CardDesigner extends Designer.DesignContext {
                     '<div style="padding: 8px; display: flex; justify-content: center;">' +
                         '<div>The Card Structure isn\'t available in Preview mode.</div>' +
                     '</div>';
-            }
-            else {
+            } else if (this._isPayloadCardArray) {
+                this.treeViewToolbox.content.innerHTML =
+                    '<div style="padding: 8px; display: flex; justify-content: center;">' +
+                        '<div>The Card Structure isn\'t available for this payload.</div>' +
+                    '</div>';
```
paulcam206, line 117: it's more of a pain, but rather than supplying content via `innerHTML` (which is considered a bad practice generally -- `""` is okay), we should be constructing the element through the DOM API... Something like:

```typescript
let toolboxDisabled = new HTMLElement("div");
toolboxDisabled.style.padding = "8px";
toolboxDisabled.style.display = "flex";
toolboxDisabled.style.justify-content = "center";
toolboxDisabled.textContent = "The card structure isn't available ...";
this._treeViewToolbox.content.appendChild(toolboxDisabled);
```

Maybe drop this in a helper method that lets you pass in the text to display :)

source/nodejs/adaptivecards-designer/src/card-designer.ts
paulcam206: We're overriding the version in the payload (`cardPayload["version"] = ...`) to respond to the user changing the target version. Presumably this is so that we get accurate parsing behavior?

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -591,22 +611,30 @@ export class CardDesigner extends Designer.DesignContext {
...
+            const currentEditorPayloadObject = JSON.parse(currentEditorPayload);
 
-            if (addToUndoStack) {
-                try {
-                    this.addToUndoStack(JSON.parse(currentEditorPayload));
-                }
-                catch {
-                    // Swallow the parse error
+            this._isPayloadCardArray = Array.isArray(currentEditorPayloadObject);
+            if (!this._isPayloadCardArray || (this._isPayloadCardArray && this.designerSurface.isPreviewMode)) {
```
paulcam206, line 618:
```suggestion
            if (!this._isPayloadCardArray || this.designerSurface.isPreviewMode) {
```

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -620,6 +648,16 @@ export class CardDesigner extends Designer.DesignContext {
...
     }
 
+    private updateDesignerSurfaceDisplay(shouldHideDesignerSurface: boolean) {
+        if (shouldHideDesignerSurface) {
+            this._designerHostElement.classList.add("acd-hidden");
+            this._designerSurfaceHiddenElement.classList.remove("acd-hidden");
+        } else {
+            this._designerSurfaceHiddenElement.classList.add("acd-hidden");
+            this._designerHostElement.classList.remove("acd-hidden");
+        }
```
paulcam206, line 658:
```suggestion
        this._designerHostElement.classList.toggle("acd-hidden", shouldHideDesignerSurface);
        this._designerSurfaceHiddenElement.classList.toggle("acd-hidden", !shouldHideDesignerSurface);
```

source/nodejs/tests/unit-tests/template-test-resources/template-array-complex.json
```diff
@@ -0,0 +1,220 @@
...
+        "type": "AdaptiveCard",
+        "$when": "${!$host.carousel}",
+        "body": [
+            {
+                "type": "TextBlock",
+                "text": "This is the regular card",
+                "wrap": true
+            },
+            {
+                "type": "TextBlock",
```
paulcam206, line 12: if we're really going for complex here, consider using the templated version of this sample and adding the `"carousel": true|false;` to the data blob...

source/nodejs/tests/unit-tests/template-test-resources/template-array.output.json
```diff
@@ -0,0 +1,12 @@
+{
```
paulcam206, line 1: I think we might want a couple more testcases here:

```json
{
    "type": "AdaptiveCard",
    "version": "1.5",
    "$when": false,
    "body": [
        {
            "type": "TextBlock",
            "text": "oh no!"
        }
    ]
}
```

```json
[
{
    "type": "AdaptiveCard",
    "version": "1.5",
    "$when": false,
    "body": [
        {
            "type": "TextBlock",
            "text": "oh no!"
        }
    ]
}
]
```

and maybe 

```json
[]
```

paulcam206 (changes requested):

source/nodejs/adaptivecards-designer/src/card-designer.ts
paulcam206: (I might have the negations reversed here)

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
licanhua, line 211: should have a better name. For example, _arrayPayload, _multipleAdapticardsPayload, _needTemplatingPalyload....

source/nodejs/adaptivecards-designer/src/card-designer.ts
anna-dingler: I guess I just don't understand why we don't grab the payload from the editor directly.

I don't think it really matters outside of the array scenario though

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
anna-dingler: I definitely like these names better! Updating now

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -290,7 +291,12 @@ export class CardDesignerSurface {
         let cardToRender: Adaptive.AdaptiveCard = this.card;
 
         if (asPreview) {
-            let inputPayload = this.card.toJSON(this._serializationContext);
+
+            let inputPayload = this._currentPayload;
+
+            if (!this._currentPayload) {
+                inputPayload = this.card.toJSON(this._serializationContext);
```
licanhua, line 298: I'm wondering if you handle errors correctly or not. For example, an invalid json. we need to warning user which line and what's the error

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -830,8 +841,13 @@ export class CardDesignerSurface {
...
         this._serializationContext.clearEvents();
 
-        this.card.parse(payload, this._serializationContext);
-
+        if (!Array.isArray(payload)) {
+            this.card.parse(payload, this._serializationContext);
+            this._currentPayload = undefined;
+        } else {
+            // No need to parse the card here since templating will need to be applied first
+            this._currentPayload = payload;
```
licanhua, line 849: Nice to have: do you want to treat only one item in array the same with only one object?

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -106,8 +110,12 @@ export class CardDesigner extends Designer.DesignContext {
                     '<div style="padding: 8px; display: flex; justify-content: center;">' +
                         '<div>The Card Structure isn\'t available in Preview mode.</div>' +
                     '</div>';
-            }
-            else {
+            } else if (this._isPayloadCardArray) {
+                this.treeViewToolbox.content.innerHTML =
+                    '<div style="padding: 8px; display: flex; justify-content: center;">' +
+                        '<div>The Card Structure isn\'t available for this payload.</div>' +
```
licanhua, line 116: Need to add the reason. For example, `The Card Structure isn\'t available for this payload because multiple adaptivecards are defined...`

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -449,9 +457,21 @@ export class CardDesigner extends Designer.DesignContext {
...
         let cardPayload = this.designerSurface.getCardPayloadAsObject();
 
-        if (typeof cardPayload === "object") {
+        if (this._isPayloadCardArray) {
+            cardPayload = JSON.parse(this.getCurrentCardEditorPayload());
+        }
+
+        if (Array.isArray(cardPayload)) {
+
+            for (let card of cardPayload) {
```
licanhua, line 469: what will happen if there are multiple cards after `$when` evaluation?

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -449,9 +457,21 @@ export class CardDesigner extends Designer.DesignContext {
     }
 
     private targetVersionChanged() {
+        // Question: is there a specific reason that we get the payload from the designerSurface first?
         let cardPayload = this.designerSurface.getCardPayloadAsObject();
 
-        if (typeof cardPayload === "object") {
+        if (this._isPayloadCardArray) {
```
licanhua, line 463: if getCardPayloadAsObject is heavy, a possible improvement 

```
let cardPayload = this._isPayloadCardArray ?  JSON.parse(this.getCurrentCardEditorPayload()) : this.designerSurface.getCardPayloadAsObject();

```

source/nodejs/adaptivecards-designer/src/card-designer.ts
anna-dingler: If there are multiple cards after expanding the template, the templating library will log an error and return undefined.

At this point in the code, we are updating the contents of the JSON editor so that each card contains the correct version; so `$when` evaluation hasn't occurred yet.

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
anna-dingler: I don't think we want to do that. I think treating a one item array the same as one object would give users the impression that a one item array is a valid payload for the renderer.

source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
anna-dingler: I believe rendering and parsing errors are handled within adaptivecards. Once the card has been parsed and rendered, we display the errors in the designer.

@licanhua Is that what you mean by warning the user?

licanhua: yes. Because there is no link to test it. I don't know there is error correctly showed or not

dclaux:
I'm not understanding this. The templating engine has always accepted an array of anything (not just cards, the template engine is 100% generic and independent from AC aside from its package name) as a template.

anna-dingler:
> I'm not understanding this. The templating engine has always accepted an array of anything (not just cards, the template engine is 100% generic and independent from AC aside from its package name) as a template.

@dclaux I misnamed this PR. The changes are focused on ensuring that the templating library always outputs an object (ideally an AdaptiveCard, but templating is independent as you noted above).

We haven't decided if we want to take this change in general, but we're keeping it here for now.

dclaux:
The templating engine should not return an object if it is fed with a template that has a top level array. It should produce an array. It should be the responsibility of the consuming application to extract what it needs to extract from the produced array. Alternatively, the consuming application could extract objects from the array and feed each one to the templating engine independently.

dclaux:
Also re: "if the given card payload is an array" - that is simply not possible: an Adaptive Card cannot be expressed as an array. It must be expressed as an object. An array would never pass validation against the AC schema.

anna-dingler:
Closing since we are not adding this feature right now.

PR #7819, opened 2022-09-08
[JS] Implement Trusted Types

### LINKED ISSUE #6692: [Feature Request] Trusted Types support in JavaScript SDK
### Problem Statement

JavaScript SDK should fully support [Trusted Types API](https://w3c.github.io/webappsec-trusted-types/dist/spec/) so that it can be seamlessly integrated in applications that enforce Trusted Types for all [DOM XSS Injection Sinks](https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-xss-injection-sinks) (such as innerHTML setters) via CSP directive [require-trusted-types-for](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for). 

On a practical side this means hopefully minor re-factoring of methods where string content is interpreted as HTML and possibly appended to DOM. At first glance there are only few locations in code where this is happening (e.g. assignments to innerHTML) so chances are this can be achieved with relatively low effort. 

### Proposed solution

Review source code of the library and identify locations that will require rewrite, groom it and do it. We could start from this list:

- Empty string assignments into .innerHTML
  - Here we should leverage [TrustedTypePolicyFactory.emptyHTML](https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-trustedtypepolicyfactory-emptyhtml)
    - https://github.com/microsoft/AdaptiveCards/blob/45bd264a7b64e7bc452961a83143a557b174a65f/source/nodejs/adaptivecards-react/src/adaptive-card.tsx#L127
    - https://github.com/microsoft/AdaptiveCards/blob/495987d1d4a3eef2e41a271111006a33b147a37b/source/nodejs/adaptivecards-controls/src/calendar.ts#L165
    - https://github.com/microsoft/AdaptiveCards/blob/495987d1d4a3eef2e41a271111006a33b147a37b/source/nodejs/adaptivecards-controls/src/calendar.ts#L285
    - https://github.com/microsoft/AdaptiveCards/blob/0ceb34c39647f0fe811e51554466a7d417855a27/source/nodejs/adaptivecards-controls/src/radiobutton.ts#L78
    - https://github.com/microsoft/AdaptiveCards/blob/0ceb34c39647f0fe811e51554466a7d417855a27/source/nodejs/adaptivecards-controls/src/textbox.ts#L150

- Arbitrary HTML assignments into .innerHTML
  - Here we probably need to create unique [TrustedTypesPolicy](https://w3c.github.io/webappsec-trusted-types/dist/spec/#create-trusted-type-policy-algorithm) for AdaptiveCards SDK and wrap content using `createHTML` method prior assignment. 
    - https://github.com/microsoft/AdaptiveCards/blob/cf1e0a81753b1240cb0325983e67398e31987497/source/nodejs/adaptivecards/src/card-elements.ts#L926
    - https://github.com/microsoft/AdaptiveCards/blob/cf1e0a81753b1240cb0325983e67398e31987497/source/nodejs/adaptivecards/src/card-elements.ts#L1063
    - https://github.com/microsoft/AdaptiveCards/blob/437e57abde755ffaac542fb52408e4eaf4994870/source/nodejs/adaptivecards/src/utils.ts#L121
    - https://github.com/microsoft/AdaptiveCards/blob/66af50085692bf26a58f10e903b3053ac8a2aef3/source/nodejs/adaptivecards-controls/src/inputwithpopup.ts#L192

Typical alternative is to rewrite the library to avoid using DOM XSS Injection Sinks completely. This is possible in some cases (e.g. when HTML is static and can be constructed by standard DOM manipulation methods) but not always. 

We also need to be careful to keep supporting browsers without Trusted Types support. This is usually done by testing whether window.trustedTypes is defined and fall-back to current behavior if it's not. 

Similar implementations in other libraries for reference:
 - [Visual Studio Code](https://github.com/microsoft/vscode/blob/469d4f6a1c1e36cf232358a8a7eb9d49a76ef56a/src/vs/editor/standalone/browser/colorizer.ts#L45)
 - [Webpack](https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/lib/runtime/CreateScriptUrlRuntimeModule.js#L55)
 - [Angular](https://github.com/angular/components/blob/e7a3f683c80601c6e023f25626cffb0aeafa9032/src/material/icon/trusted-types.ts#L71)

licanhua:
Fixes #6692
Trusted when arbitrary HTML assignments into .innerHTML

code examples
```js
const ttPolicy = window.trustedTypes?.createPolicy('adaptivecards-ui-testapp', {
    createHTML: value => value,
});

const trustedHtml = ttPolicy?.createHTML(inputsAsJson) ?? inputsAsJson;
retrievedInputsDiv.innerHTML = trustedHtml as string;
```
TT docs: https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API
A Guide to TT: https://web.dev/trusted-types/
TT w3c spec: https://w3c.github.io/webappsec-trusted-types/dist/spec/

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7819)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2018,7 +2047,7 @@ export class Image extends CardElement {
                 if (this.renderedElement) {
                     const card = this.getRootElement() as AdaptiveCard;
 
-                    this.renderedElement.innerHTML = "";
+                    this.renderedElement;
```
paulcam206, line 2050:
```suggestion
                    clearElement(this.renderedElement);
```

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -51,6 +51,15 @@ import { CardObjectRegistry, GlobalRegistry, ElementSingletonBehavior } from "./
...
 import { MenuItem, PopupMenu } from "./controls";
 
+const ttPolicy = window.trustedTypes?.createPolicy('adaptivecards', {
+    createHTML: value => value,
+});
+
+function clearElement(element: HTMLElement) : void {
+    const trustedHtml = ttPolicy?.createHTML("") ?? "";
+    element.innerHTML = trustedHtml as string;
+}
```
paulcam206, line 61: realizing a bit late that this should probably be:

```suggestion
function clearElement(element: HTMLElement) : void {
    const trustedHtml = window.trustedTypes?.emptyHTML ?? "";
    element.innerHTML = trustedHtml as string;
}
```

source/nodejs/adaptivecards-ui-testapp/src/rendering-utils.ts
```diff
@@ -129,7 +135,8 @@ export function renderCard(cardJson: any, callbackFunction: Function): void {
 
 export function cardRenderedCallback(renderedCard: HTMLElement) {
     const renderedCardDiv = document.getElementById("renderedCardSpace");
-    renderedCardDiv.innerHTML = "";
+    const trustedHtml = ttPolicy?.createHTML("") ?? "";
+    renderedCardDiv.innerHTML = trustedHtml as string;
```
paulcam206, line 139: as I note in `card-elements.ts` below, there's an `emptyHTML` member we can use:

```typescript
function clearElement(element: HTMLElement) : void {
    const trustedHtml = window.trustedTypes?.emptyHTML ?? "";
    element.innerHTML = trustedHtml as string;
}
```

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -1278,7 +1279,7 @@ export class CardDesigner extends Designer.DesignContext {
         root.style.flexDirection = "column";
         root.style.overflow = "hidden";
 
-        root.innerHTML =
+        setInnerHtml(root,
```
paulcam206, line 1282: as I noted in our chat, the designer is going to need much more extensive work than the base renderer, as all of this static HTML should be constructed through DOM APIs or some other mechanism. for now we should probably just revert the designer changes.

source/nodejs/adaptivecards-controls/src/utils.ts
```diff
@@ -138,3 +138,12 @@ export function getAttributeValueAsInt(element: HTMLElement, attributeName: stri
...
 }
+
+const ttPolicy = window.trustedTypes?.createPolicy('adaptivecards-controls', {
+    createHTML: value => value,
+});
+
+export function setInnerHtml(element: HTMLElement, htmlContent: string) : void {
+    const trustedHtml = ttPolicy?.createHTML(htmlContent) ?? htmlContent;
+    element.innerHTML = trustedHtml as string;
+}
```
paulcam206, line 149: with the `emptyHTML` property I mention below, we shouldn't need to define/create a policy -- should only need the `clearElement` helper

source/nodejs/adaptivecards-controls/src/inputwithpopup.ts
```diff
@@ -189,7 +189,7 @@ export abstract class InputWithPopup<TPopupControl extends PopupControl, TValue>
     private updateLabel() {
         if (this._labelElement) {
             if (this._value) {
-                this._labelElement.innerHTML = this.getValueAsString();
+                Utils.setInnerHtml(this._labelElement, this.getValueAsString());
```
paulcam206, line 192: looks like this use of `innerHTML` was overzealous in the first place.

```suggestion
                this._labelElement.innerText = this.getValueAsString()
```

paulcam206 (changes requested):
This change also needs to be tested -- please let me know how that's being done...

source/nodejs/adaptivecards-controls/src/inputwithpopup.ts
licanhua: I didn't get it. Need to fix this issue

image

source/nodejs/adaptivecards-controls/src/utils.ts
licanhua: the policy is shared, and we have to have policy for this

image

I can create another clearElement if you really want to go that way.

source/nodejs/adaptivecards-ui-testapp/src/rendering-utils.ts
licanhua: resolved

source/nodejs/adaptivecards-controls/src/utils.ts
licanhua: fixed

paulcam206 (commented):
changes look good -- how did you test?

licanhua:
> changes look good -- how did you test?

I did manual testing, and the designer looks good.
ideally, I should follow https://web.dev/trusted-types/#identify-trusted-types-violations, but I don't have time to do it right now

licanhua:
> > changes look good -- how did you test?
> 
> I did manual testing, and the designer looks good. ideally, I should follow https://web.dev/trusted-types/#identify-trusted-types-violations, but I don't have time to do it right now

Add this line to head of html and use the [https link](https://adaptivecardsci.z5.web.core.windows.net/pr/7819/designer)
```html
    <meta http-equiv="Content-Security-Policy" content="require-trusted-types-for 'script';" />
```

Here is the error without the fix.
image

source/nodejs/adaptivecards-designer/src/card-designer.ts
```diff
@@ -705,12 +705,12 @@ export class CardDesigner extends Designer.DesignContext {
                                 thumbnail: () => {
                                     const thumbnail = document.createElement("div");
                                     thumbnail.className = "acd-image-container";
-                                    thumbnail.innerHTML = `<div class="acd-image-upload acd-open-sample-item">
+                                    Utils.setInnerHtml(thumbnail, `<div class="acd-image-upload acd-open-sample-item">
```
paulcam206, line 739: It looks like we only use dynamic templated strings for the privacy policy link in `open-image-dialog.ts`. If possible, that should be moved over to using a static string. Everywhere else we're calling `setInnerHtml()`, we should use non-templated strings (i.e. move from using `` ` `` to using `'` + `+` to build the strings). Once that's done, `ttPolicy` should get a comment stating that it's only intended for use in hardcoded content strings.

paulcam206 (changes requested):
one last thing 😃

source/nodejs/adaptivecards-designer/src/card-designer.ts
licanhua: The correct way to address this is that we build the <div> from code. There are two solutions here:
1. I create a task to remove Utils.setInnerHtml
2. or I revert the change on adaptivecards-designer.
which way do you want to go

licanhua: I revert the change to adaptivecards-designer

paulcam206 (approved):

source/nodejs/adaptivecards-designer/src/card-designer.ts
licanhua: issue is tracked by #7842

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -1073,7 +1078,8 @@ export class TextBlock extends BaseTextBlock {
                     this._computedLineHeight * this.maxLines + "px";
             }
 
-            this.renderedElement.innerHTML = this._originalInnerHtml;
+            const originalHtml = TextBlock._ttRoundtripPolicy?.createHTML(this._originalInnerHtml) ?? this._originalInnerHtml;
+            this.renderedElement.innerHTML = originalHtml as string;
```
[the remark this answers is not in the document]
licanhua: This is not a simple change. I appended it to #7842

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -1100,6 +1106,22 @@ export class TextBlock extends BaseTextBlock {
         return false;
     }
 
+    // Markdown processing is handled outside of Adaptive Cards. It's up to the host to ensure that markdown is safely
```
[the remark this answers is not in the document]
licanhua: @paulcam206 and @dclaux, any concern for me to change to signature of onProcessMarkdown to TrustedHTML?

image

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -1100,6 +1106,22 @@ export class TextBlock extends BaseTextBlock {
...
+    // processed.
+    private static readonly _ttMarkdownPolicy = window.trustedTypes?.createPolicy(
+        "markdownPassthroughPolicy",
```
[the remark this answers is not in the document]
licanhua: this is addressed in #7860

PR #7841, opened 2022-09-13
Hariksingh/reactnative picker fix

hariksingh:
# Related Issue
This PR fixes accessibility for choice set(compact) style in Android 

Issue:
For compact style choice set, it renders custom element containing textbox with placeholder/default/selected value with dropdown image and hides the default picker element which shows its own text box in Android. On picker element selected, it shows the dropdown with list of choices.
In case of Android, the custom element was not accessible when talkback was switched on and as a result, functionality was affected.
https://user-images.githubusercontent.com/113006406/189971565-a380bf13-c040-4cb2-a356-b15c3c0373cd.mov

Fix:
https://user-images.githubusercontent.com/113006406/189971504-aed4e72d-b93b-4d2c-bf02-4049f52cca71.mov

Initialising selected value only in case of valid value present in the payload else, it used to call render function twice and change the selectedPickerValue. Due to this, wrong values were seen earlier when talkback was switched on.
Used onFocus and focus method to programmatically open and close the dropdown and disabled the picker in case of Android. onFocus and enabled props are specific to Android platform only and available as part of 1.16.3 version.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7841)


paulcam206:
@hariksingh -- Looks like this PR is targeting the wrong branch... Can you take a look?

Issue #7833, opened 2022-09-12
Test Carousel with Container as parent with selectAction in the Container.


jwoo-msft:
- [x] pagenation works
- [x] navigation works

both select action & carousel actions were triggered

bug: make sure pagenation & navigation get triggered only in this situation

it appears that it's not carousel specific issue

sample card

```json
{
    "type": "AdaptiveCard",
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "version": "1.6",
    "body": [
        {
            "type": "Container",
            "items": [
                {
                    "type": "Container",
                    "items": [
                        {
                            "type": "Carousel",
                            "rtl": false,
                            "pages": [
                                {
                                    "type": "CarouselPage",
                                    "rtl": false,
                                    "items": [
                                        {
                                            "type": "Image",
                                            "url": "https://adaptivecards.io/content/cats/1.png",
                                            "altText": "Cat"
                                        }
                                    ]
                                },
                                {
                                    "type": "CarouselPage",
                                    "items": [
                                        {
                                            "type": "Image",
                                            "url": "https://adaptivecards.io/content/cats/2.png",
                                            "altText": "Cat"
                                        }
                                    ],
                                    "rtl": false
                                }
                            ]
                        }
                    ],
                    "selectAction": {
                        "type": "Action.OpenUrl",
                        "title": "Container Level 2 Select Action"
                    }
                }
            ]
        }
    ]
}
```
another card without carousel that works
```json
{
    "type": "AdaptiveCard",
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "version": "1.6",
    "body": [
        {
            "type": "Container",
            "items": [
                {
                    "type": "Container",
                    "items": [
                        {
                            "type": "ColumnSet",
                            "columns": [
                                {
                                    "type": "Column",
                                    "width": "stretch",
                                    "items": [
                                        {
                                            "type": "Image",
                                            "url": "https://adaptivecards.io/content/cats/1.png"
                                        }
                                    ]
                                },
                                {
                                    "type": "Column",
                                    "width": "stretch",
                                    "items": [
                                        {
                                            "type": "Image",
                                            "url": "https://adaptivecards.io/content/cats/1.png"
                                        }
                                    ],
                                    "selectAction": {
                                        "type": "Action.OpenUrl",
                                        "title": "Column 2 Select Action"
                                    }
                                }
                            ]
                        }
                    ],
                    "selectAction": {
[...]

Issue #7832, opened 2022-09-12
Bug: Style only applies to outer surface of swiper. e.g, textblock within carousel won't adopt the color of style.


jwoo-msft:

jwoo-msft:
closed vis 7838

PR #7848, opened 2022-09-14
[JS Renderer & Designer] Update Carousel element to be allowed in any context

### LINKED ISSUE #7859: [JS Renderer] Allow Carousel in any context
# Platform

* JavaScript

# Issue

Related to #7779.

Per design discussion, we've decided to allow the Carousel element in any context. This change involves removing singleton logic from `adaptivecards`, updating the behavior of the designer, and ensuring that all features work as expected.

anna-dingler:
# Related Issue

Fixes #7859

# Description

Update the carousel control so that it is allowed in any context.

The changes include:
- Update carousel samples
- Do not allow an object within `body`
- Remove singleton restriction for the Carousel
- Remove singleton enum and logic from adaptivecards since there are no more singleton objects
- Treat Carousel as a normal element in the designer
- Store a parent carousel page on all DesignerPeers that belong to a Carousel. This is used to track whether or not they need their location to be updated based on the current page displayed.
- Bug fixes for #7832 and #7833

# Sample Card

Sample cards can be found at `samples/v1.6/Elements`

# How Verified

Verified manually on the AdaptiveCards site via the Designer. Testing tracked [here](https://github.com/microsoft/AdaptiveCards/projects/44).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7848)


samples/v1.6/Elements/Carousel.ScenarioCards.Timer.json
```diff
@@ -2,463 +2,465 @@
...
+	    								}
+	    							]
+	    						}
+	    					]
+	    				}
+	    			]
+	    		},
+	    		{
+	    			"type": "CarouselPage",
+	    			"id": "stockUpdate",
```
paulcam206, line 302: looks like some spaces snuck in here -- probably worth running `prettier` on all everything under `samples/v1.6`

samples/v1.6/Elements/Carousel.Header.json
```diff
@@ -0,0 +1,69 @@
+ {
```
paulcam206, line 1: given that we're being prompted to make this change because folks want to be able to apply `$when` to `Carousel` elements, we should probably have at least one sample making use of templating

for bonus points: make a templated sample that populates carousel pages from stuff in data

paulcam206 (approved):
only a couple minor formatting nitpicks and a desired testcase

:shipit:

samples/v1.6/Elements/Carousel.ScenarioCards.Timer.json
anna-dingler: I updated the formatting on all of those files

samples/v1.6/Elements/Carousel.Header.json
anna-dingler: Just added a few more test cases - let me know what you think!

jwoo-msft: @anna-dingler, can you take a stab at adding the cards?
You can use https://github.com/microsoft/AdaptiveCards/tree/main/samples/Templates/Scenarios, and you can add both scenario card and corresponding data in there.

anna-dingler: @jwoo-msft Do you mean in addition to the test cases I added to `samples/v1.6/Elements`?

samples/Templates/Scenarios/CarouselWhenShowCarousel.template.json
```diff
@@ -0,0 +1,41 @@
+{
+	"type": "AdaptiveCard",
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"version": "1.6",
+	"body": [
+		{
+			"type": "Container",
```
jwoo-msft, line 7: nit: i think it would be nicer to remove this top level Container. The inclusion of the Container in the past was it was only way to include both Carousel and Container card. Since the updated, users don't need this top level container.

jwoo-msft (approved):

PR #7861, opened 2022-09-14
Preview the difference between main branch and release/22.09-js

licanhua:
We are going to ship 22.09 js render. The difference to main branch:
1. remove Carousel
2. remove berlin config

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7861)


source/nodejs/adaptivecards-designer/src/card-designer-surface.ts
```diff
@@ -1015,45 +965,21 @@ export class CardDesignerSurface {
...
-            // Reassigning all elements to the carousel page ensures carousel's singleton behavior
-            const forbiddenChildElements = (carouselPage.cardElement as Adaptive.CarouselPage).getForbiddenChildElements();
-            this.searchAndRemoveForbiddenElements(this._rootPeer, forbiddenChildElements);
-            this._rootPeer.children.filter((child) => 
-                (!(child instanceof DesignerPeers.CarouselPeer)))
-                .forEach((e) => {carouselPage.tryAdd(e);});
-        }
-    }
-
     private searchAndRemoveForbiddenElements(peerToSearch: DesignerPeers.DesignerPeer, forbiddenElements: any) {
```
anna-dingler, line 968: `searchAndRemoveForbiddenElements` is only called by `reassignCardElementsToCarousel`, so this method can be removed as well.

jwoo-msft (commented):
why do you do this way?

if it is a patch release, we have used the last release branch we used to release and add whatever fixes that are asked for; that way we  mitigate the risk of incompatibility for the sdk users.

licanhua:
> why do you do this way?
> 
> if it is a patch release, we have used the last release branch we used to release and cherry-picked whatever fixes that are asked for; that way we mitigate the risk of incompatibility for the sdk users.

It's not a patch release, and I already aligned with @paulcam206  on this

paulcam206:
> > why do you do this way?
> > if it is a patch release, we have used the last release branch we used to release and cherry-picked whatever fixes that are asked for; that way we mitigate the risk of incompatibility for the sdk users.
> 
> It's not a patch release, and I already aligned with @paulcam206 on this

This approach isn't _usually_ what we'd want to do, but given how far `main` has moved since the last `2.x` release (remember we did code formatting passes, etc.), it looked like it'd be tricky to cherry-pick things over without dealing with tons of merge conflicts.

I think with where we are, there'd be a bigger risk in trying to hand-merge things than in taking `main`, subtracting the things we know we don't want, and then making sure we've got a solid test pass done.

PR #7885, opened 2022-09-16
Update to new `lerna` version; audit package dependencies

paulcam206:
As per usual, we do a round of dependency updates prior to releasing a new package. However, this time is a little different. Our beleaguered dependency, `lerna`, has undertaken active development once more! Lots of goodness (and it looks like there's more cool new stuff to come), but the biggest thing for us in this moment is that it's been totally updated to newer dependencies. This change includes both the conversion to the new major version of `lerna`, as well as a series of updates across our repo to get us up-to-date.

One really nice thing? Performing `npx lerna run build` will now use smarts and local caching to reduce build times:

image

Many thanks to @anna-dingler for seeing this through to the finish line (see #7818 for more info).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7885)

### DIFF
--- a/source/nodejs/adaptivecards-designer/package-lock.json
+++ b/source/nodejs/adaptivecards-designer/package-lock.json
@@ -13,7 +13,7 @@
 			},
 			"devDependencies": {
 				"adaptive-expressions": "^4.11.0",
-				"cpy-cli": "^3.1.1",
+				"cpy-cli": "^4.1.0",
 				"dotenv-webpack": "^7.0.3",
 				"monaco-editor": "^0.29.1",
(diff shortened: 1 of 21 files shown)


paulcam206:
@anna-dingler - note that 49a069eda2ca20f7e9358efe1f1d8b1a756b0c61 contains a couple changes since you last looked at this branch :)

jwoo-msft (approved):

licanhua (approved):

PR #7899, opened 2022-09-16
[Designer] Update container styling to fix overflow and sizing

### LINKED ISSUE #7883: [Designer] Overflow not working properly on Berlin Containers
# Issue

Overflow is not working properly for Berlin Containers. We need to switch from padding to margin within ac-adaptiveCard styling.

anna-dingler:
# Related Issue

Fixes #7883 

# Description

Updated the berlin styling to use margins instead of padding because overflow was not working properly.

Also adjusted the height and width because the border is not included in the given sizes.

# How Verified

Verified manually on the Adaptive Cards site.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7899)


source/nodejs/adaptivecards-designer/src/containers/berlin/berlin-container-dark.css
```diff
@@ -10,7 +10,7 @@
 
 .berlin-header {
```
anna-dingler, line 11: @paulcam206 @licanhua Should we update the header styling to use margins as well?

paulcam206: probably...

source/nodejs/adaptivecards-designer/src/containers/berlin/berlin-container-dark.css
```diff
@@ -20,36 +20,36 @@
...
-    height: 146px;
+    height: 82px;
 }
 
 .ac-adaptiveCard {
     border: 0px;
-    padding: 48px 16px 16px 16px !important;
-    border-radius: 8px !important;
+    margin: 48px 16px 16px 16px !important;
+    padding: 0px !important;
```
paulcam206, line 52: are these changes up-to-date with the latest design?

paulcam206 (approved):

source/nodejs/adaptivecards-designer/src/containers/berlin/berlin-container-dark.css
anna-dingler: @paulcam206 

Yes. I switched to margins because overflow still allows content within padding, so we would allow something like this:
image
where the highlighted text is within the margin and it should not be displayed in preview mode.

Now, we cut off the content for the margin:
image


As for the sizing changes, I confirmed with Amanda that the provided size does **not** include the border (1px), so we had to increase all of our sizes by 2px.

Before:
image

After: 
image

licanhua (approved):

anna-dingler:
@JeanRoca When is a good time to make this update?

JeanRoca:
> @JeanRoca When is a good time to make this update?

I spoke with Amanda and she suggested we can take the change as long as it does not impact our partners. I was doing some testing and noticed a couple of things with the new margin. 

- Background images are not bleeding outside the margins.
PR
image

Current designer
image

- Carousel controls should also bleed into the margins so they don't disappear. 
carouseltest

The only 3 things that should be allowed to extend past the margin are carousel pagination dots, carousel next/prev arrows, and background images (either set at the AC level or set in something like the carousel which supports bleed for background images). Let me know if you think of any others that should be able to extend past the margins.

Once these changes are made we can test to make sure it does not impact partners and go from there. Thanks Anna :)

paulcam206 (approved):

paulcam206 (approved):

PR #7910, opened 2022-09-19
Add (typeof window === 'undefined') check for trustedType

### LINKED ISSUE #7904: [ReferenceError] "ReferenceError: window is not defined" on adaptivecards@2.11.0
### Target Platforms

NodeJS

### SDK Version

2.11.0

### Application Name

Node console app

### Problem Description

My project is a node backend app using `adaptivecards` to parse/render/verify card content on server side.
It was using `adaptivecards@2.10.0` and everything worked well. After switching to the latest `adaptivecards@2.11.0`, it throws **ReferenceError: window is not defined** on `require("adaptivecards")`.

ReferenceError: window is not defined
    at Object.<anonymous> (...\node_modules\adaptivecards\lib\utils.js:149:26)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (...\node_modules\adaptivecards\lib\adaptivecards.js:22:14)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)

### Screenshots

image

### Card JSON

```JSON
{
  "type": "AdaptiveCard",
  "version": "1.4",
  "body": [
  ]
}
```

### Sample Code Language

javascript

### Sample Code

const { AdaptiveCard } = require("adaptivecards");  // error occurs at this line

const card = require("./card.json");
const ac = new AdaptiveCard();
ac.parse(card);

licanhua:
Fix #7904 
1. add (typeof window === 'undefined') for everywhere use trustedTypes
2. make unit test to use node environment instead of jsdom which will not have window object
3. bump up version with `npx lerna version --force-publish --no-push --no-git-tag-version` to make min version change

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7910)

### DIFF
--- a/source/nodejs/adaptivecards-react/src/adaptive-card.tsx
+++ b/source/nodejs/adaptivecards-react/src/adaptive-card.tsx
@@ -124,7 +124,7 @@ export const AdaptiveCard = ({
         try {
             card.parse(payload);
             const result = card.render() as HTMLElement;
-            const trustedHtml = window.trustedTypes?.emptyHTML ?? "";
+            const trustedHtml = (typeof window === 'undefined') ? "" : (window.trustedTypes?.emptyHTML ?? "");
             targetRef.current.innerHTML = trustedHtml as string;
             targetRef.current.appendChild(result);
         } catch (cardRenderError) {
--- a/source/nodejs/adaptivecards/src/utils.ts
+++ b/source/nodejs/adaptivecards/src/utils.ts
@@ -190,7 +190,7 @@ export function truncateText(element: HTMLElement, maxHeight: number, lineHeight
  * TextBlock.truncateIfSupported}), but had a bug where it might actually pass through an element
  * for which innerHTML yielded actual HTML (since fixed).
  */
-const ttDeprecatedPolicy = (typeof window === 'undefined') ? null : window.trustedTypes?.createPolicy("adaptivecards#deprecatedExportedFunctionPolicy", {
+const ttDeprecatedPolicy = window.trustedTypes?.createPolicy("adaptivecards#deprecatedExportedFunctionPolicy", {
     createHTML: (value) => value
 });
 
--- a/source/nodejs/adaptivecards/src/utils.ts
+++ b/source/nodejs/adaptivecards/src/utils.ts
@@ -190,7 +190,7 @@ export function truncateText(element: HTMLElement, maxHeight: number, lineHeight
  * TextBlock.truncateIfSupported}), but had a bug where it might actually pass through an element
  * for which innerHTML yielded actual HTML (since fixed).
  */
-const ttDeprecatedPolicy = window.trustedTypes?.createPolicy("adaptivecards#deprecatedExportedFunctionPolicy", {
+const ttDeprecatedPolicy = (typeof window === 'undefined') ? undefined : window.trustedTypes?.createPolicy("adaptivecards#deprecatedExportedFunctionPolicy", {
     createHTML: (value) => value
 });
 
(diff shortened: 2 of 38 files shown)


source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -190,7 +190,7 @@ export function truncateText(element: HTMLElement, maxHeight: number, lineHeight
  * TextBlock.truncateIfSupported}), but had a bug where it might actually pass through an element
  * for which innerHTML yielded actual HTML (since fixed).
  */
-const ttDeprecatedPolicy = window.trustedTypes?.createPolicy("adaptivecards#deprecatedExportedFunctionPolicy", {
+const ttDeprecatedPolicy = (typeof window === 'undefined') ? undefined : window.trustedTypes?.createPolicy("adaptivecards#deprecatedExportedFunctionPolicy", {
```
tosmolka, line 193: @licanhua , would it be possible to create `deprecatedExportedFunctionPolicy` before first use? Say host never uses `truncate` method. Now it still needs to allow it in trusted-types directive so that call `.createPolicy()` does not fail. 

I would prefer if policies were only created when they are needed and used. The same should be also true for other TT policies, I did not check those.

licanhua: what's your concern, performance? I think the impact is ignorable. 

If you are worried about that "window is undefined" when you import Adaptivecards, then a window object is added. I don't think it's a real use case.

I prefer to keep it simple at the beginning. otherwise, a new function should be added, and
I need to have ttDeprecatedPolicy to be undefined | null, then undefined means not initialized, and null means trusted-types are not supported.

```
let ttDeprecatedPolicy = undefined | null | PolicyType

function GetPolicy()
{
    if (ttDeprecatedPolicy === undefined) {
       ttDeprecatedPolicy  = null;
       if (typeof window == ....
   } 
}
```

source/nodejs/adaptivecards-react/src/adaptive-card.tsx
```diff
@@ -124,7 +124,7 @@ export const AdaptiveCard = ({
         try {
             card.parse(payload);
             const result = card.render() as HTMLElement;
-            const trustedHtml = window.trustedTypes?.emptyHTML ?? "";
+            const trustedHtml = (typeof window === 'undefined') ? "" : (window.trustedTypes?.emptyHTML ?? "");
```
paulcam206, line 127: any reason we can't do this (as you had originally fixed it)?

```suggestion
            const trustedHtml = window?.trustedTypes?.emptyHTML ?? "");
```

paulcam206 (approved):
approved, but please do consider my one comment :)

source/nodejs/adaptivecards-react/src/adaptive-card.tsx
licanhua: `window?` doesn't work, it throws exception. have to use typeof

image

licanhua: but `?` works fine in an existing object. For example, 
image

window is on the toplevel, and `?` doesn't work

source/nodejs/adaptivecards/src/utils.ts
tosmolka: @licanhua , my concern is that users of AC SDK now need to change `trusted-types` directive in their CSPs and allow `adaptivecards#deprecatedExportedFunctionPolicy` even when they are not using it. This goes against best practices to allow only policies that are needed.

PR #7925, opened 2022-09-26
Fix npm release pipeline

almedina-ms:
# Description

Fixes release job to push into npmjs

Added a new service connection where auth token is used rather than username + password

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7925)

### DIFF
--- a/.pipelines/js-release.yml
+++ b/.pipelines/js-release.yml
@@ -32,6 +32,12 @@ steps:
   inputs:
     versionSpec: 14.x
 
+- task: Npm@1
+  inputs:
+    command: 'custom'
+    workingDir: 'source/nodejs'
+    customCommand: 'init'
+
 - bash: |
    npm install -g npm@8.11.0
   workingDirectory: source/nodejs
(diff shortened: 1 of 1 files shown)


licanhua (approved):

.pipelines/js-release.yml
```diff
@@ -72,7 +72,8 @@ steps:
       inputs:
         command: publish
         workingDir: source/nodejs/${{ target_package }}
```
licanhua, line 74: please `cherry-pick -x` to main branch

.pipelines/js-release.yml
```diff
@@ -72,7 +72,8 @@ steps:
...
-        publishEndpoint: npmjs.com
+        publishEndpoint: auth_npmjs.com
```
jwoo-msft, line 75: can you provide more context on what auth_npmjs.com is?

almedina-ms: this is a service name that we own. In reality the change made was that the old npmjs.com used a combination of username and password, the new one uses a PAT to make the push

PR #7954, opened 2022-10-07
Added initialPage property to Carousel and updated carousel behavior for stretch height property 

jwoo-msft:
# Description
This PR addresses two issues.

1. It updates Carousel height property when it's set to 'stretch'.

Screen Shot 2022-10-06 at 5 59 16 PM

With this new change, when height is set to 'stretch' for Carousel, it will put the pagination at the end.
Screen Shot 2022-10-06 at 5 59 45 PM

2. `initialPage` property is added to Carousel. Type of this property is numeric. The property will set starting CarouselPage when AdaptiveCard is rendered, and is zero indexed.

This property can be used to make Carousel to stay on the same carousel page when card is re-renders by user's clicking on selectAction inside Carousel. CarouselPage ID and Carousel ID can be added to Action.Execute's data & verb fields. The data and verb are returned to user, and user can set its response cards' Carousel initial page using this property.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7954)

### DIFF
--- a/source/nodejs/adaptivecards/src/carousel.ts
+++ b/source/nodejs/adaptivecards/src/carousel.ts
@@ -140,9 +140,27 @@ export class Carousel extends Container {
         }
     }
 
-    static readonly initialPage = new NumProperty(Versions.v1_6, "initialPage", 0);
-    @property(Carousel.initialPage)
-    initialPage : number;
+    static readonly initialPageProperty = new NumProperty(Versions.v1_6, "initialPage", 0);
+    @property(Carousel.initialPageProperty)
+    get initialPage(): number {
+        let initialPageIndex = this.getValue(Carousel.initialPageProperty); 
+
+        if (initialPageIndex < 0) { 
+            console.warn(Strings.errors.invalidPropertyValue); 
+            initialPageIndex = 0 
+        }
+
+        return initialPageIndex;
+    }
+
+    set initialPage(value: number) { 
+        if (value && value < 0) { 
+            console.warn(Strings.errors.invalidPropertyValue); 
+            this.setValue(Carousel.initialPageProperty, 0); 
+        } else { 
+            this.setValue(Carousel.initialPageProperty, value);
+        }
+    }
 
     //#endregion
 
--- a/source/nodejs/adaptivecards/src/carousel.ts
+++ b/source/nodejs/adaptivecards/src/carousel.ts
@@ -155,12 +155,16 @@ export class Carousel extends Container {
         }
     }
     
-    isValidParsedPageIndex(index: number) : boolean {
-        return (this._pages?.length > 0 && 0 < index && index < this._pages.length);
+    private isValidParsedPageIndex(index: number) : boolean {
+        return this._pages ? this.isValidPageIndex(index, this._pages.length) : false;
     }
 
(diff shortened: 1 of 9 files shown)


source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -140,6 +140,28 @@ export class Carousel extends Container {
         }
     }
 
+    static readonly initialPageProperty = new NumProperty(Versions.v1_6, "initialPage", 0);
+    @property(Carousel.initialPageProperty)
+    get initialPage(): number {
+        let initialPageIndex = this.getValue(Carousel.initialPageProperty); 
+
+        if (initialPageIndex < 0) {
```
anna-dingler, line 148: Should we also check the upper bound here?

JeanRoca:
@jwoo-msft I tested the stretch property and it looks fine. Is there a way I can test the initialPage property?

jwoo-msft:
> @jwoo-msft I tested the stretch property and it looks fine. Is there a way I can test the initialPage property?

yes, you can add `initialPage` property to `Carousel`. e.g; 
thanks, I will also add an sample cards for these features.
```json
"{
      "type": "Carousel", 
      "initialPage" : 2
...
```

source/nodejs/adaptivecards/src/carousel.ts
jwoo-msft: I was delegating the task to swiper module to handle since it does handle the way we want, but your point is good. I updated the code to validate these, such that we can provide good info to users when things go wrong.

source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -377,6 +425,13 @@ export class Carousel extends Container {
         this.rtl = this.isRtl();
         this.applyRTL(carouselContainer);
 
+        if (this.isValidRenderedPageIndex(this.initialPageIndex)) {
```
anna-dingler, line 428: I think everything looks good, but we're seeing this issue in the designer again now:

carousel

We were using `_currentIndex` to store the state of the designer since we rerender the card every time a change is made in design mode.

With the new change, we are resetting to the initial page index each time.

Should we add a second variable to track the designer index, and use it instead of `initialPageIndex` if we're in design mode?

source/nodejs/adaptivecards/src/strings.ts
```diff
@@ -55,6 +55,8 @@ export class Strings {
             `Autoplay Delay is too short (${minAutoplayDelay}).`,
         tooManyCarouselPages: (maxCarouselPages: number) =>
             `Maximum number of Carousel pages exceeded (${maxCarouselPages}).`,
+        invalidInitialPageIndex: (initialPageIndex: number) =>
+            `InitialPage for carousel is invalid(${initialPageIndex}).`,
```
anna-dingler, line 59: nit to match the other strings :)

```suggestion
            `InitialPage for carousel is invalid (${initialPageIndex}).`,
```

anna-dingler (approved):

PR #7967, opened 2022-10-13
Markdown Fix

### LINKED ISSUE #7960: [Rendering][iOS][Android][MacOS][Windows] Shared cpp MarkDownParser goes into infinite loop and makes the client unresponsive
### Target Platforms

iOS, Android, Other

### SDK Version

master

### Application Name

Webex team in iOS, Android, macOS & Windows

### Problem Description

We are experiencing the client unresponsive for rendering cards in ios, android, macOS & windows.

The issue seems to be in Adaptive card shared MarkDownParser (`AdaptiveSharedNamespace::MarkDownParser`) attempting to transform text to html (Sample JSON attached below).

When the markdown parser tries to match the left and right emphasis in the text, it goes into infinite loop comparing.

If the text is `test***URGENT***`, the parser goes into **infinite loop**.
Whereas if the text is `test ***URGENT***`, the parser **works fine**.
**_(Notice the space between `test` and `***`)_**

Path to the class is: `https://github.com/microsoft/AdaptiveCards/tree/main/source/shared/cpp/ObjectModel)/MarkDownParsedResult.cpp`
function: `MatchLeftAndRightEmphasises()`

### Screenshots

tempsnip

### Card JSON

```JSON
JSON that goes in infinite loop
{
    "type": "AdaptiveCard",
    "version": "1.2",
    "body": [
        {
            "type": "TextBlock",
            "text": "test***URGENT***",
            "wrap": true
        }
    ]
}

JSON that works fine
{
    "type": "AdaptiveCard",
    "version": "1.2",
    "body": [
        {
            "type": "TextBlock",
            "text": "test ***URGENT***",
            "wrap": true
        }
    ]
}
```

### Sample Code Language

C++

### Sample Code

on the client side text is passed as `test***URGENT***`

Client initializes the MarkDownParser
`auto markdownParser = std::make_shared<AdaptiveSharedNamespace::MarkDownParser>(text);`
`text = markdownParser->TransformToHtml();`

jwoo-msft:
# Related Issue

Fixed #7960 

# Description

There was a bug where changing direction was not taking effect; fixed the change direction method; and the rest of algorithm worked.

# How Verified

Added a new unit test.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7967)

### DIFF
--- a/source/shared/cpp/ObjectModel/MarkDownHtmlGenerator.h
+++ b/source/shared/cpp/ObjectModel/MarkDownHtmlGenerator.h
@@ -155,9 +155,9 @@ public:
         return m_numberOfUnusedDelimiters;
     };
     bool GenerateTags(MarkDownEmphasisHtmlGenerator& token);
-    void ReverseDirectionType()
+    void ChangeDirectionToLeft()
     {
-        m_directionType = !m_directionType;
+        m_directionType = Left;
     };
 
 protected:
@@ -215,10 +215,17 @@ public:
         MarkDownEmphasisHtmlGenerator(token, sizeOfEmphasisDelimiterRun, type){};
 
     void GenerateTags(std::shared_ptr<MarkDownEmphasisHtmlGenerator>& token);
+
     bool IsRightEmphasis() const override
     {
-        return true;
+        return m_directionType == Right;
     }
+
+	bool IsLeftEmphasis() const override
+    {
+        return m_directionType == Left;
+    }
+
     std::string GenerateHtmlString() override;
     void PushItalicTag() override;
     void PushBoldTag() override;
--- a/source/shared/cpp/ObjectModel/MarkDownHtmlGenerator.h
+++ b/source/shared/cpp/ObjectModel/MarkDownHtmlGenerator.h
@@ -221,7 +221,7 @@ public:
         return m_directionType == Right;
     }
 
-	bool IsLeftEmphasis() const override
+    bool IsLeftEmphasis() const override
     {
         return m_directionType == Left;
     }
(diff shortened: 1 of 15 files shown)


source/shared/cpp/ObjectModel/MarkDownHtmlGenerator.h
```diff
@@ -215,10 +215,17 @@ class MarkDownRightEmphasisHtmlGenerator : public MarkDownEmphasisHtmlGenerator
...
 
     void GenerateTags(std::shared_ptr<MarkDownEmphasisHtmlGenerator>& token);
+
     bool IsRightEmphasis() const override
     {
-        return true;
+        return m_directionType == Right;
     }
+
+    bool IsLeftEmphasis() const override
```
anna-dingler, line 224: Should these methods from `MarkDownRightEmphasisHtmlGenerator` mirror the methods in `MarkDownLeftEmphasisHtmlGenerator`?

jwoo-msft: I'm not sure; what you mean.

anna-dingler: The **RightEmphasisHtmlGenerator** class seems very similar to this one:
https://github.com/microsoft/AdaptiveCards/blob/2359e6e5416e78005613e9fc6784b543adca6041/source/shared/cpp/ObjectModel/MarkDownHtmlGenerator.h#L179-L200

So I was curious why we wouldn't just have `IsRightEmphasis() -> return true` here as well.

jwoo-msft: oic; ChangeDirectionToLeft() will change the direction and this change takes account of it;

anna-dingler (approved):

PR #7990, opened 2022-10-25
implemented isEnabled

jwoo-msft:
# Related Issue
https://microsoft.visualstudio.com/OS/_workitems/edit/41838288

# Description
* Updated AdaptiveCollectionElement to implement IEnumerable in order to make search & iteration of AdaptiveTypedElement convenient. 
* Added a few test Utilities.
  * Added a method to find an element for given id.
  * Added a method that injects key value pair to AdaptiveCard using AdditionalProperites.
  * This method helps reduce the burden of having to write AdaptiveCard payload string which is error prone & requires much focus to write. 
  * To write an expected payload, one can write a card using objects, then inject the new properties using the method. The method will return a serialized JSON card.
* Implemented `isEnabled` property 
# Sample Card

Samples are in v1.5.

# How Verified

1. New unit tests are added

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7990)

### DIFF
--- a/source/dotnet/Library/AdaptiveCards/AdaptiveAction.cs
+++ b/source/dotnet/Library/AdaptiveCards/AdaptiveAction.cs
@@ -49,5 +49,15 @@ namespace AdaptiveCards
 #endif
         [DefaultValue("default")]
         public string Style { get; set; } = "default";
+
+        /// <summary>
+        ///    When set false, action is disabled 
+        /// </summary>
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+#if !NETSTANDARD1_3
+        [XmlAttribute]
+#endif
+        [DefaultValue(true)]
+        public bool IsEnabled{ get; set; } = true;
     }
 }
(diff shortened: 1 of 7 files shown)


source/dotnet/Library/AdaptiveCards/AdaptiveAction.cs
```diff
@@ -49,5 +49,15 @@ public abstract class AdaptiveAction : AdaptiveTypedElement
 #endif
         [DefaultValue("default")]
         public string Style { get; set; } = "default";
+
+        /// <summary>
+        ///    When set false, action is disabled 
+        /// </summary>
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
```
anna-dingler, line 56: What's the difference between `DefaultValueHandling.Ignore` and `DefaultValueHandling.IgnoreAndPopulate`?

jwoo-msft: https://www.newtonsoft.com/json/help/html/t_newtonsoft_json_defaultvaluehandling.htm

paulcam206 (approved):

PR #7994, opened 2022-10-26
Updated TextBlockStyle JSONConverter

jwoo-msft:
# Related Issue
https://microsoft.visualstudio.com/OS/_workitems/edit/41839444

# Description
* Updated the converter to create warning when invalid enumeration values are encountered.
* Added unit test.
* Created Utility class for common utility functions.

# Sample Card

samples/V1.5

# How Verified

1. New unit test is added.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/7994)

### DIFF
--- a/source/dotnet/Test/AdaptiveCards.Test/Utilities.cs
+++ b/source/dotnet/Test/AdaptiveCards.Test/Utilities.cs
@@ -1,3 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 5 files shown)


source/dotnet/Test/AdaptiveCards.Test/Utilities.cs
```diff
@@ -0,0 +1,95 @@
+using System;
```
paulcam206, line 1: missing copyright banner :)

paulcam206 (approved):

PR #8049, opened 2022-11-14
[UWP Renderer] Check radio buttons when they receive focus

### LINKED ISSUE #7630: [DesignerBug 31602479: [Functional - Adaptive Cards - Inputs.Json]: The functionality of radio buttons are not working as expected in 'Input.Text element' Pane.
### Problem Description

[Bug 31602479](https://microsoft.visualstudio.com/OS/_workitems/edit/31602479): [Functional - Adaptive Cards - Inputs.Json]: The functionality of radio buttons are not working as expected in 'Input.Text element' Pane.

User Experience:

It will impact the all the users if the functionality of the radio button is improper as the buttons are getting selected by using space key.

Test Environment:

Application: Adaptive Cards
OS version: 2004 (OS Build 21301.1000)

Application Version: 2101.28.1.0

Pre-Requisite: Launch the Adaptive Card application.

Repro steps:

1. Launch the Adaptive Card application.

2. Adaptive Card UWP Visualizer screen starts appearing. 

3. Navigate to Inputs.Json button and activate it.

4. Adaptive card 'Input.Text element'  will be appeared on right sight of the pane.

5. Navigate to Red radio button and verify the issue. 

Actual Result:

The functionality of radio buttons are not working as expected in 'Input.Text element' Pane. The radio buttons are getting selected by pressing space key on radio buttons.

Note:

This issue is repro'ing thorough out the application

Expected result:

The radio buttons should be getting selected automatically when focus lands on it by using up and down arrow key.

### Card JSON

```JSON
Designer Bug. Repro on UWP Adaptive Card visualizer
```

### Screenshot

_No response_

anna-dingler:
# Related Issue

Fixes #7630 

# Description

When receiving focus via arrow keys, Radio Buttons should also be checked.

I added logic to the `PreviewKeyDown` event for Radio Buttons that:
- Checks if the up or down arrow was pressed
- Determines which radio button will receive focus next
- Set `IsChecked(true)` for that button

# Sample Card

Can be tested with https://github.com/microsoft/AdaptiveCards/blob/45caa275735e210f27449570524d4f51a4aff40a/samples/v1.5/Scenarios/InputsWithValidation.json

# How Verified

Verified manually on the UWP Visualizer.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8049)

### DIFF
--- a/source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
+++ b/source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
@@ -171,22 +171,22 @@ namespace winrt::AdaptiveCards::Rendering::Uwp::implementation
                 }
 
                 radioButton.PreviewKeyDown(
-                    [this, stackPanel](winrt::IInspectable const& sender, winrt::KeyRoutedEventArgs const& args) -> void
+                    [stackPanel, radioButton](winrt::IInspectable const& /*sender*/, winrt::KeyRoutedEventArgs const& args) -> void
                     {
-                        std::uint32_t currentIndex;
+                        std::uint32_t currentButtonIndex;
                         auto children = stackPanel.Children();
                         auto size = children.Size();
-                        if (const auto isButtonFound = children.IndexOf(sender.as<winrt::RadioButton>(), currentIndex))
+                        if (const auto isButtonFound = children.IndexOf(radioButton, currentButtonIndex))
                         {
                             if (args.Key() == winrt::VirtualKey::Down)
                             {
-                                currentIndex = (currentIndex + 1) % size;
-                                children.GetAt(currentIndex).as<winrt::RadioButton>().IsChecked(true);
+                                auto newButtonIndex = (currentButtonIndex + 1) % size;
+                                children.GetAt(newButtonIndex).as<winrt::RadioButton>().IsChecked(true);
                             }
                             else if (args.Key() == winrt::VirtualKey::Up)
                             {
-                                currentIndex = (currentIndex + size - 1) % size;
-                                children.GetAt(currentIndex).as<winrt::RadioButton>().IsChecked(true);
+                                auto newButtonIndex = (currentButtonIndex + size - 1) % size;
+                                children.GetAt(newButtonIndex).as<winrt::RadioButton>().IsChecked(true);
                             }
                         }
                     });
--- a/source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
+++ b/source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
@@ -169,6 +169,53 @@ namespace winrt::AdaptiveCards::Rendering::Uwp::implementation
                     // Otherwise, leave all options unset
                     radioButton.IsChecked(IsChoiceSelected(values, input));
                 }
+
+                radioButton.PreviewKeyDown(
+                    [this, radioButton, stackPanel](winrt::IInspectable const& /*sender*/, winrt::KeyRoutedEventArgs const& args) -> void
+                    {
+                        if (args.Key() == winrt::VirtualKey::Down)
+                        {
+                            std::uint32_t index;
+                            auto children = stackPanel.Children();
+                            auto buttonFound = children.IndexOf(radioButton, index);
+
+                            if (buttonFound)
+                            {
+                                winrt::RadioButton nextButton;
+                                if (index < children.Size() - 1)
+                                {
+                                    nextButton = children.GetAt(index + 1).as<winrt::RadioButton>();
+                                }
+                                else
+                                {
+                                    nextButton = children.GetAt(0).as<winrt::RadioButton>();
+
+                                }
+                                nextButton.IsChecked(true);
+                            }
+                        }
+                        else if (args.Key() == winrt::VirtualKey::Up)
+                        {
+                            std::uint32_t index;
+                            auto children = stackPanel.Children();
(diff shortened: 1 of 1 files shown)


source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
```diff
@@ -169,6 +169,28 @@ namespace winrt::AdaptiveCards::Rendering::Uwp::implementation
...
+                    {
+                        std::uint32_t currentIndex;
+                        auto children = stackPanel.Children();
+                        auto size = children.Size();
+                        if (const auto isButtonFound = children.IndexOf(sender.as<winrt::RadioButton>(), currentIndex))
+                        {
+                            if (args.Key() == winrt::VirtualKey::Down)
+                            {
+                                currentIndex = (currentIndex + 1) % size;
+                                children.GetAt(currentIndex).as<winrt::RadioButton>().IsChecked(true);
```
beervoley, line 184: move this line outside of the if/else to avoid writing it twice:)

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
```diff
@@ -169,6 +169,28 @@ namespace winrt::AdaptiveCards::Rendering::Uwp::implementation
...
+                    [this, stackPanel](winrt::IInspectable const& sender, winrt::KeyRoutedEventArgs const& args) -> void
+                    {
+                        std::uint32_t currentIndex;
+                        auto children = stackPanel.Children();
+                        auto size = children.Size();
+                        if (const auto isButtonFound = children.IndexOf(sender.as<winrt::RadioButton>(), currentIndex))
+                        {
+                            if (args.Key() == winrt::VirtualKey::Down)
+                            {
+                                currentIndex = (currentIndex + 1) % size;
```
beervoley, line 183: It might be a good idea to create a new variable `nextButtonIndex` or something like that to hold this number for better readability :)

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
```diff
@@ -169,6 +169,28 @@ namespace winrt::AdaptiveCards::Rendering::Uwp::implementation
                     // Otherwise, leave all options unset
                     radioButton.IsChecked(IsChoiceSelected(values, input));
                 }
+
+                radioButton.PreviewKeyDown(
+                    [this, stackPanel](winrt::IInspectable const& sender, winrt::KeyRoutedEventArgs const& args) -> void
```
beervoley, line 174: is there any reason to pass `this` in the lambda capture?:)

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
```diff
@@ -169,6 +169,28 @@ namespace winrt::AdaptiveCards::Rendering::Uwp::implementation
...
+                            {
+                                currentIndex = (currentIndex + 1) % size;
+                                children.GetAt(currentIndex).as<winrt::RadioButton>().IsChecked(true);
+                            }
+                            else if (args.Key() == winrt::VirtualKey::Up)
+                            {
+                                currentIndex = (currentIndex + size - 1) % size;
+                                children.GetAt(currentIndex).as<winrt::RadioButton>().IsChecked(true);
+                            }
+                        }
```
beervoley, line 191: also, it'd be probably be a good idea to set `args.Handled(true)` since we've handled the event. Unless we want to bubble it up for some reason?

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
beervoley, line 174: @paulcam206, `sender` here should always be the radioButton of interest, right? RadioButton doesn't have any "children" per see that can bubble this event up, so we should be safe here?:)

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
anna-dingler: I can restructure this if you prefer (add a top level conditional that checks for both arrow keys)
Moving it outside the current if/else created issues when using the Tab key

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
anna-dingler: Just checked this out and marking this event as handled stops focus functionality :)

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
anna-dingler: Nope, just removed it :)

anna-dingler: I went ahead and passed `radioButton` in the lambda capture so that we know this won't be an issue.

source/uwp/Renderer/lib/AdaptiveChoiceSetInputRenderer.cpp
beervoley: you're right, didn't realize :)

beervoley (approved):
:shipit:

PR #8072, opened 2022-11-21
[Android] Calculate width for HorizontalFlowLayout

### LINKED ISSUE #7556: [Android][Rendering] ImageSet is occupying the full width
### Target Platforms

Android

### SDK Version

2.8.2

### Application Name

Webex, Adaptive Cards Visualizer

### Problem Description

In Android , ImageSet Component is always occupying the full width of card, because of that the adjacent elements are going out of the screen.

Even if the ImageSet's parent (Column in this case) is setting the width to "auto", imageset is behaving like "Stretch". Same Json is able to render the card correctly in Web. 

In code i can see there is no handling of width in ImageSetRenderrer.java for Android

image

I have attached the Layout inspector data as well in screen shot

### Screenshots

**On Web client the card is having Image set as WRAP_CONTENT:**

image

**On Adaptive Cards Visualizer ImageSet Component is going for MATCH_PARENT:**

image

**Layout Inspector data:**

Screenshot 2022-06-22 at 7 46 15 AM
 

**On Webex imageSet is again going for MATCH_PARENT:**

image

### Card JSON

```JSON
{
    "type": "AdaptiveCard",
    "version": "1.2",
    "body": [
        {
            "type": "ColumnSet",
            "columns": [
                {
                    "type": "Column",
                    "width": "auto",
                    "items": [
                        {
                            "type": "ImageSet",
                            "images": [
                                {
                                    "type": "Image",
                                    "url": "https://www.webex.com/content/dam/wbx/us/images/rebrand/webex-logo.png",
                                    "width": "50px",
                                    "size": "Medium"
                                }
                            ]
                        }
                    ]
                },
                {
                    "type": "Column",
                    "width": "stretch",
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "Test Card"
                        }
                    ]
                }
            ]
        }
    ]
}
```

### Sample Code Language

_No response_

### Sample Code

_No response_

anna-dingler:
# Related Issue

Fixes #7556 

# Description

Within the `onMeasure` method, we use the given `widthMeasureSpec` to get width.

I added an additional method `measureRequiredWidth` to sum the width of the view's children.
- If `MeasureSpec.getMode()` is `UNSPECIFIED`, we will use the calculated width.
- If `MeasureSpec.getMode()` is `AT_MOST`, we will use the calculated width only if it is less than the provided width.
- Otherwise, we will use the provided width.

# Sample Card

```JSON
{
    "type": "AdaptiveCard",
    "version": "1.2",
    "body": [
        {
            "type": "ColumnSet",
            "columns": [
                {
                    "type": "Column",
                    "width": "auto",
                    "items": [
                        {
                            "type": "ImageSet",
                            "images": [
                                {
                                    "type": "Image",
                                    "url": "https://www.webex.com/content/dam/wbx/us/images/rebrand/webex-logo.png",
                                    "width": "50px",
                                    "size": "Medium"
                                }
                            ]
                        }
                    ]
                },
                {
                    "type": "Column",
                    "width": "stretch",
                    "items": [
                        {
                            "type": "TextBlock",
                            "text": "Test Card"
                        }
                    ]
                }
            ]
        }
    ]
}
```

# How Verified

Verified manually on the Android Visualizer.

imageSet

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8072)

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
@@ -65,15 +65,28 @@ public class HorizontalFlowLayout extends RelativeLayout {
         // need to call super.onMeasure(...) otherwise we'll get some funny behaviour
         super.onMeasure(widthMeasureSpec, heightMeasureSpec);
 
-        final int width = MeasureSpec.getSize(widthMeasureSpec);
+        int width = MeasureSpec.getSize(widthMeasureSpec);
         int height = MeasureSpec.getSize(heightMeasureSpec);
 
+        final int requiredWidth = measureRequiredWidth(
+            getPaddingLeft(),
+            getPaddingRight());
+
+        if (MeasureSpec.getMode(widthMeasureSpec) == MeasureSpec.UNSPECIFIED) {
+            // set width as required since there's no height restrictions
+            width = requiredWidth;
+        } else if (MeasureSpec.getMode(widthMeasureSpec) == MeasureSpec.AT_MOST
+                && requiredWidth < width) {
+            // set width as required since it's less than the maximum allowed
+            width = requiredWidth;
+        }
+
         final int requiredHeight = measureRequiredHeight(
-                width,
-                getPaddingTop(),
-                getPaddingBottom(),
-                getPaddingLeft(),
-                getPaddingRight());
+            width,
+            getPaddingTop(),
+            getPaddingBottom(),
+            getPaddingLeft(),
+            getPaddingRight());
 
         if (MeasureSpec.getMode(heightMeasureSpec) == MeasureSpec.UNSPECIFIED) {
             // set height as required since there's no height restrictions
@@ -231,5 +244,45 @@ public class HorizontalFlowLayout extends RelativeLayout {
         ypos += line_height + paddingBottom;
 
         return ypos;
+    }
+
+    /**
+     * Measures the width required by this view
+     * by summing the width and padding of all of it's children.
+     * <p>
+     * (Package-private visibility for uint tests access.)
+     *
+     * @param paddingLeft left padding for this view.
+     * @param paddingRight right padding for this view.
+     * @return the height required by this view.
+     */
+    int measureRequiredWidth(final int paddingLeft,
+                             final int paddingRight) {
+        int requiredWidth = paddingLeft + paddingRight;
+        int childWidth, childMarginLeft, childMarginRight;
+        View child;
+        MarginLayoutParams childMarginLayoutParams;
+        // Go through the children to get the total width of the view
+        for (int i = 0; i < getChildCount(); i++) {
+            child = getChildAt(i);
+
+            if (child.getVisibility() != GONE) {
+                childWidth = child.getMeasuredWidth();
+
+                if (child.getLayoutParams() != null
+                    && child.getLayoutParams() instanceof MarginLayoutParams) {
+                    childMarginLayoutParams = (MarginLayoutParams) child.getLayoutParams();
+
+                    childMarginLeft = childMarginLayoutParams.leftMargin;
+                    childMarginRight = childMarginLayoutParams.rightMargin;
+                } else {
+                    childMarginLeft = 0;
+                    childMarginRight = 0;
+                }
+
+                requiredWidth += childMarginLeft + childWidth + childMarginRight;
+            }
+        }
+        return requiredWidth;
     }
 }
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
@@ -65,6 +65,7 @@ public class HorizontalFlowLayout extends RelativeLayout {
         // need to call super.onMeasure(...) otherwise we'll get some funny behaviour
         super.onMeasure(widthMeasureSpec, heightMeasureSpec);
 
+        // Width and height are initially set to be the requested size by the parent
         int width = MeasureSpec.getSize(widthMeasureSpec);
         int height = MeasureSpec.getSize(heightMeasureSpec);
 
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
@@ -72,13 +72,12 @@ public class HorizontalFlowLayout extends RelativeLayout {
         final int requiredWidth = measureRequiredWidth(
             getPaddingLeft(),
             getPaddingRight());
+        final int widthMode = MeasureSpec.getMode(widthMeasureSpec);
 
-        if (MeasureSpec.getMode(widthMeasureSpec) == MeasureSpec.UNSPECIFIED) {
-            // set width as required since there's no height restrictions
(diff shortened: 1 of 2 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
```diff
@@ -65,15 +65,28 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
         // need to call super.onMeasure(...) otherwise we'll get some funny behaviour
         super.onMeasure(widthMeasureSpec, heightMeasureSpec);
 
-        final int width = MeasureSpec.getSize(widthMeasureSpec);
+        int width = MeasureSpec.getSize(widthMeasureSpec);
```
anna-dingler, line 68: @jwoo-msft I believe on ImageSet uses this layout. Are there any other scenarios we should be testing here?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
```diff
@@ -232,4 +245,44 @@ int measureRequiredHeight(final int width,
...
+     * Measures the width required by this view
+     * by summing the width and padding of all of it's children.
+     * <p>
+     * (Package-private visibility for uint tests access.)
+     *
+     * @param paddingLeft left padding for this view.
+     * @param paddingRight right padding for this view.
+     * @return the height required by this view.
+     */
+    int measureRequiredWidth(final int paddingLeft,
```
jwoo-msft, line 259: does this ensure that images are set?
if the images are not yet loaded, will this method still work?
what is ensuring that images size are available?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
anna-dingler: Yes, the width and height here are the requested height from the parent. I can update the name.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
anna-dingler: I debugged in the visualizer, and the images are rendered before calling onMeasure(). onMeasure is called when the card is displayed (switching to the Visual tab), and images are loaded as soon as the JSON is rendered.

This is also very similar logic to the `measureRequiredHeight` method, so I don't think it is an issue.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
anna-dingler: I added a comment to explain the initial values here. We want a generic name since we reassign the variable later.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
jwoo-msft: or it could be that it works with the fast connection; and may fails with slower connection. but this should be fine since  that logic has been there for while.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
```diff
@@ -65,15 +65,29 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
...
+            getPaddingLeft(),
+            getPaddingRight());
+
+        if (MeasureSpec.getMode(widthMeasureSpec) == MeasureSpec.UNSPECIFIED) {
+            // set width as required since there's no height restrictions
+            width = requiredWidth;
+        } else if (MeasureSpec.getMode(widthMeasureSpec) == MeasureSpec.AT_MOST
+                && requiredWidth < width) {
+            // set width as required since it's less than the maximum allowed
+            width = requiredWidth;
```
jwoo-msft, line 82: since you are setting width to the same value; or seems to be more appropriate. saving the result of MeasureSpec.getMode(widthMeasureSpec) makes sense here.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
```diff
@@ -232,4 +246,44 @@ int measureRequiredHeight(final int width,
...
+        int requiredWidth = paddingLeft + paddingRight;
+        int childWidth, childMarginLeft, childMarginRight;
+        View child;
+        MarginLayoutParams childMarginLayoutParams;
+        // Go through the children to get the total width of the view
+        for (int i = 0; i < getChildCount(); i++) {
+            child = getChildAt(i);
+
+            if (child.getVisibility() != GONE) {
+                childWidth = child.getMeasuredWidth();
```
jwoo-msft, line 271: if child visibility is GONE, childWidth will use the previous child's width or undefined if it's first element.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageSetRenderer.java
```diff
@@ -94,11 +94,11 @@ public View render(
 
         if (imageSet.GetHeight() == HeightType.Stretch)
         {
-            viewGroup.addView(horizFlowLayout, new LinearLayout.LayoutParams(LinearLayout.LayoutParams.MATCH_PARENT, LinearLayout.LayoutParams.MATCH_PARENT, 1));
+            viewGroup.addView(horizFlowLayout, new LinearLayout.LayoutParams(LinearLayout.LayoutParams.WRAP_CONTENT, LinearLayout.LayoutParams.MATCH_PARENT, 1));
```
jwoo-msft, line 97: what's he justification for this?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
anna-dingler: Yes, but we only add `childWidth` to `requiredWidth` within this conditional, so we shouldn't run into any issues here?

jwoo-msft: sorry about that. LN270 looked like it matched with LN279. This still applies if child layout param is null or is not instance of MarginLayoutParam.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/layout/HorizontalFlowLayout.java
anna-dingler: I just made this change and refactored for height as well :)
Let me know what you think

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageSetRenderer.java
anna-dingler: Good call out - this was leftover from my initial investigation and doesn't have an impact. Just removed it

KesemZ (approved):

jwoo-msft (approved):

PR #8087, opened 2022-11-29
Convert tabs to spaces

krschau:
# Related Issue

N/A

# Description

Some code had a mix of tabs and spaces for whitespace. Converted these tabs to spaces for consistency.

# How Verified

Only whitespace changes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8087)


beervoley:
@krschau, I think it might be a good idea to **only** update `shared cpp model` and `uwp` source and not touch the rest to make the next changes you're planning to bring consistent with these :)

paulcam206 (approved):

jwoo-msft:
i know it's late; but it would have been better if this change was limited to UWP & object model. for example, .net templating have auto generated files with those tabs.

krschau:
@jwoo-msft Sorry about that, I got a little over-eager. I'd be happy to switch those generated ones back, or just leave it to the next person who touches that code.

tr3iwjtzq1bbopqMMH:
krschau:user/krschau/tabs

PR #8094, opened 2022-11-30
[UWP Renderer]: Make references to the ::winrt::AdaptiveCards::Rendering::Uwp namespace consistent

krschau:
# Related Issue

N/A

# Description

Current references to the `::winrt::AdaptiveCards::Rendering::Uwp` namespace are inconsistent. Sometimes it is fully qualified, other times `Rendering::Uwp`, `Uwp`, or most often just `winrt` is used. This PR changes all references to use `winrt::`, except for with `ActionMode`, since there is a type of the same name in the ObjectModel. Here, we have an alias for the namespace, `render_xaml`, that we use in addition.

# How Verified

Manually tested UWP visualizer.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8094)

### DIFF
--- a/source/uwp/Renderer/dll/CppWinRTIncludes.h
+++ b/source/uwp/Renderer/dll/CppWinRTIncludes.h
@@ -54,6 +54,7 @@ namespace winrt
     // In order to avoid "namespace not defined" errors we have to define the namespace here too.
     namespace AdaptiveCards::Rendering::Uwp{}
     using namespace ::winrt::AdaptiveCards::Rendering::Uwp;
+    namespace render_xaml = ::winrt::AdaptiveCards::Rendering::Uwp;
 
     namespace AdaptiveCards::Rendering::Uwp::implementation{}
     namespace implementation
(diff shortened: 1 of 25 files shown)


source/uwp/Renderer/dll/CppWinRTIncludes.h
```diff
@@ -54,6 +54,7 @@ namespace winrt
     // In order to avoid "namespace not defined" errors we have to define the namespace here too.
     namespace AdaptiveCards::Rendering::Uwp{}
     using namespace ::winrt::AdaptiveCards::Rendering::Uwp;
+    namespace render_xaml = ::winrt::AdaptiveCards::Rendering::Uwp;
```
beervoley, line 57: I see that in most of the places you simply use `winrt::` namespace to refer to renderer types and you've only declared `render_xaml` to use in cases where types exist in both objectModel and renderer namespaces?
like `enum ActionMode` for example?:) 

just want to make sure that is the intent:)

beervoley (approved):

paulcam206:
@beervoley - these pipelines should run automatically in the future...

PR #8101, opened 2022-12-05
[UWP Renderer] Use namespace aliases for helper namespaces

krschau:
# Related Issue

N/A

# Description

This change simplifies references to the various "helper" namespaces and minimizes instances of `using namespace`.
- **XamlHelpers**: Creates the namespace alias `XamlHelpers` for `::AdaptiveCards::Rendering::Uwp::XamlHelpers`. Since the XamlHelpers.h header is already included in pch.h, remove instances where it is essentially included twice.
- **ActionHelpers**: Creates the namespace alias `Helpers` for `::AdaptiveCards::Rendering::Uwp` and uses it for instances of `ActionHelpers` types
- **MediaHelpers**: Use the `Helpers` namespace within the `AdaptiveMediaRenderer`
- **XamlBuilder**: Use the `Helpers` namespace when referring to XamlBuilder. Since the XamlBuilder.h header is already included in pch.h, remove instances where it is essentially included twice.

# How Verified

Manually tested Adaptive Cards Renderer.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8101)

### DIFF
--- a/source/uwp/Renderer/lib/ActionHelpers.cpp
+++ b/source/uwp/Renderer/lib/ActionHelpers.cpp
@@ -689,7 +689,7 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
         auto showCardActionConfig = actionsConfig.ShowCard();
         auto showCardActionMode = showCardActionConfig.ActionMode();
         // ActionMode enum exists both in Rendering in ObjectModel namespaces. When the time permits, fix it.
-        if (showCardActionMode == winrt::render_xaml::ActionMode::Inline)
+        if (showCardActionMode == winrt::winrt_render_xaml::ActionMode::Inline)
         {
             // Get the card to be shown
             auto actionAsShowCardAction = action.as<winrt::AdaptiveShowCardAction>();
--- a/source/uwp/Renderer/lib/ActionHelpers.cpp
+++ b/source/uwp/Renderer/lib/ActionHelpers.cpp
@@ -691,7 +691,7 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
         auto showCardActionConfig = actionsConfig.ShowCard();
         auto showCardActionMode = showCardActionConfig.ActionMode();
         // ActionMode enum exists both in Rendering in ObjectModel namespaces. When the time permits, fix it.
-        if (showCardActionMode == winrt::AdaptiveCards::Rendering::Uwp::ActionMode::Inline)
+        if (showCardActionMode == winrt::render_xaml::ActionMode::Inline)
         {
             // Get the card to be shown
             auto actionAsShowCardAction = action.as<winrt::AdaptiveShowCardAction>();
--- a/source/uwp/Renderer/lib/ActionHelpers.cpp
+++ b/source/uwp/Renderer/lib/ActionHelpers.cpp
@@ -8,8 +8,6 @@
 #include "LinkButton.h"
 #include "AdaptiveHostConfig.h"
 
-using namespace AdaptiveCards::Rendering::Uwp::XamlHelpers;
-
 namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
 {
     winrt::Thickness GetButtonMargin(winrt::AdaptiveActionsConfig const& actionsConfig)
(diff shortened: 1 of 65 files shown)


source/uwp/Renderer/lib/AdaptiveExecuteActionRenderer.cpp
```diff
@@ -15,13 +15,11 @@ namespace winrt::AdaptiveCards::Rendering::Uwp::implementation
         try
         {
             renderContext.LinkSubmitActionToCard(action, renderArgs);
-            return ::AdaptiveCards::Rendering::Uwp::ActionHelpers::BuildAction(action, renderContext, renderArgs, false);
+            return Helpers::ActionHelpers::BuildAction(action, renderContext, renderArgs, false);
```
almedina-ms, line 18: is this Helpers the same as XamlHelpers?

source/uwp/Renderer/lib/pch.h
```diff
@@ -40,6 +40,10 @@
 #include "XamlBuilder.h"
 #include "XamlHelpers.h"
 
+namespace Helpers = ::AdaptiveCards::Rendering::Uwp;
```
almedina-ms, line 43: I wonder if there's a better naming we could use for this namespace as it may be confusing to read it besides ther XamlHelpers

krschau: Do you have any suggestions for a better one? I already used "render_xaml" for winrt::AdaptiveCards::Rendering::Uwp, but this is a different namespace NOT in winrt. Maybe "rendering"? The lower-case r would differentiate it from the Rendering inside it.

krschau: Alternatively, change the existing render_xaml to winrt_render_xaml and this to render_xaml.

beervoley: both `rendering `or changing render_xaml to winrt_render_xaml will work fine:)

source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -689,7 +689,7 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
         auto showCardActionConfig = actionsConfig.ShowCard();
         auto showCardActionMode = showCardActionConfig.ActionMode();
         // ActionMode enum exists both in Rendering in ObjectModel namespaces. When the time permits, fix it.
-        if (showCardActionMode == winrt::render_xaml::ActionMode::Inline)
+        if (showCardActionMode == winrt::winrt_render_xaml::ActionMode::Inline)
```
beervoley, line 692: `winrt::winrt_render_xaml` - twice winrt?:)
why don't we just use the fully qualified namespace here if it's used so infrequently :)

krschau: The fully qualified namespace contains "uwp", and I'm trying to remove those so there are fewer #ifdefs in the future. I don't love the double winrt but I think it's the most accurate and understandable alias here, and it only needs to occur once (the other instances of winrt_render_xaml don't need the winrt:: prefix). If we ever get around to the todo in the comment above, we can remove this entirely.

source/uwp/Renderer/lib/AdaptiveExecuteActionRenderer.cpp
krschau: Changed this.

source/uwp/Renderer/lib/pch.h
```diff
@@ -40,6 +40,10 @@
 #include "XamlBuilder.h"
 #include "XamlHelpers.h"
 
+namespace render_xaml = ::AdaptiveCards::Rendering::Uwp;
+
+namespace XamlHelpers = render_xaml::XamlHelpers;
```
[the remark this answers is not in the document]
krschau: I think we want to keep the winrt stuff in cppwinrtincludes.h, and the non-winrt stuff in pch.h. If we break that convention and move it anyway, it gets messy because of the order of things. The renderer headers in pch need the winrt stuff to compile, and the namespace aliases you mentioned need those headers. If we move them to cppwinrtincludes, the namespaces don't exist yet.

source/uwp/Renderer/lib/ActionHelpers.cpp
krschau: I'm not sure what that gets us? I think we're going to need to do some weird things with the Uwp namespace so I'm really trying to keep it contained.

Issue #3924, opened 2020-04-13
Input.ChoiceSet auto complete


productboard-ac:
## Summary

When there are thousands of choices in a menu (like assigning a task to someone, when there's thousands of people in your org), you need to have a searchable select menu that dynamically loads data as the user searches (since including 1,000 or more choices statically in the card/data isn't a good option).

---

**Status**: Proposal Approved

---

![](https://cdn.productboard.com/production/attachments/d9a78e01fb63f8502884f5361982c48d4d913acb491fef0ed623d958d6356bb2/portal_cover/DynamicSearchable.gif)

## Open Issues

- Need to investigate the different platform limitations and cost with regard to `isMultiSelect` support. 

## Requirements

1. P0: The input can dynamically fetch the list of choices from a remote backend as the user types

2. P0: Hosts can specify 1 or more pre-determined datasets that get the first-chance to fulfill the request. (E.g., Microsoft Teams could query the Active Directory list of users by hooking into the `graph.microsoft.com/users` dataset). If the host hasn't registered a matching dataset, the request will be sent to the backend.

3. P0: A static list of choices can be provided, and any dynamic ones will get appended to the end

4. P1: The input will support `"isMultiSelect": true` and `"style": "compact"` where possible

5. P1: Search dynamically renders a list (which has custom UI for each element) with the search result

## Input.ChoiceSet

Property  
Type  
Required  
Description  
choices.data  
Data.Query  
No  
Enables dynamic autocomplete as the user types, by fetching a remote set of choices from a backend  

## Data.Query

`Data.Query` includes the metadata necessary to fetch remote data.

Property  
Type  
Required  
Description  
dataset  
string  
Yes  
The type of data that should be fetched dynamically  

## Host-Specific Datasets (Such as the User directory)

Some Hosts, such as Microsoft Outlook and Microsoft Teams, will allow the `Input.ChoiceSet` to bind to pre-configured datasets, such as the current User directory.

Microsoft Datasets

Dataset  
Description  
graph.microsoft.com/users  
The full list of users within the organization  
graph.microsoft.com/users?scope=currentContext  
The list of users within the "current context", which will be determined by the Host. For example, in a Microsoft Teams Channel, it would be the list of users in the current channel  

## Example

Step 1: Card Payload

```
{
    "type": "Input.ChoiceSet",
    "id": "selectedUser",
    "choices": [
        { "title": "Static 1", "value": "Static 1" }
    ],
    "choices.data": {
[...]

PR #8108, opened 2022-12-07
Rename Action.Execute related interfaces

SuryanshPrabhat:
# Related Issue
https://github.com/microsoft/AdaptiveCards/issues/3924

# Description
We aim to utilize **ChannelAdapter** abstraction to handle request/response for Dynamic Typeahead in Input.Choiceset. Currently, it is only used for Action.Execute and the interfaces tied up with it, although general in name, are specific to Action.Execute. This PR introduces the changes which make it more generic to be used for other types of requests.

# Breaking Changes
1. [activity-request.ts](https://github.com/microsoft/AdaptiveCards/blob/main/source/nodejs/adaptivecards/src/activity-request.ts): Renamed interface IActivityRequest to IExecuteRequest
3. [adaptive-applet.ts](https://github.com/microsoft/AdaptiveCards/blob/main/source/nodejs/adaptivecards/src/adaptive-applet.ts): Renamed instance methods of class AdaptiveApplet: 
- onPrepareActivityRequest to onPrepareExecuteRequest
- onActivityRequestSucceeded to onExecuteRequestSucceeded
- onActivityRequestFailed to onExecuteRequestFailed

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8108)

### DIFF
--- a/source/nodejs/adaptivecards-aaf-testapp/src/app.ts
+++ b/source/nodejs/adaptivecards-aaf-testapp/src/app.ts
@@ -24,19 +24,19 @@ window.onload = function() {
     applet.channelAdapter = new TestChannelAdapter("https://acv2testbot.azurewebsites.net/aaftestbot/invoke");
 
     applet.setCard(Shared.sampleCard);
-    // applet.onActivityRequestFailed = (sender, response) => { return 2000; }
-    applet.onActivityRequestSucceeded = (sender: Adaptive.AdaptiveApplet, response: Adaptive.SuccessResponse, parsedContent: string | Adaptive.AdaptiveCard) => {
+    // applet.onExecuteRequestFailed = (sender, response) => { return 2000; }
+    applet.onExecuteRequestSucceeded = (sender: Adaptive.AdaptiveApplet, response: Adaptive.SuccessResponse, parsedContent: string | Adaptive.AdaptiveCard) => {
         if (parsedContent !== undefined && typeof parsedContent === "string") {
             alert(parsedContent);
         }
     }
-    applet.onSSOTokenNeeded = (sender: Adaptive.AdaptiveApplet, request: Adaptive.IActivityRequest, tokenExchangeResource: Adaptive.TokenExchangeResource) => {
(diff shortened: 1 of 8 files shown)


SuryanshPrabhat:
@microsoft-github-policy-service agree [company="Microsoft"]

jwoo-msft:
added @dclaux.

dclaux:
This shouldn't be done. An activity request is the "messenger" and carries an Action.Execute, just the same as it could carry a "Data.Query" or whatever else. These renames are not only unnecessary, they are not desirable.

dclaux:
If something needs to change, it is the the `action` member of IActivityRequest, which as of now is strongly typed to `ExecuteAction`. It should be updated to allow the activity request to carry a `Data.Query`

dclaux:
As for the events, again they are fine as they are:
- `onPrepareActivityRequest` prepares an activity request by setting its members etc. including its `action` member, the piece that needs to change as per my previous comment
- The other two events are also fine as they are

Again, an "activity request" is a generic "messenger", and its payload (the "action") is what types it.

SuryanshPrabhat:
@dclaux I understand that 'Activity Request' as its name suggests is meant to be generic for all types of requests. And to maintain that notion, I separated the interfaces to make a clear distinction between Activity Request and an Action.Execute request. The interface which enforces Activity Request has members which are only required for Action.Execute, such as `trigger`, `consecutiveRefreshes`, `authCode` and 'authToken'.
I know that we can make this implementation work for dynamic typeahead via the action property, but I do not feel that it is clean. And adding more properties if needed to this interface for more types of requests in the future does not do a good separation of concerns. It is better for Activity Request to be a base which can be extended by several types of requests and all those requests are handled separately. Makes it more extensible as well.

dclaux:
@SuryanshPrabhat I am sorry but I am not convinced by those arguments. I understand your goal and your intent which make sense, but I do not believe the proposed changes are the best way to achieve those. I have sent you additional thoughts offline, let's continue discussing there.

Issue #8066, opened 2022-11-20
[Designer] Toggle Visibility stopped working in show card


mukuagar:
Action.Toggle Visibility stopped working when in a showcard sometime around last Friday(18th Nov 2022).
Here is a sample json to test
```
{
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "type": "AdaptiveCard",
    "version": "1.3",
    "actions": [
        {
            "type": "Action.ShowCard",
            "title": "Action.ShowCard",
            "card": {
                "type": "AdaptiveCard",
                "version": "1.3",
                "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
                "body": [
                    {
                        "errorMessage": "error message error message error message error message ",
                        "height": "stretch",
                        "id": "11",
                        "isRequired": true,
                        "label": "Hello this is label",
                        "type": "Input.Date",
                        "isVisible": true
                    },
                    {
                        "errorMessage": "error message error message error message error message ",
                        "height": "stretch",
                        "id": "33",
                        "isRequired": true,
                        "label": "This is label",
                        "type": "Input.Time",
                        "isVisible": true
                    },
                    {
                        "actions": [
                            {
                                "targetElements": [
                                    "11",
                                    "33"
                                ],
                                "title": "Action.ToggleVisibility",
                                "type": "Action.ToggleVisibility"
                            }
                        ],
                        "id": "22",
                        "type": "ActionSet"
                    }
                ]
            }
        }
    ]
}
```

paulcam206:
Closing as duplicate of #8086

PR #8132, opened 2022-12-09
[JS Renderer] Fix toggleVisibility within showCard

### LINKED ISSUE #8086: [Authoring] Action.ToggleVisibility is not working within Action.ShowCard
### Target Application

Microsoft Teams, Outlook

### Application Operating System

Windows

### Schema Version

1.4

### Problem Description

Following an event, I contact my end-users to get answers. 
- Yes 
- No

If Yes, I want that they precise Option 1 or Option 2. 

To do this, I am using: 
- Two Action.ShowCard to display "Yes" and "No"
- Within the "Yes" Card, an Action.ToggleVisibility to select "Option 1" and "Option 2". 
- If I select "Option 1" or "Option 2" some text and a "Confirm" button shall be displayed. These elements are not visibile by default

However, when I put the Action.ToggleVisibility within a an Action.ShowCard there is no effect when I activate the Toggle. 

### Expected Outcome

When I activate an Action.ToggleVisibility action within the ShowCard, I should see the targeted elements. 

### Actual Outcome

I cannot manage to change the visibility of items within ShowCard. 

For example, in the code below I cannot change the visibility of "MyText 1"

### Card JSON

```JSON
{
    "type": "AdaptiveCard",
    "version": "1.4",
    "body": [
        {
            "type": "ActionSet",
            "actions": [
                {
                    "type": "Action.ShowCard",
                    "title": "Action.ShowCard",
                    "card": {
                        "type": "AdaptiveCard",
                        "body": [
                                {
            "type": "Container",
            "isVisibility": false,
            "id": "id11",
            "items": [
                {
                    "type": "ActionSet",
                    "actions": [
                        {
                            "title": "Action.ToggleVisibility",
                            "targetElements": [
                                "id1"
                            ]
                        }
                    ]
                },
                {
                    "type": "RichTextBlock",
                    "isVisibility": false,
                    "id": "id1",
                    "inlines": [
                        {
                            "type": "TextRun",
                            "text": "My Text 1"
                        }
                    ],
                    "isVisible": true

                }
            ]
        }
                            ]
                    }
                }
            ]
        }
    ]
}
```

### Repro Steps

_No response_

anna-dingler:
# Related Issue

Fixes #8086

# Description

We needed to override the `getElementById()` method for `ActionSet` and `ContainerWithActions`. If these elements contain a `ShowCard` action, we need to search `action.card` for the matching id.

# Sample Card

Added two samples cards. One for `ShowCard` within `actions[]` and one for `ShowCard` within an `ActionSet`.

# How Verified

Verified manually on the Designer and added two ui tests.

I also verified the test cases provided in #8086 and #8066

toggle

toggle2

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8132)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -6537,6 +6537,26 @@ export class ActionSet extends CardElement {
...
+
+                    if (result) {
+                        break;
+                    }
+                }
+            }
+        }
+
+        return result;
+    }
```
beervoley, line 6558: this code is repeated twice - could you please refactor?
see if it can be placed in the parent class of the both classes where it's used?:)

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -6537,6 +6553,16 @@ export class ActionSet extends CardElement {
...
     }
 
+    getElementById(id: string): CardElement | undefined {
+        let result = super.getElementById(id);
+
+        if (!result) {
+            result = this.getElementByIdFromAction(id);
+        }
+
+        return result;
```
beervoley, line 6563: this looks better:)
do we absolutely have to go over the tree **twice** in this scenario?
can we somehow detect this scenario in another way?

anna-dingler: Do you mean call `super.getElementById` and `this.getElementByIdFromAction`?

anna-dingler: If so, `getElementById` doesn't normally traverse the tree. It checks if the current element is the correct id :)

beervoley: oh I see... the name of the method - `getElementById` doesn't reflect what it does then :D

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -716,6 +716,22 @@ export abstract class CardElement extends CardObject {
...
         return undefined;
     }
+    
+    getElementByIdFromAction(id: string): CardElement | undefined {
+        let result = undefined;
+        for (let i = 0; i < this.getActionCount(); i++) {
+            const action = this.getActionAt(i);
+
+            if (action instanceof ShowCardAction) {
+                result = action.card.getElementById(id);
```
beervoley, line 726: do we only have the scenario when we need to pull the id of the action `only` when it's `showCardAction`? 
it's solving this particular issue, just want to make sure that we don't have any other potential scenario with other action types when we need to do the same thing:)

anna-dingler: Yup, showCardAction is the only action that has its own card as of now

beervoley (approved):

PR #8136, opened 2022-12-12
Add WinUI 3 ObjectModel

krschau:
# Related Issue

#5888

# Description

In order to eventually create a WinUI 3 Adaptive Card Renderer, this change creates a WinUI 3 version of the Object Model. 
- Completely reuse the `UWPObjectModelProjection` project, which does not have to be changed for WinUI 3. 
- Create a new WinUI 3 version of the `AdaptiveCardsObjectModel` project. We cannot completely reuse the UWP version, since it must be a WinUI 3 runtime component, but we can reuse the code in the `lib` directory, so that code moves into a new `sharedObjectModel` directory. There, everything is in the `winrt::AdaptiveCards::ObjectModel::Uwp::implementation` namespace. In `pch.h`, we check if we are compiling the WinUI 3 version of the project, and if so `#define Uwp Winui3`, so that the code compiles into an `AdaptiveCards.ObjectModel.Winui3` namespace, without having to duplicate everything or change every file to check the version.
- Create an `ObjectModelCsProjection`, so that the runtime component can be used in WinUI 3 C# projects
- Create a `WinUI3ObjectModelTest` project to test the WinUI 3 version of the Object Model. This does duplicate the one file of test code, but could be changed later. These tests are not currently added to any test pipeline but can be run locally via Test Explorer.

# How Verified

Created a WinUI 3 unit test project, called WinUI3ObjectModelTest. It duplicates all the tests from the UWP version, all tests pass.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8136)


jwoo-msft (commented):
shared is place for pure c++ codes that are shared by ios, android, and uwp.
moving projected code to shared doesn't seems to be correct.

beervoley:
2 options:
1)Rename our UWP folder to `xaml` and extract the shared code there :)
2)Put cppwinrt shared code under source/shared/cppwinrt

krschau:
> I think it would be better to extract the shared code in the `uwp` folder with something like `objectModelShared` and `rendererShared`. Then create separate projects for system xaml and winui3. After that's done - we can rename `uwp` to `xaml`.

@beervoley Let me make sure I'm understanding you correctly:

My current layout is to have a winui3 folder that basically mirrors the projects in the uwp folder, with the shared files moved out to "shared".
You’re saying have all the uwp and winui3 projects under a “xaml” folder, with the shared code in objectModelShared and rendererShared also under “xaml”? In that scenario I would still like to group the uwp projects into a “uwp” folder under “xaml” and same with winui3. I do like the idea of keeping all the xaml stuff together.
@jwoo-msft FYI.

beervoley:
> > I think it would be better to extract the shared code in the `uwp` folder with something like `objectModelShared` and `rendererShared`. Then create separate projects for system xaml and winui3. After that's done - we can rename `uwp` to `xaml`.
> 
> @beervoley Let me make sure I'm understanding you correctly:
> 
> My current layout is to have a winui3 folder that basically mirrors the projects in the uwp folder, with the shared files moved out to "shared". You’re saying have all the uwp and winui3 projects under a “xaml” folder, with the shared code in objectModelShared and rendererShared also under “xaml”? In that scenario I would still like to group the uwp projects into a “uwp” folder under “xaml” and same with winui3. I do like the idea of keeping all the xaml stuff together. @jwoo-msft FYI.

yup, this is what I mean :)

krschau:
Re-creating this PR from a branch rather than a fork: https://github.com/microsoft/AdaptiveCards/pull/8157

Issue #6341, opened 2021-09-16
[Feature Request] Feature Title Here


jonmill:
### Problem Statement

Other

### Proposed solution

Other

PR #8143, opened 2022-12-13
Snsajal/label position

baton17:
# Changes made
- Added support for "labelPosition" for input fields in the adaptive card schema starting with version 1.6.
- Added doc for labelPosition feature for input fields.
- Implementation for label width in inputs

# Description

Right now, position of the label is always above to the input field. We want to provide more options to card authors to position the label.
Now, labels can also be place inline with the input. It can provide better experience and more flexibity to card authors in different scenarios.
With `inline` label placement, card authors can also set `labelWidth`. It will help them design inline labels beautifully.

Inherited properties of all Input fields will have one more property called `labelPosition` whose type will be `InputLabelPosition`

| Property | Type | Required | Description | Version |
| -------- | ---- | -------- | ----------- | ------- |
| **labelPosition** | `InputLabelPosition?` | No | Determines the position of the label with respect to the input field. Default is "above" when not specified | 1.6 |

| Property | Type | Required | Description | Version |
| -------- | ---- | -------- | ----------- | ------- |
| **labelWidth** | `number` or `string` | No | Determines the width of the label in percent like `40` or a specific pixel width like `"40px"` when label is placed inline with the input . labelWidth would be ignored when the label is displayed above the input. | 1.6 |

### InputLabelPosition

Position for label in input fields.

* **Type**: `InputLabelPosition`
* **Required**: No
* **Allowed values**:
  * `"inline"` : should place label inline with the input field
  * `"above"` : should place label above the input field.
 
## Sample Payload

```json
     {
          "type": "Input.Text",
          "label": "Name",
	  "value": "Sneh",
	  "labelPosition" : "above"
      },
      {
          "type": "Input.Time",
          "label": "Time of Arrival",
          "value": "09:30",
          "labelPosition" : "inline",
          "labelWidth": 40
      },
      {
          "type": "Input.Text",
          "label": "Flight destination",
	  "value": "Hyderabad",
          "labelPosition" : "inline",
          "labelWidth": "30px"
      }	
```

Result post implementation of
- `inline` label position where `labelWidth` is set to `100px` (first and second text input below)
image

- `inline` label position where `labelWidth` is set to 30% (first and second text input below)
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8143)


schemas/src/enums/InputLabelPosition.json
```diff
@@ -0,0 +1,9 @@
+{
+	"$schema": "https://raw.githubusercontent.com/microsoft/AdaptiveCards/6f39aedce45864ae1067ed44a5551dc973790bb5/source/nodejs/typed-schema/schema/lib/Type.json",
+	"classType": "Enum",
+	"description": "Specifies position of the label with respect to input.",
+	"values": [
+		"inline",
+		"above"
```
sowrabh-msft, line 7: Aren't we adding `default` as an option here as well?

sowrabh-msft (approved):

schemas/src/enums/InputLabelPosition.json
baton17: "above" is the default behavior

schemas/1.6.0/adaptive-card.json
```diff
@@ -1186,8 +1187,9 @@
         "separator": {},
         "spacing": {},
         "isVisible": {},
-        "requires": {}
-      },
+        "requires": {},
+        "labelPosition": {}
+        },
```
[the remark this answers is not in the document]
baton17: fixed all the indentation issues in this file

anna-dingler (approved):
Looks good to me! @jwoo-msft Do you have anything?

specs/DesignDiscussions/InputLabelPosition.md
```diff
@@ -0,0 +1,105 @@
+# Input Label Position
+
+## Overview
+Right now position of the label is always above to the input field. We want to provide more options to card authors to position the label.
+Now, labels can also be place inline with the input. It can provide better experience and more flexibity to card authors in different scenarios.
+
+## New Schema Changes in Adaptive Card
+
+Introducing an optional new styling property in AC input fields schema which will allow developers to define `position` of the label with respect to input field. It can be both `inline` and `above`. Default would be existing `above`.
+	*Note*: When card width is small due to browser resize etc, horizontal view can shift to vertical view of input fields.
```
jwoo-msft, line 10: can you elaborate on horizontal and vertical view? How do we determine what's too small? we can wrap and still maintain label to input field host config ratio, for example, in this case, we don't necessarily need to switch to different view.

jwoo-msft (commented):
overall, it looks good.
I wouldn't add changes associated with schema in this PR. We make those changes closer to the actual release. Schema do changes, and it's for the public. We have to be unnecessarily more careful with cutting the branch if there is a schema changes, as we make website update time to time with the new branch cut, and we may not want to show the schema changes before a public announcement of the new feature. 
The spec changes in Design Discussion are always welcome, they help contributors to the features, as an agreement.
FYI, https://github.com/microsoft/AdaptiveCards/issues/6341

jwoo-msft:
Added @dclaux for visibility.

FYI, @dclaux, your alias doesn't show up in the list of reviewers.

dclaux:
While on the surface this proposed (it's still a proposal, right?) schema change looks fine, I think it's going to be very hard to implement it in such a way that resulting cards look good.

For cards that use "label on left" to look good, and per the actual requirements of the one partner I know of who's asking for this, the following need to be met when multiple "label on left" inputs appear one after another in a given container (whether the contain er is the card itself or a sub-container, including a column):
- The labels are left aligned with each other
- The actual inputs are left aligned with each other
- There isn't an unnecessary amount of space between labels and inputs. I.e., the longest label dictates the width of the label area width
  - And on top of that, there is probably a need to introduce a "maximum label width"

Implementing the above is not going to be easy given the proposal.

As I have shared in the past, I strongly believe a better model is to allow an input to refer to another element as its label (the equivalent of `aria-labelledby` if you wish) or to allow an element to specify what it is labelling (the equivalent of `for` for a HTML `<label>` element). With that approach, a developer can leverage any of the existing AC layout constructs and place labels wherever they want. The only drawback is that it requires the card author to "do the right thing", which, for that matter, is already what they have to do today anyway (they must specify a "label" property on their inputs to make them accessible)

specs/DesignDiscussions/InputLabelPosition.md
baton17: yeah makes sense. We can still maintain inline view. I will remove this line.

schemas/1.6.0/adaptive-card.json
```diff
@@ -3304,6 +3310,18 @@
...
+        "labelWidth": {
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "number"
+            }
+          ],
+          "description": "Determines the width of the label in percent like 40 or a specific pixel width like '40px' when label is placed inline with the input . labelWidth would be ignored when the label is displayed above the input.",
```
anna-dingler, line 3322: For a percentage, would the value be `40` or `40%`?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3274,11 +3274,82 @@ export class Media extends CardElement {
...
+                invalidLabelWidth = true;
+            }
+
+            if (invalidLabelWidth) {
+                context.logParseEvent(
+                    sender,
+                    Enums.ValidationEvent.InvalidPropertyValue,
+                    Strings.errors.invalidInputLabelWidth()
+                );
+                result = undefined;
```
anna-dingler, line 3320: Why don't we revert to the default value here?

anna-dingler (commented):
Left a couple of comments, but overall, looks good :)

Could you also add a couple of sample payloads to the `samples` directory?

source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -241,10 +241,11 @@ export class RequiredInputLabelTextDefinition extends BaseTextDefinition {
 
 export class InputLabelConfig {
     inputSpacing: Enums.Spacing = Enums.Spacing.Small;
+    inputLabelWidth : number = 30;
```
dclaux, line 244: Should really be just `width` as `InputLabelConfig` already suggests this is configuration for "input labels". It's also consistent with the other setting names (i.e. `requiredInputs` and not `inputLabelsForRequiredInputs`)

source/nodejs/adaptivecards/src/strings.ts
```diff
@@ -70,7 +70,9 @@ export class Strings {
         elementAlreadyParented: () => "The element already belongs to another container.",
         actionAlreadyParented: () => "The action already belongs to another element.",
         elementTypeNotStandalone: (typeName: string) =>
-            `Elements of type ${typeName} cannot be used as standalone elements.`
+            `Elements of type ${typeName} cannot be used as standalone elements.`,
+        invalidInputLabelWidth: () =>
+            `Invalid input label width. Defaulting to label width from host config"`,
```
dclaux, line 75: Period at end for consistency with other strings.

schemas/1.6.0/adaptive-card.json
baton17: for percentage, value would be a number. So it should be just 40. I think `width` in column also follows similar guidelines https://adaptivecards.io/explorer/Column.html

source/nodejs/adaptivecards/src/card-elements.ts
baton17: So I am setting default here as undefined and when we render the element and label width is undefined, I am using the default value from host config 
image

At this place, host config is not available and I want default value to be coming from host config and inside the host config, it has a default value (=30:70) when host does not provide it.

source/nodejs/adaptivecards/src/strings.ts
baton17: done

source/nodejs/adaptivecards/src/host-config.ts
baton17: updated as `width`

schemas/src/elements/inputs/Input.json
```diff
@@ -24,6 +24,16 @@
 			"type": "string",
 			"description": "Label for this input",
 			"version": "1.3"
+		},
+		"labelPosition": {
+			"type": "InputLabelPosition",
+			"description": "Label position hint for input",
```
pankajmi2, line 30: would this sentence be clearer - > "input label hint position" ? Also,  should we specify sample values like above/default like you did for labelWidth ?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3411,6 +3493,24 @@ export abstract class Input extends CardElement implements IInput {
...
+                    const defaultLabelWidth = hostConfig.inputs.label.width;
+                    this._renderedLabelElement.style.width = defaultLabelWidth.toString() + "%";
+                    this._renderedInputControlElement.style.width = (100 - defaultLabelWidth) + "%";
+                }
+                else if (this.labelWidth.unit == Enums.SizeUnit.Weight) {
+                    const labelWidthInPercent = this.labelWidth.physicalSize;
+                    this._renderedLabelElement.style.width = labelWidthInPercent.toString() + "%";
+                    this._renderedInputControlElement.style.width = (100 - labelWidthInPercent).toString() + "%";
+                }
+
```
pankajmi2, line 3507: nit: unnecessary white line

pankajmi2 (approved):

source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -241,7 +241,7 @@ export class RequiredInputLabelTextDefinition extends BaseTextDefinition {
 
 export class InputLabelConfig {
     inputSpacing: Enums.Spacing = Enums.Spacing.Small;
-    inputLabelWidth : number = 30;
+    width : number = 30;
```
dclaux, line 244: Format: no space before :

jwoo-msft (approved):

jwoo-msft:
@baton17 as we have discussed; please move changes to spec files to a separate branch and create a new PR closer to 1.6 official release.
thanks.

PR #8180, opened 2022-12-21
RevealOnHover for Inputs

baton17:
# Changes made
- Added support for "inputStyle" for input fields in the adaptive card schema starting with version 1.6.
- Added doc for input reveal on hover feature

# Description

We will allow bot developer to enhance the user experience of all `Input` fields in Adaptive card (such as `Input.Text`, `Input.Number`, `Input.Date`, `Input.Time`, `Input.Toggle` and `Input.Choiceset`) in way that, these input fields can appear just as readable fields when user is not taking any action and when user clicks or focusses or clicks on them, it allows user to update those fields and then user can use action buttons like Action.Submit/Action.Execute present on the card to send data back to the bot.

image

image

Inherited properties of all Input fields will have one more property called `inputStyle` whose type will be `InputStyle`

| Property | Type | Required | Description | Version |
| -------- | ---- | -------- | ----------- | ------- |
| **inputStyle** | `InputStyle?` | No | style hint for Input fields | 1.6 |

### inputStyle

Style hint for input fields.

* **Type**: `InputStyle`
* **Required**: No
* **Allowed values**:
  * `"revealOnHover"` : Should show input fields in read only view unless user takes any action.
  * `"default"` : Should show the existing view for input controls.

Since this property is inherited to all Input fields, it will be supported by all:

* `Input.Text`
* `Input.Number`
* `Input.Date`
* `Input.Time`
* `Input.Toggle`
* `Input.Choiceset`

### Sample Payload:

```json
     {
          "type": "Input.Text",
          "label": "Name",
          "value": "Sneh",
          "inputStyle" : "revealOnHover"
      },
      {
          "type": "Input.Date",
          "label": "Date of Arrival",
          "inputStyle" : "default"
      },
      {
          "type": "Input.Time",
          "label": "Time of Arrival",
          "value": "09:30",
          "inputStyle" : "revealOnHover"
      },
      {
          "type": "Input.Number",
          "label": "Number of Guest",
          "value": 5,
          "inputStyle" : "revealOnHover"
      }	
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8180)


specs/DesignDiscussions/InputInlineEditability.md
```diff
@@ -0,0 +1,159 @@
+# Inline Editability for Adaptive Cards
+
+## Overview
+
+Adaptive cards are used by apps to share information as well as to collect input from users to complete user scenarios.
+As of now, if we are showing some information in a card and if we also want to collect user’s input, we will have bunch of Input fields hidden behind a ShowCard button or we will launch some form using button click on the card to collect user's input.
```
anna-dingler, line 6:
```suggestion
As of now, if we are showing some information in a card and if we also want to collect user’s input, we will have a bunch of Input fields hidden behind a ShowCard button or we will launch some form using button click on the card to collect user's input.
```

specs/DesignDiscussions/InputInlineEditability.md
```diff
@@ -0,0 +1,159 @@
...
+In both cases we might have duplicate readable fields and input fields.
+
+We want to provide better experience to the user where he can read the fields and can also seamlessly update the values at the same place. 
+
+UI of showing data and taking user input will become simplistic. 
+
+ 
+## Current experience:
+
+As shown in the below picture, An adaptive card displays data for Customer name, Est. Revenue and Est. Clode Date. If user wants to update those information, he will have to click on 'Edit' button which will open some form and there he can update values for 'Est. Revenue' etc. and Click on Save to send data back to the bot service.
```
anna-dingler, line 16:
```suggestion
As shown in the below picture, An adaptive card displays data for Customer name, Est. Revenue and Est. Close Date. If user wants to update those information, he will have to click on 'Edit' button which will open some form and there he can update values for 'Est. Revenue' etc. and Click on Save to send data back to the bot service.
```

specs/DesignDiscussions/InputInlineEditability.md
```diff
@@ -0,0 +1,159 @@
...
+![img](./assets/InputInlineEdit/InputInlineEditScenario1.PNG)
+
+
+As of now, Adaptive cards support `Input` fields to collect user input. This is how current user experience of `Input` fields looks like. 
+
+ ![img](./assets/InputInlineEdit/InputInlineEditScenario2.PNG)
+
+## Proposed experience:
+
+We will allow bot developer to enhance the user experience of all `Input` fields in Adaptive card (such as `Input.Text`, `Input.Number`, `Input.Date`, `Input.Time`, `Input.Toggle` and `Input.Choiceset`) in way that, these input fields can appear just as readable fields when user is not taking any action and when user clicks or focusses or clicks on them, it allows user to update those fields and then user can use action buttons like Action.Submit/Action.Execute present on the card to send data back to the bot.
```
anna-dingler, line 27:
```suggestion
We will allow bot developer to enhance the user experience of all `Input` fields in Adaptive card (such as `Input.Text`, `Input.Number`, `Input.Date`, `Input.Time`, `Input.Toggle` and `Input.Choiceset`). These input fields can appear just as readable fields when the user is not taking any action, and when the user clicks or focusses on them, it allows the user to update those fields. The user can then use action buttons like Action.Submit/Action.Execute present on the card to send data back to the bot.
```

specs/DesignDiscussions/InputInlineEditability.md
```diff
@@ -0,0 +1,159 @@
...
+
+2. When user hovers on the card, 
+	* we will show outer outline on that card.
+	* we will show a lighter color outline to all the input fields which are inline editable.
+	
+	This will guide user to the input fields which he can go an update right on the card.
+	
+![img](./assets/InputInlineEdit/InputInlineEditScenario4.PNG)
+
+3. When user takes hovers over the input field, we will show a darker outline in that input field to guide user that the field is activated and he can now edit the field.
```
anna-dingler, line 42: Should this be "When the user clicks on the input field"?

specs/DesignDiscussions/InputInlineEditability.md
```diff
@@ -0,0 +1,159 @@
...
+4. User can go and edit the field.
+
+![img](./assets/InputInlineEdit/InputInlineEditScenario6.PNG)
+
+5. When user has updated the input field, Action.Submit or Action.Execute button such as "Save" which has  `disabledUnlessAssociatedInputsChange` property (https://github.com/microsoft/AdaptiveCards/issues/7103) will get enabled automatically. It will guide user to hit the button to commit the changes made on the card.
+
+![img](./assets/InputInlineEdit/InputInlineEditScenario7.PNG)
+
+
+6. When user clicks hits on the Action.Submit or Action.Execute button, latest input values will be sent to the bot from host and bot can then update the card with latest information which we will show in default state.
```
anna-dingler, line 55:
```suggestion
6. When the user clicks on the Action.Submit or Action.Execute button, latest input values will be sent to the bot from host and bot can then update the card with latest information which we will show in default state.
```

specs/DesignDiscussions/InputInlineEditability.md
```diff
@@ -0,0 +1,159 @@
...
+
+3. When user takes hovers over the input field, we will show a darker outline in that input field to guide user that the field is activated and he can now edit the field.
+
+![img](./assets/InputInlineEdit/InputInlineEditScenario5.PNG)
+
+4. User can go and edit the field.
+
+![img](./assets/InputInlineEdit/InputInlineEditScenario6.PNG)
+
+5. When user has updated the input field, Action.Submit or Action.Execute button such as "Save" which has  `disabledUnlessAssociatedInputsChange` property (https://github.com/microsoft/AdaptiveCards/issues/7103) will get enabled automatically. It will guide user to hit the button to commit the changes made on the card.
```
anna-dingler, line 50:
```suggestion
5. When the user has updated the input field, Action.Submit or Action.Execute button such as "Save" which has `disabledUnlessAssociatedInputsChange` property (https://github.com/microsoft/AdaptiveCards/issues/7103) will get enabled automatically. It will guide user to hit the button to commit the changes made on the card.
```

specs/DesignDiscussions/InputInlineEditability.md
```diff
@@ -0,0 +1,159 @@
...
+
+![img](./assets/InputInlineEdit/InputInlineEditScenario7.PNG)
+
+
+6. When user clicks hits on the Action.Submit or Action.Execute button, latest input values will be sent to the bot from host and bot can then update the card with latest information which we will show in default state.
+
+![img](./assets/InputInlineEdit/InputInlineEditScenario8.PNG)
+
+## New Capabilites in the card
+In order to achieve inline editable experience as mentioned above, we will add these capabilities in the card for developers and host:
```
anna-dingler, line 60: Are there supposed to be capabilities listed here? If not, it seems like we can remove the "New Capabilities in the card" section.

specs/DesignDiscussions/InputInlineEditability.md
```diff
@@ -0,0 +1,159 @@
...
+
+![img](./assets/InputInlineEdit/InputInlineEditScenario9.PNG)
+
+
+
+## Host configurable properties in card:
+1. Host can configure these style properties of `label` and `value` field: `font`, `font color` and `font weight`. 
+
+## Use Case Clarifications:
+1. If `value` is `empty` in the input field. We will show label and placeholder text in lighter color like this in the [default] state,
```
anna-dingler, line 79:
```suggestion
1. If `value` is `empty` in the input field, we will show label and placeholder text in lighter color like this in the [default] state,
```

specs/DesignDiscussions/InputInlineEditability.md
```diff
@@ -0,0 +1,159 @@
...
+
+![img](./assets/InputInlineEdit/InputInlineEditScenario10.PNG)
+
+which user can go and update the value,
```
anna-dingler, line 83:
```suggestion
which user can go and update the value.
```

specs/DesignDiscussions/InputInlineEditability.md
```diff
@@ -0,0 +1,159 @@
...
+![img](./assets/InputInlineEdit/InputInlineEditScenario11.PNG)
+
+2. If `label` is `empty`, we will just show `value` field in readable format in the default state which user can go and update the value.
+
+![img](./assets/InputInlineEdit/InputInlineEditScenario12.PNG)
+
+
+## Developer Recommendation:
+
+1. **Save Button:** Our recommendation is to have a `Save` button (Action.Submit or Action.Execute) with `disabledUnlessAssociatedInputsChange` on the card with inline editable fields. Save button will send the modified user input values to the bot and bot will respond with a card with updated input values.
```
anna-dingler, line 94:
```suggestion
1. **Save Button:** Our recommendation is to have a `Save` button (Action.Submit or Action.Execute) with `disabledUnlessAssociatedInputsChange` on the card with inline editable fields. Save button will send the modified user input values to the bot, and the bot will respond with a card with updated input values.
```

anna-dingler (changes requested):
Overall, looks good! I just added some grammar changes.

As Joseph mentioned in #8143, it would probably be best if the schema changes were added closer to adding the implementation. This will make it easier if we need to cut a release before the implementation is added.

specs/DesignDiscussions/InputInlineEditability.md
baton17: removed

specs/DesignDiscussions/InputInlineEditability.md
baton17: first experience is when user hovers anywhere on the card, we show some light outline and then second experience is when user hovers directly over the input, we want to show some darker outline to indicate that the input is active for editing.

pankajmi2 (approved):

source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -272,3 +272,21 @@ export function addCancelSelectActionEventHandler(element: HTMLElement) {
         e.cancelBubble = true;
     };
 }
+
+export function SetDateOrTimePickerVisibility(input: HTMLInputElement, value: string | undefined, show: boolean) {
```
dclaux, line 276: Function name should start with a lower case letter

source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -272,3 +272,21 @@ export function addCancelSelectActionEventHandler(element: HTMLElement) {
...
+    if (!!show) {
+        // shows the time/date picker icon
+        input.readOnly = false;
+        if (value) {
```
dclaux, line 280: Empty line before if

source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -272,3 +272,21 @@ export function addCancelSelectActionEventHandler(element: HTMLElement) {
...
+            // shows the cross button icon
+            input.required = false;
+        }
+        return;
```
dclaux, line 284: Empty line after }

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3647,6 +3675,10 @@ export class TextInput extends Input {
             result = document.createElement("input");
             result.className = this.hostConfig.makeCssClassName("ac-input", "ac-textInput");
             result.type = Enums.InputTextStyle[this.style].toLowerCase();
+
+            if (this.inputStyle !== null && this.inputStyle === Enums.InputStyle.ReadWrite && !this.inlineAction) {
```
dclaux, line 3679: This code is repeated in multiple places. It should be implemented in the base `Input` class' `render' method and the style should be applied to whatever `internalRender` returns.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3647,6 +3675,10 @@ export class TextInput extends Input {
...
+                result.classList.add(this.hostConfig.makeCssClassName("ac-textInput-readWriteStyle"));
```
dclaux, line 3680: You shouldn't use class names that are specific to each input type. So instead of `ac-textInput-readWriteStyle` you should use `ac-textInput ac-inputStyle-readWrite`. This way you can have generic code that adds the `ac-inputStyle-readWrite` class to any input regardless of its actual type.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4560,6 +4600,16 @@ export class DateInput extends Input {
...
         }
 
+        if (this.inputStyle !== null && this.inputStyle === Enums.InputStyle.ReadWrite) {
+            this._dateInputElement.classList.add(this.hostConfig.makeCssClassName("ac-dateInput-readWriteStyle"));
+            Utils.SetDateOrTimePickerVisibility(this._dateInputElement, this.defaultValue, false /*show*/);
+
+            const handleHoverOnCard = (input: HTMLInputElement, eventType: Enums.CardEventType) => 
+                Utils.SetDateOrTimePickerVisibility(input, this._dateInputElement.value, eventType === Enums.CardEventType.OnMouseEnterOnCard /*show*/);
+
+            this.registerEventsForReadWriteInputStyle( this._dateInputElement, handleHoverOnCard.bind(this));
```
dclaux, line 4610: Can you please explain what this does? I'm not getting it nor why it would be necessary?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -8582,6 +8645,18 @@ export class AdaptiveCard extends ContainerWithActions {
                 if (this.speak) {
                     renderedCard.setAttribute("aria-label", this.speak);
                 }
+
+                // register for on mouse enter on the card
+                renderedCard.onmouseenter = (ev: MouseEvent) => {
```
dclaux, line 8650: Why are these necessary?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -8612,6 +8687,14 @@ export class AdaptiveCard extends ContainerWithActions {
     get hasVisibleSeparator(): boolean {
         return false;
     }
+
+    public registerMouseEnterCallback(callback: any) {
```
dclaux, line 8691: Is the goal that all inputs are "revealed" as the mouse enters the Adaptive Card? As opposed to each input being revealed when hovered over?

If so, is that truly what partners want? When I worked with Dynamics on their requirements, they wanted individual inputs to reveal on hover.

Per the screenshots, which match what I thought we were after, that isn't the case. But then I really do not understand all that gymnastic with mouse enter/leave events.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -8612,6 +8687,14 @@ export class AdaptiveCard extends ContainerWithActions {
...
+        this._mouseEnterCallbacks.push(callback);
```
dclaux, line 8692: If the goal is to reveal all inputs when the mouse enters the card, isn't there a simpler solution, maybe even a CSS only one? Namely add a `:hover` style to the `ac-adaptiveCard` class that adds the appropriate style to anything that has the ac-input style under the card?

source/nodejs/adaptivecards/src/utils.ts
dclaux, line 276: Very confused about what this logic achieves and why it's necessary. Could you explain?

specs/DesignDiscussions/InputInlineEditability.md
```diff
@@ -0,0 +1,155 @@
...
+
+6. When the user clicks on the Action.Submit or Action.Execute button, latest input values will be sent to the bot from host and bot can then update the card with latest information which we will show in default state.
+
+![img](./assets/InputInlineEdit/InputInlineEditScenario8.PNG)
+
+## New Capabilites in the card
+In order to achieve inline editable experience as mentioned above, we will add these capabilities in the card for developers and host:
+
+## New Schema Changes in Adaptive Card
+1. Introduce an optional styling property `inputStyle` in AC input fields schema for developer to choose between existing input fields styling i.e.`default` vs inline editable styling i.e.`readWrite`.
```
dclaux, line 63: This still refers to "inline editable" and should be changed

schemas/1.6.0/adaptive-card.json
```diff
@@ -2219,6 +2225,20 @@
         }
       ]
     },
+    "InputStyle": {
+      "description": "Style hint for input.",
+      "anyOf": [
+        {
+          "enum": [
+            "readWrite",
```
dclaux, line 2233: I don't think this is the right style name:
1. `default` would be "same as today" and inputs are definitely already read-write
2. Given that inputs are read-write in `default` mode, `readWrite` makes no sense. The style is purely visual, it doesn't make inputs read-only. We need a style name that acknowledges that. I suggest `borderLess`

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3458,6 +3471,21 @@ export abstract class Input extends CardElement implements IInput {
         }
     }
 
+    protected registerEventsForReadWriteInputStyle(input: HTMLElement | undefined, inputEventHandler: (input: HTMLElement, eventType: Enums.CardEventType) => void) {
```
dclaux, line 3474: I really think this isn't/shouldn't be necessary

source/nodejs/adaptivecards/src/card-elements.ts
dclaux, line 3679: The SDK uses "undefined" across the board, and not null. The value of `this.inputStyle` will be read from the underlying property bag which will return `undefined` if `inputStyleProperty` doesn't have a value. Thus the check for `this.inputStyle !== null` is unnecessary and will always return `true`. Either change it to `this.inputStyle !== undefined` or remove it altogether (it is unnecessary).

source/nodejs/adaptivecards/src/scss/adaptivecards-base.scss
```diff
@@ -300,6 +304,36 @@ $font-family: 'Segoe UI', sans-serif;
...
+            appearance: none;
+        }
+    }
+
+    .ac-adaptiveCard:hover .ac-textInput-readWriteStyle,
+    .ac-adaptiveCard:hover .ac-numberInput-readWriteStyle,
+    .ac-adaptiveCard:hover .ac-timeInput-readWriteStyle,
+    .ac-adaptiveCard:hover .ac-dateInput-readWriteStyle,
+    .ac-adaptiveCard:hover .ac-choiceSetInput-compact-readWriteStyle {
+        border: $border-readWrite-style-oncard-hover;
```
dclaux, line 330: You already define a `:hover` style for the card, why not define it in such a way that it also updates the style of the card's input, instead of relying on JS events?

And again, don't use input-type-specific class names, use a separate `ac-inputStyle-readWrite` (I should say `ac-inputStyle-borderLess`) and use it in connjunction with `ac-input`.

source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -272,3 +272,21 @@ export function addCancelSelectActionEventHandler(element: HTMLElement) {
         e.cancelBubble = true;
     };
 }
+
+export function SetDateOrTimePickerVisibility(input: HTMLInputElement, value: string | undefined, show: boolean) {
+    if (!!show) {
+        // shows the time/date picker icon
+        input.readOnly = false;
```
dclaux, line 279: Why this?

source/nodejs/adaptivecards/src/utils.ts
dclaux, line 276: What is the `value` parameter? The function name doesn't seem to call for a value and doesn't make it clear what it is for.

source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -272,3 +272,21 @@ export function addCancelSelectActionEventHandler(element: HTMLElement) {
...
+        if (value) {
+            // shows the cross button icon
+            input.required = false;
```
dclaux, line 282: Why this?

dclaux (changes requested):

source/nodejs/adaptivecards/src/card-elements.ts
baton17: removed eventing mechanism

source/nodejs/adaptivecards/src/utils.ts
baton17: to update visibility of date and time pickers

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4450,15 +4456,24 @@ export class ChoiceSetInput extends Input {
...
         }
     }
 
+    updateVisualState(hover: boolean): void {
+        if (!this.isMultiSelect && this.isCompact) {
+            super.updateVisualState(hover);
+            if (this._selectElement && this.inputStyle === Enums.InputStyle.RevealOnHover) {
+                this._selectElement.classList.add(
+                    this.hostConfig.makeCssClassName(
+                        "ac-choiceSetInput-inputStyle-revealOnHover"
```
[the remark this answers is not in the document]
dclaux: You seem to be missing two things:
- Only add the class if `hover` is `false`
- Remove the class if `hover` is `true`

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4720,17 +4731,15 @@ export class DateInput extends Input {
...
+    }
 
-            const handleHoverOnCard = (input: HTMLInputElement, eventType: Enums.CardEventType) => 
-                Utils.SetDateOrTimePickerVisibility(input, this._dateInputElement.value, eventType === Enums.CardEventType.OnMouseEnterOnCard /*show*/);
+    updateVisualState(hover: boolean): void {
+        super.updateVisualState(hover);
 
-            this.registerEventsForReadWriteInputStyle( this._dateInputElement, handleHoverOnCard.bind(this));
+        if (this._dateInputElement && this.inputStyle === Enums.InputStyle.RevealOnHover) {
+            Utils.updateDateOrTimePickerVisibility(this._dateInputElement, hover);
```
dclaux, line 4741: This utilisty function takes an `HTMLInputElement` as a paremeter, so maybe naming it `updateInputVisibility` makes more sense?

Also, that function as part of `Utils` is exported and thus usable by developers that consume the SDK. However it doesn't make much sense to them as it's ultra specific to what you need to do. I suggest moving it to `card-elements.ts` as  non-exported function.

source/nodejs/adaptivecards/src/scss/adaptivecards-base.scss
```diff
@@ -305,32 +305,24 @@ $font-family: 'Segoe UI', sans-serif;
...
         }
     }
 
-    .ac-adaptiveCard:hover .ac-textInput-readWriteStyle,
-    .ac-adaptiveCard:hover .ac-numberInput-readWriteStyle,
-    .ac-adaptiveCard:hover .ac-timeInput-readWriteStyle,
-    .ac-adaptiveCard:hover .ac-dateInput-readWriteStyle,
-    .ac-adaptiveCard:hover .ac-choiceSetInput-compact-readWriteStyle {
-        border: $border-readWrite-style-oncard-hover;
+    .ac-adaptiveCard:hover .ac-inputStyle-revealOnHover {
```
dclaux, line 321: Nit: technically not necessary, but should probably still be `.ac-adaptiveCard:hover .ac-input .ac-inputStyle-revealOnHover`

source/nodejs/adaptivecards/src/scss/adaptivecards-base.scss
```diff
@@ -305,32 +305,24 @@ $font-family: 'Segoe UI', sans-serif;
...
+        border: $revealOnHover-subdued-border;
     }
 
-    .ac-adaptiveCard:hover .ac-choiceSetInput-compact-readWriteStyle {
+    .ac-adaptiveCard:hover .ac-choiceSetInput-inputStyle-revealOnHover {
```
dclaux, line 325: Should be `.ac-adaptiveCard:hover .ac-choiceSetInput .ac-inputStyle-revealOnHover`

source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -273,20 +273,17 @@ export function addCancelSelectActionEventHandler(element: HTMLElement) {
     };
 }
 
-export function SetDateOrTimePickerVisibility(input: HTMLInputElement, value: string | undefined, show: boolean) {
-    if (!!show) {
+export function updateDateOrTimePickerVisibility(input: HTMLInputElement, hover: boolean) {
```
dclaux, line 276: If you follow my recommendation to move this function to `card-elements.ts`, I would also suggest renaming it - you're not updating the visibility of the whole input as the current name suggests, but only of the buttons/icons that adorn the input. So maybe `updateInputAdornersVisibility`?

dclaux (changes requested):
Thanks for making those changes. Looks good overall, with a couple more suggestions.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3591,19 +3598,13 @@ export abstract class Input extends CardElement implements IInput {
...
     }
 
-    protected registerEventsForReadWriteInputStyle(input: HTMLElement | undefined, inputEventHandler: (input: HTMLElement, eventType: Enums.CardEventType) => void) {
-        if (!input) {
-            return;
+    updateVisualState(hover: boolean): void {
+        if (!hover) {
+            // when user is not hovering on the card, we will show input fields as read only view
+            if (this._renderedInputControlElement && this.inputStyle === Enums.InputStyle.RevealOnHover) {
+                this._renderedInputControlElement.classList.add(this.hostConfig.makeCssClassName("ac-inputStyle-revealOnHover"));
```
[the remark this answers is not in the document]
baton17: actually I want to add this class "ac-inputStyle-revealOnHover" at render time. and when hover happens I want to identify the input using this class. For this rule to work .ac-adaptiveCard:hover .ac-input.ac-inputStyle-revealOnHover.

So I dont want to remove the class when hover is true.

It have created two classes then , 
1. updateVisualStateOnRender: It will add common "ac-inputStyle-revealOnHover" to all inputs. Input can override to do some extra work like hiding date picker by input.date
2. On card hover this rule .ac-adaptiveCard:hover .ac-input.ac-inputStyle-revealOnHover already takes care of adding/removing certain style like border. But date/time input needs to do some extra work like shw/hide the icons for date/time picker
3. updateVisualStateOnHover: This class has empty implementation in base Input class. Date and Time input overrides to show and hide icon when hover is there.

Let me know if this sound okay to you.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -349,6 +349,11 @@ export abstract class CardElement extends CardObject {
         }
     }
 
+    protected updateInputsVisualState(hover: boolean) {
```
anna-dingler, line 352: Is there a reason why this is added at the `CardElement` level instead of at the `AdaptiveCard` or `Container` level?

It looks like this method is only called from within `AdaptiveCard`

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -8723,6 +8798,16 @@ export class AdaptiveCard extends ContainerWithActions {
                 if (this.speak) {
                     renderedCard.setAttribute("aria-label", this.speak);
                 }
+
+                // register for on mouse enter on the card
```
anna-dingler, line 8802: Do we track these events so that `updateInputAdornersVisibility` will be called?

source/nodejs/adaptivecards/src/scss/adaptivecards-base.scss
```diff
@@ -300,6 +304,32 @@ $font-family: 'Segoe UI', sans-serif;
...
+    .ac-adaptiveCard:hover .ac-input.ac-inputStyle-revealOnHover {
+        border: $revealOnHover-subdued-border;
+    }
+
+    .ac-adaptiveCard:hover .ac-input.ac-multichoiceInput.ac-choiceSetInput-compact.ac-inputStyle-revealOnHover {
+        appearance: auto;
     }
 }
 
+
```
anna-dingler, line 334: nit: remove extra lines

source/nodejs/adaptivecards/src/utils.ts
```diff
@@ -271,4 +271,4 @@ export function addCancelSelectActionEventHandler(element: HTMLElement) {
         e.preventDefault();
         e.cancelBubble = true;
     };
-}
+}
```
anna-dingler, line 274: nit: revert the changes to this file

source/nodejs/adaptivecards/src/card-elements.ts
baton17: David has suggested this. I tried to put this in AdaptiveCard class but this.getAllInputs() was not available there. So Looks like AdaptiveCard does not implement getAllInputs

source/nodejs/adaptivecards/src/card-elements.ts
baton17: I need to remove this comment. I had added for my previous implementation. Now It calls into updateInputsVisualState and that calls into Input.updateVisualState

anna-dingler:
@jwoo-msft Do you have any thoughts on this?

specs/DesignDiscussions/InputRevealOnHover.md
```diff
@@ -0,0 +1,154 @@
...
+
+2. **Cancel Button:** Our recommendation is to have a `Cancel` button (Action.Submit or Action.Execute) on the card containing inputs fields with `revealOnHover` style. If user has made some changes in the input fields but want to revert back to original state, they can click on "Cancel" button. "Cancel" button will send the card with last saved input values.
+## Out of Scope:
+
+1. Inline Action in Input.text: We will not support `revealOnHover` style for input text containing inline action. It will be shown as default input style only. Reason being, such sceanrios are reply with a comment or adding a message where user collaboration is not needed.
+
+![img](./assets/InputRevealOnHover/InputRevealOnHoverScenario13.png)
+
+2. Mobile is `out of scope` for `revealOnHover` input style design since there is no focus state on mobile and alignment will be always `vertical` as of today.
```
jwoo-msft, line 98: we do need fallback scenarios for mobile. for example, when this card is sent to mobile users, how would the experience look like? and what's the recommended changes if any?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3516,6 +3534,12 @@ export abstract class Input extends CardElement implements IInput {
                 }
             }
 
+            this.updateVisualState(true /*initialRender*/);
```
jwoo-msft, line 3537: can you use named parameter for maintainers?

samples/v1.6/Elements/InputStyle.json
```diff
@@ -0,0 +1,85 @@
+{
+	"type": "AdaptiveCard",
+	"version": "1.6",
+	"body": [
+		{
+			"type": "Input.Text",
+			"id": "inputProduct",
```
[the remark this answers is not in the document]
baton17: this file has tabs only. I checked again 
image

source/nodejs/adaptivecards/src/card-elements.ts
baton17: fixed

specs/DesignDiscussions/InputRevealOnHover.md
```diff
@@ -0,0 +1,154 @@
...
+![img](./assets/InputRevealOnHover/InputRevealOnHoverScenario3.PNG)
+
+2. When user hovers anywhere on the card,
+	* we will show a lighter color outline to all the input fields.
+	
+	This will guide user to the input fields which he can go an update right on the card.
+	
+![img](./assets/InputRevealOnHover/InputRevealOnHoverScenario4.PNG)
+
+3. When user hovers over the input field, we will show a darker outline in that input field to guide user that the field is activated and he can now edit the field.
```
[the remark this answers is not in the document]
baton17: No, cursor will behavior is same for all inputs with and without revealOnHover. We just show a different outline on the input to indicate that you can edit the field now.

specs/DesignDiscussions/InputRevealOnHover.md
baton17: Right now. this feature is out of mobile scope. But should we add parsing support in mobile and no-op implementation for this? I can take this in a subsequent PR,

jwoo-msft: it's more about documenting how this should be handled in the mobile. if the mobile experience is as good as you have designed, no changes may be needed.

source/nodejs/adaptivecards/src/card-elements.ts
jwoo-msft: I like your approach here that updateInputVisualState is called once per visual update, but it is a bit confusing that the call is implemented at the CardElement. Can you change the name of this api to more distinct from UpdateVisualState() and add a comment that this is mean to be called only at the card level?

jwoo-msft (approved):

source/nodejs/adaptivecards/src/card-elements.ts
baton17: Do you have any suggestions in mind? This "updateInputsVisualState" was suggested by David and I see similar methods names in CardElement as "updateActionsEnabledState", 
image

baton17: I have added comment on this class indicating why input.updateVisualState() needs to be called on card level on mouse events.

Also input.updateVisualState() is called from other places as well like onBlur event on input or during initial render of inputs. So this updateVisualState is not exclusive to card level events.

jwoo-msft: I meant only for updateInputVisualState. It's really confusing when the function names are different just a little bit. updateInputsVisualState vs updateVisualState

PR #8203, opened 2023-01-05
[UWP / WinUI 3 Renderer] Move code that will be shared between UWP and WinUI 3 ObjectModel

krschau:
# Related Issue

# Description

A large amount of the UWP Object Model code will be shared with the WinUI 3 Object Model. To prepare it to be used in both places, this change moves the code to a Shared Project. Separating out this change helps ensure the file histories are kept through the rename, as both renaming and editing files in the same change can cause Git to think they are different files.

# How Verified

Build/pipelines.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8203)


asklar (approved):

almedina-ms (approved):
:shipit:

PR #8226, opened 2023-01-12
[WinUI 3 Renderer] Prepare SharedRenderer code for WinUI 3 Renderer

krschau:
# Related Issue

# Description

This change removes explicit references to UWP in the SharedRenderer code, replacing them with a variable that can be changed depending on if the code is compiling for UWP or WinUI 3. For references to the Object Model, we use the Xaml_OM alias that has already been introduced. For the Renderer, we use Xaml_Rendering. This also means we can remove the render_xaml and winrt_render_xaml aliases, which were never ideal. Instances of these go back to fully qualified namespaces (ie. `::AdaptiveCards::Rendering::Xaml_Rendering`).

# How Verified

Build/pipelines.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8226)

### DIFF
--- a/source/uwp/SharedRenderer/dll/CppWinRTIncludes.h
+++ b/source/uwp/SharedRenderer/dll/CppWinRTIncludes.h
@@ -3,6 +3,12 @@
 
 #pragma once
 
+#ifndef USE_WINUI3
+#define Xaml_OM Uwp
+#else
+#define Xaml_OM Winui3
+#endif
+
 #include <winrt/base.h>
 #include <winrt/Windows.Data.Json.h>
 #include <winrt/Windows.Data.Xml.Dom.h>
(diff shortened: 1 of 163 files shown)


source/uwp/SharedRenderer/dll/CppWinRTIncludes.h
```diff
@@ -3,6 +3,14 @@
 
 #pragma once
 
+#ifndef USE_WINUI3
+#define Xaml_OM Uwp
+#define Xaml_Rendering Uwp
+#else
+#define Xaml_OM Winui3
+#define Xaml_Rendering Winui3
+#endif
```
paulcam206, line 12: The pickiest nit to nitpick:

```suggestion
#ifdef USE_WINUI3
#define Xaml_OM Winui3
#define Xaml_Rendering Winui3
#else
#define Xaml_OM Uwp
#define Xaml_Rendering Uwp
#endif
```

paulcam206 (approved):

PR #8228, opened 2023-01-13
[UWP Renderer] Add a null check to inline action rendering

### LINKED ISSUE #8227: [UWP Renderer] Regression with inline actions
# Platform

* UWP

# Issue

If an inline action is a `ShowCard` action, we are returning a `nullptr` as the expected container. This is causing an Access Violation when we attempt to reference the container.

Expected Behavior:
We should not replace the existing container with a `nullptr` if an inline action is a `ShowCard` action.

anna-dingler:
# Related Issue

Fixes #8227 

# Description

If an inline action for Input.Text was a ShowCard action, we were setting the container to a `nullptr`. This caused an AccessViolation.

I added a null check so that we do not assign the container to a nullptr for ShowCard actions.

# Sample Card

https://github.com/microsoft/AdaptiveCards/blob/release/uwp-v3.1.0/source/uwp/AdaptiveCardTestApp/Expected/SourceCards/v1.2.Tests.Input.Text.InlineAction.ShowCard.a724376.json

# How Verified

Verified manually

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8228)

### DIFF
--- a/source/uwp/Renderer/lib/AdaptiveTextInputRenderer.cpp
+++ b/source/uwp/Renderer/lib/AdaptiveTextInputRenderer.cpp
@@ -41,8 +41,14 @@ namespace winrt::AdaptiveCards::Rendering::Uwp::implementation
             bool isMultiline = adaptiveTextInput.IsMultiline();
             winrt::TextInputStyle style = adaptiveTextInput.TextInputStyle();
             isMultiline &= style != winrt::TextInputStyle::Password;
-            textBoxParentContainer = render_xaml::ActionHelpers::HandleInlineAction(
+            auto textBoxWithInlineAction = render_xaml::ActionHelpers::HandleInlineAction(
                 renderContext, renderArgs, inputUIElement, textBoxParentContainer, isMultiline, inlineAction);
+
+            // If the inlineAction is ShowCard, textBoxWithInlineAction will be null and we should not update the parent container
+            if (textBoxWithInlineAction)
+            {
+                textBoxParentContainer = textBoxWithInlineAction;
+            }
         }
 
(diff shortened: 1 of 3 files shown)


source/uwp/Renderer/lib/AdaptiveTextInputRenderer.cpp
```diff
@@ -41,8 +41,14 @@ namespace winrt::AdaptiveCards::Rendering::Uwp::implementation
             bool isMultiline = adaptiveTextInput.IsMultiline();
             winrt::TextInputStyle style = adaptiveTextInput.TextInputStyle();
             isMultiline &= style != winrt::TextInputStyle::Password;
-            textBoxParentContainer = render_xaml::ActionHelpers::HandleInlineAction(
+            auto textBoxWithInlineAction = render_xaml::ActionHelpers::HandleInlineAction(
```
almedina-ms, line 44: is a warning raised to the user that the inline action will not activate when clicked?

almedina-ms (approved):

almedina-ms (approved):

source/uwp/Renderer/lib/AdaptiveTextInputRenderer.cpp
anna-dingler: Yes, we add a warning in `HandleInlineAction`

https://github.com/microsoft/AdaptiveCards/blob/38b8f0934eeb5900f3d948a9691a94376af576b3/source/uwp/SharedRenderer/lib/ActionHelpers.cpp#L355-L358

PR #8233, opened 2023-01-17
[UWP Renderer] Update ActionSet fallback

### LINKED ISSUE #8323: [UWP Renderer] Card does not render if there is a faulty action without a fallback mechanism.
# Platform

* UWP

# Issue

When attempting to handle the fallback scenario for an action, we throw an `E_FAIL` error if no mechanism is provided (default).

https://github.com/microsoft/AdaptiveCards/blob/38b8f0934eeb5900f3d948a9691a94376af576b3/source/uwp/SharedRenderer/lib/ActionHelpers.cpp#L734-L756

This error is never handled, and an error is shown in the visualizer.

image

anna-dingler:
# Related Issue

Fixes #8323

# Description

When attempting to handle the fallback scenario for an action, we throw an `E_FAIL` error if no mechanism is provided (default).

https://github.com/microsoft/AdaptiveCards/blob/38b8f0934eeb5900f3d948a9691a94376af576b3/source/uwp/SharedRenderer/lib/ActionHelpers.cpp#L734-L756

This error is never handled, and an error is shown in the visualizer.

image

I added a try/catch block surrounding this method call so that we continue rendering even if we drop the action.

I also updated the `actionType` when getting `FallbackContent`.

# Sample Card

https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.2/Elements/ActionFallback.json

# How Verified

actionFallback2

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8233)


source/uwp/Renderer/lib/ActionHelpers.cpp
```diff
@@ -870,18 +871,30 @@ namespace AdaptiveCards::Rendering::Uwp::ActionHelpers
...
+                                                                  showCardsStackPanel,
+                                                                  columnDefinitions,
+                                                                  renderContext,
+                                                                  renderArgs);
+
+                    currentButtonIndex++;
+                }
+                catch (winrt::hresult_error const& ex)
+                {
+                    // We want to continue if the error is E_FAIL
```
jwoo-msft, line 891: even if E_FAIL is returned only when fallback mechanism is missing. this can change in the future, and could we be better off with a specific error message for missing fallback?

jwoo-msft (approved):

source/uwp/Renderer/lib/ActionHelpers.cpp
anna-dingler: @jwoo-msft I just updated this - let me know what you think

source/uwp/SharedRenderer/lib/pch.h
```diff
@@ -22,6 +22,7 @@
 #define FACILITY_ADAPTIVECARDS 0xADA
 #define ERRORBASE_ADAPTIVECARDS 0x1000
 #define E_PERFORM_FALLBACK MAKE_HRESULT(1, FACILITY_ADAPTIVECARDS, ERRORBASE_ADAPTIVECARDS)
+#define E_FALLBACK_NOT_FOUND MAKE_HRESULT(2, FACILITY_ADAPTIVECARDS, ERRORBASE_ADAPTIVECARDS)
```
anna-dingler, line 25: @beervoley First time creating an hresult, is there anything else I need here?

beervoley: nope, I think this is fine :)

source/uwp/SharedRenderer/lib/ActionHelpers.cpp
```diff
@@ -764,12 +764,13 @@ namespace AdaptiveCards::Rendering::Xaml_Rendering::ActionHelpers
 
                 case winrt::FallbackType::Content:
                     action = action.FallbackContent();
+                    actionType = action.ActionType();
```
beervoley, line 767: still needed? if not, remove:)

anna-dingler: Yup, we need this so we know what kind of button to create in 'CreateActionButtonInActionSet'

PR #8234, opened 2023-01-17
[UWP Renderer] Allow nested fallbacks

anna-dingler:
# Related Issue

[TODO]

Nested fallbacks are not working correctly

# Description

1. Added a try/catch block so that we can make the recursive call to `RenderFallback` if we catch `E_PERFORM_FALLBACK`
2. Call `RenderFallback` if no `fallbackRenderer` is found
3. If the last fallback element within nested fallback does not render, `ancestorHasFallback` will be true. This caused us to throw `E_PERFORM_CALLBACK` which is caught by the initial element. If this element did not have a fallback, it will be dropped.

   Ex: Container (does not have fallback) -> ChildElement -> Fallback1 -> Fallback2 -> Fallback3
         ChildElement, Fallback1, and Fallback2 all failed. When Fallback3 fails, it will try to get the fallback mechanism of it's ancestor (Fallback2) by throwing `E_PERFORM_FALLBACK`. However, this error will land on the Container since all of the other elements failed. Without a fallback mechanism, the Container will be dropped.

    To fix this issue, I added a try/catch block for when we call `RenderFallback` on a child element. In the catch, we can verify that the parent element actually has a fallback mechanism before throwing the error.

# Sample Card

https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.2/Tests/ContainerDoubleFallback.json
https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.2/Tests/DeepFallback.json

# How Verified

## Before

image

image

## After

image

image

### DIFF
--- a/source/uwp/Renderer/lib/XamlHelpers.cpp
+++ b/source/uwp/Renderer/lib/XamlHelpers.cpp
@@ -321,15 +321,28 @@ namespace AdaptiveCards::Rendering::Uwp::XamlHelpers
 
             if (fallbackElementRenderer)
             {
-                fallbackControl = fallbackElementRenderer.Render(fallbackElement, renderContext, renderArgs);
-
-                renderedElement = fallbackElement;
+                try
+                {
+                    fallbackControl = fallbackElementRenderer.Render(fallbackElement, renderContext, renderArgs);
+                    renderedElement = fallbackElement;
+                }
+                catch (winrt::hresult_error const& ex)
+                {
+                    if (ex.code() == E_PERFORM_FALLBACK)
+                    {
+                        std::tie(fallbackControl, renderedElement) = RenderFallback(fallbackElement, renderContext, renderArgs);
+                    }
+                    else
+                    {
+                        throw(ex);
+                    }
+                }
             }
-
-            if (!fallbackControl)
+            else
             {
                 std::tie(fallbackControl, renderedElement) = RenderFallback(fallbackElement, renderContext, renderArgs);
             }
+
             fallbackHandled = true;
             break;
         }
--- a/source/uwp/Renderer/lib/XamlBuilder.cpp
+++ b/source/uwp/Renderer/lib/XamlBuilder.cpp
@@ -302,7 +302,18 @@ namespace AdaptiveCards::Rendering::Uwp
                     throw ex;
                 }
 
-                std::tie(newControl, renderedElement) = XamlHelpers::RenderFallback(element, renderContext, renderArgs);
+                try
+                {
+                    std::tie(newControl, renderedElement) = XamlHelpers::RenderFallback(element, renderContext, renderArgs);
+                }
+                catch (winrt::hresult_error const& ex)
+                {
+                    // if we get an E_PERFORM_FALLBACK error again, we should only throw it if `ancestorHasFallBack`
+                    if (ex.code() != E_PERFORM_FALLBACK || (ex.code() == E_PERFORM_FALLBACK && ancestorHasFallback))
+                    {
+                        throw ex;
+                    }
+                }
             }
 
             // If we got a control, add a separator if needed and the control to the parent panel


source/uwp/Renderer/lib/XamlHelpers.cpp
```diff
@@ -321,15 +321,28 @@ namespace AdaptiveCards::Rendering::Uwp::XamlHelpers
...
+                {
+                    if (ex.code() == E_PERFORM_FALLBACK)
+                    {
+                        std::tie(fallbackControl, renderedElement) = RenderFallback(fallbackElement, renderContext, renderArgs);
+                    }
+                    else
+                    {
+                        throw(ex);
+                    }
+                }
```
anna-dingler, line 339: Should keep the `fallbackControl` nullcheck that triggers the recursive call to `RenderFallback` here as well?

source/uwp/SharedRenderer/lib/XamlHelpers.cpp
```diff
@@ -333,15 +333,28 @@ namespace AdaptiveCards::Rendering::Xaml_Rendering::XamlHelpers
 
             if (fallbackElementRenderer)
             {
-                fallbackControl = fallbackElementRenderer.Render(fallbackElement, renderContext, renderArgs);
-
-                renderedElement = fallbackElement;
+                try
+                {
+                    fallbackControl = fallbackElementRenderer.Render(fallbackElement, renderContext, renderArgs);
```
anna-dingler, line 338: Should we keep the fallbackControl nullcheck that triggers the call to RenderFallback here as well?

beervoley: I don't think so? at this point it will just be null(=dropped), so I think we're fine with this?

beervoley: I've checked how https://github.com/microsoft/AdaptiveCards/edit/main/samples/v1.2/Tests/ContainerDoubleFallback.json this works in nodejs renderer - and I don't think that works correct?
It will render `Hello World` but not as one of the items of the container rather as `substitution` of the container?
I thought the `notme` item should simply be dropped? since it doesn't have fallback?:)
@paulcam206 could you please clarify the desired falldown behavior?
Docs say:
image

so (if I understand this correctly) it says that if any of its children cannot be rendered -> then we execute the fallback -> meaning entire container just get substituted for a `TextBlock`? What if it was container within a container and that container had unknown type and wouldn't have a fallback? The outer container would become `TextBlock` again?

paulcam206: Ah, fallback. This is by far one of the trickiest and least-used features of AC.

In the case of this particular test card, if the author wanted to replace the Container's `notme` element with some other element, `notme` would have `fallback` set on it. Since it doesn't, we look to the parent Container's `fallback`. We see that we do indeed have fallback (our `TextBlock`), and that gets subbed in for the `Container`.

paulcam206: Also, just a heads up -- this feature was first developed in the shared cpp object model, so its implementation is closest to intended behavior (though there may be bugs). Last I remember, the NodeJS renderer isn't quite as good (particularly in the area of detecting ID collisions that occur as the result of fallback), so if their behavior differs the starting assumption should be that NodeJS doesn't quite have it right.

For reference, the id collision stuff is documented here:

https://github.com/microsoft/AdaptiveCards/blob/dd943981c76d71a8f4163b3e128b1abf17662475/source/shared/cpp/ObjectModel/ParseContext.cpp#L26-L144

anna-dingler: @paulcam206  thank you for the info!

source/uwp/SharedRenderer/lib/XamlHelpers.cpp
```diff
@@ -333,15 +333,28 @@ namespace AdaptiveCards::Rendering::Xaml_Rendering::XamlHelpers
...
+                        throw(ex);
+                    }
+                }
             }
-
-            if (!fallbackControl)
+            else
             {
                 std::tie(fallbackControl, renderedElement) = RenderFallback(fallbackElement, renderContext, renderArgs);
             }
```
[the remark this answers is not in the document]
anna-dingler: The `catch` handles it if `fallbackElementRenderer.Render` fails.

The `else` block handles it if `fallbackElementRenderer == null`

beervoley (approved):

PR #8235, opened 2023-01-17
[UWP Renderer] Add a null check to requirements in MeetsRequirements

### LINKED ISSUE #8325: [UWP Renderer] Access Violation error for Custom Inputs
# Platform

* UWP

# Issue

Access violation is being thrown for custom inputs because we are missing a null check.

anna-dingler:
# Related Issue

Fixes #8325

# Description

Added a null check to `requirements` for when `MeetsRequirements` is called by a custom input.

Without the check, we were getting an Access Violation error.

# Sample Card

https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.3/Tests/customInput.json

# How Verified

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8235)


source/uwp/Renderer/lib/Util.cpp
```diff
@@ -481,25 +481,28 @@ bool MeetsRequirements(winrt::IAdaptiveCardElement const& cardElement, winrt::Ad
...
-        if (registrationVersion.empty())
-        {
-            meetsRequirementsLocal = false;
-        }
-        else
+        for (auto req : requirements)
         {
-            std::string requirementVersionString = HStringToUTF8(req.Version());
-            if (requirementVersionString != "*")
+            // winrt::hstring name = req.Name();
```
beervoley, line 488: nit: remove

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -481,25 +481,28 @@ bool MeetsRequirements(winrt::IAdaptiveCardElement const& cardElement, winrt::Ad
     winrt::IVector<winrt::AdaptiveRequirement> requirements = cardElement.Requirements();
     bool meetsRequirementsLocal = true;
 
-    for (auto req : requirements)
+    if (requirements)
```
beervoley, line 484: what is the reason for this variable to be null?

anna-dingler: @beervoley I'm not sure. We had a null check before switching to WinRT, and I added it back.

Without it, we crash at one of our test cases.

beervoley (approved):

source/uwp/SharedRenderer/lib/Util.cpp
```diff
@@ -481,25 +481,27 @@ bool MeetsRequirements(winrt::IAdaptiveCardElement const& cardElement, winrt::Ad
     winrt::IVector<winrt::AdaptiveRequirement> requirements = cardElement.Requirements();
     bool meetsRequirementsLocal = true;
 
-    for (auto req : requirements)
+    if (requirements)
```
beervoley, line 484: not blocking, but can you see when this variable is null? I don't think it should ever be null... even if it's called on custom input it should just be of size 0. If that's an option, can we simply initialize it to empty vector where it supposed to be initialized?:)
(feel free to skip)

anna-dingler: I'm not sure why this variable would be null. The for-each loop would have worked fine if the size was 0. Maybe it has something to do with the host config?

We could update the object model to return an empty list. I'll merge this PR for now to catch this error.

PR #8236, opened 2023-01-17
[UWP Renderer] Update min/max handling for inputs

### LINKED ISSUE #8326: [UWP Renderer] Min/max handling needs to be updated for inputs.
# Platform

* UWP

# Issue

Min and max values are not working properly on all input controls.

Test cases:
https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.0/Elements/Input.Time.json
https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.3/Elements/Input.Number.ErrorMessage.json

anna-dingler:
# Related Issue

Fixes #8326 

# Description

1. NumberInputs would only check the `min` if both `min` and `max` are present. I switched from `=` to `&=` when updating the `isValid` boolean.
2. We were calculating the `minTime` value with `1000000` instead of `10000000` (missing a zero).

# Sample Card

https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.0/Elements/Input.Time.json
https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.3/Elements/Input.Number.ErrorMessage.json

# How Verified

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8236)


source/uwp/Renderer/lib/InputValue.cpp
```diff
@@ -184,13 +184,13 @@ namespace winrt::AdaptiveCards::Rendering::Uwp
...
                     double maxDouble = max.Value();
-                    isValid = (currentDouble <= maxDouble);
+                    isValid &= (currentDouble <= maxDouble);
                 }
 
                 if (min)
                 {
                     double minDouble = min.Value();
-                    isValid = (currentDouble >= minDouble);
+                    isValid &= (currentDouble >= minDouble);
```
beervoley, line 193: I don't think this change changes anything :)

anna-dingler: I added this because we reset the value at ln. 193

So say we failed the max condition (187), but we passed the min condition (193), we were returning `true` when we should return `false || true = false`

beervoley: I see :)

beervoley (approved):

PR #8237, opened 2023-01-17
[WinUI 3 Renderer] Add Renderer for WinUI 3

krschau:
# Related Issue

# Description

This change introduces an Adaptive Card Renderer for WinUI 3. Three projects are introduced:
- AdaptiveCardRenderer: a WinUI 3 runtime component version of the UWP renderer. This project reuses the code in the SharedRenderer Shared Project.
- RendererCsProjection: so that the runtime component can be used in WinUI 3 C# projects.
- SimpleVisualizer: A very simple WinUI 3 application that just displays one hardcoded Adaptive Card.

The [XamlStraddle](https://github.com/asklar/XamlStraddle) library takes care of much of the work of getting System Xaml and WinUI 3 to play nicely with each other. A few changes are made to the Renderer code to accommodate the differences:
- WinUI 3 uses `MediaPlayerElement` rather than `MediaElement` (Alias in CppWinRTIncludes.h, used in MediaHelpers.cpp)
- Since `CoreWindow` is not valid in WinUI 3, we must use a different method to get KeyState (ActionHelpers.cpp)
- The System Xaml `IButtonWithFlyout` interface is part of `IButton` in WinUI 3 (Alias in XamlStraddle, used in ActionHelpers.cpp)

# How Verified

This change includes a SimpleVisualizer project to show the WinUI 3 version of the Renderer working. A WinUI 3 version of the full AdaptiveCardVisualizer will be included in a subsequent change.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8237)


source/uwp/winui3/Renderer/idl/AdaptiveCards.Rendering.Winui3.idl
```diff
@@ -0,0 +1,668 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace AdaptiveCards
+{
+    namespace Rendering
+    {
+        namespace Winui3
```
almedina-ms, line 8: Just to verify, is this namespace name consistent with other usages of the Winui3 namespace, for example, comparing against the one defined in their repository or how the headers capitalize UI. This could be dismissible if using lowercase allows for less namespace conflicts to exist

source/uwp/winui3/SimpleVisualizer/Assets/weather.json
```diff
@@ -0,0 +1,215 @@
+{
+  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
```
almedina-ms, line 2: is there some way you can link this file from the samples folder rather than duplicating it?

almedina-ms (approved):
:shipit:

source/uwp/winui3/Renderer/idl/AdaptiveCards.Rendering.Winui3.idl
krschau: "WinUI" (with whatever capitalization) actually isn't used a lot in that repository, since the DLLs produced are in the Microsoft.UI.Xaml namespace and don't contain "winui" at all. Although when it is used, I suppose "WinUI" is more common than "Winui". If we want to make this change, I think we would want to do it in the ObjectModel as well.

paulcam206 (approved):

source/uwp/winui3/SimpleVisualizer/Assets/weather.json
krschau: Done.

source/uwp/winui3/Renderer/idl/AdaptiveCards.Rendering.Winui3.idl
krschau: https://github.com/microsoft/AdaptiveCards/pull/8255

PR #8250, opened 2023-01-20
Fixed carousel flicker issue

jwoo-msft:
# Related Issue
https://dev.azure.com/microsoft/OS/_workitems/edit/42053802

# Description

Swiper 8 has added requestAnimationFrame() call and has caused the flickering issue.
Reimplemented signaling module for resizing events to let the swiper know of the resizing event without relying on the problematic requestAnimationFrame().

# Sample Card
https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.6/Elements/Carousel.Header.json

# How Verified
1. Went through all of the Carousel cards to check for regressions.
2. Adding a space to the above sample in the designer's editor, the issue can be reproduced. has verified that flickering is no longer observed with the fix.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8250)


paulcam206 (approved):
:shipit:

PR #8254, opened 2023-01-23
[WinUI 3] Add Visualizer for WinUI 3

krschau:
# Related Issue

# Description

This change adds a WinUI 3 version of the Adaptive Card Visualizer, matching functionality of the UWP version. A few changes needed to be made to the shared code to work with both versions of Xaml:
- The `GridSplitter` control moves to code behind. In markup, we can't change which namespace it comes from (System Xaml or WinUI 3 version) based on which version we are compiling.
- In WinUI 3, we don't have the `CoreWindow.Dispatcher`, so instead use Community Toolkit's `Microsoft.UI.Dispatching.DispatcherQueueController.DispatcherQueue.EnqueueAsync()`
- In WinUI 3, `ContentDialog`s need their `XamlRoot` to be set. To have this info, we have to cache the XamlRoot when we create the MainPageViewModel.
- In WinUI 3, `MessageDialog`s need to be initialized with an HWND before they can be opened.

# How Verified

Checked each example card, plus creating and saving a new card.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8254)

### DIFF
--- a/source/uwp/SharedVisualizer/CustomElements/CustomInput.cs
+++ b/source/uwp/SharedVisualizer/CustomElements/CustomInput.cs
@@ -6,8 +6,8 @@ using System.Linq;
 using Windows.Data.Json;
 
 #if USE_WINUI3
-using AdaptiveCards.Rendering.Winui3;
-using AdaptiveCards.ObjectModel.Winui3;
+using AdaptiveCards.Rendering.WinUI3;
+using AdaptiveCards.ObjectModel.WinUI3;
 using Microsoft.UI;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
--- a/source/uwp/SharedVisualizer/CustomElements/CustomInput.cs
+++ b/source/uwp/SharedVisualizer/CustomElements/CustomInput.cs
@@ -5,7 +5,14 @@ using System.Collections.Generic;
 using System.Linq;
 using Windows.Data.Json;
 
-#if !USE_WINUI3
+#if USE_WINUI3
+using AdaptiveCards.Rendering.Winui3;
+using AdaptiveCards.ObjectModel.Winui3;
+using Microsoft.UI;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Media;
+#else
 using AdaptiveCards.Rendering.Uwp;
 using AdaptiveCards.ObjectModel.Uwp;
 using Windows.UI;
--- a/source/uwp/SharedVisualizer/ViewModel/DocumentViewModel.cs
+++ b/source/uwp/SharedVisualizer/ViewModel/DocumentViewModel.cs
@@ -10,7 +10,13 @@ using Windows.Data.Json;
 using Windows.Storage;
 using XamlCardVisualizer.CustomElements;
 
-#if !USE_WINUI3
+#if USE_WINUI3
+using AdaptiveCards.ObjectModel.Winui3;
+using AdaptiveCards.Rendering.Winui3;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Media;
+#else
 using AdaptiveCards.ObjectModel.Uwp;
 using AdaptiveCards.Rendering.Uwp;
 using Windows.UI.Xaml;
--- a/source/uwp/SharedVisualizer/ViewModel/DocumentViewModel.cs
+++ b/source/uwp/SharedVisualizer/ViewModel/DocumentViewModel.cs
@@ -11,8 +11,8 @@ using Windows.Storage;
 using XamlCardVisualizer.CustomElements;
 
 #if USE_WINUI3
-using AdaptiveCards.ObjectModel.Winui3;
-using AdaptiveCards.Rendering.Winui3;
+using AdaptiveCards.ObjectModel.WinUI3;
+using AdaptiveCards.Rendering.WinUI3;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Media;
--- a/source/uwp/SharedVisualizer/ViewModel/DocumentViewModel.cs
+++ b/source/uwp/SharedVisualizer/ViewModel/DocumentViewModel.cs
@@ -102,6 +102,9 @@ namespace AdaptiveCardVisualizer.ViewModel
                         _renderedAdaptiveCard.Action += async (sender, e) =>
                         {
                             var m_actionDialog = new ContentDialog();
+#if USE_WINUI3
+                            m_actionDialog.XamlRoot = this.MainPageViewModel._root;
+#endif
 
                             if (e.Action.ActionType == ActionType.ShowCard)
                             {
@@ -127,6 +130,9 @@ namespace AdaptiveCardVisualizer.ViewModel
                             _renderedAdaptiveCard.MediaClicked += async (sender, e) =>
                             {
                                 var onPlayDialog = new ContentDialog();
+#if USE_WINUI3
+                                onPlayDialog.XamlRoot = this.MainPageViewModel._root;
+#endif
                                 onPlayDialog.Content = "MediaClickedEvent:";
 
                                 foreach (var source in e.Media.Sources)
--- a/source/uwp/SharedVisualizer/ViewModel/DocumentViewModel.cs
+++ b/source/uwp/SharedVisualizer/ViewModel/DocumentViewModel.cs
@@ -103,7 +103,7 @@ namespace AdaptiveCardVisualizer.ViewModel
                         {
                             var m_actionDialog = new ContentDialog();
 #if USE_WINUI3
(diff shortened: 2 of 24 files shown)


source/uwp/winui3/Visualizer/Package.appxmanifest
```diff
@@ -0,0 +1,48 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+
+<Package
+  xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
+  xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
+  xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
+  IgnorableNamespaces="uap rescap">
+
+  <Identity
+    Name="9f958379-af7e-435b-a19e-3b823ef087f0"
```
beervoley, line 10:
```suggestion
    Name="AdaptiveCardVisualizer"
```

source/uwp/winui3/Visualizer/App.xaml.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <summary>
+        /// Invoked when the application is launched.
+        /// </summary>
+        /// <param name="args">Details about the launch request and process.</param>
+        protected override void OnLaunched(Microsoft.UI.Xaml.LaunchActivatedEventArgs args)
+        {
+            m_window = new MainWindow();
+            m_window.Activate();
+#if USE_WINUI3
+            Window = m_window;
```
beervoley, line 50: what is this for?:)

source/uwp/SharedVisualizer/ViewModel/DocumentViewModel.cs
```diff
@@ -96,6 +102,9 @@ protected override async void LoadPayload(string payload)
                         _renderedAdaptiveCard.Action += async (sender, e) =>
                         {
                             var m_actionDialog = new ContentDialog();
+#if USE_WINUI3
+                            m_actionDialog.XamlRoot = this.MainPageViewModel._root;
```
beervoley, line 106: Root is ambiguous and can mean the root UI element for example:)
```suggestion
                            m_actionDialog.XamlRoot = this.MainPageViewModel._xamlRoot;
```

Isn't there a way to get the root during the runtime?

source/uwp/SharedVisualizer/ViewModel/KeyboardPressTimeCounter.cs
```diff
@@ -56,7 +61,12 @@ public void CheckIfSendIsNecessary(Object stateInfo)
 
         private async Task SendShowErrorsUpdateAsync()
         {
+#if USE_WINUI3
+            var queueController = Microsoft.UI.Dispatching.DispatcherQueueController.CreateOnDedicatedThread();
+            await queueController.DispatcherQueue.EnqueueAsync(() =>
```
beervoley, line 66: Is there any specific reason you're not using
```
 var dispatcher = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
 dispatcher.TryEnqueue(..):
```

source/uwp/SharedVisualizer/CustomElements/CustomInput.cs
```diff
@@ -5,7 +5,14 @@
...
 
-#if !USE_WINUI3
+#if USE_WINUI3
+using AdaptiveCards.Rendering.Winui3;
+using AdaptiveCards.ObjectModel.Winui3;
+using Microsoft.UI;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Media;
+#else
```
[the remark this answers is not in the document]
krschau: Maybe. We can't do the thing where we alias "Microsoft.UI.Xaml" and "Windows.UI.Xaml" to just "xaml" ([source](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-directive#using-alias)). We could use `global` usings and have every namespace mentioned available to every file. Usually that's not great programming practice but in this instance it might be few enough namespaces and a small enough project. @asklar, do you have any thoughts?

source/uwp/SharedVisualizer/ViewModel/KeyboardPressTimeCounter.cs
krschau: That thread doesn't have a dispatcher, so `Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread()` returns null. 
I probably should do `DispatcherQueueController.CreateOnCurrentThread()`, but still `TryEnqueue()` returns a bool, not something awaitable. So I used DispatcherQueueExtensions to match the existing functionality. Up to you if you'd rather not use the extension, but also not await. I'm pushing a change I think is better than what I have now.

beervoley: I don't mind either way :)
If there's truly a need to enqueue it `async`, then usage of extension is correct:)

source/uwp/winui3/Visualizer/App.xaml.cs
krschau: We need to initialize the FilePicker/FolderPicker/MessageDialog with an HWND before we can open them. This seemed like the most straightforward way to get the HWND, I wasn't sure how else to from the ViewModels. (Suggested by AndrewLeader [here](https://github.com/microsoft/WindowsAppSDK/issues/1188#issuecomment-1030457077).)

source/uwp/SharedVisualizer/ViewModel/DocumentViewModel.cs
krschau: I couldn't figure out any other way to get the XamlRoot from the ViewModel. Does the ViewModel have something that knows about the xaml tree that I missed?

beervoley:
Could you also remove `SimpleVisualizer` if it's no longer needed?

source/uwp/winui3/Visualizer/Package.appxmanifest
krschau: Done

source/uwp/SharedVisualizer/ViewModel/DocumentViewModel.cs
beervoley: Nope, you're probably right..

source/uwp/SharedVisualizer/CustomElements/CustomInput.cs
```diff
@@ -5,7 +5,14 @@
 using System.Linq;
 using Windows.Data.Json;
 
-#if !USE_WINUI3
+#if USE_WINUI3
+using AdaptiveCards.Rendering.Winui3;
```
asklar, line 9: nit, can we please fix the capitalization :) so WinUI3 instead of Winui3

source/uwp/SharedVisualizer/ViewModel/KeyboardPressTimeCounter.cs
```diff
@@ -56,7 +61,18 @@ public void CheckIfSendIsNecessary(Object stateInfo)
 
         private async Task SendShowErrorsUpdateAsync()
         {
+#if USE_WINUI3
+            var dispatcher = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
+            if (dispatcher == null)
+            { 
+                var queueController = Microsoft.UI.Dispatching.DispatcherQueueController.CreateOnCurrentThread();
```
asklar, line 68: could we just take in the dispatcher queue as a param to this object or method?

source/uwp/SharedVisualizer/CustomElements/CustomInput.cs
krschau: https://github.com/microsoft/AdaptiveCards/pull/8255

source/uwp/SharedVisualizer/ViewModel/MainPageViewModel.cs
```diff
@@ -36,6 +39,10 @@ public DocumentViewModel CurrentDocument
 
         public HostConfigEditorViewModel HostConfigEditor { get; private set; }
 
+#if USE_WINUI3
+        public XamlRoot _xamlRoot;
```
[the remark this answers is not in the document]
krschau: Public to be set from the View class. Changed to a property.

source/uwp/SharedVisualizer/ViewModel/KeyboardPressTimeCounter.cs
krschau: Getting it to this ViewModel object means passing it through about four other constructors and a bunch of methods, and especially messy with the different types between uwp+winui3. It's a large enough change that I'd rather not, unless doing it this way is particularly bad.

asklar (approved):

krschau:
> Could you also remove `SimpleVisualizer` if it's no longer needed?

@beervoley Right now it's particularly useful to have a sandbox to easily play with json OR programmatically created cards. But I agree in the long term it can either go away, or have that specific feature enhanced.

beervoley (approved):

PR #8256, opened 2023-01-24
[UWP Renderer] Parse for SVG width and height

### LINKED ISSUE #8243: [Rendering] SVG Image with explicit dimensions doesn't support Size and explicit size
### Target Platforms

UWP

### SDK Version

3.1.0

### Application Name

N/A

### Problem Description

SVG image that has dimension doesn't support size: small, medium, large, and explicit size.

### Screenshots

image
image

### Card JSON

```JSON
{
  "type": "AdaptiveCard",
  "version": "0.5",
  "body": [
    {
      "type": "Image",
      "url": "https://adaptivecards.io/content/adaptive-card.svg",
      "size": "small"

    }
  ]
}
```

### Sample Code Language

_No response_

### Sample Code

_No response_

anna-dingler:
# Related Issue

Fixes #8243

# Description

`RasterizedWidth` and/or `RasterizedHeight` need to be set for all SVG images if possible. This PR addresses this in two ways:

1. Before assigning the image source, we parse the provided SVG for the height and width. If the height/width is set to a real number, we updated the `RasterizedWidth` and/or `RasterizedHeight`. Once this is complete, we continue setting the image source.
2. If an explicit height/width or height is provided in the card payload, we update the `RasterizedWidth` and/or `RasterizedHeight` with those values. 

**Note:** Values set in **1** take priority over those set in **2**.

This PR also includes refactoring for duplicate code in AdaptiveImageRenderer.

# Sample Card

## URL with size property
```JSON
{
  "type": "AdaptiveCard",
  "version": "0.5",
  "body": [
    {
      "type": "Image",
      "url": "https://adaptivecards.io/content/adaptive-card.svg",
      "size": "small"
    }
  ]
}
```
image

## URL with explicit height & width
```JSON
{
  "type": "AdaptiveCard",
  "version": "0.5",
  "body": [
    {
      "type": "Image",
      "url": "https://adaptivecards.io/content/adaptive-card.svg",
      "height": "70px",
      "width": "40px"
    }
  ]
}
```
image

## Resource resolver: fixed width & person style
```JSON
{
  "type": "AdaptiveCard",
  "version": "0.5",
  "body": [
    {
      "type": "Image",
      "url": "symbol:adaptive-card-fixed-height.svg",
      "backgroundColor": "#000000",
      "horizontalAlignment": "right",
      "width": "50px",
      "style": "person"
    }
  ]
}
```
image

## Resource resolver: size property
```JSON
{
  "type": "AdaptiveCard",
  "version": "0.5",
  "body": [
    {
      "type": "Image",
      "url": "symbol:adaptive-card-fixed-height.svg",
      "backgroundColor": "#000000",
      "horizontalAlignment": "right",
      "size": "medium"
    }
  ]
}
```
image

## Data URI: size property
```JSON
{
  "type": "AdaptiveCard",
  "version": "0.5",
  "body": [
    {
      "type": "Image",
      "url": "data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSI3MDAiICBoZWlnaHQ9IjI5NC40IiB2aWV3Qm94PSIwIDAgNzAwLjAwMDAxIDI5NC40MjI1OCI+IAo8c3R5bGU+CiAgCiAgLnRlc3QgewogICAgYW5pbWF0aW9uOiBmYWRlIDNzIGluZmluaXRlIGFsdGVybmF0ZTsgCiAgICAKICB9CiAgCiAgQGtleWZyYW1lcyBmYWRlIHsKICAgIAogICAgZnJvbSB7CiAgICAgIG9wYWNpdHk6IDAKICAgIH0KICAgIAogICAgdG8gewogICAgICAKICAgICAgb3BhY2l0eToxCiAgICB9CiAgICAKICB9CiAgCiAgPC9zdHlsZT4KICAKICA8cGF0aCBjbGFzcz0idGVzdCIgZD0iTTYwMi4yIDEyOC42Yy0xMS43IDIuNy00Mi42IDQuMi00Mi42IDQuMmwtMy44IDEyLjJzMTUuMy0xLjMgMjYuNS0uMmMwIDAgMy42LS4zIDQgNCAuMiA0LjQtLjMgOS0uMyA5cy0uMyAyLjctNCAzLjRjLTQuMy43LTMzLjMgMi0zMy4zIDJMNTQ0IDE3OXMtMS43IDMuNiAyLjIgMi42YzMuNi0xIDM0LTYuOCAzOC02IDQuMiAxIDkgNi44IDcuNiAxMi0xLjYgNi4zLTMyIDI1LjctNTAuNCAyNC40IDAgMC05LjcuNi0xNy44LTEyLjUtNy44LTEyLjUgMi43LTM2IDIuNy0zNnMtNC43LTExLTEuMi0xNWMwIDAgMi0xLjcgOC0yLjJsNy41LTE1LjRzLTguNS41LTEzLjUtNS43Yy00LjYtNi01LTguNi0xLjQtMTAuMiAzLjgtMiAzOS04LjMgNjMuMi03LjUgMCAwIDguNS0xIDE2IDEzLjcgMCAwIDMuNCA2LTIuNiA3LjRNNTExIDE4Ny44Yy0zIDcuMy0xMS4zIDE1LTIxLjMgMTAuMy0xMC4yLTQuOC0yNi4zLTM3LjYtMjYuMy0zNy42cy02LTEyLjItNy4yLTExLjhjMCAwLTEuMy0yLjQtMiAxMS0xIDEzLjMuMiAzOS4yLTUuMyA0My4zLTUgNC0xMSAyLjMtMTQuNC0yLjQtMi44LTQuNy00LTE2LTIuNC0zNS43IDEuOC0xOS43IDYuMy00MC43IDEyLTQ3LjIgNi02LjYgMTAuOC0xLjggMTIuNiAwIDAgMCA3LjcgNyAyMC43IDI3LjdsMi4yIDMuOHMxMS44IDE5LjcgMTMgMTkuNmMwIDAgMSAxIDEuOC4yIDEuMi0uMy44LTYuNy44LTYuN1M0OTMgMTQxIDQ4MiAxMDVjMCAwLTEuNi00LjYtLjUtOSAxLTQgNS4zLTIgNS4zLTJzMTYuNiA4IDI0LjcgMzVjOCAyNyAyLjYgNTEuNS0uNCA1OC44TTQyOS42IDExOC41Yy0xLjYgMi44LTIuMyA2LjctOS4yIDcuOCAwIDAtNjcgNC43LTcwLjMgOS40IDAgMC0yLjIgMi44IDEuNCAzLjUgMy44LjggMTkgMi44IDI2LjIgMy4yIDcuOCAwIDM0IC4yIDQzLjYgMTIgMCAwIDUuNSA1LjYgNS4zIDE4LjMtLjIgMTMtMi41IDE3LjYtNy42IDIyLjMtNS4zIDQuNC01MC43IDI0LjgtODAtNi40IDAgMC0xMy40LTE1IDQuNy0yNi40IDAgMCAxMy04IDQ2LjMgMS4zIDAgMCAxMCAzLjYgOS42IDcuMy0uNiA0LTguMyA4LTE5LjUgNy44LTEwLjgtLjMtMTguOC01LjUtMTcuMi00LjYgMS41LjUtMTEuNy02LjQtMTUuOC0xLjctNCA0LjQtMyA3IDEgOS43IDEwIDUuOCA0OS4zIDMuNyA2MS05LjQgMCAwIDQuNy01LjMtMi40LTkuNi03LTQtMjcuNC02LjUtMzUuMy02LjgtNy41LS40LTM1LjYgMC0zOS44LTcuMyAwIDAtNC01LjIuNC0xOS40IDQuNi0xNSAzNy4zLTIwLjggNTEuNS0yMiAwIDAgMzktMS42IDQ2LjMgNi40IDAgMCAxIDEuOC0uMiA0LjVNMzE5IDIwNi40Yy00LjcgMy41LTE0LjcgMi0xNy42LTItMi44LTMuNS0zLjgtMTcuMy0zLjMtMzkgLjctMjIuMiAxLTQ5LjQgNi01My44IDUtNC4zIDgtLjUgMTAgMi40IDIgMyA0LjYgNi4zIDUgMTMuMi42IDcgMi4zIDQzIDIuMyA0M3MyLjIgMzIuOC0yLjMgMzYuMk0zMjkgODkuNGMtMTMuOCA0LjctMjMuMiAzLjItMzEuMi0uMy0zLjUgNi4zLTUuNiA4LjItOC4yIDguNi00IC40LTcuNS02LTgtOC0uOC0xLjUtMi43LTQuMi0uNC0xMC4zLTcuOC03LTguNC0xNi40LTctMjIuNyAxLjgtNy40IDE1LTM1LjIgNTUtMzguNSAwIDAgMTkuNi0xLjQgMjMgOWguNnMxOSAwIDE4LjYgMTdjMCAxNy0yMSAzOC4yLTQyLjQgNDUuNW0xNy44LTQ4LjdjLTEyLjYgMi0zMiAxOC44LTQxLjMgMzIuNyAxNC4zIDIuNiAzOS4zIDEuNiA1MC41LTIxIDAgMCA1LjMtMTQuMi05LjItMTEuN20tNTUuMyAxMWMtNCA2LjUtNC4yIDEwLjQtMi4zIDEzIDQuNy03IDEzLTE4IDI1LjUtMjYuNi05LjYgMS0xNy43IDUtMjMuMiAxMy42TTYzMi4yIDIwNS43Yy05LjIgMjIuNi0xNyA0NS41LTIxLjUgNzkuOCAwIDAtMSA2LjctNi41IDQuNS01LjUtMi0xNC41LTExLTE2LjUtMjMuNy0yLTE2LjYgNS40LTQ0LjYgMjAuNS03Ni44LTQuNC03LTcuNS0xNy40LTUtMzIgMCAwIDQtMjcgMzEtNTEuNCAwIDAgMy4yLTIuNyA1LTEuOCAyLjIgMSAxLjMgOS42LS41IDE0LTEuNiA0LjItMTMuNiAyNS0xMy42IDI1cy03LjUgMTQuMi01LjQgMjUuM2MxNC4yLTIxLjggNDYuNS02NiA2Ni41LTUyIDEyLjcgOSAxMi43IDM4IDMuMiA1NC44LTcuNSAxMy4zLTI4LjcgNDAuOC01NyAzNC40bTQxLjYtNjhjLTcuNCA4LTIwLjYgMjMuMi0zMSA0My44IDExLTEuMiAyMS43LTcuMyAyNS0xMC40IDUuMy00LjcgMTcuNS0xNy40IDE1LjYtMzQuMiAwIDAtMS4yLTguOC05LjYuOE0yMjYgMjE3LjVjLTM1LjQgMTAuOC02OC44IDUuOC04NyAxLS41IDcuNC0xLjMgMTAuNS0yLjUgMTEuNy0xLjQgMS42LTEzIDguMi0xOS40LTEuMi0yLjgtNC41LTQuMi0xMi42LTUtMjAtNDEtMTguNi02MC00Ni02MC42LTQ3LTEtMS0xMC4zLTEwLjctMS0yMi43IDguNy0xMC44IDM3LjUtMjEuNyA2My4zLTI2IDEtMjIgMy40LTM5IDYuNS00Ni41IDMuNy05IDguNC0xIDEyLjYgNSAzLjQgNC41IDUuNSAyMy44IDUuNyAzOS4yIDE2LjgtLjggMjcgLjQgNDUuNyA0IDI0LjYgNC4yIDQxIDE2LjggMzkuNyAzMS0xLjIgMTQtMTQgMTkuOC0xOSAyMC4yLTUgLjQtMTMtMy4zLTEzLTMuMy01LjYtMi42LS41LTUgNi03LjggNy4yLTMuNSA1LjYtNyA1LjYtNy0yLjYtOC0zNC41LTEzLjMtNjYuMi0xMy4zIDAgMTcuNS43IDQ2LjUgMS4yIDYzLjQgMjIuMiA0LjIgMzguOCAzLjMgMzguOCAzLjNzODEtMi4zIDgzLjMtNTRjMi41LTUxLjgtODEtMTAxLjQtMTQyLjUtMTE3QzU2LjggMTQuNCAyMiAyNS44IDE5IDI3LjNjLTMuMyAxLjYtLjMgMi4yLS4zIDIuMlMyMiAzMCAyOCAzMmM2IDIgMS4yIDUgMS4yIDVDMTguNyA0MC42IDcgMzguNSA0LjcgMzMuN2MtMi4zLTQuNyAxLjUtOSA2LTE1LjMgNC4yLTYuNSA5LTYuMyA5LTYuMyA3Ni0yNi41IDE2OC44IDIxIDE2OC44IDIxIDg2LjggNDMuOCAxMDEuNiA5NS4zIDEwMCAxMTUuMy0xLjQgMTkuNy05IDUzLTYyLjUgNjkuMk01OSAxNDZjLTguNiA0LTIuNiAxMC40LTIuNiAxMC40IDE2LjIgMTcuMyAzNiAyOC4yIDU1IDM1IDIuMi0zMCAyLTQwLjcgMi01NS44Qzg0IDEzNy42IDY3IDE0Mi40IDU5IDE0NiIvPiA8L3N2Zz4=",
      "size": "medium"
    }
  ]
}
```
image

## Data URI: explicit height and width
```JSON
{
  "type": "AdaptiveCard",
  "version": "0.5",
  "body": [
    {
      "type": "Image",
      "url": "data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSI3MDAiICBoZWlnaHQ9IjI5NC40IiB2aWV3Qm94PSIwIDAgNzAwLjAwMDAxIDI5NC40MjI1OCI+IAo8c3R5bGU+CiAgCiAgLnRlc3QgewogICAgYW5pbWF0aW9uOiBmYWRlIDNzIGluZmluaXRlIGFsdGVybmF0ZTsgCiAgICAKICB9CiAgCiAgQGtleWZyYW1lcyBmYWRlIHsKICAgIAogICAgZnJvbSB7CiAgICAgIG9wYWNpdHk6IDAKICAgIH0KICAgIAogICAgdG8gewogICAgICAKICAgICAgb3BhY2l0eToxCiAgICB9CiAgICAKICB9CiAgCiAgPC9zdHlsZT4KICAKICA8cGF0aCBjbGFzcz0idGVzdCIgZD0iTTYwMi4yIDEyOC42Yy0xMS43IDIuNy00Mi42IDQuMi00Mi42IDQuMmwtMy44IDEyLjJzMTUuMy0xLjMgMjYuNS0uMmMwIDAgMy42LS4zIDQgNCAuMiA0LjQtLjMgOS0uMyA5cy0uMyAyLjctNCAzLjRjLTQuMy43LTMzLjMgMi0zMy4zIDJMNTQ0IDE3OXMtMS43IDMuNiAyLjIgMi42YzMuNi0xIDM0LTYuOCAzOC02IDQuMiAxIDkgNi44IDcuNiAxMi0xLjYgNi4zLTMyIDI1LjctNTAuNCAyNC40IDAgMC05LjcuNi0xNy44LTEyLjUtNy44LTEyLjUgMi43LTM2IDIuNy0zNnMtNC43LTExLTEuMi0xNWMwIDAgMi0xLjcgOC0yLjJsNy41LTE1LjRzLTguNS41LTEzLjUtNS43Yy00LjYtNi01LTguNi0xLjQtMTAuMiAzLjgtMiAzOS04LjMgNjMuMi03LjUgMCAwIDguNS0xIDE2IDEzLjcgMCAwIDMuNCA2LTIuNiA3LjRNNTExIDE4Ny44Yy0zIDcuMy0xMS4zIDE1LTIxLjMgMTAuMy0xMC4yLTQuOC0yNi4zLTM3LjYtMjYuMy0zNy42cy02LTEyLjItNy4yLTExLjhjMCAwLTEuMy0yLjQtMiAxMS0xIDEzLjMuMiAzOS4yLTUuMyA0My4zLTUgNC0xMSAyLjMtMTQuNC0yLjQtMi44LTQuNy00LTE2LTIuNC0zNS43IDEuOC0xOS43IDYuMy00MC43IDEyLTQ3LjIgNi02LjYgMTAuOC0xLjggMTIuNiAwIDAgMCA3LjcgNyAyMC43IDI3LjdsMi4yIDMuOHMxMS44IDE5LjcgMTMgMTkuNmMwIDAgMSAxIDEuOC4yIDEuMi0uMy44LTYuNy44LTYuN1M0OTMgMTQxIDQ4MiAxMDVjMCAwLTEuNi00LjYtLjUtOSAxLTQgNS4zLTIgNS4zLTJzMTYuNiA4IDI0LjcgMzVjOCAyNyAyLjYgNTEuNS0uNCA1OC44TTQyOS42IDExOC41Yy0xLjYgMi44LTIuMyA2LjctOS4yIDcuOCAwIDAtNjcgNC43LTcwLjMgOS40IDAgMC0yLjIgMi44IDEuNCAzLjUgMy44LjggMTkgMi44IDI2LjIgMy4yIDcuOCAwIDM0IC4yIDQzLjYgMTIgMCAwIDUuNSA1LjYgNS4zIDE4LjMtLjIgMTMtMi41IDE3LjYtNy42IDIyLjMtNS4zIDQuNC01MC43IDI0LjgtODAtNi40IDAgMC0xMy40LTE1IDQuNy0yNi40IDAgMCAxMy04IDQ2LjMgMS4zIDAgMCAxMCAzLjYgOS42IDcuMy0uNiA0LTguMyA4LTE5LjUgNy44LTEwLjgtLjMtMTguOC01LjUtMTcuMi00LjYgMS41LjUtMTEuNy02LjQtMTUuOC0xLjctNCA0LjQtMyA3IDEgOS43IDEwIDUuOCA0OS4zIDMuNyA2MS05LjQgMCAwIDQuNy01LjMtMi40LTkuNi03LTQtMjcuNC02LjUtMzUuMy02LjgtNy41LS40LTM1LjYgMC0zOS44LTcuMyAwIDAtNC01LjIuNC0xOS40IDQuNi0xNSAzNy4zLTIwLjggNTEuNS0yMiAwIDAgMzktMS42IDQ2LjMgNi40IDAgMCAxIDEuOC0uMiA0LjVNMzE5IDIwNi40Yy00LjcgMy41LTE0LjcgMi0xNy42LTItMi44LTMuNS0zLjgtMTcuMy0zLjMtMzkgLjctMjIuMiAxLTQ5LjQgNi01My44IDUtNC4zIDgtLjUgMTAgMi40IDIgMyA0LjYgNi4zIDUgMTMuMi42IDcgMi4zIDQzIDIuMyA0M3MyLjIgMzIuOC0yLjMgMzYuMk0zMjkgODkuNGMtMTMuOCA0LjctMjMuMiAzLjItMzEuMi0uMy0zLjUgNi4zLTUuNiA4LjItOC4yIDguNi00IC40LTcuNS02LTgtOC0uOC0xLjUtMi43LTQuMi0uNC0xMC4zLTcuOC03LTguNC0xNi40LTctMjIuNyAxLjgtNy40IDE1LTM1LjIgNTUtMzguNSAwIDAgMTkuNi0xLjQgMjMgOWguNnMxOSAwIDE4LjYgMTdjMCAxNy0yMSAzOC4yLTQyLjQgNDUuNW0xNy44LTQ4LjdjLTEyLjYgMi0zMiAxOC44LTQxLjMgMzIuNyAxNC4zIDIuNiAzOS4zIDEuNiA1MC41LTIxIDAgMCA1LjMtMTQuMi05LjItMTEuN20tNTUuMyAxMWMtNCA2LjUtNC4yIDEwLjQtMi4zIDEzIDQuNy03IDEzLTE4IDI1LjUtMjYuNi05LjYgMS0xNy43IDUtMjMuMiAxMy42TTYzMi4yIDIwNS43Yy05LjIgMjIuNi0xNyA0NS41LTIxLjUgNzkuOCAwIDAtMSA2LjctNi41IDQuNS01LjUtMi0xNC41LTExLTE2LjUtMjMuNy0yLTE2LjYgNS40LTQ0LjYgMjAuNS03Ni44LTQuNC03LTcuNS0xNy40LTUtMzIgMCAwIDQtMjcgMzEtNTEuNCAwIDAgMy4yLTIuNyA1LTEuOCAyLjIgMSAxLjMgOS42LS41IDE0LTEuNiA0LjItMTMuNiAyNS0xMy42IDI1cy03LjUgMTQuMi01LjQgMjUuM2MxNC4yLTIxLjggNDYuNS02NiA2Ni41LTUyIDEyLjcgOSAxMi43IDM4IDMuMiA1NC44LTcuNSAxMy4zLTI4LjcgNDAuOC01NyAzNC40bTQxLjYtNjhjLTcuNCA4LTIwLjYgMjMuMi0zMSA0My44IDExLTEuMiAyMS43LTcuMyAyNS0xMC40IDUuMy00LjcgMTcuNS0xNy40IDE1LjYtMzQuMiAwIDAtMS4yLTguOC05LjYuOE0yMjYgMjE3LjVjLTM1LjQgMTAuOC02OC44IDUuOC04NyAxLS41IDcuNC0xLjMgMTAuNS0yLjUgMTEuNy0xLjQgMS42LTEzIDguMi0xOS40LTEuMi0yLjgtNC41LTQuMi0xMi42LTUtMjAtNDEtMTguNi02MC00Ni02MC42LTQ3LTEtMS0xMC4zLTEwLjctMS0yMi43IDguNy0xMC44IDM3LjUtMjEuNyA2My4zLTI2IDEtMjIgMy40LTM5IDYuNS00Ni41IDMuNy05IDguNC0xIDEyLjYgNSAzLjQgNC41IDUuNSAyMy44IDUuNyAzOS4yIDE2LjgtLjggMjcgLjQgNDUuNyA0IDI0LjYgNC4yIDQxIDE2LjggMzkuNyAzMS0xLjIgMTQtMTQgMTkuOC0xOSAyMC4yLTUgLjQtMTMtMy4zLTEzLTMuMy01LjYtMi42LS41LTUgNi03LjggNy4yLTMuNSA1LjYtNyA1LjYtNy0yLjYtOC0zNC41LTEzLjMtNjYuMi0xMy4zIDAgMTcuNS43IDQ2LjUgMS4yIDYzLjQgMjIuMiA0LjIgMzguOCAzLjMgMzguOCAzLjNzODEtMi4zIDgzLjMtNTRjMi41LTUxLjgtODEtMTAxLjQtMTQyLjUtMTE3QzU2LjggMTQuNCAyMiAyNS44IDE5IDI3LjNjLTMuMyAxLjYtLjMgMi4yLS4zIDIuMlMyMiAzMCAyOCAzMmM2IDIgMS4yIDUgMS4yIDVDMTguNyA0MC42IDcgMzguNSA0LjcgMzMuN2MtMi4zLTQuNyAxLjUtOSA2LTE1LjMgNC4yLTYuNSA5LTYuMyA5LTYuMyA3Ni0yNi41IDE2OC44IDIxIDE2OC44IDIxIDg2LjggNDMuOCAxMDEuNiA5NS4zIDEwMCAxMTUuMy0xLjQgMTkuNy05IDUzLTYyLjUgNjkuMk01OSAxNDZjLTguNiA0LTIuNiAxMC40LTIuNiAxMC40IDE2LjIgMTcuMyAzNiAyOC4yIDU1IDM1IDIuMi0zMCAyLTQwLjcgMi01NS44Qzg0IDEzNy42IDY3IDE0Mi40IDU5IDE0NiIvPiA8L3N2Zz4=",
      "height": "300px",
      "width": "300px"
    }
  ]
}
```
image

# How Verified

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8256)


source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -462,7 +393,24 @@ namespace AdaptiveCards::Rendering::Uwp
 
             if (isImageSvg)
             {
-                image.as<winrt::SvgImageSource>().UriSource(imageUrl);
+                // If we have an SVG, we need to try to parse for the image size before setting the image source
+                auto svgDocumentLoadOperation = winrt::XmlDocument::LoadFromUriAsync(imageUrl);
+
+                // TODO: what do we do if this fails?
+                svgDocumentLoadOperation.Completed(
```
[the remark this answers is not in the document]
beervoley: yup, Completed is called on fail too:)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -533,12 +491,134 @@ namespace AdaptiveCards::Rendering::Uwp
...
+                        if (auto strongThis = weakThis.get())
+                        {
+                            strongThis->SetAutoSize(uiElement, parentElement, imageContainer, isVisible, false /* imageFiresOpenEvent */);
+                        }
+                    }
+                });
+        }
+    }
+
+    winrt::IAsyncAction render_xaml::XamlBuilder::SetSvgUriSource(winrt::SvgImageSource image, winrt::Uri imageUrl)
```
beervoley, line 574:
```suggestion
    winrt::IAsyncAction render_xaml::XamlBuilder::SetSvgUriSource(winrt::SvgImageSource const& image, winrt::Uri const& imageUrl)
```
these 2 **must** be alive here so that the coroutine succeeds :)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -533,12 +491,134 @@ namespace AdaptiveCards::Rendering::Uwp
...
+    {
+        try
+        {
+            co_await winrt::resume_foreground(image.Dispatcher());
+            image.as<winrt::SvgImageSource>().UriSource(imageUrl);
+        }
+        catch (winrt::hresult_wrong_thread ex)
+        {
+            // We didn't switch to the UI thread
```
beervoley, line 583: I don't think we need to fail fast here necessarily but logging the error would be nice :)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -533,12 +491,134 @@ namespace AdaptiveCards::Rendering::Uwp
...
+                                                                         bool isVisible)
+    {
+        try
+        {
+            co_await winrt::resume_foreground(image.Dispatcher());
+            auto setSourceOperation = image.SetSourceAsync(stream);
+
+            setSourceOperation.Completed(
+                [weakThis = this->get_weak(), uiElement, isAutoSize, parentElement, imageContainer, isVisible](
+                    winrt::IAsyncOperation<winrt::SvgImageSourceLoadStatus> const& operation, winrt::AsyncStatus status)
```
beervoley, line 603:
```suggestion
                   auto operation, auto status)
```
does `auto` type deduction work here?:)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -666,4 +746,80 @@ namespace AdaptiveCards::Rendering::Uwp
...
+
+    void render_xaml::XamlBuilder::ParseXmlForHeightAndWidth(winrt::XmlDocument const& xmlDoc, winrt::ImageSource const& imageSource) {
+        if (xmlDoc)
+        {
+            auto rootElement = xmlDoc.DocumentElement();
+            auto height = rootElement.GetAttribute(L"height");
+            auto width = rootElement.GetAttribute(L"width");
+
+            // We only need to set height or width, not both (fixes aspect ratio for person style)
+            bool isHeightSet = false;
```
beervoley, line 758: Is it necessary only for personal style? What if png set height and width are not the same and it's intended?:)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -407,50 +375,13 @@ namespace AdaptiveCards::Rendering::Uwp
                         if (const auto stream = dataWriter.DetachStream().try_as<winrt::InMemoryRandomAccessStream>())
                         {
                             stream.Seek(0);
-                            strongThis->SetImageSource(uiElement, image);
```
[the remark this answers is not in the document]
beervoley: I think it should include `stretch` and we need to pass it. Even if our image is acquired from `base64` encoded string, explicit dimensions can be set on it and then we need to preserve aspect ratio. This is probably a regression:)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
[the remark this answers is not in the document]
anna-dingler: Sounds good - I'll remove this comment then :)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: Yes, this seems to only be necessary for personal style.

One of the images I tested with has different height and width, and this works there as well. My guess is that as long as we set one of them, the ratio remains the same.

The only other solution I thought of was passing style to this method as well, but we'd need to pass it through a lot of different methods to get it here.

What do you think @beervoley

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: Yup this works - I'll add this throughout the file

beervoley: don't forget to use reference where needed:)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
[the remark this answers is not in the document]
anna-dingler: Updated to include stretch in all of the `SetImageSource` calls

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: I switched all of the non-primitives to `weak_ref`. Let me know what you think :)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: Added an error statement and passed the `renderContext` to these methods

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
beervoley: I don't think we should "fix" the aspect ratio for person style this way:)
what happens if the picture doesn't have the same width&height and we leave it at that? does it render badly?
I don't think it's our responsibility to "fix" aspect ratio in this case:)
how does personal style is being handled with a usual image?

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
beervoley: I didn't mean to switch all of the non-primitives to `weak_ref` :)
only those that you're unsure will be alive by the time they're used in coroutines:)
good example of it is `XamlBuilder` which is passed by `weak_ref` around: it might be destroyed by the time it's called from a coroutine :)
I'd look at the objects, thought of scenarios where they might not survive, and if that's the case and it doesn't break the intended use of the object (may be it really have to be alive for everything to fall into place) - then use `weak_ref`.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -533,12 +489,162 @@ namespace AdaptiveCards::Rendering::Uwp
...
+
+            if (imageSource && uri)
+            {
+                co_await winrt::resume_foreground(imageSource.Dispatcher());
+                imageSource.UriSource(uri);
+            }
+        }
+        catch (winrt::hresult_wrong_thread ex)
+        {
+            renderContext.AddWarning(winrt::WarningStatusCode::AssetLoadFailed,
```
beervoley, line 595: I don't think this is kind of an error that we need to log in `renderContext` :)
executing code on the wrong thread is **our** fault in the code, not that the user or system has failed:)
I don't think we really need to try/catch exception here :) 
I don't see it failing if it's done correctly:)

anna-dingler: Okay, I'll remove the try/catch block

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: For non-svg, we don't have to set the height and width of the imageSource.

Without this change, this is what the following payload renders.

```JSON
{
  "type": "AdaptiveCard",
  "version": "0.5",
  "body": [
    {
      "type": "Image",
      "url": "symbol:adaptive-card-fixed-height.svg",
      "backgroundColor": "#000000",
      "style": "person"
    }
  ]
}
```
image

I don't love this solution, but I'm not sure what the solution would be from a card author perspective.

I tested with `height > width` and `width > height`, and I didn't see a difference between this and the previous solution.

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -92,6 +92,12 @@ std::string HStringToUTF8(winrt::hstring const& in)
     return WStringToString(in);
 }
 
+double HStringToDouble(winrt::hstring const& in)
+{
+    std::string inAsUTF8 = HStringToUTF8(in);
+    return std::atof(inAsUTF8.c_str());
```
dmachaj, line 98: >atof [](http://example.com/codeflow?start=16&length=4)

Can use std::stof to avoid the need to convert wide to utf8?

[https://en.cppreference.com/w/cpp/string/basic_string/stof](https://en.cppreference.com/w/cpp/string/basic_string/stof)

source/uwp/Renderer/lib/XamlBuilder.h
```diff
@@ -65,23 +65,67 @@ namespace AdaptiveCards::Rendering::Uwp
...
+        winrt::ImageSource PopulateImageFromUrlAsync(winrt::Uri const& imageUrl,
+                                                     T const& uiElement,
+                                                     bool isAutoSize,
+                                                     winrt::IInspectable const& parentElement,
+                                                     winrt::IInspectable const& imageContainer,
+                                                     bool isVisible,
+                                                     bool isImageSvg,
+                                                     winrt::Stretch stretch);
+
+        template<typename T, typename S>
```
dmachaj, line 90: ><typename T, typename S> [](http://example.com/codeflow?start=16&length=24)

When there are multiple templated things I think it is easier to understand them with more than one-character names.  In this case perhaps it would be `<typename TElement, typename TStream>`?

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -170,6 +171,7 @@ namespace AdaptiveCards::Rendering::Uwp
         {
             if (pixelWidth)
             {
+                SetRasterizedPixelWidth(imageSource, pixelWidth);
```
dmachaj, line 174: >SetRasterizedPixelWidth [](http://example.com/codeflow?start=16&length=23)

Is it valid for the SVG to provide width but not height (and vice versa)?

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -462,7 +393,27 @@ namespace AdaptiveCards::Rendering::Uwp
...
+                        auto strongImageSource = weakImageSource.get();
+
+                        if (strongThis && strongImageSource)
+                        {
+                            if (status == winrt::AsyncStatus::Completed)
+                            {
+                                strongThis->ParseXmlForHeightAndWidth(operation.GetResults(), strongImageSource);
+                            }
+                            // Now that we've parsed the height, we can set the image source
+                            strongThis->SetSvgUriSource(strongImageSource, imageUrl);
```
dmachaj, line 414: >SetSvgUriSource [](http://example.com/codeflow?start=40&length=15)

Shouldn't this only set the source on success?

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -499,7 +452,14 @@ namespace AdaptiveCards::Rendering::Uwp
...
     template<typename T>
-    void render_xaml::XamlBuilder::PopulateImageFromUrlAsync(winrt::Uri const& imageUrl, T const& imageControl, bool const& isImageSvg)
+    winrt::ImageSource render_xaml::XamlBuilder::PopulateImageFromUrlAsync(winrt::Uri const& imageUrl,
+                                                                           T const& uiElement,
+                                                                           bool isAutoSize,
+                                                                           winrt::IInspectable const& parentElement,
+                                                                           winrt::IInspectable const& imageContainer,
+                                                                           bool isVisible,
+                                                                           bool isImageSvg,
+                                                                           winrt::Stretch stretch)
```
dmachaj, line 462: If this set of fields is passed around enough it might make sense to put them in a `struct` and pass that around instead.  If you are concerned about excess copies the struct could have the copy constructor/operator deleted.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -533,12 +492,133 @@ namespace AdaptiveCards::Rendering::Uwp
                         m_copyStreamOperations.push_back(copyStreamOperation);
 
                         copyStreamOperation.Completed(
```
dmachaj, line 494: >copyStreamOperation [](http://example.com/codeflow?start=24&length=19)

[preexisting] The completed callback pattern is hard to follow compared to coroutines.  Are there any plans to begin using `co_await` instead?

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -533,12 +492,133 @@ namespace AdaptiveCards::Rendering::Uwp
...
-                            [randomAccessStream](winrt::IAsyncOperationWithProgress<uint64_t, uint64_t> const& /*operation*/,
-                                                 winrt::AsyncStatus /*status*/) { randomAccessStream.Seek(0); });
+                            [randomAccessStream, weakThis, uiElement, parentElement, imageContainer, isAutoSize, isVisible, isImageSvg, image, stretch](
+                                auto const& /*operation*/, auto status)
```
dmachaj, line 496: >) [](http://example.com/codeflow?start=70&length=1)

What is the exception/error story for these callbacks?  There is no way to propagate errors to the containing code so they are probably dropped on the floor.  I'm not sure about this specific completion handler but there is a decent chance it will treat any exceptions that leak as a failfast.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -533,12 +492,133 @@ namespace AdaptiveCards::Rendering::Uwp
...
+            auto streamDataReader = winrt::DataReader(inputStream);
+            auto loadDataReaderOperation = streamDataReader.LoadAsync((uint32_t)streamSize);
+
+            loadDataReaderOperation.Completed(
+                [weakThis = this->get_weak(), streamDataReader, uiElementRef = winrt::make_weak(uiElement), isAutoSize, parentElementRef = winrt::make_weak(parentElement),
+                 imageContainerRef = winrt::make_weak(imageContainer), isVisible, streamRef = winrt::make_weak(stream),
+                 imageSourceRef = winrt::make_weak(imageSource.as<winrt::SvgImageSource>())](
+                    auto const& result, auto status) -> void
+                {
+                    auto strongThis = weakThis.get();
```
dmachaj, line 543: >strongThis [](http://example.com/codeflow?start=25&length=10)

That is a lot of get-strong'ing.  This is the gym lambda.

source/uwp/Renderer/lib/Util.cpp
anna-dingler: good call out - thanks!

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -533,12 +492,133 @@ namespace AdaptiveCards::Rendering::Uwp
...
+    template<typename T, typename S>
+    winrt::IAsyncAction render_xaml::XamlBuilder::SetSvgImageSourceAsync(winrt::SvgImageSource const imageSource,
+                                                                         S const stream,
+                                                                         T const uiElement,
+                                                                         bool isAutoSize,
+                                                                         winrt::IInspectable const parentElement,
+                                                                         winrt::IInspectable const imageContainer,
+                                                                         bool isVisible)
+    {
+        co_await winrt::resume_foreground(imageSource.Dispatcher());
```
dmachaj, line 603: >co_await [](http://example.com/codeflow?start=8&length=8)

Once you co_await there is nothing keeping `this` alive.  You should either capture a weakRef before the await to pass into the lambda, or call `get_strong` and save it in a variable so this is not destroyed while the await is happening.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -666,4 +744,93 @@ namespace AdaptiveCards::Rendering::Uwp
...
+            auto rootElement = xmlDoc.DocumentElement();
+            auto height = rootElement.GetAttribute(L"height");
+            auto width = rootElement.GetAttribute(L"width");
+
+            // We only need to set height or width, not both (fixes aspect ratio for person style)
+            bool isHeightSet = false;
+
+            if (!height.empty())
+            {
+                if (auto heightAsDouble = HStringToDouble(height))
```
dmachaj, line 762: >heightAsDouble [](http://example.com/codeflow?start=25&length=14)

Seems strange to put a `double` in an if block.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -666,4 +744,93 @@ namespace AdaptiveCards::Rendering::Uwp
...
+            {
+                if (auto widthAsDouble = HStringToDouble(width))
+                {
+                    SetRasterizedPixelWidth(imageSource, widthAsDouble, isHeightSet);
+                }
+            }
+        }
+    }
+
+    winrt::IAsyncAction render_xaml::XamlBuilder::SetRasterizedPixelHeight(winrt::SvgImageSource const imageSource,
```
dmachaj, line 779: >SetRasterizedPixelHeight [](http://example.com/codeflow?start=50&length=24)

Async methods should generally have `Async` as a suffix on the name.

dmachaj, line 779: >IAsyncAction [](http://example.com/codeflow?start=11&length=12)

Nobody is waiting on this.  Consider using `winrt::fire_and_forget` instead.

dmachaj: Although there are two SetRasterizedPixelHeight implementations in this class so it is really hard to tell who is calling what (not that any of them look at the return value).

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -666,4 +744,93 @@ namespace AdaptiveCards::Rendering::Uwp
...
+                                                                           double imageSize,
+                                                                           bool dropIfUnset)
+    {
+        co_await winrt::resume_foreground(imageSource.Dispatcher());
+        auto currentSize = imageSource.RasterizePixelHeight();
+        bool sizeIsUnset = isinf(currentSize);
+
+        bool dropHeight = sizeIsUnset && dropIfUnset;
```
dmachaj, line 787: >bool [](http://example.com/codeflow?start=8&length=4)

[nit] Many variables in this method and the nearby methods could be marked as `const`.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
dmachaj: And now I'm even more confused.  They differ only by return type and a 3rd parameter (which is a default parameter so callers do not require it).

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: We typically have a best effort approach to the rendering, so even though we can't set the pixel height/width, I thought we should still display the image.

Without the pixel height/width set, the image still renders, but the sizing can be off if the SVG has fixed dimensions.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: I'm not sure if we've discussed this refactor. I agree that the callbacks are difficult to follow, especially since there are so many in this class.

Maybe I can file a ticket for refactoring since this is likely an issue in other files as well?

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: @beervoley do you have any insights here?

We typically log errors through the renderContext, but like you said, we don't have access within the callbacks.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: Since height & width can be a string value (ex: auto), I wanted to make sure a valid double was returned.

With the switch to `std::stod`, I'll switch to a try/catch block

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
dmachaj: The failure doesn't seem like it would come from the width/height, though.  It would be coming from `LoadFromUriAsync`.  If that fails then it probably means that the .svg is missing from disk.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
dmachaj: Perhaps you could change `HStringToDouble` to `TryHStringToDouble` and it returns a `std::optional<double>`.  That way it has an easy way to express that it has nothing to return.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
fabiant3: Whether we crop/stretch, widescreen/letterbox, maintain the native resolution or use the display resolution, this should be part of the functional spec. At the very least we can include a DisplayMode (Fit, Stretch, Crop). @JeanRoca - any consideration?

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: The `SetRasterizedPixelHeight` implementation with `void` returns are only called from `BuildImage`, so we don't have to worry about switching threads - this is why I split the methods.

It also takes in an `ImageSource` and handles the `SvgImageSource` casting to keep `BuildImage` slightly less cluttered

Renaming to `SetRasterizedPixelHeight` and `SetRasterizedPixelHeightAsync` would make it more clear?

dmachaj: Different names would certainly make it more clear.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: Yes, that's a valid use case. SvgImageSource will inherit the non-set variable from the parent.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: I filed an issue on the repo: https://github.com/microsoft/AdaptiveCards/issues/8304

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
beervoley: yup, I agree :)
the failure here will be coming from `LoadFromUriAsync` :)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
beervoley: Anna, can you test what happens when this callback fails? Meaning, wrong URL (httpclient cannot resolve it) or it can but it's not truly an image and the stream is just garbage info?

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -533,16 +490,133 @@ namespace AdaptiveCards::Rendering::Uwp
...
-                                                 winrt::AsyncStatus /*status*/) { randomAccessStream.Seek(0); });
+                            [randomAccessStream, weakThis, imgProperties, image](
+                                auto const& /*operation*/, auto status)
+                            {
+                                if (status == winrt::AsyncStatus::Completed)
+                                {
+                                    randomAccessStream.Seek(0);
+                                    if (auto strongThis = weakThis.get())
+                                    {
+                                        strongThis->HandleAccessStreamForImageSource(
```
beervoley, line 501: see what happens to this one too when the stream is garbage

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
beervoley: @anna-dingler did you check if it negatively affects regular svgs?

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -666,4 +738,93 @@ namespace AdaptiveCards::Rendering::Uwp
...
+
+    winrt::fire_and_forget render_xaml::XamlBuilder::SetRasterizedPixelWidthAsync(winrt::SvgImageSource const imageSource,
+                                                                          double const imageSize,
+                                                                          bool const dropIfUnset)
+    {
+        co_await winrt::resume_foreground(imageSource.Dispatcher());
+        auto currentSize = imageSource.RasterizePixelWidth();
+        bool sizeIsUnset = isinf(currentSize);
+
+        bool dropWidth = sizeIsUnset && dropIfUnset;
```
beervoley, line 797: I don't seem to follow the logic here... how's that relevant if size is set or unset if you want to drop it if height is set anyway?

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -666,4 +738,93 @@ namespace AdaptiveCards::Rendering::Uwp
...
+        {
+            auto currentSize = image.RasterizePixelWidth();
+            bool sizeIsUnset = isinf(currentSize);
+
+            if (sizeIsUnset)
+            {
+                image.RasterizePixelWidth(imageSize);
+            }
+        }
+    }
```
beervoley, line 829: don't duplicate the method just to overload the parameter :)
simply add `dropIfUnset` with default value

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -666,4 +738,93 @@ namespace AdaptiveCards::Rendering::Uwp
...
+            auto rootElement = xmlDoc.DocumentElement();
+            auto height = rootElement.GetAttribute(L"height");
+            auto width = rootElement.GetAttribute(L"width");
+
+            // We only need to set height or width, not both (fixes aspect ratio for person style)
+            bool isHeightSet = false;
+
+            if (!height.empty())
+            {
+                if (auto heightAsDouble = TryHStringToDouble(height))
```
beervoley, line 756: you need to handle the exception thrown from `TryHStringToDouble` somewhere:)

source/uwp/Renderer/lib/Util.cpp
```diff
@@ -92,6 +92,19 @@ std::string HStringToUTF8(winrt::hstring const& in)
     return WStringToString(in);
 }
 
+std::optional<double> TryHStringToDouble(winrt::hstring const& in)
+{
+    try
+    {
+        return std::stod(winrt::to_string(in));
```
beervoley, line 99: check if `_wtof` is available and may be use that instead?
you can use it like:
```suggestion
        return _wtof(in.c_str());
```

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: @beervoley Yes, I've tested with all of the cases listed in this PR (have height and width set) and in my original PR (uses auto)

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: These two methods are only called from `BuildImage` and do not require switching to the foreground thread. So we can remove that logic and keep them sync.

I also include an imagesource check here, since it isn't confirmed that they are svgs at this point.

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: I can add a comment here explaining this.

If an image has an explicit size, we have already set it from within `BuildImage`. If that is the case, we _must_ update the value.

Say the image is 100x100, but the explicit size is 40x40. Without replacing the value here, the rasterized values would end up being 100x40. This would look something like this:

image

Instead of:
image

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: If the URL cannot be resolved, the status is `Error` and the image source is not set.

If the info is garbage, the errors will be handled by `HandleAccessStreamForImageSource`

source/uwp/Renderer/lib/AdaptiveImageRenderer.cpp
anna-dingler: I added a few changes here.

- Catch error if the dataStream is not read properly or if xml string cannot be loaded
- Ensure that the root element is an svg
- Null check for svgString

Let me know what you think!

beervoley (approved):

PR #8260, opened 2023-01-27
added forceLoad option for image

jwoo-msft:
# Related Issue
https://microsoft.visualstudio.com/OS/_workitems/edit/42514115

# Description
Unlike static image, animation has state. Due to caching, animation state cannot be reset. Added a forceLoad option that directs browser to fetch from the source instead of relying on its cache. 

# Sample Card
Included in PR.

# How Verified
The first Image element doesn't have forceLoad property, and its animation doesn't get refreshed vs 2n image element that does.
https://user-images.githubusercontent.com/4112696/214984279-c58f8d7b-08fe-4735-80f7-f782eeadab37.mp4

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8260)

### DIFF
--- a/samples/v1.6/Elements/Image.ForceLoad.json
+++ b/samples/v1.6/Elements/Image.ForceLoad.json
@@ -1,0 +1,21 @@
+{
+	"type": "AdaptiveCard",
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"version": "1.6",
+	"body": [
+		{
+			"type": "Image",
+			"height": "32px",
+			"width": "32px",
+			"url": "https://live-tile-images.scdn.co/dw-dark-previewl.gif",
+			"horizontalAlignment": "Center"
+		}, {
+			"type": "Image",
+			"height": "32px",
+			"width": "32px",
+			"url": "https://live-tile-images.scdn.co/dw-dark-previewl.gif",
+			"forceLoad": true,
+			"horizontalAlignment": "Center"
+		}
+	]
+}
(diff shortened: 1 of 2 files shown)


samples/v1.6/Elements/Image.ForceLoad.json
```diff
@@ -0,0 +1,21 @@
+{
+	"type": "AdaptiveCard",
+	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+	"version": "1.6",
+	"body": [
+		{
+			"type": "Image",
+			"height": "32px",
+			"width": "32px",
+			"url": "https://live-tile-images.scdn.co/dw-dark-previewl.gif",
```
jwoo-msft, line 10: we may need to update the url later.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2140,6 +2144,8 @@ export class Image extends CardElement {
                 imageElement.style.backgroundColor = backgroundColor;
             }
 
+            this.configureImageForForceLoading();
+
             imageElement.src = <string>this.preProcessPropertyValue(Image.urlProperty);
```
anna-dingler, line 2149: Does `Image.urlProperty` hold the same value as `this.url`?

jwoo-msft: No. This is really confusing. the property does a few things. it serializes and deserializes. it also serves as a key to a property bag of Adaptive element. The property bag holds the value for the url. 
We can make it more clear by using named parameter for preProcessPropertyValue as it takes two parameters, prop which is key that used here and propertyValue, intended value, but the preProcessPropertyValue is used in many places other than Image, so it's out of scope for this PR request.

anna-dingler (approved):

PR #8267, opened 2023-01-31
JS SDK - Dynamic Typeahead in Input.ChoiceSet

SuryanshPrabhat:
The PR covers code changes to add Dynamic Type Ahead support to Input.ChoiceSet in adaptive cards in JS SDK.

**Feature Details/Design**: 
https://microsoftapc-my.sharepoint.com/:w:/g/personal/suprabhat_microsoft_com/EZ9Hhv0Gls5OlxxH02r47LIBH3Xp-mRlnwlYXvr30NmN9g?e=s5awDy

**UX** for the feature:

1. Default State of the Input.ChoiceSet
image

2. On-hover/Click mouse experience.

https://user-images.githubusercontent.com/39309315/218469337-39d460e7-4145-471c-ab0c-cb3c2043b472.mp4

4. Dropdown navigation/Enter via arrow keys.

https://user-images.githubusercontent.com/39309315/218469445-00f80cac-7433-4332-9502-6c6555654274.mp4

6. Loading Experience when there are no static choices or choices do not match the filter.
image

7. Loading Experience when there are static choices and they do match the filter.
image

8. Error experience when the Data.Query request returns an error response. 
image

9. Error experience when there are no results matching the filter. 
image

10. Error message if validation fails.

https://user-images.githubusercontent.com/39309315/220511254-56d094fd-ffe0-4a9c-ba5c-fab360d9637b.mp4


SuryanshPrabhat:
@dclaux  Can you please take a look?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4196,6 +4196,35 @@ export class Choice extends SerializableObject {
     }
 }
 
+export class ChoiceSetInputDataQuery extends SerializableObject {
+    //#region Schema
+
+    static readonly typeProperty = new StringProperty(
```
anna-dingler, line 4202: @jwoo-msft Do we normally include the `type` property here?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4196,6 +4196,35 @@ export class Choice extends SerializableObject {
...
+        Versions.v1_6,
+        "type",
+        true,
+        new RegExp("/^Data.Query$/")
+      );      
+    static readonly datasetProperty = new StringProperty(Versions.v1_6, "dataset");
+
+	@property(ChoiceSetInputDataQuery.typeProperty)
```
anna-dingler, line 4210: nit: spacing

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4205,6 +4234,11 @@ export class ChoiceSetInput extends Input {
         "choices",
         Choice
     );
+    static readonly choicesDataProperty = new SerializableObjectProperty(
```
anna-dingler, line 4237: nit: spacing

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4245,6 +4279,9 @@ export class ChoiceSetInput extends Input {
     @property(ChoiceSetInput.choicesProperty)
     choices: Choice[] = [];
 
+    @property(ChoiceSetInput.choicesDataProperty)
+	choicesData?: ChoiceSetInputDataQuery;
```
anna-dingler, line 4283: spacing

SuryanshPrabhat: Fixed.

source/nodejs/adaptivecards/src/card-elements.ts
SuryanshPrabhat: Fixed.

source/nodejs/adaptivecards/src/card-elements.ts
SuryanshPrabhat: Fixed.

source/nodejs/adaptivecards/src/card-elements.ts
SuryanshPrabhat: Type here needs to be defined since I am not extending from CardObject but SerializableObject. . The intention here is to not regard "choices.data" if the type is not "Data.Query".  I am not sure what is the best way to do it. Currently, I am checking the type and dataset to allow dynamic typeahead supported choiceset to be rendered. Other ways to handle could be doing something with internalparse or internalvalidateproperties method but I feel that is unnecessary. Let me know your thoughts. @jwoo-msft @anna-dingler

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4196,6 +4196,40 @@ export class Choice extends SerializableObject {
...
+    dataset: string;
+
+    //#endregion
+
+    protected getSchemaKey(): string {
+        return "choices.data";
+    }
+}
+
+export type FetchedChoice = {
```
anna-dingler, line 4228: @jwoo-msft Could we use `Choice` here instead? Or do we want to avoid that since it extends `SerializableObject`?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4650,6 +4751,221 @@ export class ChoiceSetInput extends Input {
...
+                }
+            } else if (event.key === "Enter") {
+                choice.click();
+            }
+        };
+        return choice;
+    }
+
+    filterChoices(isDynamic?: boolean) {
+        const filter = this._textInput?.value.toLowerCase() || "";
```
anna-dingler, line 4860: Can the card author provide static and dynamic choices? If so, would we want to filter both arrays here?

SuryanshPrabhat: We filter the static choices and dynamic choices at separate times. Notice that we call the filterChoices method the first time (for filtering only the static choices) in the flow when we show the loading indicator. This is done because we want to show the static choices as soon as possible to the user. Dynamic choices will take time to be fetched via the network call. And when they do come, we will filter and add them to the list. It is for a better UX, hence the design.

source/nodejs/adaptivecards/src/card-elements.ts
SuryanshPrabhat: Precisely. We can do it using `Choice` as well. But this will require additional processing upon receiving the raw response from the host, since `Choice` is not just about `title` and `string` but has the properties of a `SerializableObject`, and we do not need that here.

SuryanshPrabhat:
> Hi @SuryanshPrabhat,
> 
> Can you add samples to this folder: https://github.com/microsoft/AdaptiveCards/tree/main/samples/v1.6/Elements, and also include some gifs/videos of this working in your PR?

Sure @anna-dingler 
Thanks for pointing it out. I've done the same.

anna-dingler (approved):
Looks good! I have also added @jwoo-msft

pankajmi2 (approved):

jwoo-msft (commented):
Your change looks great. I have just a few questions.
Will error label displayed if the input is required, and not provided or validation is set but it doesn't meet the validation?
property name type is confusing. Since the element is not part AdaptiveCard element yet having the same name is confusing.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4188,6 +4188,40 @@ export class Choice extends SerializableObject {
...
+    static readonly typeProperty = new StringProperty(
+        Versions.v1_6,
+        "type",
+        true,
+        new RegExp("^Data.Query$")
+    );
+    static readonly datasetProperty = new StringProperty(Versions.v1_6, "dataset");
+
+    @property(ChoiceSetInputDataQuery.typeProperty)
+    type: string;
```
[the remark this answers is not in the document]
dclaux: @jwoo-msft this class extends `SerializableObject`, not `BaseCardElement`. `SerializableObject` does not have a `type` property. It is completely fine (and the right thing to do) to name this property `typeProperty`/`type`.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4237,6 +4276,9 @@ export class ChoiceSetInput extends Input {
     @property(ChoiceSetInput.choicesProperty)
     choices: Choice[] = [];
 
+    @property(ChoiceSetInput.choicesDataProperty)
```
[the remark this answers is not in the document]
dclaux: @jwoo-msft I'm not sure why this particular name would need to be clarified. There are many property definitions throughout the code, and they each are named using the format `<name of the property in the AC schema>property`. This is the property definition for the `choices.data` schema property.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5587,6 +5932,33 @@ export class ExecuteAction extends SubmitActionBase {
     }
 }
 
+export class DataQuery extends UniversalAction {
```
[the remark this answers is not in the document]
dclaux: @jwoo-msft that is because an action isn't necessarily something that happens as a result of a button click. There is no reason why an action shouldn't be executed as a result of different triggered, such as, in this case, the fact that the input's value has changed and new values need to be retrieved from the backend. Besides, an "action" merely defines a contrac, and here the contract is "a Data.Query action contains information on where to retrieve data from (the dataset) and the response must be a set of key/value pairs"

Defining `Data.Query` as an action fits beautifully into the overall model and architecture of the SDK.

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4405,8 +4472,44 @@ export class ChoiceSetInput extends Input {
...
+                            this._textInput.removeAttribute("placeholder");
+                            this._textInput.removeAttribute("aria-label");
+                        } else if (this.placeholder) {
+                            this._textInput.placeholder = this.placeholder;
+                            this._textInput.setAttribute("aria-label", this.placeholder);
+                        }
+                    }
+                }, this.hostConfig.inputs.debounceTime);
+            }
+            filteredChoiceSet.parent = this;
```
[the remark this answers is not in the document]
SuryanshPrabhat: Yes, and yes. FilteredChoiceSet is simply concerned with just the UI part.
Error label and input validation works as expected because `textInput` of `ChoiceSetInput` instance is being set with the `textInput` in the `FilteredChoiceSet` instance. This way the validation logic works in this case as well. If you look at line 4697, the `isValid` method, I'll added logic there to handle dynamic choices as well. Let me add a video in the PR description as well to show the error label.

source/nodejs/adaptivecards/src/adaptive-applet.ts
```diff
@@ -714,6 +748,52 @@ export class AdaptiveApplet {
...
+        this._choiceSet?.showLoadingIndicator();
+        let response = undefined;
+        try {
+            response = await this.channelAdapter.sendRequestAsync(request);
+        } catch (error) {
+            logEvent(Enums.LogLevel.Error, "Activity request failed: " + error);
+        }
+        this._choiceSet?.removeLoadingIndicator();
+        if (response) {
+            if (response instanceof SuccessResponse) {
```
baton17, line 764: can these two if blocks be combined into one?

source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -266,11 +266,13 @@ export class InputConfig {
     readonly errorMessage: BaseTextDefinition = new BaseTextDefinition({
         color: Enums.TextColor.Attention
     });
+    readonly debounceTime: number = 0;
```
baton17, line 269: is it milliSeconds? Would this be better name debounceTimeInMilliSeconds?

baton17 (approved):
Added some code readability related changes. Overall looks good to me. Signing off. Please resolve the comments before merge.

source/nodejs/adaptivecards/src/adaptive-applet.ts
```diff
@@ -714,6 +748,52 @@ export class AdaptiveApplet {
...
+                const rawResponse = response.rawContent;
+                if (rawResponse) {
+                    let parsedResponse = rawResponse;
```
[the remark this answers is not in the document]
SuryanshPrabhat: It does not make any difference. I believe it's fine.

source/nodejs/adaptivecards/src/adaptive-applet.ts
SuryanshPrabhat: Different blocks are inside.

PR #8272, opened 2023-02-01
Feature - [Android] Dynamic Type Ahead

dipja:
# Feature - [Android] Dynamic Type Ahead
The feature is divided into 3 PR's for easier reviews and merged into 1 after reviews. Below are the details.

This is the **first** PR covers code changes to add IChoicesResolver in BaseElementRender. This resolver can be used by any element or input element, in this case will be used by ChoiceSetRender in order to resolve dynamic choices from the host.

https://github.com/microsoft/AdaptiveCards/pull/8273 - This is the second PR covers code changes to add drawables as UI resources in dynamic type ahead new screen. Also, provides support for custom icons from host.

https://github.com/microsoft/AdaptiveCards/pull/8274 - This is the third PR covers code changes to add new screen for dynamic type ahead. When user clicks on the Input.Choiceset which has choices.data, a new screen is opened. The choices in this screen are fetched dynamically from the host and allows user to make selections and clear then as well.

Feature Details: https://github.com/microsoft/AdaptiveCards/blob/dipja/dynamic-type-ahead-doc-android/specs/DesignDiscussions/DynamicTypeAhead.md

Design Doc: https://github.com/microsoft/AdaptiveCards/blob/dipja/dynamic-type-ahead-doc-android/specs/DesignDiscussions/DynamicTypeAheadAndroid.md


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
```diff
@@ -60,17 +62,42 @@ public RenderedAdaptiveCard render(
...
         @Nullable IOverflowActionRenderer overflowActionRenderer,
+        @Nullable IChoicesResolver choicesResolver,
+        @Nullable ActivityResultRegistry activityResultRegistry,
         HostConfig hostConfig)
     {
         RenderedAdaptiveCard result = new RenderedAdaptiveCard(adaptiveCard);
         CardRendererRegistration.getInstance().registerOverflowActionRenderer(overflowActionRenderer);
-        View cardView = internalRender(result, context, fragmentManager, adaptiveCard, cardActionHandler, hostConfig, false, View.NO_ID);
+        CardRendererRegistration.getInstance().registerActivityResultRegistry(activityResultRegistry);
+        View cardView = internalRender(result, context, fragmentManager, adaptiveCard, cardActionHandler, choicesResolver, hostConfig, false, View.NO_ID);
```
almedina-ms, line 100: >choicesResolver [](http://example.com/codeflow?start=106&length=15)

why is choicesResolver not registered as the other two new parameters in this method?

dipja: @almedina-ms - Let's think of a scenario when there are multiple cards in 1 activity/screen (teams or outlook).

In such a case if we register a paramter then it shall be common for all cards in the activity/screen. Other parameters such as activity register or custom overflow menu renderer are common for all such cards. 

But when we talk about action handling or fetching dynamic choices, it can be different for different cards. For example there are two cards in an activity card 1 and card 2. Now actions clicked on card 1 can be handled differently than how there are handled on card 2 (in case of teams invokes are sent to different bots who have sent the card).

That's why ICardActionHanlder and IChoicesResolver are per card and need to be sent as parameters in the render() method. Since this is an interface method, we cannot overloading is not possible which makes this a breaking change. But as I explained earlier, this is a required change.

Let me know if you have any further questions

dipja: I have introduced a `ChannelAdaptor` which is responsible for any host/SDK communication. Currently it will contain `ICardActionHandler` and `IChoicesResolver`.

This will enable us to prevent any breaking changes in the future if new host/SDK interactions are introduced.

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/IActionLayoutRenderer.java
```diff
@@ -10,14 +10,15 @@
 import io.adaptivecards.objectmodel.BaseActionElementVector;
 import io.adaptivecards.objectmodel.HostConfig;
 import io.adaptivecards.renderer.actionhandler.ICardActionHandler;
+import io.adaptivecards.renderer.typeaheadsearch.IChoicesResolver;
```
almedina-ms, line 13: I think this import may not be needed anymore

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/IBaseActionElementRenderer.java
```diff
@@ -10,6 +10,7 @@
 import io.adaptivecards.renderer.actionhandler.ICardActionHandler;
 import io.adaptivecards.objectmodel.BaseActionElement;
 import io.adaptivecards.objectmodel.HostConfig;
+import io.adaptivecards.renderer.typeaheadsearch.IChoicesResolver;
```
almedina-ms, line 13: same here

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/DateInputRenderer.java
```diff
@@ -15,13 +15,15 @@
...
 import io.adaptivecards.renderer.AdaptiveWarning;
+import io.adaptivecards.renderer.ChannelAdaptor;
 import io.adaptivecards.renderer.RenderArgs;
 import io.adaptivecards.renderer.RenderedAdaptiveCard;
 import io.adaptivecards.renderer.TagContent;
 import io.adaptivecards.renderer.Util;
 import io.adaptivecards.renderer.actionhandler.ICardActionHandler;
 import io.adaptivecards.renderer.inputhandler.DateInputHandler;
 import io.adaptivecards.renderer.readonly.RendererUtil;
+import io.adaptivecards.renderer.typeaheadsearch.IChoicesResolver;
```
almedina-ms, line 26: same here

source/android/mobile/src/main/java/io/adaptivecards/adaptivecardssample/CustomObjects/Actions/CustomActionListener.java
```diff
@@ -9,6 +9,7 @@
 import io.adaptivecards.objectmodel.BaseActionElement;
```
almedina-ms, line 9: I think you may be missing the update to the sample applications or a way to exercise this API manually

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
```diff
@@ -67,10 +69,34 @@ public RenderedAdaptiveCard render(
...
+    }
+
+    public RenderedAdaptiveCard render(Context context,
+                                       FragmentManager fragmentManager,
+                                       AdaptiveCard adaptiveCard,
+                                       @NonNull ChannelAdaptor channelAdaptor,
+                                       @Nullable IOverflowActionRenderer overflowActionRenderer,
+                                       @Nullable ActivityResultRegistry activityResultRegistry)
+    {
+        return render(context, fragmentManager, adaptiveCard, channelAdaptor, overflowActionRenderer, activityResultRegistry, defaultHostConfig);
```
almedina-ms, line 84: why is defaultHostConfig used here?

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/IActionLayoutRenderer.java
dipja: yes, removed

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/DateInputRenderer.java
dipja: done

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/IBaseActionElementRenderer.java
dipja: done

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/AdaptiveCardRenderer.java
dipja: Thanks for pointing out, somehow got it wrong during resolving conflicts. Updated and fixed.

source/android/mobile/src/main/java/io/adaptivecards/adaptivecardssample/CustomObjects/Actions/CustomActionListener.java
dipja: The sample app will use this API. The change is present in the 3rd PR for the same feature

almedina-ms (approved):

PR #8277, opened 2023-02-02
[WinUI 3 Renderer] Change target platform to match WindowsAppSDK support

krschau:
# Related Issue

#5888

# Description

The Adaptive Cards WinUI 3 renderer should support all platforms that WinUI 3 / WindowsAppSDK can run on. This change supports that by explicitly targeting the lowest supported version. This requires linking a few additional libraries, as well as setting a specific target on the ObjectModelProjection projct.

# How Verified

Build pipelines

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/AdaptiveCards/pulls/8277)

### DIFF
--- a/source/uwp/ObjectModelProjection/ObjectModelProjection.vcxproj
+++ b/source/uwp/ObjectModelProjection/ObjectModelProjection.vcxproj
@@ -196,7 +196,7 @@
     <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
     <AppContainerApplication>true</AppContainerApplication>
     <ApplicationType>Windows Store</ApplicationType>
-    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.15063.0</WindowsTargetPlatformVersion>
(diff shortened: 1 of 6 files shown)


source/uwp/ObjectModelProjection/ObjectModelProjection.vcxproj
```diff
@@ -196,7 +196,7 @@
     <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
     <AppContainerApplication>true</AppContainerApplication>
     <ApplicationType>Windows Store</ApplicationType>
-    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
```
krschau, line 199: @beervoley I originally tried changing this to match the WindowsTargetPlatformMinVersion, but the pipeline VMs didn't have that SDK installed. Is 17763 ok? It can't be 10.0 because targeting a newer version makes it incompatible with the CsWinRT projection.

beervoley (approved):

PR #8299, opened 2023-02-09
Jwoo/js keyboard fix

jwoo-msft:
# Related Issue
https://microsoft.visualstudio.com/OS/_sprints/taskboard/OWL%20(Open%20Windows%20Platform).SnaX.Team/OS/2302?workitem=43065527

# Description
Navigating through carousel pages using arrow keys will turn the pages of all carousels within a browser.
To fix the issue.
1. disable navigation via arrow keys
2. add keyboard event handler to bullets.

# Sample Card

N/A

# How Verified

Created a test app to test this use case.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8299)


source/nodejs/adaptivecards/src/carousel.ts
```diff
@@ -605,6 +605,17 @@ export class Carousel extends Container {
             this.raiseCarouselEvent(Enums.CarouselInteractionEvent.Autoplay);
         });
 
+        carousel.on('paginationRender', (swiper, paginationEl) => {
```
anna-dingler, line 608: nit/question: Did we want to enable the keyboard arrow keys on the carousel container?

anna-dingler (approved):

source/nodejs/adaptivecards/src/carousel.ts
jwoo-msft: arrow key is enabled, is just that pressing arrow key won't turn carousel pages.

anna-dingler: @jwoo-msft yeah I meant that if we implemented turning the carousel pages ourselves, would it work as expected (not turn all widgets)?

I like the feature in general, so I was just curious :)

jwoo-msft: @anna-dingler, we probably run into the same issue they are facing of figuring out where the event occurred.

PR #8306, opened 2023-02-13
[BugFix] Fixing input container's width when label width is specified in percentage

baton17:
# Description

Title: Fixing input container's width when label position is inline and width is specified in percentage .

Issue: Earlier, when label is placed inline and if `labelWidth` is specified to let's say 50 means 50%, then we were assigning 50% width of the outer input component to label element and we should have assigned 50% of the outer input component to the input control element but instead we were setting input element's width as 50% width of its inner input container element not the outer input component.

Fix: Fixed by setting width on the appropriate input container.

# Sample Card

```
{
	"type": "AdaptiveCard",
	"version": "1.6",
	"body":
         [
		{
			"type": "Input.Text",
			"id": "inputProduct",
			"value": "Cosmetics",
			"label": "Product Name",
			"labelPosition": "inline",
			"labelWidth": 50
		}
        ]
}
```
Before: 
input element was getting less width:
image

After:
input element was getting expected width:
image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8306)


anna-dingler (approved):
Looks good!
Can you add before and after screenshots to the PR in case we look back this?

PR #8328, opened 2023-02-24
Add host parameter to AdaptiveCards.Templating

paulcam206:
Adds a `$host` parameter to the .NET Templating library (achieving parity with the `adaptivecards-templating` NPM package).

Basic idea: The app using the templating library can provide its own separate data blob to be used in the binding process. The structure of this data is not specified (similar to the root data blob that can already be supplied). This allows a host application to provide some environmental information for use by the template.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8328)


anna-dingler (approved):

beervoley (approved):
lgtm

PR #8352, opened 2023-03-08
[UWP] Update visualizer for high contrast

### LINKED ISSUE #8040: [Adaptive Cards>ActivityUpdatewithLabel.json]: Unselected text look like a selected text in HC mode
### Target Platforms

UWP

### SDK Version

3.0.2210.26001

### Application Name

Adaptive cards

### Problem Description

**Test Environment:**
OS version: 22H2 (Build 25182.1010)
Application: AdaptiveCardVisualizer
Application Version: 3.0.2210.26001
Screen Reader: Narrator

**Repro Steps:**

1.Open the Adaptive Card Visualizer application.
2.Turn on the HC mode.
3.Navigate to the 'ActivityUpdate.Json' control through tab key and activate it.
4.Now, navigate through ActivityUpdatewithLabel.json card and observe the issue.

**Actual Result:**
When the low vision user applied HC mode, unselected text look like a selected text.

**Expected Result:**
For better low vision user experience, unselected text should not look like a selected text.

**Observation:**
Same issue repro throughout the application.

**User Impact:**
Low vision users will get confusion, if the text is looks like a selected text, even if the text is not selected.

**MAS Reference:**
https://liquid.microsoft.com/Web/Object/Read/MS.Accessibility/Requirements/04.03.01

### Screenshots

In HC mode normal text is visible as selected text

### Card JSON

```JSON
NA
```

### Sample Code Language

_No response_

### Sample Code

_No response_

anna-dingler:
# Related Issue

Fixes #8040 

# Description

The card background was not being updated when it is in high contrast.

I resolved this on the visualizer by:
- Switching host configs when we are in high contrast mode. These config files do not have many updates now, but we can work with designers to update other color selections as needed. **Note: ** there is a config file for each of the 4 high contrast themes.
- Apply a border around the card while in high contrast mode. Otherwise, the card and the card viewer background are the same color.

# How Verified

Verified manually on the UWP Visualizer

## Before
image

## After
image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8352)


source/uwp/SharedVisualizer/ViewModel/HostConfigEditorViewModel.cs
```diff
@@ -58,9 +59,10 @@ protected override void LoadPayload(string payload)
 
         public static async Task<HostConfigEditorViewModel> LoadAsync(MainPageViewModel mainPageViewModel)
         {
+            string fileName = PickHostConfigFile();
             try
             {
-                StorageFile file = await StorageFile.GetFileFromApplicationUriAsync(new Uri("ms-appx:///HostConfigs/DefaultHostConfig.json"));
+                StorageFile file = await StorageFile.GetFileFromApplicationUriAsync(new Uri("ms-appx:///" + fileName));
```
paulcam206, line 65: super nitpicky, but this might be a little cleaner if you left in `HostConfigs/` and only appended the filename... feel free to not take this suggestion. :)

source/uwp/SharedVisualizer/ViewModel/HostConfigEditorViewModel.cs
```diff
@@ -73,5 +75,26 @@ public static async Task<HostConfigEditorViewModel> LoadAsync(MainPageViewModel
                 return new HostConfigEditorViewModel(mainPageViewModel);
             }
         }
+
+        protected static string PickHostConfigFile()
+        {
+            AccessibilitySettings accessibilitySettings = new AccessibilitySettings();
```
paulcam206, line 81: While you're doing this, it's probably worth wiring up a [`HighContrastChanged` event handler](https://learn.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.accessibilitysettings.highcontrastchanged?view=winrt-22621) so that user contrast theme changes are automatically applied.

paulcam206 (approved):

source/uwp/SharedVisualizer/ViewModel/HostConfigEditorViewModel.cs
paulcam206, line 81: should we also be setting [`HighContrastAdjustment` to `None`](https://learn.microsoft.com/en-us/windows/apps/design/accessibility/high-contrast-themes#setting-highcontrastadjustment-to-none)?

paulcam206 (approved):

source/uwp/SharedVisualizer/ViewModel/HostConfigEditorViewModel.cs
anna-dingler: I didn't make this change because `HighContrastAdjustment` handles all text colors by default.

PR #8414, opened 2023-03-27
[Input][RevealOnHover]Fixes for reveal on hover on inputs

baton17:
# Description
It consists of some fixes related to revealOnHover property on inputs.
1. css fix to add onfocus to revealOnHover style class. 
2. reading correct css className for revealOnHover coming from hostConfig.
3. add allowRevealOnHover boolean in host config. Host can choose to selectively disable this styling feature while using latest sdk version.

# How Verified
1. manual scenario verification.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8414)

### DIFF
--- a/source/nodejs/adaptivecards/src/host-config.ts
+++ b/source/nodejs/adaptivecards/src/host-config.ts
@@ -268,10 +268,13 @@ export class InputConfig {
     });
     readonly debounceTimeInMilliSeconds: number = 0;
 
+	allowRevealOnHoverStyle: boolean = false;
+
     constructor(obj?: any) {
         if (obj) {
             this.label = new InputLabelConfig(obj["label"]);
             this.errorMessage = new BaseTextDefinition(obj["errorMessage"]);
+			this.allowRevealOnHoverStyle = obj["allowRevealOnHoverStyle"] || this.allowRevealOnHoverStyle;
             this.debounceTimeInMilliSeconds = obj.debounceTimeInMilliSeconds;
         }
     }
--- a/source/nodejs/adaptivecards/src/host-config.ts
+++ b/source/nodejs/adaptivecards/src/host-config.ts
@@ -268,7 +268,7 @@ export class InputConfig {
     });
     readonly debounceTimeInMilliSeconds: number = 0;
 
-    allowRevealOnHoverStyle: boolean = true;
+    allowRevealOnHoverStyle: boolean = false;
 
     constructor(obj?: any) {
         if (obj) {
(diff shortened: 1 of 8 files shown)


source/nodejs/adaptivecards/src/host-config.ts
```diff
@@ -268,10 +268,13 @@ export class InputConfig {
     });
     readonly debounceTimeInMilliSeconds: number = 0;
 
+    allowRevealOnHoverStyle: boolean = true;
+
     constructor(obj?: any) {
         if (obj) {
             this.label = new InputLabelConfig(obj["label"]);
             this.errorMessage = new BaseTextDefinition(obj["errorMessage"]);
+            this.allowRevealOnHoverStyle = obj["allowRevealOnHoverStyle"] || this.allowRevealOnHoverStyle;
```
anna-dingler, line 277: Why is this an `or` statement?

anna-dingler: Is this meant to be an "opt-in" feature?

jwoo-msft: why do you need 'or'?

jwoo-msft (commented):
could you address the anna-dingler's comment?

source/nodejs/adaptivecards/src/host-config.ts
baton17: incase obj["allowRevealOnHoverStyle"] is null, it will resolve to false. I followed similar to allowCustomStyle 

export class AdaptiveCardConfig {
    allowCustomStyle: boolean = false;

    constructor(obj?: any) {
        if (obj) {
            this.allowCustomStyle = obj["allowCustomStyle"] || this.allowCustomStyle;
        }
    }
}

baton17: Yes its an opt-in feature. revealOnHover is only defined for web/win32 experiences where mouse hover is there. Its not intended to be lighten up for mobile. There are apps like Outlook mobile which uses adaptive card js sdk inside webview. So, we dont want to enable this feature for them, as mobile view should directly show inputs as editable field as there is no hover experience.

Thus, adding this flag where host can choose to enable/disable it.

anna-dingler: Makes sense, thank you for explaining!

anna-dingler (approved):

PR #8491, opened 2023-04-25
[UWP Renderer] Validate text inputs dynamically and make errors live regions

### LINKED ISSUE #8191: {Freeway-Bug}[Adaptive Cards> InputFormWithLabels.json> Phone Number]: Screen reader does not announce the information of 'Maximum character limit reached' on the 'Phone number' edit field.
### Target Platforms

UWP

### SDK Version

3.0.2210.26001

### Application Name

Adaptive Cards

### Problem Description

**Test Environment:**
OS version: 22H2 (Build 25267.1000)
Application: AdaptiveCardVisualizer
Application Version: 3.0.2210.26001
Screen reader: Narrator

**Repro Steps:**

1. Open the Adaptive Card Visualizer application.
2. Turn on the Narrator through Ctrl+ Windows+ Enter key.
3. Navigate to the 'InputFormWithLabels.json' through tab key and activate it.
4. Now, navigate to 'Phone Number' through tab key.
5. Observe the issue.

**Actual Result:**
Screen reader does not notify about the "maximum character limit reached " to the user when user provide input in the 'phone number' edit field.

**Expected Result:**
Screen reader should notify about the "**maximum character limit reached** " to the user when user provide input in the 'phone number' edit field.

**Notes:**
Validation is not implemented on 'Phone Number' edit field so that user can't be able to type alphabet  in the same.

**User Impact:**
Screen reader user will not be able to track or will continue to enter the data in the edit field if maximum limit reached information is not announced.

**MAS Reference link:**
https://microsoft.sharepoint.com/:w:/s/accessibility/EXdDUZzldv5MgwDaMHnTzcMBcT8opKGncYjU-gCJ4uqQdg?e=9nuVmt

“Have feedback to share on Bugs? Please help fill [Trusted Tester Bug Feedback (office.com)](https://nam06.safelinks.protection.outlook.com/?url=https://forms.office.com/pages/responsepage.aspx?id%3Dv4j5cvGGr0GRqy180BHbR5dUqGQgJ9lItv-jqY8F09lURVdYTENIRkY2UzlCR0NKV1E2VUE5OVpDVy4u&data=04%7c01%7cv-vaibgu%40microsoft.com%7c551850210d444be9b28308d9691e3dde%7c72f988bf86f141af91ab2d7cd011db47%7c1%7c0%7c637656400980259288%7cUnknown%7cTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7c1000&sdata=TrcBaCyAPKg0WAQfdSZJP2i9pghMEDg59i2NeUUs5EQ%3D&reserved=0) “

### Screenshots

https://user-images.githubusercontent.com/104483005/209768239-cc109e12-45a1-4c17-9f6b-73c41d1176bc.mp4

### Card JSON

```JSON
NA
```

### Sample Code Language

_No response_

### Sample Code

_No response_

anna-dingler:
# Related Issue

Fixes #8191

# Description

We need to add dynamic input validation so that all necessary information is announced by Narrator.

In `AdaptiveTextInputRenderer`:
- Validate the text each time it changes

In `InputValue`:
- Make the errorMessage a live region. I used `Assertive` so that the error is announced as soon as possible
- Whenever the errorMessage appears, we raise a `LiveRegionChanged` event so that Narrator will announce it.

# Sample Card

https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.5/Scenarios/InputForm.json

# How Verified

Verified manually on the UWP Visualizer.

https://user-images.githubusercontent.com/98650930/234417261-20ddb46b-8dfd-4d65-bdd5-cf6754431d13.mp4

**Note:**  This should be excluded from the v1.6 release since it is only implemented on UWP.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8491)

### DIFF
--- a/source/uwp/SharedRenderer/lib/InputValue.cpp
+++ b/source/uwp/SharedRenderer/lib/InputValue.cpp
@@ -49,8 +49,7 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering
             uiElementDescribers.Append(uiValidationErrorAsDependencyObject);
 
             // We also want to raise a LiveRegionChanged event so that the error text is announced by narrator when it appears
-            auto errorAutomationPeer = winrt::FrameworkElementAutomationPeer::FromElement(m_validationError);
-            if (errorAutomationPeer != nullptr)
+            if (const auto errorAutomationPeer = winrt::FrameworkElementAutomationPeer::FromElement(m_validationError))
             {
                 errorAutomationPeer.RaiseAutomationEvent(winrt::AutomationEvents::LiveRegionChanged);
             }
--- a/source/uwp/SharedRenderer/lib/InputValue.cpp
+++ b/source/uwp/SharedRenderer/lib/InputValue.cpp
@@ -37,6 +37,7 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering
     {
         auto uiElementDescribers = winrt::AutomationProperties::GetDescribedBy(m_uiInputElement);
 
+        winrt::AutomationProperties::SetLiveSetting(m_validationError, winrt::AutomationLiveSetting::Polite);
         auto uiValidationErrorAsDependencyObject = m_validationError.as<winrt::DependencyObject>();
 
         uint32_t index;
@@ -46,6 +47,13 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering
         if (!isInputValid && !found)
         {
             uiElementDescribers.Append(uiValidationErrorAsDependencyObject);
+
+            // We also want to raise a LiveRegionChanged event so that the error text is announced by narrator when it appears
+            auto errorAutomationPeer = winrt::FrameworkElementAutomationPeer::FromElement(m_validationError);
+            if (errorAutomationPeer != nullptr)
+            {
+                errorAutomationPeer.RaiseAutomationEvent(winrt::AutomationEvents::LiveRegionChanged);
+            }
         }
         else if (isInputValid && found)
         {
--- a/source/uwp/SharedRenderer/lib/InputValue.cpp
+++ b/source/uwp/SharedRenderer/lib/InputValue.cpp
@@ -37,7 +37,7 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering
     {
         auto uiElementDescribers = winrt::AutomationProperties::GetDescribedBy(m_uiInputElement);
 
-        winrt::AutomationProperties::SetLiveSetting(m_validationError, winrt::AutomationLiveSetting::Polite);
+        winrt::AutomationProperties::SetLiveSetting(m_validationError, winrt::AutomationLiveSetting::Assertive);
         auto uiValidationErrorAsDependencyObject = m_validationError.as<winrt::DependencyObject>();
 
         uint32_t index;
--- a/source/uwp/SharedRenderer/lib/AdaptiveTextInputRenderer.cpp
+++ b/source/uwp/SharedRenderer/lib/AdaptiveTextInputRenderer.cpp
@@ -113,7 +113,7 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
 
         auto inputValue = winrt::make_self<winrt::TextInputValue>(adaptiveTextInput, textBox, validationBorder);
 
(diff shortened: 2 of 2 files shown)


source/uwp/SharedRenderer/lib/InputValue.cpp
```diff
@@ -37,6 +37,7 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering
     {
         auto uiElementDescribers = winrt::AutomationProperties::GetDescribedBy(m_uiInputElement);
 
+        winrt::AutomationProperties::SetLiveSetting(m_validationError, winrt::AutomationLiveSetting::Polite);
         auto uiValidationErrorAsDependencyObject = m_validationError.as<winrt::DependencyObject>();
```
beervoley, line 41:
```suggestion
```

source/uwp/SharedRenderer/lib/InputValue.cpp
```diff
@@ -46,6 +47,13 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering
         if (!isInputValid && !found)
         {
             uiElementDescribers.Append(uiValidationErrorAsDependencyObject);
+
+            // We also want to raise a LiveRegionChanged event so that the error text is announced by narrator when it appears
+            auto errorAutomationPeer = winrt::FrameworkElementAutomationPeer::FromElement(m_validationError);
+            if (errorAutomationPeer != nullptr)
+            {
+                errorAutomationPeer.RaiseAutomationEvent(winrt::AutomationEvents::LiveRegionChanged);
+            }
```
beervoley, line 56:
```suggestion
            if (const auto errorAutomationPeer = winrt::FrameworkElementAutomationPeer::FromElement(m_validationError))
            {
                errorAutomationPeer.RaiseAutomationEvent(winrt::AutomationEvents::LiveRegionChanged);
            }
```

beervoley:
Good stuff :) A couple of question though:
1) In the a11y ticket they asked to make the narrator announce when the max number of characters is reached in the textbox. This change doesn't fix it (at least I think it doesn't?) and we need to raise an automation event to notify the user that max number of characters is reached while they're still typing (no need on submit I think since it will be announced during typing?)
2) Why shouldn't it be included in 1.6? This is an accessibility change for one of our renderers and it's alright that others don't  (yet) follow suit?:) Do you suggest we wait until it's in every renderer and then include it in the release?

anna-dingler:
> 

1. Yup, I agree that this doesn't directly address the a11y issue. However, there isn't a way (at least that I know of) to force an announcement without a visual indicator on screen. IMO this change addresses the issue, just in a different solution than the one they proposed

A couple of things we _could_ try:
 - Update the description to say that the max number of characters is reached. Then call `LiveRegionChanged` on the entire textbox. I think this would announce the whole textbox again though
 - Change the error message to "Max number of characters reached" and set the Visibility to `Gone`. TBD if this has the expected behavior


2. Joseph and I have been discussing releasing v1.6, and then immediately releasing v1.7 beta. This would go in v1.7 beta because there are no plans to implement this on other renderers.

At best, it will make it in JS, but we are no longer supporting mobile once decentralization is complete and the new owners will have no obligation to maintain parity.

anna-dingler:
@beervoley I synced with the a11y team, and they said that this approach should be fine as long as we make the live regions assertive (most recent commit).

What do you think?

beervoley:
> @beervoley I synced with the a11y team, and they said that this approach should be fine as long as we make the live regions assertive (most recent commit).
> 
> What do you think?

I think that's fine as long as a11y teams agree with the solution :)

beervoley (approved):

anna-dingler:
> > @beervoley I synced with the a11y team, and they said that this approach should be fine as long as we make the live regions assertive (most recent commit).
> > What do you think?
> 
> I think that's fine:) assertive works great here because it will interrupt the input once regex doesn't match (right?) and that seems reasonable.

Yup, assertive will interrupt as soon as we raise the event.

Polite (what I had before) would have waited until current narrations were done.

PR #8516, opened 2023-05-22
iOS - Language skip validation to prevent crashes

GabrielMedAlv:
# Related Issue

A Pull Request should close a **single** issue; multiple issues can be closed when the issues are **small and related**, but that should be an exception not the rule. Please keep Pull Requests small and targeted; large 'code drops' with dozens of files will be closed and asked to split into reviewable pieces. Reviews that need to be large due to dependencies will be
reviewed on a case-by-case basis.

Please use one of the well-known [github fixes keywords](https://help.github.com/en/articles/closing-issues-using-keywords) to reference
the issue fixed with this PR (eg Fixes #<github issue number>). If an issue doesn't yet exist please create one to aid
in issue tracking.

page above.

# Description

There is a crash happening on our app due to a parsing issue with the language field. Looking into the solution, we found that the problem is in the C++ code of SharedAdaptiveCard, specifically with the language validation, which always fail into a runtime error. Our investigation concludes that is because of differences on the implementation of the standard library, specifically the functions std::locale, on different platforms (In this case, Apple platforms).

Screenshot 2023-04-27 at 17 36 11

The proposed solution is to skip the language validation on Apple devices. This doesn't affect these devices because, currently is failing this validation every time a language field is included.

# Sample Card

{"body":[{"type":"TextBlock","text":"Hello World!","wrap":true},{"height":"auto","type":"Image","url":"https:\/\/us-prod.asyncgw.teams.microsoft.com\/urlp\/v1\/url\/content?url=https%3a%2f%2fcataas.com%2fcat","spacing":"none"}],"lang":"en","type":"AdaptiveCard","version":"1.2"}

# How Verified

Verified manually using multiple cards with language field.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8516)


source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
```diff
@@ -185,9 +185,11 @@ std::shared_ptr<ParseResult> AdaptiveCard::Deserialize(const Json::Value& json,
     std::string fallbackText = ParseUtil::GetString(json, AdaptiveCardSchemaKey::FallbackText);
     std::string language = ParseUtil::GetString(json, AdaptiveCardSchemaKey::Language);
     std::string speak = ParseUtil::GetString(json, AdaptiveCardSchemaKey::Speak);
-
+    
+#if !defined(__APPLE__)
```
jwoo-msft, line 189: nit: I suggest changing it to #ifndef for consistency with the rest of the codes.

jwoo-msft (commented):
@GabrielMedAlv, in order to fix the clang-format error. you can run clang-format command line tool from Apple terminal.

if not installed, you can grab it from homebrew. after that, do clang-format -i \<your file name\>
replace string inside angular brackets including brackets with the file name.

jwoo-msft (approved):

PR #8545, opened 2023-06-09
Use `makeCssClassName` to apply/remove CSS classes for Filtered Input.Choiceset

SuryanshPrabhat:
# Related Issues
1. Hosts might add unique identifiers to CSS class names during webpack/minification/build. To maintain the correct mapping of the classes which the SDK might add or remove on its end, `makeCssClassName` method in the HostConfig can be implemented by the host.
2. Scroll error indicator into view if present.
3. Better naming for CSS classes.
4. Input validation cues for Dynamically filtered choiceset.


anna-dingler (approved):

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -4914,8 +4918,12 @@ export class FilteredChoiceSet {
                 `ac-choiceSetInput-${this._choiceSetId}-choice-${id}`
             );
             if (nextHighlightedChoice) {
-                currentHighlightedChoice?.classList.remove("focused");
-                nextHighlightedChoice.classList.add("focused");
+                currentHighlightedChoice?.classList.remove(
```
jwoo-msft, line 4921: We will need to remove this constant literal and replace it with a variable. Then, we should link the variable to a relevant property to avoid any unexpected issues. By doing so, we can save a significant amount of effort in re-engineering.

jwoo-msft (commented):
https://github.com/microsoft/AdaptiveCards/issues/8556

jwoo-msft (approved):

SuryanshPrabhat:
Will merge PR post further testing with the team to see if any other fix needs to be done.

source/nodejs/adaptivecards/src/card-elements.ts
SuryanshPrabhat: Will resolve this conversation and merge PR post further testing with the team to check if any other fixes need to be made.

Issue #7160, opened 2022-03-22
[Designer] Running with `npm start` doesn't work correctly


paulcam206:
# Description

Running the Designer locally via `npm start` yields a broken experience.

## Repro

From a clean clone of our repo:
1. `cd source/nodejs`
2. `npm i && npx lerna bootstrap && npx lerna run build`
3. `cd adaptivecards-designer`
4. `npm start`

Designer fails to load complaining that `getAllDeclaredVersions()` isn't a function.

## Explanation

This is actually a somewhat interesting problem. Some context here around packaging:

* `adaptivecards` - contains the Adaptive Cards renderer
* `adaptivecards-designer` - contains the implementation of the Designer
* `adaptivecards-designer-app` - contains a stripped-down version of the Designer for embedding use where our defaults (e.g. Host Apps) aren't desirable (only mentioning this for completeness)

When `npm run start` is issued in the `adaptivecards-designer` directory, `webpack` is invoked in `dev` mode, which builds Designer stuff without optimizations, kicks off a dev server, and opens a browser tab pointing to it. `webpack` pulls things together based on configuration, including generating some HTML outputs:

https://github.com/microsoft/AdaptiveCards/blob/bcd3cd1f6470eb1f3596aa25da3d94c0af9bcef1/source/nodejs/adaptivecards-designer/webpack.config.js#L58-L63

Because `adaptivecards-designer` can be used in places that already depend on the `adaptivecards` package, `adaptivecards` is referenced as an external dependency:

https://github.com/microsoft/AdaptiveCards/blob/bcd3cd1f6470eb1f3596aa25da3d94c0af9bcef1/source/nodejs/adaptivecards-designer/webpack.config.js#L113-L130

This means that it's up to the consumer of the `adaptivecards-designer` package to satisfy the `adaptivecards` dependency. This is accomplished in the `npm run start` case by referencing `adaptivecards` like this:

https://github.com/microsoft/AdaptiveCards/blob/bcd3cd1f6470eb1f3596aa25da3d94c0af9bcef1/source/nodejs/adaptivecards-designer/index.html#L6

[...]

PR #8548, opened 2023-06-10
Added host container query param to designer site

Diegorro98:
# Description
This PR allows to set a host container by using a query param. This query param should be a number that represents the index on `CardDesigner._hostContainers` array (and the index at Host App Choice Picker at the designer's UI).

To do that I created a function that checks that the index is on the valid range and then applies it.

# How Verified

   No tests were added, I didn't know where are `CardDesigner` class tests.

I was able to test that the query param works with this url: [https://adaptivecards.io/designer?hostContainerIndex=1](https://adaptivecardsci.z5.web.core.windows.net/pr/8548/designer?hostContainerIndex=1) but I wasn't able to make work the designer because although I tried to set up the development environment with this [instructions](https://github.com/microsoft/AdaptiveCards/blob/main/source/nodejs/README.md) and this [workaround](https://github.com/microsoft/AdaptiveCards/issues/7160)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8548)


Diegorro98:
Closing this PR because I made a new PR (#8850) with more query params.

PR #8559, opened 2023-06-14
Render SVG images in WinUI 3

### LINKED ISSUE #8558: [Rendering] Can't render SVGs in WinUI 3
### Target Platforms

Other

### SDK Version

1.5

### Application Name

Dev Home

### Problem Description

Rendering an SVG in WinUI 3 throws an exception. In the code, there are references to SvgImageSource.Dispatcher(). In WinUI 3, Dispatcher() always returns null and we need to use DispatcherQueue() instead.
https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/migrate-to-windows-app-sdk/guides/threading#migrate-winrtresume_foreground-cwinrt

### Screenshots

_No response_

### Card JSON

```JSON
{
	"type": "AdaptiveCard",
	"version": "1.5",
	"body": [
		{
			"type": "Image",
			"backgroundColor": "#eeeeee",
			"url": "data:image/svg+xml;base64,PHN2ZyBoZWlnaHQ9IjEwMCIgd2lkdGg9IjI1MCI+CiAgPHBvbHlsaW5lIHBvaW50cz0iMjAsMjAgNDAsNDAgNjAsNjAiIHN0eWxlPSJmaWxsOm5vbmU7c3Ryb2tlOmJsYWNrO3N0cm9rZS13aWR0aDozIiAvPgo8L3N2Zz4="
		}
	]
}
```

### Sample Code Language

C#

### Sample Code

_No response_

krschau:
# Related Issue

Fixes #8558

# Description

Rendering an SVG in WinUI 3 is currently broken, since an access violation exception is thrown. WinUI 3 needs to use `SvgImageSource.DispatcherQueue()`, rather than UWP's `SvgImageSource.Dispatcher()`, which in WinUI 3 will always return `null`.
https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/migrate-to-windows-app-sdk/guides/threading#migrate-winrtresume_foreground-cwinrt

Also fixes an issue where the list of cards in AdaptiveCardsVisualizer wouldn't load.

# Sample Card

# How Verified

Verified with AdaptiveCards/samples/v1.5/Tests/Image.Svg.json card as well as other cards containing svgs in base64.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8559)

### DIFF
--- a/source/uwp/SharedRenderer/lib/AdaptiveImageRenderer.cpp
+++ b/source/uwp/SharedRenderer/lib/AdaptiveImageRenderer.cpp
@@ -35,6 +35,15 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
 
 namespace AdaptiveCards::Rendering::Xaml_Rendering
 {
+    auto inline GetDispatcher(winrt::SvgImageSource const &imageSource)
+    {
+#ifdef USE_WINUI3
+        return imageSource.DispatcherQueue();
+#else
+        return imageSource.Dispatcher();
+#endif
+    }
+
     //////////////////////////////////////////////////////////////////////////////////////////////////////////////////
     //
     // IMPORTANT! Methods below here are actually XamlBuilder methods. They're defined here because they're only used
@@ -599,11 +608,7 @@ namespace AdaptiveCards::Rendering::Xaml_Rendering
     winrt::fire_and_forget XamlBuilder::SetSvgUriSource(winrt::SvgImageSource const imageSource,
                                                         winrt::Uri const uri)
     {
-#ifdef USE_WINUI3
-        co_await wil::resume_foreground(imageSource.DispatcherQueue());
-#else
-        co_await winrt::resume_foreground(imageSource.Dispatcher());
-#endif
+        co_await wil::resume_foreground(GetDispatcher(imageSource));
         imageSource.UriSource(uri);
     }
 
@@ -614,11 +619,7 @@ namespace AdaptiveCards::Rendering::Xaml_Rendering
     {
         auto weakThis = this->get_weak();
 
-#ifdef USE_WINUI3
-        co_await wil::resume_foreground(imageSource.DispatcherQueue());
-#else
-        co_await winrt::resume_foreground(imageSource.Dispatcher());
-#endif
+        co_await wil::resume_foreground(GetDispatcher(imageSource));
         auto setSourceOperation = imageSource.SetSourceAsync(stream);
 
         setSourceOperation.Completed(
@@ -809,11 +810,7 @@ namespace AdaptiveCards::Rendering::Xaml_Rendering
                                                                       double const imageSize,
                                                                       bool const dropIfUnset)
     {
-#ifdef USE_WINUI3
-        co_await wil::resume_foreground(imageSource.DispatcherQueue());
-#else
-        co_await winrt::resume_foreground(imageSource.Dispatcher());
-#endif
+        co_await wil::resume_foreground(GetDispatcher(imageSource));
         auto currentSize = imageSource.RasterizePixelHeight();
         bool sizeIsUnset = isinf(currentSize);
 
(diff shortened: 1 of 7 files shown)


source/uwp/SharedRenderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -599,7 +603,11 @@ namespace AdaptiveCards::Rendering::Xaml_Rendering
     winrt::fire_and_forget XamlBuilder::SetSvgUriSource(winrt::SvgImageSource const imageSource,
                                                         winrt::Uri const uri)
     {
+#ifdef USE_WINUI3
+        co_await wil::resume_foreground(imageSource.DispatcherQueue());
+#else
         co_await winrt::resume_foreground(imageSource.Dispatcher());
+#endif
```
jwoo-msft, line 610: Can these lines of code be consolidated into a routine? That way, if we need to make changes, they can be made in one location.

krschau: @jwoo-msft Coroutines are not my strong suit... do you know what this would look like? I played around a bit but am not sure the correct way to do this with the co_awaits.

krschau: I think we still need to detect running winui3, since we still need to use Dispatcher vs DispatcherQueue (DQ is not available on uwp [link](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.media.imaging.svgimagesource?view=winrt-22621)).

As for using WIL for both, UWP really doesn't seem to want to find that wil/cppwinrt_helpers.h file. I included the WIL nuget package, but still nothing. If we have to do the winui3 check anyway, it may not be worth the trouble.

jwoo-msft: @krschau, I'm pretty sure you've already tried. what kind of issue did you ran when the wrapper routine returns a co-await able object, and co-await on the returned object?

jwoo-msft: We can also create a macro or an inline function that takes no arguments. As we all agree, it is important to keep it in one place to avoid unnecessary time spent on maintenance. In this situation, using a macro seems appropriate. It is simply a single line of code, and it does not affect debuggability since it involves an API call to an external library. ACResumeForeGround();

krschau: Originally I was getting something like [this](https://devblogs.microsoft.com/oldnewthing/20210809-00/?p=105539) but the suggested solution wasn't working for me. I think I'm getting farther this time...

krschau: @jwoo-msft Very weird -- I tried adding this method and co_await-ing it in the rest of the code. But with this change, I get a `System.AccessViolationException: 'Attempted to read or write protected memory. This is often an indication that other memory is corrupt.'` when rendering a card with an SVG.

```
inline auto ACResumeForeground(winrt::SvgImageSource const imageSource)
{
#ifdef USE_WINUI3
    return wil::resume_foreground(imageSource.DispatcherQueue());
#else
    return winrt::resume_foreground(imageSource.Dispatcher());
#endif
}
```

jwoo-msft: Today is Windows Learning day :-) I think the quick solution here is using the macro. we just let the preprocess expand it. 
for the inline, what if the compiler did inline expansion, and wonder what it would look like after the expansion.

jwoo-msft: @krschau 
you can also try something like this.
```
    auto inline GetDispatcher(winrt::SvgImageSource const &imageSource)
    {
	#ifdef USE_WINUI3
		return imageSource.DispatcherQueue();
	#else
		return imageSource.Dispatcher();
	#endif
    }
```
and call the inline where resume_foreground is called
```
co_await winrt::resume_foreground(GetDispatcher(imageSource));
```
this works.

jwoo-msft (approved):

krschau:
@jwoo-msft Do you know what's up with the android-ci pipeline? I've re-run it a few times this morning and it keeps failing.

jwoo-msft:
> @jwoo-msft Do you know what's up with the android-ci pipeline? I've re-run it a few times this morning and it keeps failing.

let me check on it

jwoo-msft:
@krschau, go ahead and force merge if you'd like. It has nothing to do with your change.

krschau:
@jwoo-msft I don't have authority to force merge, would you mind doing it?

PR #8564, opened 2023-06-16
[WinUI3 Visualizer] Add if block for high contrast changed event

### LINKED ISSUE #8565: [WinUI3 Visualizer] App does not load when launched
# Platform

* UWP

# Issue

The WinUI3 visualizer is throwing a `COMException` on launch. The app never loads because of this exception.

image

anna-dingler:
# Related Issue

Fixes #8565

# Description

We cannot call AccessibilitySettings.HighContrastChanged with WinUI3 because it does not work without a core window.

Added an if block to resolve this error:
image

# Sample Card

N/A

# How Verified

Verified manually in the WinUI3 and System.XAML visualizers

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8564)

### DIFF
--- a/source/uwp/SharedVisualizer/ViewModel/HostConfigEditorViewModel.cs
+++ b/source/uwp/SharedVisualizer/ViewModel/HostConfigEditorViewModel.cs
@@ -8,7 +8,6 @@ using AdaptiveCards.Rendering.Uwp;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Windows.Storage;
 using Windows.UI.ViewManagement;
--- a/source/uwp/SharedVisualizer/ViewModel/HostConfigEditorViewModel.cs
+++ b/source/uwp/SharedVisualizer/ViewModel/HostConfigEditorViewModel.cs
@@ -8,6 +8,7 @@ using AdaptiveCards.Rendering.Uwp;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Windows.Storage;
 using Windows.UI.ViewManagement;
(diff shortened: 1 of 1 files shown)


source/uwp/SharedVisualizer/ViewModel/HostConfigEditorViewModel.cs
```diff
@@ -17,7 +17,10 @@ namespace AdaptiveCardVisualizer.ViewModel
     public class HostConfigEditorViewModel : GenericDocumentViewModel
     {
         private HostConfigEditorViewModel(MainPageViewModel mainPageViewModel) : base(mainPageViewModel) {
+#if !USE_WINUI3
```
jwoo-msft, line 20: wouldn't it be better to check if core windows exist rather than disabling this feature for all WinUI3 use cases?

anna-dingler: We could do either? I don't think the core window will ever exist in this scenario

anna-dingler: As a side note, I found a work around for this in the RNW repo, but I'm not sure if its worth implementing for a "nice to have" feature. High contrast selection still works on launch, and the contrast handling is the responsibility of the host app, not the renderer

krschau: I think using the #if vs null checking CoreWindow comes down to personal preference. The fix on the WinUI side would be removing the underlying reliance on a CoreWindow, rather than putting in a CoreWindow -- which is to say, using the null check wouldn't get us the advantage of suddenly starting to work one day when we update to a newer version of WinUI.

jwoo-msft: Why will a CoreWindow never exist in this previously mentioned scenario?

anna-dingler: Its unsupported: https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/desktop-to-uwp-supported-api

And the event in question is also listed as unsupported: https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/desktop-to-uwp-supported-api#events

jwoo-msft (approved):

PR #8611, opened 2023-07-11
Add support for .NET 6.0 to .NET nugets 

almedina-ms:
# Related Issue

Enables the creation of AdaptiveCards dotnet packages with support for .NET 6.0
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/AdaptiveCards/pull/8611)


source/dotnet/Library/AdaptiveCards.Net6/AdaptiveCards.csproj
```diff
@@ -0,0 +1,66 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net6.0</TargetFrameworks>
```
paulcam206, line 3: I'm curious -- did you try adding `net6.0` to the existing .csproj's `TargetFrameworks`? It would be nice if we could ship this as an additive thing rather than a separate package.

almedina-ms: we're releasing in the same package, we're just adding both dlls to the same nuget package

paulcam206: if it's _possible_ to get `net6.0` building in the existing project, it would be nice to not have to duplicate all of this scaffolding. I'll resolve this and sign off, but please do give it a shot 👍🏼

paulcam206 (approved):

PR #8704, opened 2023-10-10
[Site] Set HSTS header

paulcam206:
Set [HTTP `Strict-Transport-Security`](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) header for website.

### DIFF
--- a/source/nodejs/adaptivecards-site/themes/adaptivecards/source/web.config
+++ b/source/nodejs/adaptivecards-site/themes/adaptivecards/source/web.config
@@ -4,6 +4,7 @@
 		<httpProtocol>
 			<customHeaders>
 				<add name="X-Content-Type-Options" value="nosniff" />
+				<add name="Strict-Transport-Security" value="max-age=63113852; includeSubDomains; preload" />
 			</customHeaders>
 		</httpProtocol>
 		<staticContent>


anna-dingler (approved):

source/nodejs/adaptivecards-site/themes/adaptivecards/source/web.config
```diff
@@ -4,6 +4,7 @@
 		<httpProtocol>
 			<customHeaders>
 				<add name="X-Content-Type-Options" value="nosniff" />
+				<add name="Strict-Transport-Security" value="max-age=63113852; includeSubDomains; preload" />
```
anna-dingler, line 7: Would this have any impact on the PWA portion?

jwoo-msft (approved):

source/nodejs/adaptivecards-site/themes/adaptivecards/source/web.config
paulcam206: I couldn't find any indication that it would make a difference... the short version is that the presence of this header means that requests that _might_ be made via plain HTTP will get transparently upgraded to HTTPS without even attempting to contact the server, so this should theoretically only impact the PWA if it's pulling from HTTP URLs (and even then, it _should_ Just Work™️)

PR #8746, opened 2023-11-08
Remove unnecessary `aria-required` property from container element

paulcam206:
# Description

The `aria-required` attribute is only valid on elements (or elements with a role matching an element) that have a `required` HTML property. To specify `aria-required` on other elements violates WCAG rules.

# How Verified

Local build/test and Accessibility Insights verification

### DIFF
--- a/source/nodejs/adaptivecards/src/card-elements.ts
+++ b/source/nodejs/adaptivecards/src/card-elements.ts
@@ -3557,7 +3557,6 @@ export abstract class Input extends CardElement implements IInput {
             this._renderedInputControlElement.style.minWidth = "0px";
 
             if (this.isNullable && this.isRequired) {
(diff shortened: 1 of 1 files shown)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -3557,7 +3557,6 @@ export abstract class Input extends CardElement implements IInput {
             this._renderedInputControlElement.style.minWidth = "0px";
 
             if (this.isNullable && this.isRequired) {
-                this._renderedInputControlElement.setAttribute("aria-required", "true");
```
anna-dingler, line 3560: Does this still announce that it's required?

paulcam206: narrator reads it the same way :)

anna-dingler (approved):

PR #8753, opened 2023-11-14
Fix site schema explorer render version

paulcam206:
# Description

This was a pretty interesting issue to track down... the [`Input.Text`](https://adaptivecards.io/explorer/Input.Text.html) `inlineAction` example's inline submit button didn't have its tooltip applied properly. However, the sample example worked just fine in the designer.

Turns out the examples are rendering with the version specified in the sample itself. I updated all of the samples referencing `tooltip` to be at least version `1.5`.

Future reference for anyone wanting to answer a question like "what is the version of all samples with the word 'tooltip' in them", here's something useful in WSL (you'll need to have `ripgrep` and `jq` available):

```sh
# from the samples/ directory
rg --files-with-matches tooltip | xargs -I _ sh -c "echo -n '_: ' && cat _ | jq .version"
```

sample output for above:

image

# How Verified

Local build/devtools

### DIFF
--- a/samples/v1.1/Elements/Column.SelectAction.json
+++ b/samples/v1.1/Elements/Column.SelectAction.json
@@ -1,7 +1,7 @@
 {
 	"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
 	"type": "AdaptiveCard",
-	"version": "1.1",
+	"version": "1.5",
 	"body": [
 		{
 			"type": "TextBlock",
--- a/samples/v1.1/Elements/ColumnSet.SelectAction.json
+++ b/samples/v1.1/Elements/ColumnSet.SelectAction.json
@@ -1,6 +1,6 @@
 {
 	"type": "AdaptiveCard",
-	"version": "1.1",
+	"version": "1.5",
 	"body": [
 		{
 			"type": "TextBlock",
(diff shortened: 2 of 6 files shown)


jwoo-msft (approved):

source/nodejs/adaptivecards-site/themes/adaptivecards/source/js/script.js
```diff
@@ -458,16 +458,14 @@ $(function () {
...
 		var renderedCard;
 
-		var version = json.version.split(".");
-
 		if (dataJson) {
 			var template = new ACData.Template(json);
 			adaptiveCard.parse(template.expand({
 				$root: dataJson
-			}), new AdaptiveCards.SerializationContext(new AdaptiveCards.Version(version[0], version[1])));
+			}), new AdaptiveCards.SerializationContext(AdaptiveCards.Versions.Latest));
```
anna-dingler, line 465: What is the version specified in the sample? I think we parse for the version specified here.

anna-dingler: image

We could go either way, but updating the version in the sample would fix this I think

paulcam206: good point -- less scary fix too :)

paulcam206: hmm... looking at this a bit more, we have a few samples that need updates for version...

jwoo-msft (approved):

Issue #8783, opened 2023-12-22
[Android][Rendering] Memory leak issue while rendering Textblock render


trickcharan:
### Target Platforms

Android

### SDK Version

3.1.1

### Application Name

Webex

### Problem Description

The issue at hand involves the rendering of an adaptive card within a fragment or activity in Android. As expected, once we navigate away from the fragment or activity, all memory associated with the adaptive card should be released. However, this isn't happening as it should. Instead, the text block renderer in the adaptive card is retaining the reference to the fragment or activity(code reference: https://github.com/microsoft/AdaptiveCards/blob/76cec36cf063009260f3dc5642d29a7624255ea5/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TextBlockRenderer.java#L239C6-L256C6), preventing the full release of memory. This memory leak could potentially lead to performance issues in the application.

In addition to this, we've identified another potential memory leak in the following places 

1. [DateInputHandler.java](https://github.com/microsoft/AdaptiveCards/blob/76cec36cf063009260f3dc5642d29a7624255ea5/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/DateInputHandler.java#L24)
2. [TimeInputHandler.java](https://github.com/microsoft/AdaptiveCards/blob/76cec36cf063009260f3dc5642d29a7624255ea5/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/TimeInputHandler.java#L26)

The problem arises from the storage of the fragment manager from the application which can potentially cause memory leaks.

### Screenshots

<img width="486" alt="Screenshot 2023-12-22 at 10 35 00 AM" src="https://github.com/microsoft/AdaptiveCards/assets/29520819/adf7ae79-6b6a-4733-ac20-ee725c324a28">

### Card JSON

```JSON
{
"$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
"type": "AdaptiveCard",
"version": "1.0",
"body": [
{
"type": "TextBlock",
"text": "This is some text",
"size": "large"
},
{
"type": "TextBlock",
"text": "It doesn't wrap by default",
"weight": "bolder"
},
{
"type": "TextBlock",
"text": "So set wrap to true if you plan on showing a paragraph of text",
"wrap": true
},
{
"type": "TextBlock",
"text": "You can also use maxLines to prevent it from getting out of hand",
"wrap": true,
"maxLines": 2
},
{
[...]

PR #8784, opened 2023-12-22
Fixed memory leak issue while rendering Text view (TextBlockRenderer)

trickcharan:
#8783 
# Description
 The AccessibilityManager in text block rendere was retaining a strong reference to the TextView even when we navigate away from the current fragment or activity. This was causing a memory leak as the memory associated with the TextView wasn't being properly released.  
 
To fix the issue, a WeakReference to the TextView was introduced. This allows the garbage collector to reclaim memory from the TextView when it is no longer in use, preventing a memory leak. Also, a mechanism to remove the listener when the TextView is no longer in memory or when it is detached from the window was added, ensuring that we don't have any dangling listeners causing unnecessary memory retention.

In addition to this, we've identified another potential memory leak in the following places 

1. [DateInputHandler.java](https://github.com/microsoft/AdaptiveCards/blob/76cec36cf063009260f3dc5642d29a7624255ea5/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/DateInputHandler.java#L24)
2. [TimeInputHandler.java](https://github.com/microsoft/AdaptiveCards/blob/76cec36cf063009260f3dc5642d29a7624255ea5/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/inputhandler/TimeInputHandler.java#L26)

The problem arises from the storage of the fragment manager from the application which can potentially cause memory leaks. To mitigate this, the code has been modified to accept the fragment manager as a weak reference. This change will prevent unnecessary memory retention, further enhancing the overall performance of the application

# Sample Card
https://adaptivecards.io/explorer/TextBlock.html

# How Verified

1. The fix was first verified by building the library, ensuring that the changes made didn't cause any issues during this process.
3. Following that, the updated library was integrated into the application and tested. The application functioned as expected, confirming the successful resolution of the memory leak issue.

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/DateInputRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/input/DateInputRenderer.java
@@ -9,6 +9,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.EditText;
 
+import java.lang.ref.WeakReference;
 import java.text.DateFormat;
 
 import io.adaptivecards.objectmodel.BaseCardElement;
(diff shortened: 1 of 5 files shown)


trickcharan:
microsoft-github-policy-service agree

trickcharan:
@microsoft-github-policy-service agree

jwoo-msft:
we don't service android anymore; can your create this PR to [AdaptiveCards-Mobile](https://github.com/microsoft/AdaptiveCards-Mobile)?

trickcharan:
Sure Will raise there.

PR #8806, opened 2024-01-29
[UWP] Add new property to the renderer to remove margins from selectAction

anna-dingler:
# Related Issue

https://dev.azure.com/microsoft/OS/_workitems/edit/46854074/

# Description

For selectAction, we add an additional margin to the button. This causes the hover state to have two different colors, and it looks overdrawn. 

To resolve this, I added a boolean to the renderer `AddSelectActionMargin` that will determine if we should add the additional margin. The boolean defaults to `true`.
# Sample Card

N/A

# How Verified

Verified manually with the visualizer.


source/uwp/winui3/Renderer/AdaptiveCardRenderer.vcxproj
```diff
@@ -124,6 +124,7 @@
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
     <ClCompile>
       <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <DebugInformationFormat Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">ProgramDatabase</DebugInformationFormat>
```
anna-dingler, line 127: This change resolved an error that I was seeing, but can remove if we don't need it in main

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
```diff
@@ -54,6 +54,8 @@ namespace AdaptiveCards
                 String OverflowButtonText;
                 String OverflowButtonAccessibilityText;
 
+                Boolean AddSelectActionMargin;
```
anna-dingler, line 57: General note: I tried looking for other ways we could override this behavior (custom styling to override the onHover foreground and background colors?), but didn't see anything. Open to any ideas here :)

source/uwp/SharedRenderer/lib/AdaptiveCardRendererComponent.cpp
```diff
@@ -78,6 +78,16 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
         GetHostConfig()->OverflowMaxActions = overflowMaxActions;
     }
 
+    bool AdaptiveCardRenderer::AddSelectActionMargin()
+    {
+        return GetHostConfig()->AddSelectActionMargin;
+    }
+
+    void AdaptiveCardRenderer::AddSelectActionMargin(bool addSelectActionMargin)
```
jwoo-msft, line 86: do we need this method since the value is read from host config?

anna-dingler: Yes, I believe so. The usage is like this:

`_renderer.AddSelectActionMargin = false;`

I'm actually not sure why we store the values in the hostConfig for this type of property. Maybe because we don't have access to renderer properties when we construct the objects? @paulcam206

jwoo-msft (approved):

source/uwp/Renderer/idl/AdaptiveCards.Rendering.Uwp.idl
anna-dingler: Resolving reason: I investigated using custom styling instead, but this does not work because programmatically set styles take precedence over resource dictionaries.

source/uwp/SharedRenderer/lib/AdaptiveCardRendererComponent.cpp
beervoley: Do we maintain parity of HostConfig capabilities between the platforms today? Because if we do, adding this property to HostConfig would break that. @paulcam206

anna-dingler: @jwoo-msft Do you have any additional context here?

From my understanding, this doesn't maintain parity; however, it isn't an issue for the other platfroms because it wasn't added to the _shared_ host config

anna-dingler: Somewhat related - do we have a plan for decoupling from the mobile renderers?

jwoo-msft: we don't put much emphasis on parities across platforms anymore. We only support UWP/WinRT and Type Script renderers, and we have allowed them to diverge.

jwoo-msft: > Somewhat related - do we have a plan for decoupling from the mobile renderers?

the coupling doesn't exist today. Mobile renderers have their own copy of shared code.

PR #8835, opened 2024-02-26
[NodeJS] Update hashing algorithm for webpack

paulcam206:
Building with NodeJS >16 yields errors complaining about deprecated hashing functions (`ERR_OSSL_EVP_UNSUPPORTED`). This is a known and intentional change in NodeJS to prevent using insecure hashing functions. It's a little overzealous in *this* case because the hashing function is just being used to break up chunks of code, but rather than having to opt everything into allowing insecure hashing functions, it's probably better to just make Webpack use a more recent function (i.e. [`xxhash64`](https://webpack.js.org/configuration/output/#outputhashfunction), the default in Webpack's next version).

### DIFF
--- a/source/nodejs/adaptivecards-aaf-testapp/webpack.config.js
+++ b/source/nodejs/adaptivecards-aaf-testapp/webpack.config.js
@@ -3,48 +3,48 @@ const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 
 module.exports = (env, argv) => {
-	const mode = argv.mode || 'development';
-	const devMode = mode === "development";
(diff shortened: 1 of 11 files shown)


source/nodejs/adaptivecards-aaf-testapp/webpack.config.js
```diff
@@ -15,6 +15,7 @@ module.exports = (env, argv) => {
 		},
 		output: {
 			path: path.resolve(__dirname, "dist"),
+            hashFunction: "xxhash64",
 			filename: devMode ? "[name].js" : "[name].min.js",
```
paulcam206, line 19: ugh... need to fix tabs/spaces?

jwoo-msft (approved):

PR #8838, opened 2024-02-27
[NodeJS] Add `ImageSet` class to enable styling

paulcam206:
Simple addition of class to `ImageSet` element. As-is, this doesn't change anything, but it _does_ allow host applications to apply special styling to `ImageSet`s without a bunch of effort.


jwoo-msft (approved):

beervoley (approved):
:shipit:

PR #8839, opened 2024-02-28
[XAML] Fix for exception thrown in `WholeItemsPanel`

paulcam206:
Looks like a simple mistake in the conversion to cppwinrt (`as` vs. `try_as`). The converted version would throw instead of quietly skipping the check...

**original**
https://github.com/microsoft/AdaptiveCards/blob/d6084118fe516aa1793370bda7378b7a7ccd1616/source/uwp/Renderer/lib/WholeItemsPanel.cpp#L469-L478

**converted**
https://github.com/microsoft/AdaptiveCards/blob/1dcef8c69715bd56cc4384ab2b1a80906e696922/source/uwp/Renderer/lib/WholeItemsPanel.cpp#L357-L363


anna-dingler (approved):

jwoo-msft (approved):

beervoley:
https://github.com/microsoft/AdaptiveCards/pull/8841 - this will have to go in before this PR :)

PR #8845, opened 2024-03-02
[NodeJS] Add `grid` `ImageSet` style

paulcam206:
Adding a new `ImageSet` style to allow for a cleaner presentation. Also improve default `ImageSet` rendering.

## Old `ImageSet` render (default style)

image

## New `ImageSet` render (default style)

image
(a small tweak -- just added vertical padding)

### Schema
```json
{
  "type": "ImageSet",
  "style": "grid",
  "images": [ ]
}
```

#### Small
image

#### Medium
image

#### Large
image

#### Large (rtl)
image


jwoo-msft:
@paulcam206, can we create a tag for new features to be attached to the commit, so that we can easily pull those features when we want to bring each platform to parity?

source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -2518,38 +2501,64 @@ export class ImageSet extends CardElementContainer {
         let element: HTMLElement | undefined = undefined;
 
         if (this._images.length > 0) {
+            const imageSetIsGrid = (this.presentationStyle === Enums.ImageSetPresentationStyle.Grid);
```
jwoo-msft, line 2504: do we need version check here since renderer can work with different version of Enums?

jwoo-msft (approved):

paulcam206:
> @paulcam206, Do you think we need to create a tag for new features to be attached to the commit, so that we can easily pull those features when we want to bring each platform to parity?

good idea, I created the [`NodeJS-2403`](https://github.com/microsoft/AdaptiveCards/pulls?q=label%3ANodeJS-2403) label to track.

source/nodejs/adaptivecards/src/card-elements.ts
paulcam206: I believe that should be enforced here:

https://github.com/microsoft/AdaptiveCards/blob/67f7109761bd6b4fdf565b121fc7c64472f3d0bc/source/nodejs/adaptivecards/src/card-elements.ts#L2472-L2477

Issue #8627, opened 2023-07-25
[Feature Request] Distribute ESM alongside CJS, add "modules" entry.


dzearing:
### Problem Statement

The `adaptivecards` package can't load in node. Try creating a test package with this dependency, run node, run `require('adaptivecards')`. The CJS you distribute can not import the ESM-only peer dependency "swiper". This causes numerous downstream consequences:

1. You can never server render.
2. You can't unit test it in an ESM-only environment.
3. You get no tree-shaking from the bundler.
4. Internal projects to speed up inner loops are blocked and must work around this package.

There's an existing issue: https://github.com/microsoft/AdaptiveCards/issues/6765. Hasn't been touched in a year and change.

We tried using this in a major M365 project, and ran into major blockers. Please address this with high priority.

### Proposed solution

Like most browser packages in npm, you should be distributing ESM and pointing to your entry with the `module` key in your package.json. This prevents the scenario where your library can't load ESM dependencies. Even better, use an `exports` map as well which has priority over the legacy ways of declaring entries.

```js
"exports": {
  ".": {
    "types": "./path/to/any/index.d.ts",
    "require": "./path/to/cjs/index.js",
    "import": "./path/to/esm/index.js"
  }
}
```

AnilGayakwad:
Hi Team, 
Any updates on this issue? It's preventing us from moving forward with implementation. It's a complete roadblock for us. 
We are working on creating a user interface on https://nextjs.org/ with all latest version.

Issue #8319, opened 2023-02-22
Must use import to load ES Module: swiper.esm.js


mectors:
# Platform
* JavaScript

# Author or host
Yes

# Version of SDK
    "adaptive-expressions": "^4.19.1",
    "adaptivecards": "^2.11.2",
    "adaptivecards-templating": "^2.3.1",
    "swiper": "^9.0.5"

# Issue
I am trying for adaptivecards templates to work consistently and I get really strange issues with some cards. So I just tried to make a very simple js script to test different templates with data. Unfortunately I get problems with carousel.js trying to use swiper which for some reason I had to add manually.

internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: = ./adaptivetemplating/node_modules/swiper/swiper.esm.js
require() of ES modules is not supported.
require() of ./adaptivetemplating/node_modules/swiper/swiper.esm.js from ./adaptivetemplating/node_modules/adaptivecards/lib/carousel.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename swiper.esm.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from ./adaptivetemplating/node_modules/swiper/package.json.

    at new NodeError (internal/errors.js:322:7)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1131:13)
    at Module.load (internal/modules/cjs/loader.js:979:32)
    at Function.Module._load (internal/modules/cjs/loader.js:819:12)
    at Module.require (internal/modules/cjs/loader.js:1003:19)
    at require (internal/modules/cjs/helpers.js:107:18)
    at Object.<anonymous> (./adaptivetemplating/node_modules/adaptivecards/lib/carousel.js:42:16)
    at Module._compile (internal/modules/cjs/loader.js:1114:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
    at Module.load (internal/modules/cjs/loader.js:979:32) {
  code: 'ERR_REQUIRE_ESM'
}

The code is simple and although not fully tested always gives the swiper error:
import * as ACData from "adaptivecards-templating";
import * as Adaptive from "adaptivecards";
var fs = require("fs");

console.log("template:",process.argv[2])
console.log("data:",process.argv[3])

[...]

Issue #7983, opened 2022-10-24
[Unit Tests] Configure Jest to allow `swiper` module


anna-dingler:
# Platform

* Unit test project

# Issue

Jest does not allow `swiper` to be present in `node_modules` since it is an ES Module. We need to configure the unit-test project to allow `swiper` so that we can update the project to v3.x.x of adaptivecards

Potentially helpful links:
https://jestjs.io/docs/29.0/ecmascript-modules
https://jestjs.io/docs/29.0/configuration
https://nodejs.org/api/esm.html#esm_enabling
https://kulshekhar.github.io/ts-jest/docs/guides/esm-support/

PR #8871, opened 2024-04-08
[NodeJS] Updates to streamline package size

paulcam206:
# Description

There's a lot going on in this PR, but unfortunately there's no easy way to break some of this stuff down. The main goal is to reduce the size of the `adaptivecards` bundle. Today, version 3.0.2 of `adaptivecards.min.js` is 442428 bytes, and contains a fair amount of effectively dead code pulled in from `swiper`. The non-minified version contains code for things like effects that we don't use:

```js
/***/ "./node_modules/swiper/modules/effect-cube/effect-cube.js":
/*!****************************************************************!*\
  !*** ./node_modules/swiper/modules/effect-cube/effect-cube.js ***!
  \****************************************************************/
/***/ ((__unused_webpack___webpack_module__, __webpack_exports__, __webpack_require__) => {

   /// 200 lines of code implementing the "cube" slide transition style

});
```

`webpack` is capable of stripping out unused or unreachable code via a process called ["tree shaking"](https://webpack.js.org/guides/tree-shaking/), but there are some limitations. The main limitation in our case is that since the `require` directive can lead to side-effects, `webpack` can't strip out `require`d code... and as currently configured, we're feeding ES5 JS (which doesn't support `import`/`export` natively) in CommonJS format to `webpack`.

The fix is to build to more modern ECMAScript up until the part where the final bundle is being emitted.

This of course leads to follow-on effects where different parts of the project need to be updated or configured to consume this newer style of JS. As part of this, the project needed to be moved over to using [`npm` Workspaces](https://docs.npmjs.com/cli/v10/using-npm/workspaces). A follow-on benefit of this is that we no longer need to run `npx lerna bootstrap` to build. Now you just:
1. `cd source/nodejs`
2. `npm i`
3. `npm run build` or `npm run release` or whatever :)

**Also of note: Node v14 is ancient and out of support at this point, so I've done development against Node 20, the most recent LTS version. You will likely need to update in order to build properly.**

## Summary per-file

- `.gitignore`: the `.nx` directory is a build cache - don't want that checked in
- `.pipelines/*`: updates to account for new build commands
- `.husky/pre-commit`: we weren't utilizing husky hooks any more, so might as well remove them
- `.npmrc`: change default for how npm hoists packages to more closely align with our monorepo structure
- various `package.json` and `package-lock.json`s: updated versions of various dependencies
  - of note, the root `package.json` had various `clang-format`-related stuff removed, as that is no longer in use, and shouldn't have been there in the first place
- various `tsconfig` files:
  - explicitly reference types
- various `webpack.config.js` files:
  - update file paths as needed for new `node_modules` structure
- `source/nodejs/adaptivecards-react-testapp`: remove unused test app
- `adaptivecards-site`:
  - `/package.json`:
    - update build structure to pull `workbox` libraries in using a more official mechanism
  - `themes/adaptivecards/scripts/generator-adaptiveassets.js`
    - add optional debug spew
    - clean up implementation a bit to make maintenance easier
    - add mechanism to deal with new `node_modules` structure
- `adaptivecards-templating/src/json-schema-card.ts`
  - remove `JSONSchemaCardTime` as the mechanism is unused (and non-standard)
- `adaptivecards`:
  - `/package.json`: updated `swiper` dependency
  - `*carousel*`: adapt to updated `swiper` dependency
  - `utils.ts`: fix for node version update
- `lerna.json`: fixups to use npm workspaces
- `nx.json`: generated config for `nx` (a caching build accelerator used by lerna)


paulcam206:
Just a heads up -- I know that the `adaptivecards` version needs to be bumped... I'll add a commit to do that once I work out the build kinks :)

paulcam206:
> Just a heads up -- I know that the `adaptivecards` version needs to be bumped... I'll add a commit to do that once I work out the build kinks :)

AC version is bumped.

paulcam206:
@jwoo-msft / @anna-dingler / @almedina-ms -- this should be ready to review now

.pipelines/js-release.yml
```diff
@@ -26,37 +26,21 @@ parameters:
...
-  displayName: 'Bash - npm install -g npm 8.11.0'
-
-- bash: |
-   npm ci 
+   npm ci --maxsockets=1
   workingDirectory: source/nodejs
   displayName: 'Bash - npm ci'
 
-- bash: |
-   npx lerna bootstrap --ci
```
anna-dingler, line 46: We don't need to bootstrap anymore?

anna-dingler: Oh I see, we are using npm now isntead of lerna?

anna-dingler:
General question - Seems like these changes might help resolve the following issues?
https://github.com/microsoft/AdaptiveCards/issues/8627
https://github.com/microsoft/AdaptiveCards/issues/8319
https://github.com/microsoft/AdaptiveCards/issues/7983

source/nodejs/README.md
```diff
@@ -13,13 +13,16 @@ This directory contains all the JS packages for adaptive cards:
 Removed JS packages:
 * `adaptivecards-fabric` - A set of Fabric UI extensions for the Adaptive Cards JS renderer. This package has been deprecated. Please migrate to [`adaptivecards-fluentui`](https://www.npmjs.com/package/adaptivecards-fluentui). The source code is available in [21.09](https://github.com/microsoft/AdaptiveCards/releases/tag/21.09) or before releases.
 
+## Authenticate with npm registry
+
+1. `vsts-npm-auth -config .npmrc` in `source/nodejs`
```
anna-dingler, line 18: Are the new auth requirements part of the switch to npm?

source/nodejs/adaptivecards-controls/package.json
```diff
@@ -7,6 +7,7 @@
 	"main": "lib/adaptivecards-controls.js",
 	"types": "lib/adaptivecards-controls.d.ts",
 	"style": "dist/adaptivecards-controls.css",
+	"sideEffects": false,
```
anna-dingler, line 10: What does this do?

source/nodejs/adaptivecards-site/themes/adaptivecards/scripts/generator-adaptiveassets.js
```diff
@@ -2,169 +2,203 @@
...
-  "node_modules/adaptive-expressions/lib/*.*",
-  "node_modules/adaptivecards-designer/dist/*.*",
-  "node_modules/adaptivecards-templating/dist/*.*",
-  "node_modules/@fortawesome/fontawesome-free/css/all.min.css",
-  "node_modules/@fortawesome/fontawesome-free/webfonts/*.*",
-  "node_modules/highlightjs/highlight.pack.min.js",
-  "node_modules/highlightjs/styles/default.css",
-  "node_modules/jquery/dist/jquery.min.js",
-  "node_modules/markdown-it/dist/markdown-it.min.js"
+    "node_modules/adaptive-expressions/lib/*.*",
```
anna-dingler, line 24: Were some of these unused?

source/nodejs/lerna.json
```diff
@@ -1,16 +1,15 @@
 {
 	"version": "independent",
 	"useNx": true,
-	"packages": [
```
anna-dingler, line 4: What does this do?

source/nodejs/package.json
```diff
@@ -29,63 +41,28 @@
...
 		"rimraf": "^3.0.2",
 		"sass": "^1.43.4",
 		"style-loader": "^3.3.1",
-		"ts-jest": "^27.0.7",
-		"ts-loader": "^9.2.6",
-		"typescript": "^4.4.4",
+		"ts-jest": "^29.1.2",
+		"ts-loader": "^9.5.1",
+		"typescript": "^5.4.2",
+        "uuid": "^9",
```
anna-dingler, line 59: spacing?

.pipelines/js-release.yml
paulcam206: that's correct -- I updated the README, but I should've noted that in the description above as well... I'll update that now. :)

source/nodejs/README.md
paulcam206: We're required to use our internal artifact feed with public `npm` as an upstream (hit me up for a link if you want more info). This work should have been done some time ago, but it wasn't until I started messing with things that CI tooling started complaining. :)

source/nodejs/adaptivecards-controls/package.json
paulcam206: You can find docs [here](https://webpack.js.org/guides/tree-shaking/#clarifying-tree-shaking-and-sideeffects), but the short version is that it tells webpack that simply loading the `adaptivecards-controls` library doesn't cause other side effects (e.g. doing `require("adaptivecards-controls");` in a `<script>` block doesn't create a bunch of DOM elements or override other things that have already been loaded). This allows webpack to better reason about what pieces of JS can be lazy loaded as well as which pieces of JS can be elided altogether.

source/nodejs/adaptivecards-site/themes/adaptivecards/scripts/generator-adaptiveassets.js
paulcam206: they weren't unused, but part of the switch to `npm workspaces` shuffled around where some of the node packages were placed on disk (e.g. instead of `source/nodejs/adaptivecards-site/node_modules/adaptivecards/...` the package is at `source/nodejs/node_modules/adaptivecards/...`). these types of changes are _usually_ transparent to tooling (node walks up parent directories to the project root when trying to resolve packages), but since we have our own hand-rolled stuff for the site, I needed to make these logic changes -- if you look a bit further down, you'll see that we are still including these :)

https://github.com/microsoft/AdaptiveCards/blob/02c2427da2e99f9d60f7d899fddd263888d7ea45/source/nodejs/adaptivecards-site/themes/adaptivecards/scripts/generator-adaptiveassets.js#L36-L45

source/nodejs/lerna.json
paulcam206: pre-`workspaces`, this was how you told `lerna` what directories contained packages it was responsible for... now it lets `npm` do all of that heavy lifting. `npm` itself knows what to build from here:

https://github.com/microsoft/AdaptiveCards/blob/02c2427da2e99f9d60f7d899fddd263888d7ea45/source/nodejs/package.json#L10-L28

source/nodejs/README.md
jwoo-msft: This is for ci pipelines, for example, if I want to work on JS, I don't need to do this step?

source/nodejs/package.json
paulcam206: damn, thought I got 'em all 😓

source/nodejs/adaptivecards-designer/webpack.config.js
```diff
@@ -87,8 +87,8 @@ module.exports = (env, argv) => {
                     {
                         dest: 'dist/adaptivecards-designer.css',
                         src: [
-                            './node_modules/adaptivecards-controls/dist/adaptivecards-controls.css',
-                            './node_modules/adaptivecards/dist/adaptivecards-carousel.css',
+                            '../node_modules/adaptivecards-controls/dist/adaptivecards-controls.css',
```
jwoo-msft, line 90: how did it work before? it was broken?

source/nodejs/adaptivecards-site/package.json
```diff
@@ -3,13 +3,15 @@
...
 	"private": true,
 	"scripts": {
+		"build": "npm run generate && npm run generate-sw",
 		"clean": "hexo clean",
-		"build": "hexo generate && npm run generate-sw",
-		"release-deps": "npx lerna run --scope @microsoft/adaptivecards-site --include-dependencies release --stream",
 		"generate-md-ac": "node node_modules/marked-schema/bin/marked-schema ../../../schemas/adaptive-card.json ./schema-explorer-toc.yml -r AdaptiveCard",
 		"generate-md-hc": "node node_modules/marked-schema/bin/marked-schema ../../../schemas/host-config.json ./schema-hostconfig-toc.yml -r AdaptiveCardConfig",
-		"release": "hexo clean && hexo generate && npm run generate-sw",
+		"release": "npm run generate && npm run generate-sw",
```
jwoo-msft, line 10: shouldn't we add `hexo clean` before  `npm run generate`?

source/nodejs/package.json
```diff
@@ -5,22 +5,34 @@
 	"scripts": {
 		"start:designer": "lerna run start --scope=adaptivecards-designer --stream",
 		"start:designer-app": "lerna run start --scope=adaptivecards-designer-app --stream",
-		"start:all": "lerna run --parallel watch && lerna run --parallel start",
-		"audit-all": "lerna-audit",
-		"format": "clang-format-launcher --verbose",
-		"verify": "clang-format-launcher -verify --verbose",
-		"prepare": "cd ../.. && husky install source/nodejs/.husky"
+		"start:all": "lerna run --parallel watch && lerna run --parallel start"
```
jwoo-msft, line 8: will doing auditing be same?

.pipelines/js-ci.yml
```diff
@@ -41,28 +40,34 @@ parameters:
...
-  workingDirectory: source/nodejs
-  displayName: 'Bash - lerna bootstrap'
+- task: npmAuthenticate@0
+  displayName: 'npm authenticate'
+  inputs:
+    workingFile: 'source/nodejs/.npmrc'
 
 - bash: |
-   npx lerna run test
+   npm ci --maxsockets=1
```
jwoo-msft, line 54: nit: wonder if it's too restrictive.

.pipelines/js-ci.yml
```diff
@@ -41,28 +40,34 @@ parameters:
...
-  name: NodeTool1
-  displayName: Use Node 14.x
+- task: UseNode@1
+  displayName: Use Node 20.x
   inputs:
-    versionSpec: 14.x
+    version: '20.x'
 
-- bash: |
-   npm install -g npm@8.11.0
```
jwoo-msft, line 51: we don't need npm install step anymore?

.pipelines/js-release.yml
```diff
@@ -5,7 +5,7 @@ trigger: none
 
 pool:
   name: Azure Pipelines
-  vmImage: ubuntu-20.04
+  vmImage: ubuntu-22.04
```
jwoo-msft, line 8: I'm ccurious why ci is running on windows and release on linux.

jwoo-msft (approved):
Thank you, Paul!

source/nodejs/adaptivecards-designer/webpack.config.js
paulcam206: as part of the move to `workspaces`, the `adaptivecards-controls` package started getting linked in the project root directory :)

source/nodejs/adaptivecards-site/package.json
paulcam206: hmm... let me look into this :)

jwoo-msft (dismissed):

source/nodejs/package.json
paulcam206: auditing should be simpler now -- just `npm audit`

.pipelines/js-ci.yml
paulcam206: this is a workaround for a bug in npm that affects `npm ci` on Windows machines (like our CI box). I wasn't able to repro locally...

.pipelines/js-ci.yml
paulcam206: we were doing some non-standard things held over from when `npm` wasn't directly included in the `nodejs` release packages... but it's included now, so we don't need to manually install it.

.pipelines/js-release.yml
paulcam206: it's an excellent question... I'm not entirely sure why. let me look :)

source/nodejs/README.md
```diff
@@ -13,13 +13,16 @@ This directory contains all the JS packages for adaptive cards:
...
+
 ## Bootstrap the repo
```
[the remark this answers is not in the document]
paulcam206: their requirements are a _bit_ different because they're actually running their code server-side inside node. we only use node as a toolchain to generate JS to run in the browser. that said, it's probably worth a shout-out...

.pipelines/js-release.yml
paulcam206: it's using ubuntu now :)

jwoo-msft (approved):

Issue #8874, opened 2024-04-15
[Feature Request] Speak property should be used by Narrator for UWP/WinUI renderers 


sosharm:
### Problem Statement

The speak property is not hooked up in the UWP renderer. When the adaptive card gets focus, the speak prop is not read by the Narrator.

### Proposed solution

Set the automation name of the root framework element

PR #8875, opened 2024-04-15
Set automation name for root AC element in UWP/WinUI3 renderer

sosharm:
# Related Issue
#8874 

# Description
The speak property is not being hooked up for UWP and WinUI3 renderers.
Setting the automation name for the card root using the speak property in the template.

# How Verified
Verified the fix by using the CalendarReminder sample in the Visualizer which has a speak property.
Narrator reads the speak property in the UWP and WinUI3 Visualizers.


sosharm:
@jwoo-msft @paulcam206

jwoo-msft (approved):

paulcam206 (approved):

paulcam206:
Looks good -- I'll run CI :)

PR #8878, opened 2024-04-18
Fix v1.6 issues for AC object model

tomlm:
# Description
1. **Card.Metadata definition is missing** - Added AdaptiveMetadata class
2.  **ChoiceSet missing DataQuery property** - Added AdaptiveDataQuery class and modified ChoiceSet to serialize it.
4. **Implied types for TableRow/Cell broken** - added attribute 
5. **Visitor class doesn't support all element types** - Fixed visitor class to visit new schema types
6. **WPF renderer has bogus methods added which break the WPF visualizer** - removed those methods, they never should have been added.
7. **Bumped Supported card version to 1.6**  
8. Implement AdaptiveWidth and AdaptiveWidthConverter for parity with AdaptiveHeight
9. add missing horizontal aligment for collections
10.  Property name VerticalAndHorizontal cell content alignment to align with docs 
11.  Add missing XmlElement to Background image.
11.  Fix Refresh, it can be ActionExecute OR Submit. (Teams will barf on preview if you actionexecute)
11.  add Rtl that was missing on Table
11.  Fix Table Column Width that should allow double values, not integers for parity with javascript designer.
11. removed duplicate Style property from TableRow
11. Added missing XmlElement(AdaptiveExecuteAction)  to inline actions
11. Removed duplicate ientries in typed element types
11. Added TableColumnWidth with converter
11. added unit tests
11. Disabled XmlSerializationTests so it is clear that they are not running, they have been broken for years and silently "succeeding'.  XmlSerialization is very broken, but it's not me. 

# How Verified
**Unit Tests disabled for 1.4, 1.5, .16** -- enabled and fixed all issues


tomlm:
@microsoft-github-policy-service agree

source/dotnet/Library/AdaptiveCards/AdaptiveRefresh.cs
```diff
@@ -23,8 +23,9 @@ public class AdaptiveRefresh
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
 #if !NETSTANDARD1_3
         [XmlElement(typeof(AdaptiveExecuteAction))]
+        [XmlElement(typeof(AdaptiveSubmitAction))]
 #endif
-        public AdaptiveExecuteAction Action { get; set; }
+        public AdaptiveAction Action { get; set; }
```
jwoo-msft, line 28: Has this been changed? https://microsoft.visualstudio.com/AdaptiveCards/_build/results?buildId=96490456&view=results

Cryptojen17 (approved):

source/dotnet/Library/AdaptiveCards/AdaptiveDataQuery.cs
```diff
@@ -0,0 +1,67 @@
+using Newtonsoft.Json;
```
paulcam206, line 1: missing copyright banner here and elsewhere

source/dotnet/Library/AdaptiveCards/AdaptiveHeight.cs
```diff
@@ -51,6 +51,38 @@ public AdaptiveHeight()
         {
         }
 
+        /// <summary>
+        /// Initializes an AdaptiveHeight instance with the given string "auto" |"stretch" or "100px"
+        /// </summary>
+        /// <param name="value">enumeration value or pixel .</param>
+        public AdaptiveHeight(string value)
+        {
+            if (value == null)
```
paulcam206, line 60:
```suggestion
            if (String.IsNullOrEmpty(value))
```

same for `AdaptiveWidth`

source/dotnet/Library/AdaptiveCards/AdaptiveHeight.cs
```diff
@@ -51,6 +51,38 @@ public AdaptiveHeight()
         {
         }
 
+        /// <summary>
+        /// Initializes an AdaptiveHeight instance with the given string "auto" |"stretch" or "100px"
```
paulcam206, line 55: nitpick:

```suggestion
        /// Initializes an AdaptiveHeight instance with the given string "auto", "stretch", or "100px"
```

source/dotnet/Library/AdaptiveCards/AdaptiveHeight.cs
```diff
@@ -51,6 +51,38 @@ public AdaptiveHeight()
...
+                Unit = uint.Parse(value);
+            }
+            else if (uint.TryParse(value, out var val))
+            {
+                HeightType = AdaptiveHeightType.Pixel;
+                Unit = val;
+            }
+            else
+            {
+                HeightType = (AdaptiveHeightType)Enum.Parse(typeof(AdaptiveHeightType), value, ignoreCase: true);
```
paulcam206, line 77: `Enum.TryParse` and gracefully fall back to `Auto` (this is what JS renderer does)?

same for `AdaptiveWidth`...

source/dotnet/Library/AdaptiveCards/AdaptiveHeightConverter.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace AdaptiveCards
+{
+    internal class AdaptiveHeightConverter : JsonConverter<AdaptiveHeight>, ILogWarnings
+    {
+        public List<AdaptiveWarning> Warnings { get; set; } = new List<AdaptiveWarning>();
+
+        public AdaptiveHeightConverter()
+        {
+        }
+
```
paulcam206, line 18: super nit: extra space

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveVisitor.cs
```diff
@@ -1,56 +1,67 @@
...
     {
+        public HashSet<AdaptiveTypedElement> Elements { get; set; } = new HashSet<AdaptiveTypedElement>();
+
         public virtual void Visit(AdaptiveElement adaptiveElement)
         {
             if (adaptiveElement is AdaptiveImage image)
                 Visit(image);
 
-            if (adaptiveElement is AdaptiveTextBlock textBlock)
+            else if (adaptiveElement is AdaptiveTextBlock textBlock)
```
paulcam206, line 20: appreciate the optimization here -- I wonder if we can go further with something like:

```c#
switch (adaptiveElement)
{
case AdaptiveTextBlock textBlock:
    Visit(textBlock);
    break;
case AdaptiveContainer container:
    Visit(container);
    break;
/// ...
};
```

paulcam206 (changes requested):
looks great! only found a few minor things...

source/dotnet/Library/AdaptiveCards/AdaptiveRefresh.cs
tomlm: Teams will reject adaptive cards with Action.Execute when using TaskModule. It must be Action.Submit.  This means that this can be either ActionExecute or ActionSubmit, and Action is the base class.

source/dotnet/Library/AdaptiveCards.Rendering.Wpf/AdaptiveCardRenderer.cs
```diff
@@ -288,15 +288,5 @@ private async Task<RenderedAdaptiveCardImage> RenderCardToImageInternalAsync(Ada
...
-
-        public void UseXceedElementRenderers()
-        {
-            throw new NotImplementedException();
-        }
-
-        public void UseDefaultElementRenderers()
-        {
-            throw new NotImplementedException();
-        }
```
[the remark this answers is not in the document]
tomlm: yes. The default element renderers are already configured.

It is only the Xceed renderer library which has 2 extension methods that allow you to Explicitly set the input renderers to use the default or the xceed renderers.

See AdaptiveCards.Rendering.Wpf.Xceed\XceedAdaptiveCardRenderer.cs

Defining the methods UseDefaultElementRenderers() UseXceedElementRenderers() makes no sense (because the WPF library has no dependency on Xceed) and worse defining those methods on the class disabled the extension methods, breaking the WPF visualizer because it is no longer calling the Xceeed extension methods.

source/dotnet/Library/AdaptiveCards/AdaptiveDataQuery.cs
```diff
@@ -0,0 +1,67 @@
...
+        public const char Separator = '#';
+
+        /// <summary>
+        /// Specifies that it's a Data.Query object.
+        /// </summary>
+        [JsonProperty]
+#if !NETSTANDARD1_3
+        [XmlIgnore]
+#endif
+        public string Type { get; set; } = "Data.Query";
```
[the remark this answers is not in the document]
tomlm: Well, the documentation states it should be there:
image

source/dotnet/Library/AdaptiveCards/AdaptiveHeightConverter.cs
paulcam206: not entirely clear from the diff given here, but I was referring to line 19 :D

source/dotnet/Library/AdaptiveCards/AdaptiveRefresh.cs
tomlm: Hmmm....I don't have access to the link. My belief was that there is a teams bug where they reject cards with refresh of Action.Execute if you are doing a task module, but the documentation clearly says Action.Execute, so I will revert this.

source/dotnet/Library/AdaptiveCards/AdaptiveDataQuery.cs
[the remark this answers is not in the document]
jwoo-msft: You're right :-) It was a while back when the feature was first introduced. I remember asking the same question about the JS changes, but I apparently didn't update my brain with the decision.

jwoo-msft (dismissed):
Thank you!

source/dotnet/Library/AdaptiveCards/AdaptiveHeight.cs
tomlm: We want to throw the exception so the converter can add warning text. Added a comment to this effect.

source/dotnet/Library/AdaptiveCards/AdaptiveDataQuery.cs
tomlm: added comment

source/dotnet/Library/AdaptiveCards/AdaptiveHeight.cs
tomlm: done

source/dotnet/Library/AdaptiveCards/AdaptiveHeightConverter.cs
tomlm: done

source/dotnet/Library/AdaptiveCards/Rendering/AdaptiveVisitor.cs
tomlm: yeah, I considered this but it doesn't save anything on spacing or typing so ...?shrug?

paulcam206: I was thinking that it might be quicker, but ultimately probably not terribly important :)

paulcam206 (dismissed):

paulcam206 (approved):

paulcam206 (approved):

jwoo-msft:
[like]  Joseph Woo reacted to your message:

Issue #8879, opened 2024-04-18
[Feature Request] Support for svg xml data in images


sosharm:
### Problem Statement

The UWP/WinUI3 renderer only supports base64 svg images. Requesting support for `<svg>...</svg>` format.

### Proposed solution

If the svg image is not in a base6 format, skip the decoding code and extract the svg data from the image url.

PR #8880, opened 2024-04-18
Add support for svg xml in the data scheme for images

sosharm:
# Related Issue
#8879 

# Description
The UWP/WinUI3 renderer only supports base64 svg images. Adding support for `<svg>...</svg>` format.
If the svg image is not in a base6 format, skip the decoding code and extract the svg data from the image url.

# Sample Card
Updated [Image.Svg.json](https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.5/Tests/Image.Svg.json)

# How Verified
Updated [Image.Svg.json](https://github.com/microsoft/AdaptiveCards/blob/main/samples/v1.5/Tests/Image.Svg.json)
image


sosharm:
@paulcam206 @jwoo-msft

paulcam206 (dismissed):
looks great!

source/uwp/SharedRenderer/lib/AdaptiveImageRenderer.cpp
```diff
@@ -356,14 +357,23 @@ namespace AdaptiveCards::Rendering::Xaml_Rendering
...
         {
-            // Decode base 64 string
-            winrt::hstring dataPath = imageUrl.Path();
-            std::string data = AdaptiveBase64Util::ExtractDataFromUri(HStringToUTF8(dataPath));
-            std::vector<char> decodedData = AdaptiveBase64Util::Decode(data);
-
             winrt::DataWriter dataWriter{winrt::InMemoryRandomAccessStream{}};
 
-            dataWriter.WriteBytes(std::vector<byte>{decodedData.begin(), decodedData.end()});
+            auto foundBase64 = HStringToUTF8(imageUrl.Path()).find("base64");
```
jwoo-msft, line 362: Can you first check `if (imgProperties.isImageSvg)` and, only if that fails, proceed with decoding the base64 string? In this way, we don’t call HstringToUTF8 unnecessarily.

Also, can you cache the result of `HstringToUTF8(imageUrl.Path())` as dataPath, and reuse it for the foundBase64 check and data extraction, since they are the same string?

jwoo-msft (changes requested):

source/uwp/SharedRenderer/lib/AdaptiveImageRenderer.cpp
sosharm: The image can be base64 and svg. It can also be bitmap and base64. We follow the same code path for base64. If that is not a match and the image was svg, we try the svg xml code path.

jwoo-msft: if so, please disregard the first paragraph.

PR #8891, opened 2024-04-24
[Website][A11y] Update keyboard nav for blog posts

### LINKED ISSUE #8883: [Adaptive Cards-Blog page]: Under the Blog page, Blogs are not defined in List.
### Target Platforms

Other

### SDK Version

1.6

### Application Name

Adaptive Cards

### Problem Description

**Test environment:**
OS: Win11
OS Version: 24H2 (OS Build 26100.1)
URL:[Adaptive Cards](https://adaptivecards.io/)
Edge dev Browser version: 125.0.2506.0 (Official build) dev (64-bit)
Screen reader: Narrator.

**Repro Steps:** 
1. Open URL in the Edge browser & sign in with valid credentials. 
2. It will open the Adaptive cards home page 
3.Activate the blog control, navigate through tab key, and observe the issue. 

**Actual Result:**  
The cards in the Blog posts are navigable by the tab keys instead of up/down arrow keys and those cards are not defined in the list. 
 
**Expected Result:** 
The cards in the blog page should be defined in the list that should be navigable by the up/down arrow keys. 

**User Experience:**  
The keyboard users will be impacted if the Blog posts navigable by the tab key instead of arrow keys as if there are multiple posts the users will have to do multiple tabbing on the page to reach the desired location which will be confusing task. 

**WCAG Reference:** 
[MAS 2.1.1 Keyboard 
](https://www.w3.org/WAI/WCAG21/Understanding/keyboard)

Have feedback to share on Bugs ? please tag bug as “A11yRCA” and add your feedback in the comment section

### Screenshots

https://github.com/microsoft/AdaptiveCards/assets/95913748/26ac2dfc-7727-49a4-8dd7-dc6cf144bca6

Blogs are not defined in list

### Card JSON

```JSON
NA
```

### Sample Code Language

_No response_

### Sample Code

_No response_

anna-dingler:
# Related Issue

Fixes #8883 

# Description

Add keyboard navigation for the blog posts on the AC website.

I added support for the left and right arrows to navigate each row of blog posts. When we reach the end of the row, we continue to the next/previous row, depending on which direction we are going.

There is also up/down navigation support that will jump to the same position in the next/previous row.

# Sample Card

N/A

# How Verified

Verified manually with the adaptive cards site.


source/nodejs/adaptivecards-site/themes/adaptivecards/layout/archive.ejs
```diff
@@ -22,7 +64,7 @@ layout: layout
 	<% if(i % 3 === 0) { /* start a new row every 3 posts */ %>
 	<div class="w3-cell-row w3-padding-16">
 		<% } %>
-		<article class="w3-cell w3-margin-bottom w3-padding-16 w3-container w3-mobile article" style="width: 328px">
+		<article class="w3-cell w3-margin-bottom w3-padding-16 w3-container w3-mobile article" style="width: 328px" onkeyup="onKeyUp(event)" tabindex="0">
```
anna-dingler, line 67: I wasn't quite sure how we should handle this, but I assigned tab order to the main article element so that it can receive focus like so:
image

It isn't a clickable element, but it will help jump through the blog posts more quickly.

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/archive.ejs
```diff
@@ -3,6 +3,74 @@ layout: layout
...
+					if (prev != null) {
+						prev.focus();
+					} else {
+						prevRow?.lastElementChild?.focus();
+					}
+					break;
+				case "ArrowUp":
+					if (prevRow != null) {
+						// Special case - only one element in the row
+						if (parentRow.children.length === 1) {
```
anna-dingler, line 35: The up/down cases are slightly more complex so that we handle the case where the last row does not have 3 elements

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/archive.ejs
```diff
@@ -51,6 +119,10 @@ layout: layout
 	<% } %>
 	<% }); %>
 
+	<% if(page.posts.length % 3 !== 0) { /* if the posts were not divisible by 3, we need to close the last row */ %>
+	</div>
+	<% } %>
+
 	<!-- Pager -->
```
anna-dingler, line 126: QQ @paulcam206 What is the Pager for? Is it added for when we eventually have more blogs?

jwoo-msft (dismissed):

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/archive.ejs
paulcam206: how hard would it be to make it invokable? might be nice :)

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/archive.ejs
paulcam206: looks like an [option](https://hexo.io/docs/configuration#Pagination) we're not currently using:

https://github.com/microsoft/AdaptiveCards/blob/09b0c8adaa2da5d2f5d8879733f6eabd1d85608b/source/nodejs/adaptivecards-site/_config.yml#L70-L73

paulcam206 (dismissed):

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/archive.ejs
anna-dingler: Ah interesting. Should we remove it or just keep it around in case we want it one day?

source/nodejs/adaptivecards-site/themes/adaptivecards/layout/archive.ejs
anna-dingler: I _think_ it should be reasonable :)

anna-dingler: Okay I made it so that you can navigate to the article when you click enter. I didn't want to expose the onclick - that seemed like too much noise

paulcam206 (approved):

PR #8899, opened 2024-04-30
[NodeJS] Fix site generation from Windows hosts

paulcam206:
Turns out that some of my changes to site generation didn't work properly on Windows builders (I authored using WSL). Mostly this was due to path munging interacting poorly with `glob` in `generators-adaptiveassets.js`.

[Here's](https://microsoft.visualstudio.com/AdaptiveCards/_build/results?buildId=97502826&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=7cc137c2-61f2-5d65-b5f8-0e03d1af3170) a clean run showing that it works on Windows 👍🏼


anna-dingler (approved):
Looks good to me - thank you!

PR #8913, opened 2024-05-14
Add a style override to the root container named "Adaptive.RootContainer"

Apxov:
Add a style override to the root container named "Adaptive.RootContainer" so that custom margins can be applied

# Description
This style enables a host to set the margins of the container to be other than uniform.  This enables the background of the panel to bleed up into the frame for a card that may be contained in UI that provides a title bar.

# How Verified
Manually verified in a host that sets an OverrideStyles that contains a style with this name in its dictionary.

### DIFF
--- a/source/uwp/SharedRenderer/lib/XamlBuilder.cpp
+++ b/source/uwp/SharedRenderer/lib/XamlBuilder.cpp
@@ -105,15 +105,6 @@ namespace AdaptiveCards::Rendering::Xaml_Rendering
 
             bool isInShowCard = renderArgs.IsInShowCard();
 
-            if (isInShowCard)
-            {
-                XamlHelpers::SetStyleFromResourceDictionary(renderContext, L"Adaptive.ShowCard.Card", rootAsFrameworkElement);
-            }
-            else
-            {
-                XamlHelpers::SetStyleFromResourceDictionary(renderContext, L"Adaptive.Card", rootAsFrameworkElement);
-            }
-
             xamlTreeRoot = rootAsFrameworkElement;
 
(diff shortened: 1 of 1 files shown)


source/uwp/SharedRenderer/lib/XamlBuilder.cpp
```diff
@@ -213,7 +213,13 @@ namespace AdaptiveCards::Rendering::Xaml_Rendering
             bodyElementHostImpl->SetAdaptiveHeight(true);
             winrt::WholeItemsPanel bodyElementHost = *bodyElementHostImpl;
 
-            XamlHelpers::ApplyMarginToXamlElement(hostConfig, bodyElementHost);
+            // Override the root styles
+            XamlHelpers::SetStyleFromResourceDictionary(renderContext, L"Adaptive.RootContainer", bodyElementHost);
```
almedina-ms, line 217: I think we should keep the name centered around elements, other styles have names as Adaptive.Column, I think in this case it could be Adaptive.Card or something similar

@paulcam206 , @jwoo-msft , what do you think?

Apxov: It looks like you already have such a style used by BuildXamlTreeFromAdaptiveCard however because ApplyMarginToXamlElement is called which calls .Margin() directly, the style override of Margin won't work.

Apxov: I updated the approach so that a new style name is not required.

paulcam206 (approved):

PR #8916, opened 2024-05-15
fixed base64 decoding issue

jwoo-msft:
# Related Issue
N/A

# Description
Spacing were not removed in base64 decoding.
Size calculation was wrong since it compute for chunk size not actual printable 4 bytes sized type space.

# How Verified
Manually verified.
image


source/shared/cpp/ObjectModel/AdaptiveBase64Util.cpp
```diff
@@ -249,6 +249,17 @@ std::string AdaptiveBase64Util::Encode(const std::vector<char>& decodedBase64)
 // Format for DataURI is data:[<MediaType>][;base64],data with MediaType and base64 being optional and data is composed of [A-Z a-z 0-9 + /] characters
 std::string AdaptiveBase64Util::ExtractDataFromUri(const std::string& dataUri)
 {
-    size_t comaPosition = dataUri.find_last_of(",");
-    return dataUri.substr(comaPosition + 1);
+    size_t commaPosition = dataUri.find_last_of(",");
```
Apxov, line 252: That is a long way to go for a big image.  If it is 100k, that is a long walk.  Since we know it is a data: scheme why can't we parse if from the front.

jwoo-msft: This isn't the part of my change.

Apxov (approved):

source/shared/cpp/ObjectModel/AdaptiveBase64Util.cpp
Apxov: Yes, just noting that it wasn't a very performant approach when originally written not a regression.

PR #8919, opened 2024-05-22
Make AdaptiveCards.Templating AOT compatible

jevansaks:
# Description

This change moves AdaptiveCards.Templating to be built on System.Text.Json and the flavor of AdaptiveExpressions that also uses System.Text.Json. It was necessary to move off of Newtonsoft to be able to support AOT.

I have preserved the old API, but added AOT compatibility attributes and added AOT compatible versions instead. The AdaptiveCards.Templating library fundamentally works against json strings, and the APIs which took object would immediately serialize that object to a json string. The main changes are:
* All APIs taking `object` from the application now take `string`, making it the caller's responsibility to serialize to json.
* `object`-based APIs remain for compatibility but are flagged as `RequiresDynamicCode` and `RequiresUnreferencedCode`.
* I changed the unit tests to compile as AOT compatible and changed them where needed.

# How Verified

Ran all existing unit tests and they passed.

Note: I had to make one minor change to an expected test result because System.Text.Json does not round-trip literal double quotes in a string, they come back escaped not as `\"` but as `\u0022`. It seems that Newtonsoft had different behavior around this.


paulcam206 (approved):

source/dotnet/Test/AdaptiveCards.Templating.Test/TestTransform.cs
```diff
@@ -13123,8 +13126,8 @@ public void TestSerialization()
                 "\"body\": [" + 
                 "{" + 
                 "\"id\": \"stringWithJson\"," + 
-                "\"type\": \"TextBlock\"," + 
-                @"""text"": ""String With JSON - \""{\\\""number1\\\"":23}\""""," + 
+                "\"type\": \"TextBlock\"," +
+                @"""text"": ""String With JSON - \u0022{\\u0022number1\\u0022:23}\u0022""," +
```
jwoo-msft, line 13130: I'm little bit concerned about this change.
will this text produce the correct AC? it's only comparing if two jsons are equal. Will .Net AC SDK render a correct AdaptiveCards?

jevansaks: The change here is just that Newtonsoft was preserving escaped quotes in strings as `\"` which is not to spec. System.Text.Json is escaping these to \u0022. That's why this changed. Functionally it's the same.

jwoo-msft (approved):

jwoo-msft:
azp /run

PR #8940, opened 2024-06-21
[Designer] Update data binding 

### LINKED ISSUE #8013: [Designer] Data binding button produces invalid binding for arrays
### Problem Description

Clicking the `...` or `Bind...` data binding button and selecting a property which is an array, or is contained in an array produces data binding strings in the form `${$root['propertyName[0]'}`, which are not valid.

### Card JSON

```JSON
{
    "type": "AdaptiveCard",
    "version": "1.5",
    "body": [
        {
            "type": "TextBlock",
            "text": "${$root['employees[0]']['name']}",
            "wrap": true
        }
    ],
    "data": {
        "employees": [
            {
                "name": "Gabriela Leticia"
            },
            {
                "name": "Natalia Tercera"
            },
            {
                "name": "Waleska Cristobal"
            }
        ]
    }
}
```

### Screenshot

image

anna-dingler:
# Related Issue

Fixes #8013 

# Description

When generating the templating variable from the `Bind...` button in the designer, we have an edge case for data variables that contain special characters. In this scenario, we typically want to use index syntax and wrap the var in braces.

However, this does not work for arrays. For arrays, we index into the first element by appending `[0]` to the name. This forces the index syntax incorrectly because the name now includes special characters.

I updated the logic to remove `[0]` from the end of array type names before checking for special characters.

# Sample Card

```JSON
{
    "type": "AdaptiveCard",
    "version": "1.5",
    "body": [
        {
            "type": "TextBlock",
            "text": "${$root['employees[0]']['name']}",
            "wrap": true
        }
    ],
    "data": {
        "employees": [
            {
                "name": "Gabriela Leticia"
            },
            {
                "name": "Natalia Tercera"
            },
            {
                "name": "Waleska Cristobal"
            }
        ]
    }
}
```

# How Verified

Verified on the CI site.

### DIFF
--- a/source/nodejs/adaptivecards-designer/src/data.ts
+++ b/source/nodejs/adaptivecards-designer/src/data.ts
@@ -302,9 +302,17 @@ export class FieldDefinition {
         let forceIndexerSyntax: boolean = false;
 
         for (let i = path.length - 1; i >= 0; i--) {
-            let qualifiedName = path[i].qualifiedName(false)
+            let qualifiedName = path[i].qualifiedName(false);
 
-            if (shouldUseIndexerSyntax(qualifiedName) || forceIndexerSyntax) {
+            let modifiedName = qualifiedName;
+
+            // If we include `[0]` in the qualified name while determining if we should use indexer syntax,
+            // We could incorrectly wrap the name in []
+            if (this.dataType instanceof ArrayData && qualifiedName.endsWith("[0]")) {
+                modifiedName = qualifiedName.substring(0, qualifiedName.length - 3);
+            }
+
+            if (shouldUseIndexerSyntax(modifiedName) || forceIndexerSyntax) {
                 qualifiedName = "['" + qualifiedName + "']";
 
                 forceIndexerSyntax = true;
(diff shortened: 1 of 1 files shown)


jwoo-msft:
Hi Anna,
This looks odd:
```
['employees[0]']

```
I don't think this is valid syntax to access into an array.

anna-dingler:
> Hi Anna, This looks odd:
> 
> ```
> ['employees[0]']
> ```
> 
> I don't think this is valid syntax to access into an array.

Hey @jwoo-msft! That sample is confusing - sorry! That is the previous behavior that the PR fixes. If you try to use the `Bind...` feature right now, that is the result. The issue occurs because we use the indexer syntax (wrap in []) if the json key contains special characters. However, for arrays, we asses the key _after_ we add `[0]` and it is incorrectly wrapped.

jwoo-msft:
Hi @anna-dingler , 

[$root.employees[0]] works, but 'employee[0]' is just a string in AEL expression.

source/nodejs/adaptivecards-designer/src/data.ts
```diff
@@ -302,9 +302,17 @@ export class FieldDefinition {
...
 
         for (let i = path.length - 1; i >= 0; i--) {
-            let qualifiedName = path[i].qualifiedName(false)
+            let qualifiedName = path[i].qualifiedName(false);
 
-            if (shouldUseIndexerSyntax(qualifiedName) || forceIndexerSyntax) {
+            let modifiedName = qualifiedName;
+
+            // If we include `[0]` in the qualified name while determining if we should use indexer syntax,
+            // We could incorrectly wrap the name in []
```
jwoo-msft, line 310: can you add example string?

PR #8946, opened 2024-06-28
Remove old ReactNative implementation

paulcam206:
The ReactNative project has long been spun out to https://github.com/Bigthinkcode/AdaptiveCards/

This change removes the existing (out of date) version.


paulcam206:
whoops - need to remove the github action as well

dmachaj:
Shouldn't the link out be to https://github.com/microsoft/AdaptiveCards instead of BigThinkCode (whatever that is)?

jwoo-msft (dismissed):

paulcam206:
> Shouldn't the link out be to https://github.com/microsoft/AdaptiveCards instead of BigThinkCode (whatever that is)?

This project was a "community" project -- basically, a separate group built the ReactNative renderer alongside us and shipped as a no-official-support package. After a time, it was decided that it would be best if they continued their development in their own repo, which is located in the BigThinkCode repo :)

jwoo-msft (approved):

PR #8947, opened 2024-06-28
added invalid numeric width

jwoo-msft:
# Related Issue
It's a bug if there is only one column with numeric width, since we need more than one numeric values for ratio.

# Description
if this case is found, we replace the column with 'stretch'

# Sample Card
```
{
    "type": "AdaptiveCard",
    "body": [
        {
            "type": "Container",
            "items": [
                {
                    "type": "ColumnSet",
                    "columns": [
                        {
                            "type": "Column",
                            "items": [
                                {
                                    "type": "TextBlock",
                                    "text": "Top Notifications",
                                    "size": "Medium",
                                    "height": "stretch"
                                }
                            ],
                            "width": 80
                        },
                        {
                            "type": "Column",
                            "items": [
                                {
                                    "type": "TextBlock",
                                    "text": "See all",
                                    "size": "Small",
                                    "color": "Accent"                                }
                            ],
                            "horizontalAlignment": "Right",
                            "verticalContentAlignment": "Bottom",
                            "selectAction": {
                                "type": "Action.Execute",
                                "verb": "openNotificationsPage"
                            },
                            "width": "stretch"
                        }
                    ],
                    "spacing": "None"
                }
            ]
        }
    ],
    "version": "1.6"
}
```

# How Verified
Manually verified.
Before
image
After
image


source/shared/cpp/ObjectModel/ColumnSet.cpp
```diff
@@ -53,6 +53,33 @@ void ColumnSet::DeserializeChildren(ParseContext& context, const Json::Value& va
...
+
+    // If there is one column with numeric width,
+    // set it to stretch since the ratio requires at least two columns
+    // with numeric width.
+    for (auto const& column : m_columns)
+    {
+        const auto width {column->GetWidth()};
+        const auto ch = width.at(0);
+        if (column->GetPixelWidth() == 0 &&
+            (ch != 'a' && ch != 's') &&
```
paulcam206, line 68: what happens in this case? (note `A` in `Auto`)

```json
{
  "type": "ColumnSet",
  "columns": [
    {
      "width": "Auto",
      "items": [
        {}
      ]
    }
  ]
}
```

paulcam206 (dismissed):

source/shared/cpp/ObjectModel/ColumnSet.cpp
jwoo-msft: we put lower case width when we deserilize.

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ExplicitDimensionTest.cpp
```diff
@@ -455,6 +455,12 @@ namespace AdaptiveCardsSharedModelUnitTest
                     {\
                         \"type\":\"ColumnSet\",\
                         \"columns\": [\
+                            {\
```
jwoo-msft, line 458: since this is positive test; i added additional column to make the payload valid payload.

source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/ExplicitDimensionTest.cpp
```diff
@@ -531,7 +537,7 @@ namespace AdaptiveCardsSharedModelUnitTest
             std::shared_ptr<ColumnSet> columnSet = std::static_pointer_cast<ColumnSet>(element);
             std::shared_ptr<Column> column = columnSet->GetColumns().front();
             Assert::AreEqual<int>(0, column->GetPixelWidth());
-            Assert::AreEqual<std::string>("20px20", column->GetWidth());
+            Assert::AreEqual<std::string>("stretch", column->GetWidth());
```
jwoo-msft, line 540: at LN 524, width is invalid, and in the invalid case, we set it to 'stretch', and that's correct behavior.

Apxov (approved):

Issue #8967, opened 2024-07-23
[Rendering] Bug Title Here


beervoley:
### Target Platforms

UWP

### SDK Version

10.0.17763.0

### Application Name

N/A

### Problem Description

We need to check `RootElement` for nullptr here https://github.com/microsoft/AdaptiveCards/blob/4745cb65ae8d57529811f7a0ba13f04afbadcacc/source/uwp/SharedRenderer/lib/XamlBuilder.cpp#L58 to avoid various crashes (AV most common) if we failed to render rootelement.

### Card JSON

```JSON
[]
```

PR #8968, opened 2024-07-23
[UWP/WinUI3 Renderer]: Discard Background Image if URI is not valid

beervoley:
# Related Issue
For MSFT folks: https://microsoft.visualstudio.com/OS/_workitems/edit/52712745

# Description
Background Image with invalid URI would crash the entire app when we attempted to render it. This change addresses that and brings the behaviour on par with NodeJS renderer and drops the image in this case.  + added warning for text sources for mediasource for the similar problem.

# Sample Card
```
{
    "type": "AdaptiveCard",
    "version": "1.6",
    "backgroundImage" : "invalidURI",
    "body": []
}
```

# How Verified
Verified locally that the BackgroundImage is being discarded as well as added unit test for this scenario.

TODO: follow up on https://github.com/microsoft/AdaptiveCards/issues/8967

### DIFF
--- a/source/uwp/SharedRenderer/lib/Util.cpp
+++ b/source/uwp/SharedRenderer/lib/Util.cpp
@@ -536,27 +536,26 @@ bool IsBackgroundImageValid(winrt::AdaptiveBackgroundImage const& backgroundImag
     return false;
 }
 
-winrt::Uri UriTryCreate(winrt::hstring const &uri, winrt::hstring const& baseUri = L"")
+winrt::Uri UriTryCreate(winrt::hstring const &uriString, winrt::hstring const& baseUriString = L"")
 {
     auto factory = winrt::
         get_activation_factory<winrt::Windows::Foundation::Uri, winrt::Windows::Foundation::IUriRuntimeClassFactory>();
     auto abiFactory = static_cast<ABI::Windows::Foundation::IUriRuntimeClassFactory *>(winrt::get_abi(factory));
 
-    const winrt::hstring &localUri = uri;
-    winrt::Windows::Foundation::Uri returnValue{nullptr};
+    winrt::Windows::Foundation::Uri uri{nullptr};
     HRESULT hr = S_OK;
-    if (baseUri.empty())
+    if (baseUriString.empty())
     {
         hr = abiFactory->CreateUri(
-            static_cast<HSTRING>(winrt::get_abi(localUri)),
-            reinterpret_cast<ABI::Windows::Foundation::IUriRuntimeClass**>(winrt::put_abi(returnValue)));
+            static_cast<HSTRING>(winrt::get_abi(uriString)),
+            reinterpret_cast<ABI::Windows::Foundation::IUriRuntimeClass**>(winrt::put_abi(uri)));
     }
     else
     {
         hr = abiFactory->CreateWithRelativeUri(
-            static_cast<HSTRING>(winrt::get_abi(baseUri)),
-            static_cast<HSTRING>(winrt::get_abi(uri)),
-            reinterpret_cast<ABI::Windows::Foundation::IUriRuntimeClass**>(winrt::put_abi(returnValue)));
+            static_cast<HSTRING>(winrt::get_abi(baseUriString)),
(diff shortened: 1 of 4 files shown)


source/uwp/SharedRenderer/lib/Util.cpp
```diff
@@ -535,19 +536,50 @@ bool IsBackgroundImageValid(winrt::AdaptiveBackgroundImage const& backgroundImag
     return false;
 }
 
+winrt::Uri UriTryCreate(winrt::hstring const &uri, winrt::hstring const& baseUri = L"")
+{
+    auto factory = winrt::
+        get_activation_factory<winrt::Windows::Foundation::Uri, winrt::Windows::Foundation::IUriRuntimeClassFactory>();
```
jwoo-msft, line 542: @beervoley, are there any reasons we have to do this way instead of creating using winrt::Uri?

source/uwp/SharedRenderer/lib/Util.cpp
```diff
@@ -535,19 +536,50 @@ bool IsBackgroundImageValid(winrt::AdaptiveBackgroundImage const& backgroundImag
...
+    auto abiFactory = static_cast<ABI::Windows::Foundation::IUriRuntimeClassFactory *>(winrt::get_abi(factory));
+
+    const winrt::hstring &localUri = uri;
```
jwoo-msft, line 545: nit can we change uri to uriString, and returnValue to uri?

source/uwp/SharedRenderer/lib/Util.cpp
```diff
@@ -535,19 +536,50 @@ bool IsBackgroundImageValid(winrt::AdaptiveBackgroundImage const& backgroundImag
...
     }
     else
     {
         winrt::hstring imageBaseUrl = hostConfig.ImageBaseUrl();
 
-        if (const auto uriFromRelativeUri = winrt::Uri{imageBaseUrl, urlString})
+        if (const auto uriFromRelativeUri = UriTryCreate(urlString, imageBaseUrl))
         {
             return uriFromRelativeUri;
         }
```
jwoo-msft, line 585: do you think it could be better to throw, then returning the nullptr? I think that we catch exceptions from rendering. if it's too much work, i don't think it's worth it, but if the catch block is there, then the error handling code would be simpler and slightly better.

jwoo-msft: I meant at LN 587

source/uwp/SharedRenderer/lib/Util.cpp
beervoley: in cpp winrt we don't have a constructor for `winrt::Uri` that doesn't throw if the `uri` is invalid thus every tiem we try to create `winrt::Uri` and there's something wrong with the `uri` -> it throws

jwoo-msft: thanks @beervoley

source/uwp/SharedRenderer/lib/Util.cpp
beervoley: I prefer exceptions for cases that more fatal in nature and they lead to program termination or some fundamental routine exit. In this case, we're simply parsing a `URI` and throwing an exception (which will capture entire stack as well) looks like an overkill to me :)
In some of the places where we've been using `GetUrlFromString` like here: https://github.com/microsoft/AdaptiveCards/blob/4745cb65ae8d57529811f7a0ba13f04afbadcacc/source/uwp/SharedRenderer/lib/AdaptiveImageRenderer.cpp#L57-L59 -> we already expect it to return `nullptr` on failure. Before this PR, this would never happen and the method would throw, leading to exception bubbling up and we wouldn't know what's wrong. In this case, it will work as expected.

jwoo-msft: that's fine. all i'm saying is if we throw, the exception will get caught by elemental renderer, and will get skipped which is the desired behavior.

jwoo-msft (approved):

PR #8992, opened 2024-08-23
Add adaptivecards#markdownPassthroughPolicy and adaptivecards#restore…

Farcemolina:
# Related Issue

page above.

# Description

- The adaptivecards-v1 version we are using on outlook web at some point should be updated and use adaptivecards-v3 that already has the trusted types support, this is to unblock the enabling of the Trusted Types CSP on outlook.
- I copied exactly what Anna did on adaptivecards-v3 to have consistency on the fix.
- Using dompurify inside of this function could create more issues since the html can get tags or attributes needed on outlook.

# Sample Card

# How Verified

2. Existing relevant unit/regression tests that you ran
3. Manual scenario verification if any; ***Do include .gif's or screenshots of the testing you performed here if you think that it

### DIFF
--- a/source/nodejs/adaptivecards/package.json
+++ b/source/nodejs/adaptivecards/package.json
@@ -56,8 +56,5 @@
 			"jsx",
 			"json"
 		]
-	},
-	"dependencies": {
-		"dompurify": "^3.1.6"
 	}
 }
--- a/source/nodejs/adaptivecards/package.json
+++ b/source/nodejs/adaptivecards/package.json
@@ -56,5 +56,8 @@
 			"jsx",
 			"json"
 		]
+	},
+	"dependencies": {
+		"@types/trusted-types": "^2.0.7"
 	}
 }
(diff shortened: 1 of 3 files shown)


source/nodejs/adaptivecards/src/card-elements.ts
```diff
@@ -5,11 +5,10 @@ import * as Shared from "./shared";
 import * as Utils from "./utils";
 import * as HostConfig from "./host-config";
 import * as TextFormatters from "./text-formatters";
-import { sanitize } from "dompurify";
```
anna-dingler, line 8: Should this be removed from `package.json` and `package-lock.json` as well?

anna-dingler (approved):
Looks good to me. @paulcam206 @jwoo-msft Any thoughts?

source/nodejs/adaptivecards/package.json
```diff
@@ -56,8 +56,5 @@
 			"jsx",
 			"json"
 		]
-	},
```
jwoo-msft, line 59: why are you reverting your changes? Can you leave descriptions?

jwoo-msft: @Farcemolina,
Will this change be good once and for all?

jwoo-msft (approved):

PR #9002, opened 2024-09-06
Update CsWinRT for WinUI 3 renderer

### LINKED ISSUE #9001: [Feature Request] Update CsWinRT for WinUI 3 renderer
### Problem Statement

Dev Home would like to update to WindowsAppSDK 1.6. In order to do that, dependencies must update to CsWinRT version 2.1.0 or later.

### Proposed solution

Update CsWinRT

_No response_

krschau:
# Related Issue

Closes #9001

# Description

Any WinUI 3 apps that use Adaptive Cards and would like to update to WindowsAppSDK 1.6 will need Adaptive Cards to support CsWinrt 2.1.0+. This change updates that version. It also adds the WindowsSdkPackageVersion property, as described in   https://github.com/microsoft/CsWinRT/releases/tag/2.1.1.240807.2

# Sample Card

N/A

# How Verified

Manually verified with AdaptiveCardVisualizer

### DIFF
--- a/source/uwp/winui3/ObjectModelCsProjection/ObjectModelCsProjection.csproj
+++ b/source/uwp/winui3/ObjectModelCsProjection/ObjectModelCsProjection.csproj
@@ -15,7 +15,7 @@
     <BaseOutputPath>$(SolutionDir)\$(Platform)\$(MSBuildProjectName)</BaseOutputPath>
   </PropertyGroup>
   <ItemGroup>
-      <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.0.1" />
+      <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.1.1" />
       <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.2.221109.1" />
       <PackageReference Include="Microsoft.Windows.SDK.BuildTools" Version="10.0.22621.1" />
   </ItemGroup>
--- a/source/uwp/winui3/ObjectModelCsProjection/ObjectModelCsProjection.csproj
+++ b/source/uwp/winui3/ObjectModelCsProjection/ObjectModelCsProjection.csproj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <TargetFramework>net6.0-windows10.0.17763.0</TargetFramework>
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
-    <WindowsSdkPackageVersion>10.0.18362.38</WindowsSdkPackageVersion>
+    <WindowsSdkPackageVersion>10.0.17763.38</WindowsSdkPackageVersion>
     <!-- Set Platform to AnyCPU to allow consumption of the projection assembly from any architecture. -->
     <Platform>AnyCPU</Platform>
     <RootNamespace>AdaptiveCards.ObjectModel.WinUI3</RootNamespace>
--- a/source/uwp/winui3/ObjectModelCsProjection/ObjectModelCsProjection.csproj
+++ b/source/uwp/winui3/ObjectModelCsProjection/ObjectModelCsProjection.csproj
@@ -1,7 +1,8 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net6.0-windows10.0.17763.0</TargetFramework>
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
+    <WindowsSdkPackageVersion>10.0.18362.38</WindowsSdkPackageVersion>
     <!-- Set Platform to AnyCPU to allow consumption of the projection assembly from any architecture. -->
     <Platform>AnyCPU</Platform>
     <RootNamespace>AdaptiveCards.ObjectModel.WinUI3</RootNamespace>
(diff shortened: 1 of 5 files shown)


beervoley (dismissed):

source/uwp/winui3/ObjectModelCsProjection/ObjectModelCsProjection.csproj
```diff
@@ -1,7 +1,8 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net6.0-windows10.0.17763.0</TargetFramework>
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
+    <WindowsSdkPackageVersion>10.0.18362.38</WindowsSdkPackageVersion>
```
manodasanW, line 5: The OS version here should probably match the TFM OS version.  In this case being 17763.  Same for others.

beervoley: I don't think that matters in this case match, right? `TargetPlatformMinVersion` is set to `17763` and (iirc) if we build successfully, we're not using any of the >=17763 APIs anyway :)
but I agree, it might be better to do it anyway.

krschau: When I tried changing it to 10.0.17763.38 I got errors:
```
warning MSB3277: Found conflicts between different versions of "Microsoft.Windows.SDK.NET" that could not be resolved.
warning MSB3277: There was a conflict between "Microsoft.Windows.SDK.NET, Version=10.0.17763.38, Culture=neutral, PublicKeyToken=31bf3856ad364e35" and "Microsoft.Windows.SDK.NET, Version=10.0.18362.24, Culture=neutral, PublicKeyToken=31bf3856ad364e35".
warning MSB3277:     "Microsoft.Windows.SDK.NET, Version=10.0.17763.38, Culture=neutral, PublicKeyToken=31bf3856ad364e35" was chosen because it was primary and "Microsoft.Windows.SDK.NET, Version=10.0.18362.24, Culture=neutral, PublicKeyToken=31bf3856ad364e35" was not.

error NETSDK1148: A referenced assembly was compiled using a newer version of Microsoft.Windows.SDK.NET.dll. Please update to a newer .NET SDK in order to reference this assembly.
```

jwoo-msft (approved):

source/uwp/winui3/ObjectModelCsProjection/ObjectModelCsProjection.csproj
beervoley: does this happen when you set `WindowsSdkPackageVersion` to `10.0.17763.38` in **all** the project simultaneously?

manodasanW: You are right that with respect to usage of APIs, `TargetPlatformMinVersion` will ensure you are using only the compatible ones.  But the projections for the different OS versions do have different assembly versions which is kind of seen above in the error.  This means if you have your projections shipping as compatible with 17763 TFM but use the 18362 Windows SDK projection package, and there was someone targeting 17763 that consumed your package, they will see the error similar to above.  Then they would need to move their TFM up to 18362.

@krschau which project did you see that error in?

krschau: @beervoley I did make the change in all the projects simultaneously
@manodasanW I saw the error in the AdaptiveCardVisualizer project. If I keep only that one up on version 10.0.18362.38 then everything builds -- since that's just the test app, that's probably the way to go?

source/uwp/winui3/Visualizer/AdaptiveCardVisualizer.csproj
```diff
@@ -3,6 +3,7 @@
     <OutputType>WinExe</OutputType>
     <TargetFramework>net6.0-windows10.0.22000.0</TargetFramework>
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
+    <WindowsSdkPackageVersion>10.0.18362.38</WindowsSdkPackageVersion>
```
manodasanW, line 6:
```suggestion
    <WindowsSdkPackageVersion>10.0.22000.38</WindowsSdkPackageVersion>
```

The TFM here is 22000, so it would use that respective SDK package version by default.

source/uwp/winui3/ObjectModelCsProjection/ObjectModelCsProjection.csproj
manodasanW: The TFM for that project looks like to be targeting 22000, so that probably makes sense why you got that error..

source/uwp/winui3/Visualizer/AdaptiveCardVisualizer.csproj
krschau: Whoops, I was misunderstanding what "TFM" stood for.

manodasanW (approved):
Changes look good to me.  As an FYI, the .38 version of the Windows SDK projection is still not included by default in the .NET SDK yet.  It is upcoming in the next servicing release.  Until then any consumers would need to manually use `WindowsSdkPackageVersion` to make sure they got that version when consuming your projection.  So due to that, I am not sure whether you want to publish to nuget.org before the .NET SDK update is out or not.  In WinAppSDK we chose to but there we also put in custom warnings, release notes, and so on to catch this.

krschau:
Should I change this to use the 2.1.3 version that just came out? Does it need to be added to the feed? I don't see it available for updating in VS.

manodasanW:
> Should I change this to use the 2.1.3 version that just came out? Does it need to be added to the feed? I don't see it available for updating in VS.

I just published it so might take a bit to index.  But 2.1.3 shouldn't impact generated projections like here.  Fixes were in WinRT.Runtime and the source generator and it doesn't seem the test apps here were impacted by those issues.

jwoo-msft (approved):

PR #9032, opened 2024-10-04
Add Support for Carousel to UWP

jwoo-msft:
# Description
Added Carousel support for UWP
This is done trough taking dependency on WinUI2.
This may impact host app & host app needs to take exact version of WinUI2 in its package graph.
The consequence of this is not substantial; it will simply not render carousel.

# How Verified

manual verified for both winui3 and uwp


source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
```diff
@@ -36,23 +37,25 @@
...
     <DefaultLanguage>en-US</DefaultLanguage>
-    <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+    <MinimumVisualStudioVersion>15.0</MinimumVisualStudioVersion>
     <AppContainerApplication>true</AppContainerApplication>
     <ApplicationType>Windows Store</ApplicationType>
-    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
-    <WindowsTargetPlatformMinVersion>10.0.15063.0</WindowsTargetPlatformMinVersion>
     <ApplicationTypeRevision>10.0</ApplicationTypeRevision>
+    <WindowsTargetPlatformVersion Condition=" '$(WindowsTargetPlatformVersion)' == '' ">10.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformMinVersion>10.0.22621.0</WindowsTargetPlatformMinVersion>
```
paulcam206, line 53: is this correct? I think the min version should be lower than this

paulcam206 (approved):
code looks fine, but double check that min version :)

source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
jwoo-msft: I will double check that.

jwoo-msft: updated the minVersion

paulcam206 (approved):

PR #9044, opened 2024-10-22
add new preprocess macro def

jwoo-msft:
# Description
AC is compiled with a newer vclib that may not be compatible with older vclibs.
When AC is run in Shufan's app, it runs against the older vclibs, causing a crash. 
The added preprocessor macro is an escape hatch.
The more detail can be found in [this link](https://stackoverflow.com/questions/78598141/first-stdmutexlock-crashes-in-application-built-with-latest-visual-studio)

# Sample Card
N/A

# How Verified
Run sample cards.

### DIFF
--- a/source/uwp/AdaptiveCardsObjectModel/AdaptiveCardsObjectModel.vcxproj
+++ b/source/uwp/AdaptiveCardsObjectModel/AdaptiveCardsObjectModel.vcxproj
@@ -118,7 +118,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup>
     <ClCompile>
-      <PreprocessorDefinitions>_WINRT_DLL;_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_WINRT_DLL;_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING;_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <AdditionalIncludeDirectories>
         ..\..\shared\cpp\ObjectModel;
         ..\..\shared\cpp\ObjectModel\json;
--- a/source/uwp/AdaptiveCardsObjectModel/AdaptiveCardsObjectModel.vcxproj
+++ b/source/uwp/AdaptiveCardsObjectModel/AdaptiveCardsObjectModel.vcxproj
@@ -118,7 +118,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup>
     <ClCompile>
-      <PreprocessorDefinitions>_WINRT_DLL;_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING;_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR%(PreprocessorDefinitions)</PreprocessorDefinitions>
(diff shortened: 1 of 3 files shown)


source/uwp/AdaptiveCardsObjectModel/AdaptiveCardsObjectModel.vcxproj
```diff
@@ -118,7 +118,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup>
     <ClCompile>
-      <PreprocessorDefinitions>_WINRT_DLL;_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_WINRT_DLL;_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING;_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
shufan, line 121: Should there be a ";" between `_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR` and `%(PreprocessorDefinitions)`

shufan: Same for other file.

shufan (approved):

source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
```diff
@@ -120,7 +120,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup>
     <ClCompile>
-      <PreprocessorDefinitions>_WINRT_DLL;_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_WINRT_DLL;_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING;_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
dmachaj, line 123: This vcpxroj statically links ObjectModelProjection.vcxproj.  That vcxproj does not define `_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR`.  Is that going to create ODR violations?

source/uwp/ObjectModelProjection/ObjectModelProjection.vcxproj
```diff
@@ -246,6 +246,14 @@
       <TreatWarningAsError>true</TreatWarningAsError>
       <DisableSpecificWarnings>4702</DisableSpecificWarnings>
       <ControlFlowGuard Condition="'$(Configuration)'=='Release'">Guard</ControlFlowGuard>
+      <PreprocessorDefinitions Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">_UNICODE;UNICODE;_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
dmachaj, line 249: These doesn't need conditions.  Every version is the same.

jwoo-msft: updated it

source/uwp/Renderer/AdaptiveCardRenderer.vcxproj
jwoo-msft: updated it

dmachaj (approved):

PR #9057, opened 2024-11-14
Switch WinUI3 + UWP AdaptiveCard projects to use HybridCRT

### LINKED ISSUE #9056: UWP/WinUI3 AdaptiveCard projects need to switch to HybridCRT for consumption by Windows system components
It is forbidden for Windows OS system components to take a dependency on the Desktop CRT, as it is present by default in the OS and cannot be serviced with the OS.

By switching to HybridCRT, the UWP/WinUI3 AdaptiveCards projects will be able to dynamically link to the Universal CRT, shipped in the OS and statically link to the rest of the CRT such that it will always be available.

shufan:
Fixes #9056

# Description

This PR follows the guidance at aka.ms/hybridcrt to switch the UWP/WinUI3 AdaptiveCard projects to use HybridCRT. This is accomplished simply by copying over the [HybridCRT.props](https://github.com/microsoft/WindowsAppSDK/blob/main/HybridCRT.props) and importing them to the appropriate projects, and specifying 	`<UseCrtSDKReferenceStaticWarning>false</UseCrtSDKReferenceStaticWarning>`

# Sample Card

N/A

# How Verified

- [x] Built projects and used Depends.exe to validate that the DLLs produced no longer depended upon the Desktop CRT or UWP CRT DLLs
- [x] Tested UWP and WinUI3 rendering in AdaptiveCardVisualizer with various sample Widgets
  - [x] In both these cases, was able to confirm the DLLs being used in the Visualizer for rendering were using HybridCRT, as Depends.exe didn't show any dependency on the Desktop or UWP CRT
- [x] Replaced the current Client.CBS package DLLs for Widgets with these (AdaptiveCards.Rendering.Uwp.dll and AdaptiveCards.ObjectModel.Uwp.dll) and confirmed they were able to be loaded and used to render Widgets and resolved the current issue encountered with the current DLLs.
- [x] Replaced the DLLs for WidgetBoard with these (AdaptiveCards.Rendering.WinUI3.dll and AdaptiveCards.ObjectModel.WinUI3.dll) and confirmed they were able to be loaded and used to render Widgets

### DIFF
--- a/source/uwp/AdaptiveCardsObjectModel/AdaptiveCardsObjectModel.vcxproj
+++ b/source/uwp/AdaptiveCardsObjectModel/AdaptiveCardsObjectModel.vcxproj
@@ -53,8 +53,11 @@
     <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>
     <CppWinRTFastAbi>true</CppWinRTFastAbi>
     <CppWinRTOptimized>true</CppWinRTOptimized>
+	<UseCrtSDKReferenceStaticWarning>false</UseCrtSDKReferenceStaticWarning>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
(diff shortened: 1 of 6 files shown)


paulcam206 (approved):
Looks great!

PR #9069, opened 2024-12-09
refactor: only use PolySharp as private asset

theknut:
# PolySharp adds transitive build behavior

## Description

PolySharp is used in `AdaptiveCards.Templating` and adds transitive build behavior which introduces a security risk for package consumers. Furthermore PolySharp is not required for the usage of AdaptiveCards.

## Solution
Add `PrivatAssets = all` to PolySharp package reference.


fowl2:
hi @paulcam206 you recently updated this file, could you please take a look?

PR #9088, opened 2025-01-15
[Regression Fix]: Use FrameworkElement instead of empty statics for dependency property

beervoley:
# Description
One of the partner teams discovered a bug that leads to a crash due to an invalid pointer read. This is regression that occurred during WRL -> Cpp/WinRT migration of the UWP codebase.

# How Verified
Verified locally and mentioned functionality in WholeItemsPanel works as expected now.

### DIFF
--- a/source/uwp/SharedRenderer/lib/WholeItemsPanel.cpp
+++ b/source/uwp/SharedRenderer/lib/WholeItemsPanel.cpp
@@ -104,8 +104,7 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
 
                                     // Text did not reach its minHeight property and we have to clear it otherwise
                                     // ellipses won't be displayed correctly...
-                                    winrt::IFrameworkElementStatics frameworkElementStatics;
-                                    child.ClearValue(frameworkElementStatics.MinHeightProperty());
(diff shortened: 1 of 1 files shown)


anna-dingler (approved):

source/uwp/SharedRenderer/lib/WholeItemsPanel.cpp
```diff
@@ -104,8 +104,7 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
 
                                     // Text did not reach its minHeight property and we have to clear it otherwise
                                     // ellipses won't be displayed correctly...
-                                    winrt::IFrameworkElementStatics frameworkElementStatics;
-                                    child.ClearValue(frameworkElementStatics.MinHeightProperty());
+                                    child.ClearValue(winrt::FrameworkElement::MinHeightProperty());
```
jwoo-msft, line 107: curious, have you searched MinHeightProperty() in other places if they use this syntax?

jwoo-msft: great. just to be sure; I meant we searched & removed the old syntax.

beervoley: agreed, I did that before checking in -> no other instances found using the old syntax.

PR #9249, opened 2025-08-06
User/satwiksharma/check for strong this

SatwikKrSharma:
# Description
This PR addresses a security vulnerability in the SerializationTests.cs file related to the use of TypeNameHandling.All in JSON serialization. The TypeNameHandling.All setting in Newtonsoft.Json can lead to type-based deserialization attacks if used with untrusted data.
While this is a test project and not production code, it's still important to demonstrate secure coding practices. This PR modifies the Test_TypeHandling method to implement a safer approach by using a custom SerializationBinder that restricts which types can be deserialized.
The SafeTypeBinder class maintains a whitelist of allowed types and only permits those specific types to be deserialized, preventing potential injection of malicious types. This approach maintains the functionality needed for testing while implementing a more secure pattern.

# Sample Card

N/A - This change is to test code only and doesn't affect any sample cards.

# How Verified

1.	Existing unit tests continue to pass with the same functionality
2.	The Test_TypeHandling method now uses a safer approach while maintaining identical test behavior
3.	The implementation follows security best practices for JSON deserialization with type information


pchaurasia14 (approved):

JanviMahajan14 (approved):

JanviMahajan14:
Need to check why one of the checks is failing

pchaurasia14 (approved):

PR #9349, opened 2026-02-18
Carousel layout cycle renderer bug

SatwikKrSharma:
# Description
**Root Cause**
The root cause of the "Layout cycle detected" error:

StackPanel (Vertical) → gives FlipView infinite height
  └─ FlipView (MaxH=inf) → sizes to content, asks pages for height
      └─ Page LayoutUpdated → calls SetFlipViewMaxHeight()
          └─ SetFlipViewMaxHeight() calls page.Measure() + sets flipView.MaxHeight()
              └─ This invalidates FlipView layout → triggers LayoutUpdated again → ∞ cycle

The problem is in AdaptiveCarouselRenderer.cpp. When HeightInPixels is 0, the LayoutUpdated handler calls Measure() on the page and sets MaxHeight on the FlipView during layout, which re-triggers layout — creating an infinite cycle.
Two issues compound this:
1.	LayoutUpdated fires during every layout pass — it's not a one-shot event
2.	The condition previousMaxHeight < maxHeight doesn't prevent oscillation because Measure() with infinity can return varying results

The carousel renderer was using LayoutUpdated handlers inside the per-page loop to dynamically compute the FlipView MaxHeight. Because LayoutUpdated fires during the layout pass, calling Measure() and setting MaxHeight inside it would re-invalidate layout, triggering an infinite cycle. This was the direct cause of the "Layout cycle detected" error.
Additionally, event handlers captured carouselUI by value (strong WinRT reference) and were never unregistered, leading to potential memory leaks when carousels were dynamically created and destroyed.
Changes
Replaced LayoutUpdated with safe, post-layout events:

- Loaded — fires once after the element is in the visual tree and has been measured. Used to compute the initial MaxHeight safely outside of a layout pass.

- SizeChanged — fires after layout completes (not during). Handles window resize or container width changes without causing layout invalidation.

- SelectionChanged for height update — different carousel pages may have different content heights, so MaxHeight is recalculated when the selected page changes.

- Added re-entrancy guard (isUpdatingHeight):

- A std::shared_ptr<bool> flag is captured by all three lambdas. This prevents SetFlipViewMaxHeight → MaxHeight change → SizeChanged → SetFlipViewMaxHeight feedback loops.

- Added width <= 0 guard in SetFlipViewMaxHeight:

If ActualWidth is 0 (element not yet laid out), Measure with zero width produces meaningless results. The early return avoids unnecessary work and incorrect height values.

- Added Unloaded handler for event cleanup:

An Unloaded handler revokes the Loaded, SizeChanged, and SelectionChanged event tokens when the FlipView is removed from the visual tree. This breaks the strong reference cycles created by the capturing lambdas and prevents memory leaks in scenarios with many carousel instances.

**JSON which is creating the issue -**
{
  "type": "AdaptiveCard",
  "version": "1.6",
  "body": [
    {
      "type": "Carousel",
      "rtl": false,
      "pages": [
        {
          "type": "CarouselPage",
          "rtl": false,
          "items": [
            {
              "type": "Container",
              "items": [
                {
                  "type": "ColumnSet",
                  "spacing": "None",
                  "columns": [
                    {
                      "type": "Column",
                      "width": "100px",
                      "items": [
                        {
                          "type": "Image",
                          "url": "https://picsum.photos/seed/screen-01/200/200",
                          "altText": "Open screen recorder",
                          "horizontalAlignment": "Center",
                          "width": "76px",
                          "height": "64px"
                        }
                      ],
                      "horizontalAlignment": "Center",
                      "verticalContentAlignment": "Center",
                      "spacing": "None"
                    },
                    {
                      "type": "Column",
                      "width": "auto",
                      "items": [
                        {
                          "type": "TextBlock",
                          "text": "Record from screen",
                          "weight": "Bolder",
                          "spacing": "Small",
                          "wrap": true
                        },
                        {
                          "type": "TextBlock",
                          "text": "Capture your screen with audio in one tap.",
                          "wrap": true,
                          "spacing": "None",
                          "isSubtle": false,
                          "size": "Small",
                          "style": "default",
                          "fontType": "Default"
                        }
                      ],
                      "verticalContentAlignment": "Center",
                      "horizontalAlignment": "Left"
                    }
                  ],
                  "minHeight": "110px",
                  "horizontalAlignment": "Left",
                  "height": "stretch"
                }
              ],
              "backgroundImage": {
                "url": "https://images.unsplash.com/photo-1522199710521-72d69614c702?auto=format&fit=crop&w=1200&q=60",
                "horizontalAlignment": "Center",
                "verticalAlignment": "Center"
              },
              "minHeight": "110px",
              "selectAction": {
                "type": "Action.OpenUrl",
                "tooltip": "Open screen recorder",
                "url": "https://example.com/record/screen",
                "role": "Link"
              },
              "spacing": "None",
              "height": "stretch",
              "horizontalAlignment": "Center",
              "verticalContentAlignment": "Center"
            },
            {
              "type": "Container",
              "items": [
                {
                  "type": "ColumnSet",
                  "spacing": "None",
                  "columns": [
                    {
                      "type": "Column",
                      "width": "100px",
                      "items": [
                        {
                          "type": "Image",
                          "url": "https://picsum.photos/seed/camera-02/200/200",
                          "altText": "Open camera recorder",
                          "horizontalAlignment": "Center",
                          "width": "76px",
                          "height": "64px"
                        }
                      ],
                      "horizontalAlignment": "Center",
                      "verticalContentAlignment": "Center",
                      "spacing": "None"
                    },
                    {
                      "type": "Column",
                      "width": "auto",
                      "items": [
                        {
                          "type": "TextBlock",
                          "text": "Record from camera",
                          "weight": "Bolder",
                          "spacing": "Small",
                          "wrap": true
                        },
                        {
                          "type": "TextBlock",
                          "text": "Use your webcam to record a quick clip.",
                          "wrap": true,
                          "spacing": "None",
                          "isSubtle": false,
                          "size": "Small",
                          "style": "default",
                          "fontType": "Default"
                        }
                      ],
                      "verticalContentAlignment": "Center",
                      "horizontalAlignment": "Left"
                    }
                  ],
                  "minHeight": "110px",
                  "horizontalAlignment": "Left",
                  "height": "stretch"
                }
              ],
              "backgroundImage": {
                "url": "https://images.unsplash.com/photo-1526498460520-4c246339dccb?auto=format&fit=crop&w=1200&q=60",
                "horizontalAlignment": "Center",
                "verticalAlignment": "Center"
              },
              "minHeight": "110px",
              "selectAction": {
                "type": "Action.OpenUrl",
                "tooltip": "Open camera recorder",
                "role": "Link",
                "url": "https://example.com/record/camera"
              },
              "spacing": "Small",
              "horizontalAlignment": "Center",
              "verticalContentAlignment": "Center"
            }
          ]
        },
        {
          "type": "CarouselPage",
          "rtl": false,
          "items": [
            {
              "type": "Container",
              "items": [
                {
                  "type": "ColumnSet",
                  "spacing": "Medium",
                  "columns": [
                    {
                      "type": "Column",
                      "width": "100px",
                      "items": [
                        {
                          "type": "Image",
                          "url": "https://picsum.photos/seed/editor-03/200/200",
                          "altText": "Open editor mode",
                          "horizontalAlignment": "Center",
                          "width": "76px",
                          "height": "64px"
                        }
                      ],
                      "horizontalAlignment": "Left",
                      "verticalContentAlignment": "Center",
                      "height": "stretch"
                    },
                    {
                      "type": "Column",
                      "width": "auto",
                      "items": [
                        {
                          "type": "TextBlock",
                          "text": "Record from editor",
                          "wrap": true,
                          "weight": "Bolder",
                          "spacing": "Small"
                        },
                        {
                          "type": "TextBlock",
                          "text": "Create a recording while editing your document.",
                          "wrap": true,
                          "spacing": "None",
                          "isSubtle": false,
                          "style": "default",
                          "fontType": "Default",
                          "size": "Small",
                          "weight": "Default"
                        }
                      ],
                      "verticalContentAlignment": "Center",
                      "horizontalAlignment": "Left"
                    }
                  ],
                  "minHeight": "110px",
                  "horizontalAlignment": "Left",
                  "verticalContentAlignment": "Top",
                  "height": "stretch"
                }
              ],
              "backgroundImage": {
                "url": "https://images.unsplash.com/photo-1521737604893-d14cc237f11d?auto=format&fit=crop&w=1200&q=60"
              },
              "minHeight": "110px",
              "selectAction": {
                "type": "Action.OpenUrl",
                "tooltip": "Open editor mode",
                "url": "https://example.com/record/editor",
                "role": "Link"
              }
            },
            {
              "type": "Container",
              "height": "stretch",
              "minHeight": "110px"
            }
          ],
          "verticalContentAlignment": "Top"
        }
      ]
    }
  ]
}

# How Verified
Verified carousel renders without "Layout cycle detected" errors

### DIFF
--- a/source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
+++ b/source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
@@ -123,16 +123,29 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
                 {
                     carouselUI.Items().Append(carouselPageUI);
                 }
+            }
 
-                // If no fixed height was specified, compute the max height from page content
-                // after the FlipView has been loaded (not during layout) to avoid layout cycles.
-                if (fixedHeightInPixel == 0)
-                {
-                    auto isUpdatingHeight = std::make_shared<bool>(false);
+            // If no fixed height was specified, compute the max height from page content
+            // after the FlipView has been loaded (not during layout) to avoid layout cycles.
+            if (fixedHeightInPixel == 0)
+            {
+                auto isUpdatingHeight = std::make_shared<bool>(false);
 
-                    // Use Loaded event to set initial height — fires once after the element
-                    // is in the visual tree and has been measured, but not during a layout pass.
-                    auto loadedToken = carouselUI.Loaded([carouselUI, isUpdatingHeight](auto&&, auto&&) {
+                // Use Loaded event to set initial height — fires once after the element
+                // is in the visual tree and has been measured, but not during a layout pass.
+                auto loadedToken = carouselUI.Loaded([carouselUI, isUpdatingHeight](auto&&, auto&&) {
+                    if (*isUpdatingHeight)
+                        return;
+                    *isUpdatingHeight = true;
+                    SetFlipViewMaxHeight(carouselUI);
+                    *isUpdatingHeight = false;
+                });
+
+                // Use SizeChanged to adjust if the FlipView width changes (e.g. window resize).
+                // SizeChanged fires after layout completes, so it won't cause a layout cycle
+                // as long as we guard against re-entrancy.
+                auto sizeChangedToken =
+                    carouselUI.SizeChanged([carouselUI, isUpdatingHeight](auto&&, winrt::SizeChangedEventArgs const&) {
                         if (*isUpdatingHeight)
                             return;
                         *isUpdatingHeight = true;
@@ -140,34 +153,20 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
                         *isUpdatingHeight = false;
                     });
 
-                    // Use SizeChanged to adjust if the FlipView width changes (e.g. window resize).
-                    // SizeChanged fires after layout completes, so it won't cause a layout cycle
-                    // as long as we guard against re-entrancy.
-                    auto sizeChangedToken =
-                        carouselUI.SizeChanged([carouselUI, isUpdatingHeight](auto&&, winrt::SizeChangedEventArgs const&) {
-                        if (*isUpdatingHeight)
-                            return;
-                        *isUpdatingHeight = true;
-                        SetFlipViewMaxHeight(carouselUI);
-                        *isUpdatingHeight = false;
-                    });
+                // Update height when the selected page changes, since pages may differ in size.
+                auto selectionChangedToken = carouselUI.SelectionChanged([carouselUI, isUpdatingHeight](auto&&, auto&&) {
+                    if (*isUpdatingHeight)
+                        return;
+                    *isUpdatingHeight = true;
+                    SetFlipViewMaxHeight(carouselUI);
+                    *isUpdatingHeight = false;
+                });
 
-                    // Update height when the selected page changes, since pages may differ in size.
-                    auto selectionChangedToken = carouselUI.SelectionChanged([carouselUI, isUpdatingHeight](auto&&, auto&&) {
-                        if (*isUpdatingHeight)
-                            return;
-                        *isUpdatingHeight = true;
-                        SetFlipViewMaxHeight(carouselUI);
-                        *isUpdatingHeight = false;
-                    });
-
-                    carouselUI.Unloaded([ carouselUI, loadedToken, sizeChangedToken, selectionChangedToken ](auto&&, auto&&) {
-                        carouselUI.Loaded(loadedToken);
-                        carouselUI.SizeChanged(sizeChangedToken);
-                        carouselUI.SelectionChanged(selectionChangedToken);
-                    });
-
-                }
+                carouselUI.Unloaded([carouselUI, loadedToken, sizeChangedToken, selectionChangedToken](auto&&, auto&&) {
+                    carouselUI.Loaded(loadedToken);
+                    carouselUI.SizeChanged(sizeChangedToken);
+                    carouselUI.SelectionChanged(selectionChangedToken);
+                });
             }
 
             if (carousel.InitialPage())
--- a/source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
+++ b/source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
@@ -11,240 +11,200 @@
 
 namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
 {
-winrt::UIElement AdaptiveCarouselRenderer::Render(
-    winrt::AdaptiveCards::ObjectModel::Xaml_Rendering::IAdaptiveCardElement const& element,
-    winrt::AdaptiveCards::Rendering::Xaml_Rendering::AdaptiveRenderContext const& context,
-    winrt::AdaptiveCards::Rendering::Xaml_Rendering::AdaptiveRenderArgs const& renderArgs)
-{
-    auto carousel = element.as<winrt::IAdaptiveCarousel>();
-    auto styledCollection = element.as<winrt::IAdaptiveContainerBase>();
-
-    StackPanel stackPanel{};
-    FlipView carouselUI{};
-    bool loopEnabled{false};
-
-    if (carousel.AutoLoop())
+    winrt::UIElement AdaptiveCarouselRenderer::Render(
+        winrt::AdaptiveCards::ObjectModel::Xaml_Rendering::IAdaptiveCardElement const& element,
+        winrt::AdaptiveCards::Rendering::Xaml_Rendering::AdaptiveRenderContext const& context,
+        winrt::AdaptiveCards::Rendering::Xaml_Rendering::AdaptiveRenderArgs const& renderArgs)
     {
-        loopEnabled = carousel.AutoLoop().GetBoolean();
-    }
+        auto carousel = element.as<winrt::IAdaptiveCarousel>();
+        auto styledCollection = element.as<winrt::IAdaptiveContainerBase>();
 
-    PipsPager pipsPager{};
-    pipsPager.HorizontalAlignment(winrt::HorizontalAlignment::Center);
-    pipsPager.NumberOfPages(carousel.Pages().Size());
-    auto hostConfig = context.HostConfig();
+        StackPanel stackPanel{};
+        FlipView carouselUI{};
+        bool loopEnabled{false};
 
-    // FlipView has its own background color property, so we need to clear the background color
-    carouselUI.Background(winrt::SolidColorBrush{GetColorFromString("#00FFFFFF")});
-
-    // auto adaptiveCarouselContainer = element.as<winrt::IAdaptiveContainerBase>();
-    //  Get any RTL setting set on either the current context or on this container. Any value set on the
-    //  container should be set on the context to apply to all children
-    auto previousContextRtl = context.Rtl();
-    auto currentRtl = previousContextRtl;
-    auto containerRtl = carousel.Rtl();
-
-    bool updatedRtl = false;
-    if (containerRtl)
-    {
-        currentRtl = containerRtl;
-        context.Rtl(currentRtl);
-        updatedRtl = true;
-    }
-
-    if (currentRtl)
-    {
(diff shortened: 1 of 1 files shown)


source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
```diff
@@ -192,11 +231,22 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
     void SetFlipViewMaxHeight(FlipView const& flipView)
     {
         auto selectIndex = flipView.SelectedIndex();
+
+        // FIX: Validate selectedIndex is within bounds
+        if (selectIndex < 0 || static_cast<uint32_t>(selectIndex) >= flipView.Items().Size())
```
dipeshmsft, line 235: Is this necessary here ?

If the SelectedIndex is set by the user via the UI, the checks should already be present there right ? Otherwise the selection won't make sense.

I am not able to find the code for FlipView so I can't verify what I am thinking, at the moment.

source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
```diff
@@ -192,11 +231,22 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
...
+            return;
+        }
+
         auto carouselPageUI = flipView.Items().GetAt(selectIndex).try_as<FrameworkElement>();
 
-        if (flipView.IsLoaded())
+        if (carouselPageUI != nullptr && flipView.IsLoaded())
         {
             float width = static_cast<float>(flipView.ActualWidth());
+            if (width <= 0)
```
dipeshmsft, line 245: ActualWidth can be 0 in cases where the UI control is collapsed as well ( assuming we have collapsed functionality present here )

Moreover ActualWidth can never be less than 0 right ? It returns the effective computed value which always boils down to >= 0.

source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
```diff
@@ -122,13 +122,52 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
...
+                    });
+
+                    // Use SizeChanged to adjust if the FlipView width changes (e.g. window resize).
+                    // SizeChanged fires after layout completes, so it won't cause a layout cycle
+                    // as long as we guard against re-entrancy.
+                    auto sizeChangedToken =
+                        carouselUI.SizeChanged([carouselUI, isUpdatingHeight](auto&&, winrt::SizeChangedEventArgs const&) {
+                        if (*isUpdatingHeight)
+                            return;
+                        *isUpdatingHeight = true;
```
dipeshmsft: Probably, we can move the isUpdatingHeight value setting to the  SetFlipViewMaxHeight method, will prevent redundancy.

source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
SatwikKrSharma: SetFlipViewMaxHeight is invoked from three event handlers:
 
1.	Loaded — fires when the FlipView is first loaded. At this point, SelectedIndex defaults to 0 if there are items, or -1 if the Items collection is empty.
2.	SizeChanged — fires when the FlipView is resized. The selected index is already established by the control.
3.	SelectionChanged — fires when the user or code changes the selection.

The key issue is the Loaded event: if the carousel has zero pages (e.g., all pages failed to render), the Items collection would be empty, and SelectedIndex would be -1. The FlipView control does not guarantee that SelectedIndex is always valid — it returns -1 when there's no selection. So the selectIndex < 0 check is genuinely needed.

Additionally, there's a subtle race with the loop logic higher up in Render:
if (loopEnabled &&
    static_cast<unsigned int>(val) == carouselUI.Items().Size())

This code compares the selected index to the item count, suggesting the index can temporarily be at an out-of-bounds position when looping is enabled, which could also trigger SelectionChanged → SetFlipViewMaxHeight with an invalid index.

---
The bounds check is necessary for two reasons:
1.	Empty items case: If all carousel pages fail to render (e.g., fallback), Items is empty and SelectedIndex returns -1. The Loaded and SizeChanged events will still fire, and without the < 0 check, the code would call Items().GetAt(-1) which would throw.
2.	Loop edge case: The SelectionChanged handler in Render explicitly handles SelectedIndex == Items().Size() for loop mode, meaning the index can momentarily be at the size boundary. Since the same SelectionChanged event also triggers SetFlipViewMaxHeight, the >= Size() check prevents an out-of-bounds access before the loop handler resets the index to 0.

source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
SatwikKrSharma: ActualWidth is 0 before the first layout pass completes or if the control is collapsed.

If the FlipView's Visibility is set to Collapsed:
1.	ActualWidth returns 0, collapsed elements have no rendered size.
2.	IsLoaded() still returns true, a collapsed element remains in the visual tree; Loaded/Unloaded are about tree membership, not visibility.
3.	The current code hits width == 0 and returns early, so there's no crash, but no MaxHeight is set either.

The real question is: should we even attempt to measure when the FlipView is collapsed?
No, calling Measure on a collapsed element's children is meaningless because the available width is 0, producing incorrect DesiredSize values. If the FlipView later becomes visible, SizeChanged will fire with the correct dimensions and SetFlipViewMaxHeight will run again with a valid width.

dipeshmsft: As long as SizeChanged is fired on visibility, then it's fine.

source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
dipeshmsft: Can you please point me to the FlipView's code? I want to verify index out of range situation. Index < 0 makes complete sense.

dipeshmsft: Anyhow there is no harm in having an extra check, hence resolving the comment.

source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
```diff
@@ -122,13 +122,51 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
...
+
+                // Use SizeChanged to adjust if the FlipView width changes (e.g. window resize).
+                // SizeChanged fires after layout completes, so it won't cause a layout cycle
+                // as long as we guard against re-entrancy.
+                auto sizeChangedToken =
+                    carouselUI.SizeChanged([carouselUI, isUpdatingHeight](auto&&, winrt::SizeChangedEventArgs const&) {
+                        if (*isUpdatingHeight)
+                            return;
+                        *isUpdatingHeight = true;
+                        SetFlipViewMaxHeight(carouselUI);
```
godlytalias, line 152: Probably this may cause flickers as this will lead to mismatch of timing where the size change happens for CarouselUI and FlipView

source/uwp/SharedRenderer/lib/AdaptiveCarouselRenderer.cpp
```diff
@@ -122,13 +122,51 @@ namespace winrt::AdaptiveCards::Rendering::Xaml_Rendering::implementation
...
+            {
+                auto isUpdatingHeight = std::make_shared<bool>(false);
+
+                // Use Loaded event to set initial height — fires once after the element
+                // is in the visual tree and has been measured, but not during a layout pass.
+                auto loadedToken = carouselUI.Loaded([carouselUI, isUpdatingHeight](auto&&, auto&&) {
+                    if (*isUpdatingHeight)
+                        return;
+                    *isUpdatingHeight = true;
+                    SetFlipViewMaxHeight(carouselUI);
```
godlytalias, line 140: SetFlipViewMaxHeight seems to be an operation that figure out the size for flipview by calling Measure operations on the child contents. So ideally this should get called during the Measure operation of FlipView to avoid such cycles.

PR #9384, opened 2026-07-07
[Android ][iOS] Fix unsafe AdaptiveBase64 decode handling

karthikgopal36:
Guard malformed base64 padding and invalid input before decoding. Decode into a temporary vector with bounded push_back writes so malformed data cannot write past the allocated output buffer.

# Description

Fixes unsafe Base64 decoding in  AdaptiveBase64Util  by validating malformed padding/input before decoding and avoiding unchecked writes into a pre-sized output buffer.

Changes

• Decode into a temporary vector using bounded  push_back  writes.
• Reject malformed Base64 inputs such as excessive padding, invalid length, invalid characters, and padding in the middle.
• Preserve valid padded and unpadded decode behavior.
• Add regression tests for valid unpadded inputs and malformed padding cases.

# Sample Card

Using Image Elements cards

### DIFF
--- a/source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/Base64Test.cpp
+++ b/source/shared/cpp/AdaptiveCardsSharedModel/AdaptiveCardsSharedModelUnitTest/Base64Test.cpp
@@ -60,14 +60,42 @@ namespace AdaptiveCardsSharedModelUnitTest
             }
         }
 
+        TEST_METHOD(DecodingUnpaddedTest)
+        {
+            std::vector<std::string> expectedDecodedData{ "f", "fo", "foobar" };
+            std::vector<std::string> encodedData{ "Zg", "Zm8", "Zm9vYmFy" };
+
+            for (size_t i{}; i < encodedData.size(); ++i)
+            {
+                Assert::IsTrue(ContainSameCharacters(expectedDecodedData[i],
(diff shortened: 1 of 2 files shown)


hggzm:
confirming this is cross-platform, not android only. the change is entirely in the shared cpp ObjectModel (AdaptiveBase64Util.cpp), and iOS decodes data:image URIs through the same path: ACRView.mm calls AdaptiveBase64Util::ExtractDataFromUri + AdaptiveBase64Util::Decode (around line 518), and android goes through the JNI binding to the same code. there is no separate swift base64 impl on main, so this one fix hardens the data-uri image path on iOS and android both.

might be worth dropping the [Android] tag from the title so security/reviewers can see it protects iOS too.

fwiw i validated the same diff on the ios side out of band: ported it into our teams adaptivecards fork, it compiles clean in the ios build, and a set of ios UI tests rendering data:image backgroundImage cards (including a deliberately malformed base64 one) all pass without crashing. happy to share the run if useful.

hggzm (approved):
verified this hardens the shared cpp base64 decode used by both ios (ACRView.mm) and android. malformed data-uri base64 is now rejected safely instead of writing past the output buffer. validated on ios out of band (compiles + UI tests with malformed base64 cards pass without crashing).

GabrielMedAlv (approved):

ardlank (approved):

PR #9396, opened 2026-08-25
Add Dependabot version updates config for AdaptiveCards ecosystems

Sandeep-622:
# Add Dependabot version updates configuration

## What
Adds `.github/dependabot.yml` to enable Dependabot version-update PRs across the package ecosystems actually used in AdaptiveCards.

## Ecosystems covered
| Ecosystem | Directory / Directories |
| --- | --- |
| `npm` | `/source/nodejs` (lerna + npm-workspaces monorepo — its single top-level `package.json` + lockfile covers every workspace package) |
| `nuget` | `/source/dotnet`, `/source/uwp`, `/source/uwp/winui3`, `/source/ios/tools/IOSFeed` |
| `gradle` | `/source/android` (its `settings.gradle` includes the adaptivecards, mobile, mobilechatapp, and uitestapp subprojects) |
| `swift` | `/` (Package.swift) |
| `github-actions` | `/` (workflows) |

## Settings
- **Schedule:** weekly for every ecosystem
- **Open PR limit:** 1 per ecosystem (keeps noise low)
- Ecosystem-specific labels applied

## Scoping rationale
Directories were verified against the actual repo layout rather than using broad globs:
- **npm** is scoped to `/source/nodejs` only. The repo-root and `/schemas` `package-lock.json` files are empty stubs (`{"lockfileVersion": 1}`) with **no `package.json`** — Dependabot's npm updater would error on them, so they are intentionally excluded.
- **nuget** targets the concrete solution roots (each has its own `.sln` with real `PackageReference`/`packages.config` dependencies) instead of broad `**` globs, which the Dependabot guidance warns can be slow / time out for NuGet.
- `source/shared/cpp` C++ projects have **no NuGet dependencies**, so they are not listed.
- The `source/dotnet/.../package.json` is a **Unity (UPM)** manifest, not npm — Dependabot cannot process it, so it is excluded.

## Notes
- CocoaPods (iOS `Podfile`) is intentionally omitted — it is not a Dependabot-supported ecosystem.
- The `swift` entry is currently inert (`Package.swift` declares no external SPM dependencies) but is kept for forward-compatibility; it is a harmless no-op until an SPM dependency is added.
- This is a configuration file, not a pipeline; it enables version-update PRs. Dependabot alerts / security updates are managed separately in repo settings.

## Related
Part of the Component Governance onboarding effort for `github:microsoft/AdaptiveCards`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,62 @@
+version: 2
+updates:
+  # ---- JavaScript / TypeScript (npm) ----
+  # /source/nodejs is a lerna + npm-workspaces monorepo; its single top-level
+  # package.json + lockfile covers every workspace package (adaptivecards,
+  # -templating, -react, etc.). The repo-root and /schemas lockfiles are empty
+  # stubs with no package.json, so they are intentionally not listed.
+  - package-ecosystem: "npm"
+    directory: "/source/nodejs"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 1
+    labels:
+      - "dependencies"
+      - "javascript"
+
+  # ---- .NET (NuGet) ----
+  # Target the concrete solution roots (each has its own .sln that references
+  # its projects) rather than broad "**" globs, which the Dependabot guidance
+  # warns can be slow / time out for NuGet.
+  - package-ecosystem: "nuget"
+    directories:
+      - "/source/dotnet"
+      - "/source/uwp"
+      - "/source/uwp/winui3"
+      - "/source/ios/tools/IOSFeed"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 1
+    labels:
+      - "dependencies"
+      - "dotnet"
+
+  # ---- Android (Gradle) ----
+  - package-ecosystem: "gradle"
+    directory: "/source/android"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 1
+    labels:
+      - "dependencies"
+      - "android"
+
+  # ---- Swift Package Manager (Package.swift at repo root) ----
+  - package-ecosystem: "swift"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 1
+    labels:
+      - "dependencies"
+      - "swift"
+
+  # ---- GitHub Actions workflows ----
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


raghavmadan1034 (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,62 @@
...
+updates:
+  # ---- JavaScript / TypeScript (npm) ----
+  # /source/nodejs is a lerna + npm-workspaces monorepo; its single top-level
+  # package.json + lockfile covers every workspace package (adaptivecards,
+  # -templating, -react, etc.). The repo-root and /schemas lockfiles are empty
+  # stubs with no package.json, so they are intentionally not listed.
+  - package-ecosystem: "npm"
+    directory: "/source/nodejs"
+    schedule:
+      interval: "weekly"
```
harshit7962, line 11: can we have it as every build?

.github/dependabot.yml
```diff
@@ -0,0 +1,62 @@
...
+    open-pull-requests-limit: 1
```
harshit7962, line 12: Is it batched on to same pull request in case a PR already exist? Or does it not create a PR itself?

.github/dependabot.yml
Sandeep-622: Dependabot's  schedule.interval  only supports daily , weekly , or monthly and there's no per-build/per-commit option, since Dependabot runs as a GitHub-scheduled service rather than as a step in our CI build. On each run it checks the registries and opens an update PR only when a newer version is available (so it already "runs" regardless of open PRs).
If we want to stay as current as possible, I'm happy to switch the interval from weekly to daily. I kept it weekly to limit PR noise but daily is fine if the team prefers faster updates let me know and I'll update it.

.github/dependabot.yml
Sandeep-622: **open-pull-requests-limit** isn't batching i.e it caps how many version-update PRs stay open at once for this ecosystem (here, 1). 
Dependabot still creates a PR, but it just won't open a second one while the first is open and additional updates are queued and opened one at a time as each is merged/closed. It does not combine multiple bumps into a single existing PR (that's the separate groups feature), and it doesn't skip creating a PR.

rashmi-thakurr (approved):

SatwikKrSharma:
Remove swift and /source/ios/tools/IOSFeed, /source/android as we no longer support them

SatwikKrSharma (approved):
Added one comment

